PR #2, opened 2021-02-21
Incorporating new schema

charliepnnl:
### Schema changes affecting Transformers:
- The units for transformer capacity changed from kVA to V-A
- The capacity_ratio field was eliminated

### Code changes that will affect Juan's testing:
- The `invalid_rmrs` attribute in the `evaluate_rule` return value is now a dictionary whose keys are the invalid RMR names and whose values are the corresponding validation errors. Note: Empty dictionaries and lists are Falsey in Python, so if there are no invalid RMRs, then the `invalid_rmrs` attribute will be empty and therefore Falsey.


JasonGlazer (approved):
Looks good to me. I'm still unfamiliar with most of your code but what I understood looks fine.

PR #13, opened 2021-03-28
CH RDDS 15-3 and -4 logic changes

charliepnnl:


supriyagoel (approved):
Discussed with Charlie, the list functionality doesn't require the applicability check anymore. Changes look good

PR #15, opened 2021-03-29
updated lighting rules

xingcx:


mcneillj:
@xingcx does this PR update the RDS to the current format?  i.e. adds Rule Logic

Is there any reason this cannot be merged, or can we go ahead with that?

PR #17, opened 2021-04-09
Ruleset 11 rules 15 5 15 6

charliepnnl:
This PR includes:
- Rules 15-5 and 15-6
- Improvements to Rules `15-3 and 15-4
- Passing a data object to the workflow methods of RuleDefiningBase
- Simplified matching of RMR lists; also ensures that the order of the list outcomes is the same as the index_rmr
- Other miscellaneous improvements


rct229/rule_engine/engine.py
```diff
@@ -27,6 +27,7 @@ def evaluate_all_rules(user_rmr, baseline_rmr, proposed_rmr):
 
     # Get reference to rule functions in rules model
     AvailableRuleDefinitions = rules.__getrules__()
+    print('len(AvailableRuleDefinitions):', len(AvailableRuleDefinitions))
```
charliepnnl: I've removed this print statement in a commit after this PR was created.

mcneillj (approved):

PR #21, opened 2021-04-16
Ruletest engine development

jugonzal07:
Hi all,

Sorry to not have ironed this out beforehand-- I will do so in the future. I made the small corrections needed to get the `pipenv run pytest` tests to succeed as well.


charliepnnl:
Can you change this PR to point to develop instead of master?  That will make the diffs more appropriate.

charliepnnl (approved):

mcneillj (approved):

PR #1, opened 2021-02-18
Schema from early feb

JasonGlazer:
Replace older schema files with latest from early February. Untested.

PR #22, opened 2021-04-18
Update Rule6-2.md

xingcx:


JasonGlazer:
Original discussion at:

https://github.com/pnnl/ruleset-checking-tool/pull/19/files/251b7f0445e13beff0663242b5a784202acc40e9#r615442469

JasonGlazer (approved):

supriyagoel (approved):

charliepnnl:
I think we need a way of specifying the level of items that should appear in the report.  For this rule, if I understand the intent, we need to report PASS, FAIL, etc for each `building`.  For the transformers, there was only one list so the "For each .." construct seemed to work for that as well as indicating looping. Here I see that there are multiple lists involved that we need to loop over beside the one that we need to report on.

I propose a variation on the "For each ..." construct, namely:
- Report for each `building` in `P_RMR.buildings`:

charliepnnl:
I favor indenting under a "For each ..." construct to indicate steps that logically occur inside the loop.

charliepnnl:
I think we can make the "For each ..." constructs more concise by combining the parts before and after the colon. For example, the beginning of this rule's logic could be:
- Report for each `building` in `P_RMR.buildings`:
  - For each `buliding_segment` in `building.building_segments`:

charliepnnl:
It doesn't really matter, but for this particular rule, we don't really need to append `_proposed` to all the variables because it is already clear from the context that on the Proposed Model is used throughout.

docs/section9/Rule6-ja.md
```diff
@@ -16,28 +16,42 @@
...
 ## Rule Logic: 
 
-- Get the building area lighting area type of the building segment in the Proposed model: ```building_light_area_type = building_segment.lighting_building_area_type``` (Note XC, assuming this field lists values from T-9.5.1 or T-G3.8, for "Building Area Method", or "None" if the building goes for "Space-by-Space Method")
+- For each building_segment in the Proposed Model: ```building_segment_proposed in P_RMR.building.building_segments```  
 
-- Get thermal_block from building segment: ```thermal_block in building_segment.thermal_blocks:```
+- Get the lighting_building_area_type for the building_segment: ```lighting_building_area_type_proposed = building_segment_proposed.lighting_building_area_type```  
 
-- Get thermal_zone from thermal block: ```thermal_zone in thermal_block.zones:```
+  - Determine if the building segment uses Building Area Method: ```if lighting_building_area_type_proposed is "None": bam_flag = FALSE else bam_flag = TRUE```
```
charliepnnl, line 23: Lines 21 and 23 could be combined by replacing `lighting_building_area_type_proposed` in line 23 with `buliding_segment_proposed.lighting_buildling_area_type`.

docs/section9/Rule6-ja.md
```diff
@@ -16,28 +16,42 @@
...
+- For each building_segment in the Proposed Model: ```building_segment_proposed in P_RMR.building.building_segments```  
 
-- Get thermal_block from building segment: ```thermal_block in building_segment.thermal_blocks:```
+- Get the lighting_building_area_type for the building_segment: ```lighting_building_area_type_proposed = building_segment_proposed.lighting_building_area_type```  
 
-- Get thermal_zone from thermal block: ```thermal_zone in thermal_block.zones:```
+  - Determine if the building segment uses Building Area Method: ```if lighting_building_area_type_proposed is "None": bam_flag = FALSE else bam_flag = TRUE```  
 
-- Get space from thermal zone: ```space in thermal_zone.spaces:```
+- For each thermal_block in building_segment: ```thermal_block_proposed in lighting_building_area_type_proposed.thermal_blocks:```
```
charliepnnl, line 25: `lighting_building_area_type_proposed.thermal_blocks` should be `building_segment_proposed.thermal_blocks`.

docs/section9/Rule6-ja.md
```diff
@@ -16,28 +16,42 @@
...
+- Get the lighting_building_area_type for the building_segment: ```lighting_building_area_type_proposed = building_segment_proposed.lighting_building_area_type```  
 
-- Get thermal_zone from thermal block: ```thermal_zone in thermal_block.zones:```
+  - Determine if the building segment uses Building Area Method: ```if lighting_building_area_type_proposed is "None": bam_flag = FALSE else bam_flag = TRUE```  
 
-- Get space from thermal zone: ```space in thermal_zone.spaces:```
+- For each thermal_block in building_segment: ```thermal_block_proposed in lighting_building_area_type_proposed.thermal_blocks:```  
 
-- Get floor_area from space: ```space_floor_area = space.floor_area```
+- For each thermal_zone in thermal_block: ```thermal_zone_proposed in thermal_block_proposed.zones:```
```
charliepnnl, line 27: Shouldn't this be `zone` instead of `thermal_zone`?  Thermal zone is not in the schema.

charliepnnl (changes requested):

charliepnnl:
Here's are some potentially thorny issues that must be resolved:

What should we do if a building has no building segments, or a building segment has no thermal_blocks, etc?  At this point, the schema allows any of these lists to be empty.

What if a building segment has no lighting_building_area_type field or an interior_lighting has no power_per_area field, etc?  These are not currently required by the schema.

docs/section9/Rule6-ja.md
JasonGlazer: I agree.

JasonGlazer:
@charliepnnl the two small fixes that you found should be fixed. Concerning the other suggestions, some were discussed on the last RCT meeting and the rest I defer to @mcneillj

docs/section9/Rule6-ja.md
xingcx: Updated.

docs/section9/Rule6-ja.md
xingcx: Updated.

mcneillj:
> @charliepnnl the two small fixes that you found should be fixed. Concerning the other suggestions, some were discussed on the last RCT meeting and the rest I defer to @mcneillj

I added more detailed comments here:  https://github.com/pnnl/ruleset-checking-tool/pull/23

We need to focus on developing RDS to prove out the specific rules.  We will need to come back to address any general issues that may arise in the future.  If the RDS accurately describes our method for testing total building interior lighting power from the current RMR schema then we should accept this PR.

charliepnnl (approved):

mcneillj:
> In row 23, we should be checking whether building area type is known, not whether project "uses" building area method. Projects do not have to declare which method they use. Also, I believe compliance should be established for project as a whole, not for each building segment individually. So I think the logic should be as follows:
> 
> 1. Calculate BAM limit:
>    a. For each building segment: calculate the total area of spaces within that segment and multiply by allowance for this building area type from Table G3.8 to calculate the BAM allowance for that segment
>    b. Add up BAM allowances for individual segments to determine the total project allowance with BAM method.
> 2. Calculate Space-by-Space limit:
>    a. For each building segment: multiply floor area of each space by the allowance for this space type from Table G3.7. (We only need to account for building segment here because some of the space-by-space allowances depend on building area type - e.g., corridors in hospitals have different allowances than in other building types).
>    b. Add up space-by-space allowances for all spaces in thee project to determine the total project allowance with space-by-space method.
> 3. Calculate the total specified (P-RMR) lighting wattage for the project as the sum of specified lighting across all spaces in all segments.
> 4. If total specified wattage from (3) does not exceed the greater of BAM allowance (from #1) or Space-by-space allowance (from #2), project passes.

We will be addressing this interpretation of the rule through a separate conversation.

mkarpman:
@JasonGlazer and @supriyagoel , @mcneillj suggested to address my comments above as a separate conversation. Please look over my comments above and let me know if you have any thoughts.

supriyagoel:
@xingcx this was discussed by the team on 06.28.2021. Please revise the RDS based on this guidance-

If lighting_building_area_type and lighting_space_type are know known, then calculate allowance based on both, pass if less than max
If lighting_building_area_type is not specified, lighting_space_type is included: then determine allowance based on lighting space type.
If PASS- Output should say that project passed based on space by space method. Reviewer should verify if the project uses space by space method.
If FAIL- then CAUTION and say that it fails space by space method and lighting_building_area_type is not known to determine allowance based on BAT.
If lighting_building_area_type provided but lighting_space_type not included: 
If PASS - Output should say that project passed based on building area method. Reviewer should verify if the project uses building area method.
If FAIL- then CAUTION and say that it fails building area method and lighting_space_type is not known to determine allowance based on space by sspace method.

mkarpman:
> @xingcx this was discussed by the team on 06.28.2021. Please revise the RDS based on this guidance-
> 
> If lighting_building_area_type and lighting_space_type are know known, then calculate allowance based on both, pass if less than max
> If lighting_building_area_type is not specified, lighting_space_type is included: then determine allowance based on lighting space type.
> If PASS- Output should say that project passed based on space by space method. Reviewer should verify if the project uses space by space method.
MK: I don't think any verification is necessary in this case, it just passes. (I.e., it effectively meets the first conditioned up above),
> If FAIL- then CAUTION and say that it fails space by space method and lighting_building_area_type is not known to determine allowance based on BAT.
> If lighting_building_area_type provided but lighting_space_type not included:
> If PASS - Output should say that project passed based on building area method. Reviewer should verify if the project uses building area method.
MK: I think this should just be pass, similar to scenario when it only passes SBS. The requirement is that proposed design passes either SBS or BAT. So it just needs to pass one of the two. 
> If FAIL- then CAUTION and say that it fails building area method and lighting_space_type is not known to determine allowance based on space by sspace method.

PR #23, opened 2021-04-18
Update Rule6-jb.md

xingcx:


JasonGlazer:
Original discussion at the link below. Please give this link time to load. This page might take a few seconds to load and another 4-8 seconds to land at the correct comment:

https://github.com/pnnl/ruleset-checking-tool/pull/19#discussion_r611812717

supriyagoel:
This looks good to me. Adding Charlie as a reviewer

JasonGlazer (approved):

supriyagoel (approved):

charliepnnl:
If I am understanding the intent, we need to report something for each space in each building in the proposed RMR.  To uniquely identify each space, we'll need to report out for all the levels above spaces as well. So I think the rule logic needs to be something like:

- Report for each `building_proposed` in `P_RMR`:
  - Find `building_user` in `U_RMR` for which `building_user.id == building_proposed.id`
  - Report for each `building_segment_proposed` in `building_proposed`:
    - Find `building_segment_user` in `building_proposed` for which `building_segment_user.id == building_segment_proposed.id`
    - Get `lighting_building_area_type_proposed = building_segment_proposed.lighting_building_area_type`
    - Get `lighting_building_area_type_user = building_segment_user.lighting_building_area_type`
    - Report for each `thermal_block_proposed` in `building_segment_proposed`:
      - Find `thermal_block_user` in `building_segment_user` for which `thermal_block_user.id == thermal_block_proposed`
      - Report for each `zone_proposed` in `thermal_block_proposed`:
        - Find `zone_user` in `thermal_block_user` for which `zone_user.id == zone_proposed.id`
        - Report for each `space_proposed` in `therm_block_proposed`:
          - Find `space_user` in `thermal_block_user` for which `space_user.id == space_proposed.id`
          - Calculate `space_lighting_power_proposed = calculate_space_lighting_power(... whatever is depend on ... can we pull the logic of this function out so it can also be used in 6-ja and maybe others?)
          - Calculate `space_lighting_power_user = calculate_space_lighting_power(... whatever is depend on ...)
          - **Rule Assertion**: `space_lighting_power_proposed == space_lighting_power_user`

charliepnnl:
In may comment above, I assumed that we match up the pieces in the Proposed and User RMRs by their `id` field.  Is that right? Should it be `name`, or should it be `id` if it exists and `name` otherwise?

charliepnnl:
We have some of the same issues here as I brought up for 6-ja.   
Namely, 
- What do we do if a building segment doesn't have a lighting_building_area_type or an interior lighting doesn't have a power_per_area, etc?  I am hoping that these issues can be address by the schema or as separate high-level validation checks on the RMRs.

mcneillj:
> We have some of the same issues here as I brought up for 6-ja.
> Namely,
> 
> * What do we do if a building segment doesn't have a lighting_building_area_type or an interior lighting doesn't have a power_per_area, etc?  I am hoping that these issues can be address by the schema or as separate high-level validation checks on the RMRs.

This is a general issue that affects all Rule Definitions that we will need to address in the future.

mcneillj:
> In my comment above, I assumed that we match up the pieces in the Proposed and User RMRs by their `id` field. Is that right? Should it be `name`, or should it be `id` if it exists and `name` otherwise?

This is another general issue that we will address in the future, as the committee may make changes.  For now we will use ID as Xing has chosen.  We can change in the future, if required due to schema changes.  I will work on making sure all RDS use ID for the matching identifier.


> Also in the comment above, I indicated the idea of pulling out the logic of calculating the lighting power for a space into a separate function that could be documented separately and could be then used by other rules also.

You can feel free to make functions in the Rule Definitions whenever you see repeated code.  They Rule Definitions do not need to match this exactly.

mkarpman:
Names vs ID was discussed by the schema working group at the last meeting. Some people think that both are necessary, others that just one of the two will do as long as it is unique and descriptive to allow associating the object with the actual specified system (e.g. RTU1).

mkarpman:
Some ruleset - specific descriptors such as lighting_building_area_type have to be specified by user when entering data into BEM tool and then written by the BEM tool to the RMR. If it is missing from RMR, some of the RCT logic cannot be applied to the project being checked, so perhaps the associated tests should be tagged as not performed (as opposed to pass or fail) due to missing information in the RMR. But it's important to differentiate the ruleset specific descriptors that can only come from user, such as lighting space types, from descriptors that can be derived from physics-based inputs such as surface type. For the latter, I think we decided to stick with physics-based logic so descriptors such as surface type or space conditioning category should not be used by the RCT even if they are included in the RMR.

supriyagoel:
> Some ruleset - specific descriptors such as lighting_building_area_type have to be specified by user when entering data into BEM tool and then written by the BEM tool to the RMR. If it is missing from RMR, some of the RCT logic cannot be applied to the project being checked, so perhaps the associated tests should be tagged as not performed (as opposed to pass or fail) due to missing information in the RMR. But it's important to differentiate the ruleset specific descriptors that can only come from user, such as lighting space types, from descriptors that can be derived from physics-based inputs such as surface type. For the latter, I think we decided to stick with physics-based logic so descriptors such as surface type or space conditioning category should not be used by the RCT even if they are included in the RMR.

@mkarpman Yes, we have decided to go ahead with this approach. If some data elements are missing from the RMR, we will identify that in the RCT output report and proceed with other tests. Regarding physics based inputs- i agree that we shouldn't be checking for parameters at two places and it tilt/azimuth are missing, then those rules would have a similar outcome- i.e. cannot be tested.

charliepnnl (approved):

docs/section9/Rule6-jb.md
```diff
@@ -13,24 +13,36 @@
...
+  - For each thermal_block in building segment: ```thermal_block_proposed in building_segment_proposed.thermal_blocks:```
 
-  **Rule Assertion:** The total lighting power in each space for U_RMR and P_RMR are the same: ```space_lighting_power_user = space_lighting_power_proposed```
+  - For each zone in thermal block: ```zone_proposed in thermal_block_proposed.zones:```
+
+  - For each space in thermal zone: ```space_proposed in zone_proposed.spaces:```  
+
+    - Get floor_area from space: ```floor_area_proposed = space_proposed.floor_area```  
+
+    - Get interior_lighting in space: ```interior_lighting_proposed = space_proposed.interior_lightings```
```
charliepnnl, line 42: I believe it should be `space_proposed.interior_lighting`

docs/section9/Rule6-jb.md
```diff
@@ -13,24 +13,36 @@
...
+  - For each zone in thermal block: ```zone_user in thermal_block_user.zones:```
 
-  - Get space from thermal zone: ```space in thermal_zone.spaces:```
+  - For each space in thermal zone: ```space_user in zone_user.spaces:```  
 
-  - Get the total lighting power for each space: ```space_lighting_power_user = space.lighting_power``` (Note XC, depending on how rmr reports the lighting power for each space, e.g. report fixtures, or report total wattage, this needs to be adjusted.)
+    - Get floor_area from space: ```floor_area_user = space_user.floor_area```  
 
-- Calculate the total lighting power for each space under each building segment in the Proposed RMR: ```For building_segment in P_RMR.building.building_segments:```
+    - Get interior_lighting in space: ```interior_lighting_user = space_user.interior_lightings```
```
charliepnnl, line 26: I believe it should be `space_user.interior_lighting`

charliepnnl (changes requested):

docs/section9/Rule6-jb.md
xingcx: Thanks Charlie. You are right it's space_proposed.interior_lighting. This is now corrected.

docs/section9/Rule6-jb.md
xingcx: Thanks Charlie. This is now corrected.

charliepnnl: Thanks.

charliepnnl (approved):

mkarpman (approved):

PR #49, opened 2021-05-10
Update Rule5-14_TCD_v8.md

xingcx:


docs/section5/Rule5-6.md
```diff
@@ -24,20 +24,20 @@
...
     - Get space conditioning type: ```space_conditioning_type_baseline = space_baseline.conditioning_type```  
 
-      - Get baseline contruction from Table G3.4-1 to G3.4-8 based on space conditioning type, status type and function type: ```surface_performance_target = data_lookup(climate_zone,space_conditioning_type_baseline,"Slab-on-Grade Floors")```  
+      - Get baseline contruction from Table G3.4-1 to G3.4-8 based on climate zone, space conditioning type and function type: ```surface_performance_target = data_lookup(table_G3_4,climate_zone,space_conditioning_type_baseline,"Slab-on-Grade Floors")```  
 
     - For each surface in space: ```for surface_baseline in space_baseline.surfaces:```  
 
       - Get the surface construction if the surface is slab-on-grade floor: ```if ( surface_baseline.classification == "FLOOR" ) AND ( surface_baseline.adjacent_to == "GROUND" ): surface_construction_baseline = surface_baseline.construction```  
 
       - Get the performance values for the construction: ```surface_performance_value_baseline = surface_construction_baseline.f_factor```  (Note XC, assumes for slab on grade, RMR reports f-factor not u-factor)
```
JasonGlazer, line 41: Yes, the data group construction has f_factor.

The same two issues as other similar rules:

a. I think this and others when the f_factor is not provided should alternatively compute the f_factor based on the rest of the construction.

b. It needs to confirm the layer-by-layer description if it is provided that matches what is provided in appendix A6.

JasonGlazer (changes requested):

docs/section5/Rule5-6.md
mkarpman: @JasonGlazer I think the requirement should be that (1) project uses layers method and that (2) F-factor included in the RMR is correct. If we expect RCT to calculate the F-factor based on info provided in layers object as you suggest above, then we need to update schema to include additional details such as conductivity, density, specific heat and thickness of each material in the assembly.

supriyagoel (approved):

mkarpman (approved):

JasonGlazer (approved):

PR #54, opened 2021-05-10
Add Rule5-34

xingcx:


docs/section5/Rule5-12.md
```diff
@@ -0,0 +1,51 @@
...
+
+      - Calculate the total skylight area in surface: ```skylight_area_proposed = sum(skylight.area for skylight in roof_surface_proposed.fenestration_subsurfaces)```  
+
+      - Calculate the building total roof area: ```total_roof_area_proposed += roof_surface_proposed.area```  
+
+      - Calculate the building total skylight area: ```total_skylight_area_proposed += skylight_area_proposed```  
+
+      - Calculate the building skylight to roof ratio: ```skylight_roof_ratio_proposed = total_skylight_area_proposed / total_roof_area_proposed```  
+
+    - Get matching space from Baseline RMR: ```space_baseline = match_data_element(B_RMR, spaces, space_proposed.name)```  (Note XC, getting space in baseline, or the surface or subsurface name will match? do we need to check skylight one to one, or only checking total skylight area in B_RMR?)
```
JasonGlazer, line 39: I think the baseline skylights should be the same as in the proposed skylights, so yes, they should match. It doesn't say that here but it is part of Table G3.1 part 1 baseline, 2nd paragraph.

docs/section5/Rule5-12.md
```diff
@@ -0,0 +1,51 @@
...
+
+      - Check if surface is roof: ```if ( surface_baseline.classification == "CEILING" ) AND ( surface_baseline.adjacent_to == "AMBIENT" ): roof_surface_baseline = surface_baseline```  
+
+      - Calculate the total skylight area in Baseline RMR surface: ```skylight_area_baseline = sum(skylight.area for skylight in roof_surface_baseline.fenestration_subsurfaces)```  
+
+      **Rule Assertion:** Baseline skylight area is equal to that in the Proposed design or 3%, whichever is smaller.  
+
+      - Case 1: The skylight to roof ratio in the Proposed model is less than or equal to 3%: ```if skylight_roof_ratio_proposed <= 3%: skylight_area_baseline == skylight_area_proposed```  
+
+      - Case 2: The skylight to roof ratio in the Proposed model is more than 3%: ```else: skylight_area_baseline == skylight_area_proposed * ( 3% / skylight_roof_ratio_proposed )```
```
JasonGlazer, line 51: I'm not sure it is enough to check the total area is reduced, I think the requirement is that each individual skylight needs to be reduced by an identical ratio so I think it needs to go back to each skylight to check.

JasonGlazer (changes requested):

JasonGlazer:
@xingcx I think this one is still be debated by the team.

docs/section5/Rule5-34.md
```diff
@@ -0,0 +1,41 @@
...
+## Rule Logic:
+
+- Get building skylight roof areas dictionary for B_RMR: `skylight_roof_areas_dictionary_b = get_building_segment_skylight_roof_areas(B_RMR)`
+
+- Get building skylight roof areas dictionary for P_RMR: `skylight_roof_areas_dictionary_p = get_building_segment_skylight_roof_areas(P_RMR)`
+
+- For each building segment in B_RMR: `for building_segment_b in B_RMR.building.building_segments:`  
+
+  - Calculate skylight roof ratio for building segment: `skylight_roof_ratio_b = skylight_roof_areas_dictionary_b[building_segment_b.id][0] / skylight_roof_areas_dictionary_b[building_segment_b.id][1]`
+
```
mkarpman, line 30: I do not understand the nomenclature of [0] and [1], but trust that you know what you are doing here :-)

mkarpman (approved):

supriyagoel (approved):

docs/section5/Rule5-34.md
xingcx: The get_skylight_roof_areas() function will return a skylight_roof_areas_dictionary in the format of: 
building_segment_id: [total skylight area, total roof area]
[0] calls for the first value, which is the total skylight area for that building segment
[1] calls for the second, which is the total roof area for that building segment

mkarpman: Got it, thanks!

supriyagoel (approved):

mkarpman (approved):

PR #56, opened 2021-05-10
Add Rule5-38

xingcx:


JasonGlazer (approved):

JasonGlazer:
@mkarpman do you want to look at this before I hand it over to @charliepnnl?

docs/section5/Rule5-14.md
```diff
@@ -0,0 +1,45 @@
...
+
+  - For each space in thermal zone: ```for space_baseline in zone_baseline.spaces:```  
+
+    - Get surface in space: ```for surface_baseline in space_baseline.surfaces:```  
+
+      - Check if surface is roof: ```if ( surface_baseline.classification == "CEILING" ) AND ( surface_baseline.adjacent_to == "AMBIENT" ): roof_surface_baseline = surface_baseline```  
+
+        - Calculate the total roof area: ```total_roof_area_baseline += roof_surface_baseline.area```  
+
+        - Get the U-factor for skylightss: ```skylight_performance_value_baseline.append(skylight.solar_heat_gain_coefficient for skylight in roof_surface_baseline.fenestration_subsurfaces)```
```
supriyagoel, line 35: 'Get the U-factor for skylightss' should say 'Get the SHGCr for skylights'

docs/section5/Rule5-14.md
```diff
@@ -0,0 +1,45 @@
...
+        - Calculate the total roof area: ```total_roof_area_baseline += roof_surface_baseline.area```  
+
+        - Get the U-factor for skylightss: ```skylight_performance_value_baseline.append(skylight.solar_heat_gain_coefficient for skylight in roof_surface_baseline.fenestration_subsurfaces)```  
+
+        - Calculate the total skylight area: ```total_skylight_area_baseline += sum(skylight.area for skylight in roof_surface_baseline.fenestration_subsurfaces)```  
+
+        - Calculate the skylight to roof ratio: ```skylight_roof_ratio_baseline = total_skylight_area_baseline / total_roof_area_baseline```  
+
+    - Get space conditioning type: ```space_conditioning_type_baseline = space_baseline.conditioning_type```  
+
```
supriyagoel, line 42: need to determine if the space is residential or non residential. Residential space types include guest rooms, living quarters, private living space, and sleeping quarters. Other building and space types are considered nonresidential.
The baseline U factor is also dependent on the space type

supriyagoel (changes requested):

JasonGlazer:
@xingcx I think you have enough guidance to revise this one.

JasonGlazer (approved):

JasonGlazer:
@supriyagoel and @mkarpman I think this addresses all your comments and it ready for @charliepnnl. Please let me know if you want to review it further.

xingcx:
Updated. Check for CEILING is now removed.

JasonGlazer:
@mkarpman repeating the physics based criteria for heating and cooling capacity to determine the space conditioning category for every rule that uses the space conditioning category seems unnecessarily repetitious. Can we check once and assume that it is correct after that?

mkarpman:
That would require that BEM tools have compliance shell and apply 90.1
rules when writing the RMRs. I think the plan was to not require compliance
shells. What’s the big deal with repeating it?

JasonGlazer:
Let me restate. All I'm suggesting is that if the space conditioning category is present it get compared with the heating and cooling capacities to make sure it is correct. If it is not present then we use the heating and cooling capacities to determine the space conditioning category. We do this once in the RCT. After that we use the space conditioning category in all the rules. 

We haven't figured out what to call this process but "precomputed rules" or something like that.

mkarpman:
I think it makes total sense to have what you call "precomputed rules", but that would require RCT to remember the result - e.g., once conditioning category for a space is determined, it gets associated with that SpaceID and may be used for all other rules that need it. (That would come very handy for many things - surface types, HVAC system types, etc.) However, based on my discussions with Mike T months ago, James (or someone else involved with RCT implementation) didn't want to rely on anything being "remembered". So I assumed that we are stuck with repetition. Can't this just involve the same piece of code called multiple times? 

Once this is figured out, why would we want to use different verification logic for projects that have space conditioning categories written to RMR vs those that don't? Applying the same logic to all seems to allow simpler implementation.

JasonGlazer:
Mostly I just want to document the algorithm once. I don't care if it is "precomputed" or a "function call" I just want the logic documented in one place that gets used multiple times. If it ends up in five different rules, that is five times we need to make sure it is correct instead of just once. 

If we don't want to "remember" (which I don't remember that decision), then lets just make a "function call" that is called something like:  Determine_Space_Conditioning_Category()

supriyagoel:
> Mostly I just want to document the algorithm once. I don't care if it is "precomputed" or a "function call" I just want the logic documented in one place that gets used multiple times. If it ends up in five different rules, that is five times we need to make sure it is correct instead of just once.
> 
> If we don't want to "remember" (which I don't remember that decision), then lets just make a "function call" that is called something like: Determine_Space_Conditioning_Category()

Having a function call makes sense. We might need to have similar function calls for other things (determine is a surface is a roof/floor/wall). i completely agree with this approach.

mkarpman:
Function call makes total sense - we don't want to repeat things over and over. 
Aside from that, I think it is important to be consistent about not using secondary ruleset-specific descriptors such as space conditioning category, wall type or HVAC system type irrespective of whether they are part of the RMR. These descriptors do not factor into simulations performed by the engines, so may be wrong even when simulation inputs and compliance outcomes are correct. For example, compliance shell may not know that space conditioning categories are based on space types and not building types, so may tag corridors in MF buildings as residential spaces. Even though this is an error, with physics-based testing it would not be flagged by RCT if for the given climate zones U-values for residential and non-residential occupancies are the same.

mkarpman:
> I do not know enough about software development to weigh in on why we cannot "remember" precomputed things- that is something @mcneillj can work on with the RCT team.

When I was working on the envelope TCDs, I suggested to Mike T that we treat definitions such as space conditioning categories or wall types as separate rules, remember the outcome for each space/surface, and then apply it to other rules. But he told me that nothing must be remembered (or at least that is how I understood him) so was assuming this was the case ever since. I always do as I am told, as you well know :-)

supriyagoel (approved):

mkarpman (approved):

docs/section5/Rule5-38.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Rule Logic:  
+
+- Get building climate zone: `climate_zone = B_RMR.weather.climate_zone`  
+
+- Get surface conditioning category dictionary for B_RMR: `scc_dictionary_b = get_surface_conditioning_category(B_RMR)`  
+
+- Get building skylight roof ratios dictionary: `building_scc_skylight_roof_ratios_dictionary = get_building_scc_skylight_roof_ratios(B_RMR)`
+
+- For each building in the Baseline model: `for building_b in B_RMR.building:`
```
supriyagoel, line 31: small fix required to remove the check for each building.

supriyagoel: a high level applicability check to verify that RMR has a single building. If a RMR has multiple buildings then it should be flagged early on by the RCT and user needs to submit a modified RMR

charliepnnl:
I noticed your discussion about functions and remembering results and thought I would weigh in.
- Whenever you find that a piece of logic is used multiple times, it is best to abstract that logic into a separate function. That will simplify RDS development and the resulting software development.
- As for "remembering"  results, I think it is best if you assume that doesn't happen and instead assume the function is called whenever that information is needed. The software development team can then decide whether "remembering" the results is needed for the sake of efficiency. It becomes an implementation detail that the RDS team need not worry over.

docs/section5/Rule5-38.md
```diff
@@ -0,0 +1,89 @@
...
+
+- Check if B_RMR has exterior mixed type skylight: `if rmr_scc_skylight_roof_ratios_dictionary["EXTERIOR MIXED"] > 0:`
+
+  - Check if residential and non-residential type skylight SHGC requirements for different skylight-roof-ratio are the same, get skylight SHGC requirements: `if ( data_lookup(table_G3_4, climate_zone, "RESIDENTIAL, "SKYLIGHT", "0%-2.0%", "ASSEMBLY MAX. SHGC") == data_lookup(table_G3_4, climate_zone, "RESIDENTIAL, "SKYLIGHT", "2.1%+", "ASSEMBLY MAX. SHGC") ) AND ( data_lookup(table_G3_4, climate_zone, "NON-RESIDENTIAL, "SKYLIGHT", "0%-2.0%", "ASSEMBLY MAX. SHGC") == data_lookup(table_G3_4, climate_zone, "NON-RESIDENTIAL, "SKYLIGHT", "2.1%+", "ASSEMBLY MAX. SHGC") ) AND ( data_lookup(table_G3_4, climate_zone, "RESIDENTIAL, "SKYLIGHT", "0%-2.0%", "ASSEMBLY MAX. SHGC") == data_lookup(table_G3_4, climate_zone, "NON-RESIDENTIAL, "SKYLIGHT", "0%-2.0%", "ASSEMBLY MAX. SHGC") ): target_shgc_mixed = data_lookup(table_G3_4, climate_zone, "RESIDENTIAL, "SKYLIGHT", "0%-2.0%", "ASSEMBLY MAX. SHGC")`
+
+  - Else, request manual review: `else: manual_review_flag = TRUE`
+
+- Else, B_RMR does not have exterior mixed type roof surface: `else:`
```
mkarpman, line 37: Suggest removing line 37

docs/section5/Rule5-38.md
```diff
@@ -0,0 +1,89 @@
...
+
+          - Case 4: Else if roof is exterior non-residential type and skylight SHGC matches Table G3.4 requirement: `else if ( scc_dictionary_b[surface_b] == "EXTERIOR NON-RESIDENTIAL" ) AND ( subsurface_b.solar_heat_gain_coefficient == target_shgc_nonres ): PASS`
+
+          - Case 5: Else if roof is semi-exterior type and skylight SHGC matches Table G3.4 requirement: `else if ( scc_dictionary_b[surface_b] == "SEMI-EXTERIOR" ) AND ( subsurface_b.solar_heat_gain_coefficient == target_shgc_semiheated ): PASS`
+
+          - Case 6: Else: `else: FAIL and NUMBER_OF_FAIL_COMPONENTS++`
+
+**Rule Assertion - RMR:**
+
+- Case 1: If any subsurface in B-RMR is ruled as "UNDETERMINED": `UNDETERMINED and raise_message "MANUAL REVIEW IS REQUESTED TO VERIFY SKYLIGHT MEETS SHGC REQUIREMENT AS PER TABLE G3.4."`
```
mkarpman, line 83: I don't think we should require manual review of all surfaces because there is one surface that requires a manual review

PR #59, opened 2021-05-17
Rct/ch/GitHub actions

charliepnnl:
The Github actions are applied to all pull requests and are triggered when the PR is created an whenever a push is made to the branch in the PR.
- Python syntax and formatting errors are checked with Black.
- All tests are run with Pytest.

The results appear at the bottom of the PR.
The PR owner should receive an email if either of the actions fails.

I have tested this and it seems to work as expected.
Note: 
- This reports results but does not prevent creating the PR or pushing to it.
- I believe that the actions will only be triggered if the .github/workflows/python-app.yml file provided here is part of the PR code.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -2,7 +2,7 @@
 # and run tests with pytest
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-# Apply this workflow when pushing or creating a pull request to any branch
+# Apply this workflow when creating a pull request or pushing a commit to a PR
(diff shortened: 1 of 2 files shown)


mcneillj:
@charliepnnl I expect that we can just leave as "on: pull_request" and that will cover all branches, including develop and master, right?

charliepnnl:
Yes, that's right.  "on: pull_request" specifies that this workflow should run for **all pull requests, on any branch**.  If we decide to use another working branch in the future, say for the RDS, we are already covered.

mcneillj (approved):

PR #61, opened 2021-05-24
Update Rule5-18_TCDv8.md

xingcx:
Original code by James.


docs/section5/Rule5-7-1.md
```diff
@@ -2,7 +2,7 @@
 **Rule ID:** 5-7-1  
 **Rule Description:** Baseline WWR should be equal to User WWR or 40%, whichever is smaller.
```
supriyagoel, line 3: @xingcx this rule description is incorrect, apologies for that. This used to be the requirement in App G 90.1-2010 but the new App G requires vertical fenestration areas for new buildings and additions shall equal that in Table G3.1.1-1
based on the area of gross above-grade walls that separate conditioned spaces and semiheated spaces from the exterior.

docs/section5/Rule5-7-1.md
```diff
@@ -33,4 +33,4 @@
                 - Repeat WWR calculation for User RMR space: `_WWR_user = _calculate_wwr(_user_space)`
                 - **Rule Assertion:** `_WWR_baseline == min(_WWR_user, _allowable_baseline_wwr)`
```
supriyagoel, line 34: Line 26 where we're determining the space conditioning category, please use the methodology discussed for 5-2. It should be based on the heating and cooling capacities.

supriyagoel (changes requested):

docs/section5/Rule5-7-1.md
supriyagoel: the 40% limit is incorrect for allowable baseline WWR. It is based on Building area type, and is specified in Table G3.1.1-1. Please see the TCD spreadsheet for additional guidance

JasonGlazer: @supriyagoel I think the line that says:

if _building_area_type in table_G3_1_1_1: _allowable_baseline_wwr = data_lookup(table_G3_1_1_1, _building_area_type) else: _allowable_baseline_wwr = 0.40

is already looking up from that table.

mcneillj: @supriyagoel description was updated to reflect latest spreadsheet naming.

mcneillj: @supriyagoel @JasonGlazer @mkarpman 

I'm a bit confused what we are doing with zones and spaces in the functions related to envelope. The RMR schema still has space_conditioning_category and status_type on Space, but it appears we are using the zone_conditioning_category calculated from Zone.  Since status_type is still on Space how can we confidently determine the Status Type at the zone level?  Does status_type need to be moved to Zone in the schema?

image

mkarpman: @mcneillj we are not using conditioning_type for any RDSs - it does not apply to 90.1. Instead, we determine zone conditioning category via a function by applying 90.1 definition to physical elements included in the RMR. 

I don't believe we are using status_type for any RDSs either. I am not sure what it'll be used for. There is also Building.is_all_new which may or may not be useful for anything.

mcneillj: @mkarpman That makes sense.  Just clarifying.  I think I'm understanding 5.baseline.c as it states "for new buildings and additions".  Currently, the schema guidance is to model additions as two buildings (one is_all_new==True and one is_all_new==False).  Wouldn't we would need to calculate WWR based on the existing and the addition?  If so, I'm not sure how we are intending to test whether the building has an addition.  Maybe @supriyagoel @JasonGlazer have an idea of this. 

image

mkarpman: I see - I missed that note. I wonder it it makes sense to have this on building level, building segment level or space level. Now we have it in two places - on building level (is_all_new) and space level (status_type). If 90.1 RCT looks at one and not the other, we need to communicate this to the vendors. But ideally I think it should only be in one place. Maybe bring it up with schema WG.

docs/section5/Rule5-7-1.md
xingcx: Updated.

xingcx:
Do we need to check if each window in B_RMR are proportion to P_RMR or just check the overall WWR for the building_segment?

supriyagoel (approved):

supriyagoel:
> Do we need to check if each window in B_RMR are proportion to P_RMR or just check the overall WWR for the building_segment?

@xingcx we need to do both. The updated list of envelope TCDs (V2) has the proportionality test under a separate RDS (5-17). 5-16 is only dealing with the WWR at the building segment level.

docs/section5/Rule5-16.md
```diff
@@ -0,0 +1,41 @@
...
+**Appendix G Section Reference:**  
+
+- Table G3.1-5. Building Envelope, Baseline Building Performance, c. Vertical Fenestration Areas  
+- Table G3.1.1-1  
+
+**Applicability:** All required data elements exist for B_RMR
+**Applicability Checks:** 
+
+1. Building has spaces that are NEW, ADDITION or ALTERATION.  
+
```
mkarpman, line 17: This rule does not apply to alterations - alterations must be modeled with the same window area as existed prior to retrofit.

docs/section5/Rule5-16.md
```diff
@@ -0,0 +1,41 @@
...
+**Evaluation Context:**  Each Data Element  
+**Data Lookup:** Table G3.1.1-1  
+**Function Call:**  
+
+  1. get_overall_building_segment_wwr()
+
+## Rule Logic:  
+
+- **Applicability Check 1:** `length( [ if space.status_type for space in U_RMR...spaces is in [NEW, ADDITION, ALTERATION] ] ) > 0:`  
+
```
mkarpman, line 28: Suggest dropping this applicability check.

docs/section5/Rule5-16.md
```diff
@@ -0,0 +1,41 @@
...
+- For each building segment in the Baseline model: `for building_segment_b in B_RMR.building.building_segments:`
+
+  - Check if building segment area type is included in Table G3.1.1-1: `if data_lookup(table_G3_1_1_1, building_segment_b.area_type_vertical_fenestration):`
+
+    **Rule Assertion:**
+
+      - Case 1: If building segment window-wall-ratio matches Table G3.1.1-1 allowance: `if building_wwr_dictionary_b[building_segment_b.id] == data_lookup(table_G3_1_1_1, building_segment_b.area_type_vertical_fenestration): PASS`
+
+      - Case 2: Else: `else: CAUTION and raise_warning "BASELINE BUILDING SEGMENT AREA TYPE IS INCLUDED IN TABLE G3.1.1-1. BUT BUILDING SEGMENT FENESTRATION AREA DOES NOT MATCH TABLE G3.1.1-1. CHECK IF BUILDING SEGMENT HAS EXISTING ENVELOPE THAT CAN BE EXCLUDED FROM THE WINDOW-WALL-RATIO CALCULATION."`
+
```
mkarpman, line 40: I suggest changing them to FAIL and keep the existing message.

mkarpman (changes requested):

docs/section5/Rule5-16.md
supriyagoel: this approach works because in each 'FAIL' case, a jurisdiction will require a modeler to explain the FAIL.

docs/section5/Rule5-18.md
```diff
@@ -0,0 +1,37 @@
...
+- For each building segment in the Baseline model: `for building_segment_b in B_RMR.building.building_segments:`
+
+  - Check if building segment area type is included in Table G3.1.1-1: `if data_lookup(table_G3_1_1_1, building_segment_b.area_type_vertical_fenestration):`
+
+    **Rule Assertion:**
+
+      - Case 1: If building segment window-wall-ratio matches Table G3.1.1-1 allowance: `if building_wwr_dictionary_b[building_segment_b.id] == data_lookup(table_G3_1_1_1, building_segment_b.area_type_vertical_fenestration): PASS`
+
+      - Case 2: Else: `else: FAIL and raise_warning "BASELINE BUILDING SEGMENT AREA TYPE IS INCLUDED IN TABLE G3.1.1-1. BUT BUILDING SEGMENT FENESTRATION AREA DOES NOT MATCH TABLE G3.1.1-1. CHECK IF BUILDING SEGMENT HAS EXISTING OR ALTERED ENVELOPE THAT CAN BE EXCLUDED FROM THE WINDOW-WALL-RATIO CALCULATION."`
+
```
mkarpman, line 36: I think we need to additionally check for whether the existin gbuilding rules apply, as follows:
a. Case 1 - additionally check if building:is_all_new=TRUE. If so, the outcome should be PASS with no message
b. Case 2 - additionally check if building:is_all_new=TRUE. If so, the outcome should be FAIL with no message
c. Case 3 - if building:is_all_new=FALSE AND WWR matches the look-up value, the outcome should be CAUTION with the message: "Baseline WWR matches values prescribed in Table G3.1.1-1. However, the fenestration area prescribed in Table G3.1.1-1 does not apply to the existing envelope per Table G3.1 Baseline Column #5 (c). For existing evelope, the baseline fenestration area must equal the existing fenestration area prior to the proposed work. A manual check is required to verify compliance. "
d. Case 4 - if building:is_all_new=FALSE AND WWR does not match the look-up value, the outcome should be CAUTION with the message: "Baseline WWR does not match values prescribed in Table G3.1.1-1. However, the fenestration area prescribed in Table G3.1.1-1 does not apply to the existing envelope per Table G3.1 Baseline Column #5 (c). For existing evelope, the baseline fenestration area must equal the existing fenestration area prior to the proposed work. A manual check is required to verify compliance. "

xingcx: Updated.

docs/section5/Rule5-18.md
```diff
@@ -0,0 +1,37 @@
...
+**Applicability:** All required data elements exist for B_RMR  
+**Applicability Checks:** None  
+
+**Manual Checks:** None  
+**Evaluation Context:**  Each Data Element  
+**Data Lookup:** Table G3.1.1-1  
+**Function Call:**  
+
+  1. get_overall_building_segment_wwr()
+
```
mkarpman, line 22: I didn's see function under that name. Did you mean "get_window_wall_areas" function?

mkarpman (approved):

PR #62, opened 2021-05-24
Update Rule5-17.md

xingcx:
Original code by James.

### DIFF
--- a/docs/section5/Rule5-7-2.md
+++ b/docs/section5/Rule5-7-2.md
@@ -2,7 +2,7 @@
 **Rule ID:** 5-7-2  
 **Rule Description:** Baseline RMR WWR is distributed in the same proportion as User RMR.  
 **Rule Assertion:** Baseline RMR = expected value  
-**Appendix G Section:** Envelope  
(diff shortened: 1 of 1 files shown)


docs/section5/Rule5-7-2.md
```diff
@@ -32,4 +32,4 @@
             - Repeat previous steps to calculate the User WWR for each orientation and building area type: `_WWR_user = _calculate_wwr(U_RMR)`  
             - **Rule Assertion** For each building area type and orientation the WWR is equal between the Baseline and User RMRs: `_WWR_baseline_ == _WWR_user`
```
supriyagoel, line 33: for space conditioning category, we should use the heating and cooling capacity, as clarified in 5-2
This check is to verify that the windows are distributes in the correct proportions- please see example calculations in the envelope TCD spreadsheet, tab 'Calcs_Rule5.7'. Step 2 calculates the window proportions in the U-RMR and Step 4 does the same for the B-RMR.

xingcx:
This PR can be closed. The RDS numbering and content is outdated. New PR for this rule (now Rule 5-21 as per TCDv10) is here 
https://github.com/pnnl/ruleset-checking-tool/pull/130.

PR #64, opened 2021-05-24
Update Rule5-19.md

xingcx:
Original code by James.


docs/section5/Rule5-7-4.md
```diff
@@ -27,4 +27,4 @@
             - Repeat calculating the total fenestration area for each orientation and building area type for the Proposed RMR: `_proposed_fenestration_area = = sum(_fenestration.area for _fenestration in _exterior_surface.fenestration_subsurfaces)`  
             - **Rule Assertion** The fenestration area for each building area type and orientation is equal between the Proposed and User RMRs: `_proposed_fenestration_area == _user_fenestration_area`
```
supriyagoel, line 28: @JasonGlazer and @mkarpman  i see that this check is verifying the WWR at the building area type level. Is that sufficient or do we need to verify that the WWR for each surface in the P-RMR = the WWR for the corresponding surface in the U-RMR?

mkarpman: @supriyagoel Why can't we just check that each surface in P_RMR has the same windows area as the corresponding surface in U_RMR? Is this because we can't reliably identify P-RMR and U-RMR pairs?

supriyagoel: @mkarpman we could do that using surface IDs. those should be the same between the P an U RMR. 
@xing - please revise this test to check that the WWR of the P RMR is the as the WWR of the U RMR for each surface. We can use the surface IDs for these checks

mkarpman: > check that the WWR of the P RMR is the as the WWR of the U RMR for each surface

@supriyagoel why do you think we should be checking WWR and not just fenestration area on each surface? Either would work I think, but I am curious why you have that preference.

supriyagoel: @mkarpman I don't have a preference. fenestration area works just fine.

mkarpman (approved):

supriyagoel (approved):

mkarpman (approved):

PR #65, opened 2021-05-26
Add RDS for rule 15-7

mcneillj:


JasonGlazer:
@mcneillj I don't understand the rule description. Appendix G says that the baseline should exceed the proposed not that they should be the same.

PR #67, opened 2021-05-26
Add RDS for rule 15-9

mcneillj:
Note that capacity ratio is capacity (kVA) and peak_load (W).  I would assume we are going to consider the ratio in these units to be acceptable, as we don't have a power factor for a building.

### DIFF
--- a/docs/section15/Rule15-9.md
+++ b/docs/section15/Rule15-9.md
@@ -17,7 +17,7 @@
     - Get maximum regulated capacity: `if _user_transformer.phase == SINGLE_PHASE: _max_capacity_limit = 333.0 else if _user_transformer.phase == THREE_PHASE: _max_capacity_limit = 1000.0`  
     - **Applicability Check 1:** `_user_transformer.type == DRY_TYPE`  
     - **Applicability Check 2:** `_user_transformer.capacity >= 15.0 and _user_transformer.capacity <= _max_capacity_limit:`  
-    - Get baseline transformer matching name of user transformer: `_baseline_transformer = match_data_element(B_RMR, transformers, _user_transformer.name)`  
+    - Get baseline transformer matching ID of user transformer: `_baseline_transformer = match_data_element(B_RMR, transformers, _user_transformer.id)`  
     - Calculate user transformer capacity ratio: `_user_transformer_capacity_ratio = _user_transformer.capacity / _user_transformer.peak_load`  
     - Calculate baseline transformer capacity ratio: `_baseline_transformer_capacity_ratio = _baseline_transformer.capacity / _baseline_transformer.peak_load`  
     - **Rule Assertion:** `_user_transformer_capacity_ratio == _baseline_transformer_capacity_ratio`  


docs/section15/Rule15-9.md
```diff
@@ -0,0 +1,28 @@
...
+
+**Manual Checks:** None  
+
+## Rule Logic:
+- For each transformer `_user_transformer` in `U_RMR.transformers`:  
+    - Get maximum regulated capacity: `if _user_transformer.phase == SINGLE_PHASE: _max_capacity_limit = 333.0 else if _user_transformer.phase == THREE_PHASE: _max_capacity_limit = 1000.0`  
+    - **Applicability Check 1:** `_user_transformer.type == DRY_TYPE`  
+    - **Applicability Check 2:** `_user_transformer.capacity >= 15.0 and _user_transformer.capacity <= _max_capacity_limit:`  
+    - Get baseline transformer matching name of user transformer: `_baseline_transformer = match_data_element(B_RMR, transformers, _user_transformer.name)`  
+    - Calculate user transformer capacity ratio: `_user_transformer_capacity_ratio = _user_transformer.capacity / _user_transformer.peak_load`
```
supriyagoel, line 21: @mcneillj the only change required here is that the matching should be based on ID and not name.

docs/section15/Rule15-9.md
```diff
@@ -0,0 +1,28 @@
+# Transformers - Rule 15-9
+**Rule ID:** Rule 15-9  
+**Rule Description:** Transformer capacity ratio reported in Baseline RMR equals User RMR  
+**Rule Assertion:** User RMR = Baseline RMR
```
mkarpman, line 4: For envelope and lighting, we express such rules as B-RMR=P-RMR, and also have rule that P-RMR=U-RMR. Even though it's mathematically the same, I would stick with the existing framework throughout the RDSs for consistency, in stead of doing U-RMR = B-RMR.

docs/section15/Rule15-9.md
```diff
@@ -0,0 +1,28 @@
...
+**Rule Description:** Transformer capacity ratio reported in Baseline RMR equals User RMR  
+**Rule Assertion:** User RMR = Baseline RMR  
+**Appendix G Section:** Transformers  
+**Appendix G Section Reference:**  Table G3.1 15. Distribution Transformers  
+**Data Lookup:** Table 8.4.4  
+**Evaluation Context:**  Each Data Element  
+**Applicability Checks:**
+1. Transformer is DRY_TYPE
+2. U_RMR transformer capacity is within range of Table 8.4.4  
+
```
mkarpman, line 12: If we change it to B-RMR = P-RMR as suggested above, then we can skip applicability since it would already be checked as part of P-RMR=U-RMR. If we keep it as is (B-RMR=U-RMR, then also need to add a check that U-RMR efficiency exceeds minimums in Table 8.4.4.

docs/section15/Rule15-9.md
```diff
@@ -0,0 +1,28 @@
...
+**Manual Checks:** None  
+
+## Rule Logic:
+- For each transformer `_user_transformer` in `U_RMR.transformers`:  
+    - Get maximum regulated capacity: `if _user_transformer.phase == SINGLE_PHASE: _max_capacity_limit = 333.0 else if _user_transformer.phase == THREE_PHASE: _max_capacity_limit = 1000.0`  
+    - **Applicability Check 1:** `_user_transformer.type == DRY_TYPE`  
+    - **Applicability Check 2:** `_user_transformer.capacity >= 15.0 and _user_transformer.capacity <= _max_capacity_limit:`  
+    - Get baseline transformer matching name of user transformer: `_baseline_transformer = match_data_element(B_RMR, transformers, _user_transformer.name)`
```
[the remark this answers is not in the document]
mkarpman: Agreed

docs/section15/Rule15-9.md
mkarpman: Suggest to change to refer to P_RMR instead of U-RMR

docs/section15/Rule15-9.md
```diff
@@ -0,0 +1,28 @@
...
+## Rule Logic:
+- For each transformer `_user_transformer` in `U_RMR.transformers`:  
+    - Get maximum regulated capacity: `if _user_transformer.phase == SINGLE_PHASE: _max_capacity_limit = 333.0 else if _user_transformer.phase == THREE_PHASE: _max_capacity_limit = 1000.0`  
+    - **Applicability Check 1:** `_user_transformer.type == DRY_TYPE`  
+    - **Applicability Check 2:** `_user_transformer.capacity >= 15.0 and _user_transformer.capacity <= _max_capacity_limit:`  
+    - Get baseline transformer matching name of user transformer: `_baseline_transformer = match_data_element(B_RMR, transformers, _user_transformer.name)`  
+    - Calculate user transformer capacity ratio: `_user_transformer_capacity_ratio = _user_transformer.capacity / _user_transformer.peak_load`  
+    - Calculate baseline transformer capacity ratio: `_baseline_transformer_capacity_ratio = _baseline_transformer.capacity / _baseline_transformer.peak_load`  
+    - **Rule Assertion:** `_user_transformer_capacity_ratio == _baseline_transformer_capacity_ratio`  
+
```
mkarpman, line 24: Same comment as above about change from U-RMR to P-RMR. Also, I think we need to stick with the PASS/FAIL outcome framework that we use for lighting and envelope.

weilixu:
No longer relevant - close it now.

PR #68, opened 2021-05-26
Add RDS for rule 15-10

mcneillj:
Note that capacity ratio is capacity (kVA) and peak_load (W).  I would assume we are going to consider the ratio in these units to be acceptable, as we don't have a power factor for a building.


docs/section15/Rule15-10.md
```diff
@@ -0,0 +1,28 @@
...
+**Rule Description:** Transformer capacity ratio reported in Proposed RMR equals User RMR  
+**Rule Assertion:** User RMR = Proposed RMR  
+**Appendix G Section:** Transformers  
+**Appendix G Section Reference:**  Table G3.1 15. Distribution Transformers  
+**Data Lookup:** Table 8.4.4  
+**Evaluation Context:**  Each Data Element   
+**Applicability Checks:**
+1. Transformer is DRY_TYPE
+2. U_RMR transformer capacity is within range of Table 8.4.4  
+
```
mkarpman, line 12: Another applicability check is that the proposed design transformers exceed the efficiency requirements of Table 8.4.4

docs/section15/Rule15-10.md
```diff
@@ -0,0 +1,28 @@
...
+## Rule Logic:
+- For each transformer `_user_transformer` in `U_RMR.transformers`:
+    - Get maximum regulated capacity: `if _user_transformer.phase == SINGLE_PHASE: _max_capacity_limit = 333.0 else if _user_transformer.phase == THREE_PHASE: _max_capacity_limit = 1000.0`  
+    - **Applicability Check 1:** `_user_transformer.type == DRY_TYPE`  
+    - **Applicability Check 2:** `_user_transformer.capacity >= 15.0 and _user_transformer.capacity <= _max_capacity_limit:`
+    - Get proposed transformer matching name of user transformer: `_proposed_transformer = match_data_element(P_RMR, transformers, _user_transformer.name)`
+    - Calculate user transformer capacity ratio: `_user_transformer_capacity_ratio = _user_transformer.capacity / _user_transformer.peak_load`
+    - Calculate proposed transformer capacity ratio: `_proposed_transformer_capacity_ratio = _proposed_transformer.capacity / _proposed_transformer.peak_load`
+    - **Rule Assertion:** `_user_transformer_capacity_ratio == _proposed_transformer_capacity_ratio`
+
```
mkarpman, line 24: This RDS is not written using the typical framework with PASS, FAIL or CAUTION outcome.

docs/section15/Rule15-10.md
```diff
@@ -0,0 +1,28 @@
+# Transformers - Rule 15-10
+**Rule ID:** Rule 15-10  
+**Rule Description:** Transformer capacity ratio reported in Proposed RMR equals User RMR
```
mkarpman, line 3: I wonder if this is the right rule. Shouldn't this instead be that the regulated aspects of the transformers (capacity, efficiency, phase) are the same in U-RMR and P-RMR? I am not sure we need to require peak_load equivalency, because in theory there are any differences in loads between U-RMR and P-RMR.

weilixu:
No longer relevant - close it now.

PR #69, opened 2021-06-01
Update Rule5-1.md

xingcx:
Split Rule 5-1 into 2 RDS to keep one rule assertion per RDS.

### DIFF
--- a/docs/section5/Rule5-1.md
+++ b/docs/section5/Rule5-1.md
@@ -1,8 +1,8 @@
 
-# Envelope - Rule 5-1
+# Envelope - Rule 5-1-1  
 
-**Rule ID:** 5-1  
-**Rule Description:** Baseline Performance is the average of 4 rotations if vertical fenestration area per each orientation differ by more than 5%  
+**Rule ID:** 5-1-1  
+**Rule Description:** Baseline Performance is the average of 4 rotations if vertical fenestration area per each orientation differ by more than 5%.  
 **Rule Assertion:** Baseline RMR Building:is_rotated = expected value  
 **Appendix G Section:** Section G3.1-5(b) Building Envelope Modeling Requirements for the Baseline building  
 **Appendix G Section Reference:** Table G3.1 Section 5a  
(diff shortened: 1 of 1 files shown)


docs/section5/Rule5-1.md
```diff
@@ -20,44 +20,42 @@
 
 - **Applicability Check 1:** ```if (rule-5-7  == PASS):```
```
supriyagoel, line 22: Applicability check 1 (5-7) will have to be renumbered when all envelope TCDs are renumbered

docs/section5/Rule5-1.md
```diff
@@ -20,44 +20,42 @@
...
+  - For each thermal_block in building segment: ```thermal_block_b in building_segment_b.thermal_blocks:```  
   
-  - For each zone in thermal block: ```zone_baseline in thermal_block_baseline.zones:```  
+  - For each zone in thermal block: ```zone_b in thermal_block_b.zones:```  
 
-  - For each space in zone: ```space_baseline in zone_baseline.spaces:```  
+  - For each space in zone: ```space_b in zone_b.spaces:```  
 
-    - For each surface in space: ```for surface_baseline in space_baseline.surfaces:```  
+    - For each surface in space: ```for surface_b in space_b.surfaces:```
```
supriyagoel, line 34: @xingcx and @JasonGlazer i believe the schema is going to be changed so that surfaces are associated with zones instead of spaces, correct? If yes, the RDS will have to be revised to check for surfaces at the zone level.

docs/section5/Rule5-1.md
```diff
@@ -20,44 +20,42 @@
...
+  - For each zone in thermal block: ```zone_b in thermal_block_b.zones:```  
 
-  - For each space in zone: ```space_baseline in zone_baseline.spaces:```  
+  - For each space in zone: ```space_b in zone_b.spaces:```  
 
-    - For each surface in space: ```for surface_baseline in space_baseline.surfaces:```  
+    - For each surface in space: ```for surface_b in space_b.surfaces:```  
 
-      - Check that surface is exterior and vertical: ```if ( surface_baseline.adjacent_to == "AMBIENT" ) AND ( 60 <= _surface.tilt <= 90 ): exterior_vertical_surface = surface_baseline```  
+      - Check that surface is exterior and vertical: ```if ( surface_b.adjacent_to == "EXTERIOR" ) AND ( 60 <= _surface.tilt <= 90 ): exterior_vertical_surface_b = surface_b```
```
supriyagoel, line 36: instead of checking the surface type, reference the function for defining surface type.

docs/section5/Rule5-1.md
```diff
@@ -20,44 +20,42 @@
...
       - Summarize the total fenestration area and the total surface area for each orientation for the building:  
 
-      ```if ( surface_azimuth_baseline > 315 ) OR ( surface_azimuth_baseline <= 45 ): total_fenestration_area_north += fenestration_area_baseline```  
+      ```if ( surface_azimuth_b > 315 ) OR ( surface_azimuth_b <= 45 ): total_fenestration_area_north += fenestration_area_b```  
 
-      ```elsif ( surface_azimuth_baseline > 45 ) AND ( surface_azimuth <= 135 ): total_fenestration_area_east += fenestration_area_baseline```  
+      ```elsif ( surface_azimuth_b > 45 ) AND ( surface_azimuth_b <= 135 ): total_fenestration_area_east += fenestration_area_b```  
 
-      ```elsif ( surface_azimuth > 135 ) AND ( surface_azimuth <= 225 ): total_fenestration_area_south += fenestration_area_baseline```  
+      ```elsif ( surface_azimuth_b > 135 ) AND ( surface_azimuth_b <= 225 ): total_fenestration_area_south += fenestration_area_b```
```
supriyagoel, line 48: @mkarpman and @JasonGlazer 90.1 is not very clear about what is north/south/east/west orienattion. Section 5.5.4.5 has the following-

- **East oriented**- oriented within 45 degrees of true east to the south and within 22.5 degrees of true east to the north in the northern hemisphere; oriented within 45 degrees of true east to the north and within 22.5 degrees of true east to the south in the southern hemisphere
- West Oriented- oriented within 45 degrees of true west to the south and within 22.5 degrees of true west to the north in the northern hemisphere; oriented within 45 degrees of true west to the north and within 22.5 degrees of true west to the south in the southern hemisphere
- **north-oriented** - facing within 45 degrees of true north in the northern hemisphere (however,
facing within 45 degrees of true south in the southern hemisphere).
- south oriented- there is no definition of this.

if N is within 45 deg of true north and East is within 22.5 degrees of ttrue east to the North, then what happens to the 22.5 degrees remaining to the North? What orientation is that. Same question for W. 

Let me know what I'm missing.

docs/section5/Rule5-1.md
```diff
@@ -20,44 +20,42 @@
...
+- Case 1, the total vertical fenestration area per exposure differ by less than 5% and the baseline building is not rotated: ```( is_rotation_required == FALSE ) AND ( is_rotated_b == FALSE ): PASS```  
 
-- Case 2, the total vertical fenestration area per exposure differ by less than 5% and the baseline building is rotated: ```( is_rotation_required == FALSE ) AND ( is_rotated_baseline == TRUE ): FAIL```  
+- Case 2, the total vertical fenestration area per exposure differ by less than 5% and the baseline building is rotated: ```( is_rotation_required == FALSE ) AND ( is_rotated_b == TRUE ): FAIL```  
 
-- Case 3, the total vertical fenestration area per exposure differ by 5% or more and the baseline building is not rotated: ```( is_rotation_required == TRUE ) AND ( is_rotated_baseline == FALSE ): MANUAL_CHECK```  
+- Case 3, the total vertical fenestration area per exposure differ by 5% or more and the baseline building is not rotated: ```( is_rotation_required == TRUE ) AND ( is_rotated_b == FALSE ): MANUAL_CHECK```
```
supriyagoel, line 60: @xingcx this is quite well written up- thank you!

supriyagoel (changes requested):

docs/section5/Rule5-1.md
JasonGlazer: Yes.

docs/section5/Rule5-1.md
JasonGlazer: @supriyagoel I don't see a definition for north oriented, where is that?

supriyagoel: @JasonGlazer the definition of north-oriented is actually in the 'definitions' section. I have edited my comment to clarify that

JasonGlazer: In addition, Table G3.1 part 8 proposed exception 2 and Table G3.1 part 8 proposed item (b). Both of these assume a simple 45 binning of orientation. Although to me, the language implies four quadrants of 90 degrees each.

mkarpman: This exception is that rotation is not required for "Buildings where the vertical fenestration area on each orientation varies by less than 5%." 

There is also the following in the definition section:  "orientation: the direction an envelope element faces, i.e., the direction of a vector perpendicular to and pointing away from the surface outside of the element."

So I wonder if, instead separating surfaces by exposure (N, NE, NW, etc.), we should just use azimuth as criteria for totation.

JasonGlazer: That might be the simplest approach and it depends only on the defined term of "orientation"

mkarpman: So the logic might be to just calculate the total fenestration area for each unique azimuth, find max and min per azimuth, and if they differ by more than 5% then the baseline must be rotated.

supriyagoel: excellent- makes sense. Thank you. Summarized for @xingcx below.

- we need not find the total fenestration area for N, S, E, W (see the reason explained in MK comemnt above)
- Instead, we find the total fenestration area for each unique azimuth within a project
- Identify the min and max fenestration area for each azimuth
- if the min/max differ by >5%, then the baseline must be rotated.

docs/section5/Rule5-1.md
```diff
@@ -20,44 +20,42 @@
...
+- Case 2, the total vertical fenestration area per exposure differ by less than 5% and the baseline building is rotated: ```( is_rotation_required == FALSE ) AND ( is_rotated_b == TRUE ): FAIL```  
 
-- Case 3, the total vertical fenestration area per exposure differ by 5% or more and the baseline building is not rotated: ```( is_rotation_required == TRUE ) AND ( is_rotated_baseline == FALSE ): MANUAL_CHECK```  
+- Case 3, the total vertical fenestration area per exposure differ by 5% or more and the baseline building is not rotated: ```( is_rotation_required == TRUE ) AND ( is_rotated_b == FALSE ): MANUAL_CHECK```  
 
-- Case 4, the total vertical fenestration area per exposure differ by 5% or more and the baseline building is rotated: ```( is_rotation_required == TRUE ) AND ( is_rotated_baseline == TRUE ): PASS```  
+- Case 4, the total vertical fenestration area per exposure differ by 5% or more and the baseline building is rotated: ```( is_rotation_required == TRUE ) AND ( is_rotated_b == TRUE ): PASS```
```
mkarpman, line 61: •	Uses is_rotated schema element that has been eliminated, should replace with ASHRAE229: building_rotation_angles
•	Rule assertion does not align with the RDS logic, which focuses on whether the building was rotated and does not confirm that results were averaged.  
•	Surfaces are now attributes of zones, not spaces, need to change accordingly
•	RDS currently considers fenestration on any walls adjacent to exterior. Instead, it needs to consider fenestration on exterior and semi-exterior walls only.  Need to update to reference Function 3: Surface Conditioning Category. See definitions below for reference. 
fenestration: an assembly, including the frame, in the building envelope that allows light to pass.  
building envelope: the exterior plus the semiexterior portions of a building. For the purposes of determining building envelope requirements, the classifications are defined as follows:
exterior building envelope: the elements of a building that separate conditioned spaces from the exterior.
semiexterior building envelope: the elements of a building that separate  conditioned space from unconditioned space or that enclose semiheated spaces through which thermal energy may be transferred to or from the exterior, to or from unconditioned spaces, or to or from conditioned spaces.
•	RDS should aggregate fenestration with azimuths that exactly match, and then determine whether max and min across all azimuths differ by more than 5%. Currently, it aggregates fenestration area that has azimuths within 90 degrees before running the 5% check.

mkarpman (changes requested):

docs/section5/Rule5-1.md
xingcx: RDS is updated but not completed for the following: @JasonGlazer @mkarpman @supriyagoel 
1. Not sure how we can check if the baseline performance is averaged. Does OverallSimulationOutputs contain separate result set for each ASHRAE229.building_rotation_angles plus the average result?
2. Right now the code is updated to check exact azimuth. Sometimes the wall azimuth can be slightly different in the model (e.g. 0.5deg, etc) depending on the software snapping function even though they are the same on arch drawing, which could easily cause difference >5%. Not sure if this will be a concern.

mkarpman: @xingcx , to your first question - I suggested some updates to schema to support the baseline energy use calculations that would be necessary for this RDS. Perhaps you should hold off with it till it schema WG chimes in. 

For the second question, we discussed early on using rounding to allow values to not be exactly the same. @supriyagoel , do we have decision on this? Seems like number of decimal digits that we want to keep depends on the attribute - e.g. it may be different for U-factors vs LPDs vs azimuths.

supriyagoel: @mkarpman and @xingcx regarding #1, yes- we might need to wait for the schema WG to discuss that. we could address all other comments on this RDS and then come back to it once this aspect is discussed.
Regarding #2, lets discuss this during the team call tomorrow. For the number of decimal digits, the RCT team discussed this and thought it would be best to follow 90.1 guidance on this, i.e. if 90.1 uses 3 digits, we would round to that etc. However, there is also the question of tolerance. Do we need number to match exactly after rounding off, or is a certain % variation acceptable?

JasonGlazer: Regarding (1), please describe what is needed in OverallSimulationOutput as a new issues here: 

https://github.com/open229/ruleset-model-report-schema/issues

mkarpman: @JasonGlazer I added new issues to the schema and created a shared googedoc with the proposed changes. @xingcx let's wait with this portion of the RDS till proposed changes to the schema are vetted by the WG. You are shared on the googedoc with the proposed schema, so please disregard the Schema Data Element tab in the Envelope TCD v8 spreadsheet.

@supriyagoel , concerning #2, I agree that where there is a corresponding value in 90.1, we should stick with the same number of decimal digits. But in some cases it will not be the case - e.g. with azimuths, envelope dimentions or temperatures.

JasonGlazer (approved):

weilixu:
This is redrafting in #1163. Close this PR.

PR #72, opened 2021-06-02
Rt/jg/transformer json

jugonzal07:
Hi all,

I worked with @charliepnnl and we resolved a few issues identified by our tests. The pytest passes for all Rules that have been written, but ultimately fails as Section15Rule7 and Section15Rule8 have not been written yet. Once those are written, I believe transformers would be ready to go.


jugonzal07:
Hi all,

I think the transformer JSON work is ready for review. Note, I have merged in the old allow-templates-in-ruletest_engine branch into this so that PR is now redundant.  Let me know if I should just close that one.

mcneillj (changes requested):
It looks like there are a number of changes that may have been updated in develop branch.  You should merge develop in to make sure this branch is starting from the most up to date version.

jugonzal07:
@charliepnnl I had two comments for you.

1. It looks like merging in development did not quite fix the issue of other code from unrelated parts of the code (e.g., the changes to json_utils_test.py). Looking through the commit history, some others were due to changes we made when we met a few weeks back to get existing Rules to pass (see [this commit](https://github.com/pnnl/ruleset-checking-tool/pull/72/commits/1d133d4647da3e99727b9e3ba4d7640b230802f6)). Some other minor ones were due to formatting to black (see [this commit]( https://github.com/pnnl/ruleset-checking-tool/pull/72/commits/fc96c3433e36a21a45cd9636e674c4c7381eb2a2) for example). 

2. I had to make two small changes to the table_8_4_4_eff logic to pass Pytest results. One was I had to use an assert_approx_equal to handle funky float division. The other was table_8_4_4_eff function no longer requires dividing by 100.

rct229/data_fns/table_8_4_4_eff.py
```diff
@@ -66,4 +66,5 @@ def table_8_4_4_eff(phase, kVA):
         ],
     }
 
-    return strict_list_linear_interpolation(table_lists[phase], kVA)
+    # Round to 4 figures to match Table 8.4.4.
+    return round(strict_list_linear_interpolation(table_lists[phase], kVA), 4)
```
charliepnnl, line 70: Makes sense.

rct229/rules/section15.py
```diff
@@ -220,7 +220,7 @@ def get_calc_vals(self, context, data=None):
             return {
                 "baseline_transformer_efficiency": context.baseline["efficiency"],
                 "required_baseline_transformer_efficiency": table_8_4_4_eff(
-                    phase=baseline_phase, kVA=baseline_kVA
+                    phase=baseline_transformer_phase, kVA=baseline_transformer_kVA
```
charliepnnl, line 223: Good.  This is the same as the branch I'm currently working in. Not sure why it wasn't already this way.

charliepnnl (approved):

PR #77, opened 2021-06-21
Update Rule 6-mno to 6-4

xingcx:


supriyagoel:
@xingcx  this was discussed by the team on 06/28/2021. We will simply check if U-PRM =P-RMR for existing lighting. As MK pointed out, an existing space doesn't imply an existing lighting.
The Schema WG was not supportive about adding an input which would identify existing lighting, as this is not a tool that BEM tools report.

Hence, the RDS should simply check if U-RMR = P-RMR. The output should be a CAUTION as the reviewer will have to manually verify that the lighting is existing.

mkarpman:
@supriyagoel I wonder if it should be PASS instead of CAUTION.

mcneillj:
We will continue with CAUTION for now.  We will need to do a pass through all of these sections where we coordinate the message enumerations, so that can be a single set of work in the future once we complete the basic logic for the RDS.

supriyagoel (approved):

docs/section6/Rule6-4.md
```diff
@@ -0,0 +1,26 @@
...
+
+  - For each thermal_block in building segment: ```thermal_block_p in building_segment_p.thermal_blocks:```  
+
+    - For each zone in thermal block: ```zone_p in thermal_block_p.zones:```  
+
+      - For each space in zone: ```space_p in zone_p.spaces:```  
+
+        - Check if status_type is existing: ```if space_p.status_type == "EXISTING":```
+
+          **Rule Assertion:** Lighting power in the proposed RMR is as existing and cannot be verified by RCT: ```CAUTION```
```
mkarpman, line 26: I think for this rule the check should be that P-RMR = U-RMR? It's conceptually the same as for NC, except with NC the U-RMR must reflect design documents, and for EB U-RMR must reflect existing conditions.

mkarpman, line 26: I think the check should be that P-RMR = U-RMR. Isn't it the same as for NC, except for NC U-RMR must reflect design documents, and for EB U-RMR must reflect existing lighting that remains?

docs/section6/Rule6-4.md
```diff
@@ -0,0 +1,26 @@
...
+- For each building_segment in the proposed model: ```building_segment_p in P_RMR.building.building_segments:```  
+
+  - For each thermal_block in building segment: ```thermal_block_p in building_segment_p.thermal_blocks:```  
+
+    - For each zone in thermal block: ```zone_p in thermal_block_p.zones:```  
+
+      - For each space in zone: ```space_p in zone_p.spaces:```  
+
+        - Check if status_type is existing: ```if space_p.status_type == "EXISTING":```
+
```
mkarpman, line 25: I don't know if status_type is a good check because the space may be existing (e.g. this may be a renovation project) but new lighting system may be designed.

xingcx: Updated. Please let me know if the new assertion works. @supriyagoel @mkarpman

docs/section6/Rule6-4.md
xingcx: Updated. Now it is written to check if P_RMR == U-RMR for both Rule 6-4 (vs existing) and 6-5 (vs design).

xingcx: Updated. Now it is written to check if P_RMR == U-RMR for both Rule 6-4 (vs existing) and 6-5 (vs design).

docs/section6/Rule6-4.md
supriyagoel: @xingcx - this looks good to me.

supriyagoel (approved):

docs/section6/Rule6-4.md
```diff
@@ -0,0 +1,30 @@
...
+    - For each zone in thermal block: ```zone_p in thermal_block_p.zones:```  
+
+      - For each space in zone: ```space_p in zone_p.spaces:```  
+
+        - Get interior lighting in U_RMR: ```interior_lighting_b = match_data_element(U_RMR, InteriorLightings, space_p.interior_lighting.id)```
+
+          **Rule Assertion:** 
+
+          - Case 1: Lighting power in the proposed RMR is as existing and cannot be verified by RCT: ```if space_p.interior_lighting == interior_lighting_b: CAUTION```
+
```
mkarpman, line 29: I don't think this assertion works because there are no attributes in the RMR that allow differentiating between projects with existing lighting vs newly designed lighting vs lighting that is not designed. So all steps prior to row 28 would be identical for these three cases. So then how can RCT say that "Lighting power in the proposed RMR is as existing and cannot be verified by RCT"? I think in all three cases it should say "Lighting power in the proposed RMR is the same as in the baseline RMR". Then RCT can throw either "Pass" or "Caution" flag.

xingcx: Updated with manual check and new rule assertion. @mkarpman @supriyagoel @JasonGlazer 
Same question here on if we want all fields in InteriorLighting be the same between U_RMR and P_RMR or just power_per_area, lighting_multiplier_schedule, has_occupancy_control, has_daylighting_control.

docs/section6/Rule6-4.md
```diff
@@ -8,11 +8,17 @@
 
 **Applicability:** All required data elements exist for P_RMR  
 **Applicability Checks:** None  
-**Manual Check:** None  
+**Manual Check:** 
+
+1. Where a complete lighting system exists, the actual lighting power cannot be verified by the RCT. Manual review is required.  
+
```
mkarpman, line 14: @xingcx could you please remind me the significance of the Manual Check? How does it address the issue of RCT not being able to tell whether lighting is newly designed, existing to remain or not designed? Since we don't have the attribute(s) in the RMR that would allow us to differentiate between these three scenario, should all three be covered by the same RDS? 

This decision would apply to all building systems and components, not just lighting. For example, for envelope G3.1 #5a says "a. All components of the building envelope in the proposed design shall be modeled as shown on architectural drawings or as built for existing building envelopes."  Also, Table G3.1 #1c says
c. When the Performance Rating Method is applied to buildings in
which energy-related features have not yet been designed
(e.g., a lighting system), those yet-to-be-designed features
shall be modeled in the proposed design to comply with but not
exceed the requirements of this standard as described in Table
G3.1, Nos. 6, 10, 11, and 12. 
I suggest that in all such cases RCT checks if U-RMR=P-RMR, and perhaps additionally check if P-RMR reflects minimum requirements of 90.1 Sections 5-10, as applicable. @supriyagoel please take a look also.

JasonGlazer (approved):

docs/section6/Rule6-4.md
supriyagoel: @mkarpman and @xingcx the 3 RDS for as existing, as designed and not yet designed - would all require manual check.
Maybe we have a CAUTION outcome for when P-RMR = U-RMR and a FAIL outcome for when P-RMR is not equal to U-RMR. @mkarpman - would that work?
Xing, please have a look out here-
https://docs.google.com/document/d/1KNEe0Igcv8CA6e9u4rlKCCGsvAsNNvjN1_mmpPQeuQc/edit?usp=sharing

supriyagoel: @mkarpman its good to have 3 different RDS for this, even though the outcome is the same. This is because these are 3 different rules in 90.1  and we want these to be separate rule definitions with separate outcomes. Hope that makes sense

mkarpman: @supriyagoel I don't mind having different RDSs for existing, as designed and not yet designed, but wouldn't they all be exactly the same? We may have the following scenarios:
1. U-RMR = P-RMR 
a. correct if component is existing or as designed
b. may or may not be correct if component is yet to be designed because U-RMR may use assumptions other than what is prescribed in App G for such scenarios.  

2. U-RMR != P-RMR
a. wrong if component is existing or as designed
b. may or may not be correct depending on whether P-RMR matches what is prescribed in App G for such scenarios.  

You suggest to show Caution for (1) and Fail for (2), but maybe it should be more nuanced. I think as James suggested, we should also check if P-RMR reflects BAM LPDs and if they do show caution + "P-RMR reflects yet-to-be designed default LPD" in cases 1b and 2b.

I think we will have similar situations with systems other than lighting as well.

supriyagoel (approved):

docs/section6/Rule6-4.md
```diff
@@ -0,0 +1,46 @@
...
+    - For each zone in thermal block: `zone_p in thermal_block_p.zones:`  
+
+      - For each space in zone: `space_p in zone_p.spaces:`  
+
+        - For each interior lighting in space: `interior_lighting_p in space_p.interior_lightings:`  
+
+          **Rule Assertion:**  
+
+          - Case 1: If lighting status type in space is not-yet designed, and interior lighting power density matches Table 9.5.1: `if ( space_lighting_status_type_dict[space.id] == "NOT-YET DESIGNED" ) and ( interior_lighting_p.power_per_area == data_lookup(table_9_5_1, lighting_building_area_type_p) ): PASS and raise_warning "LIGHTING NEITHER EXISTS OR IS SUBMITTED WITH DESIGN DOCUMENTS. LIGHTING POWER DENSITY IN P_RMR MATCHES TABLE 9.5.1."`  
+
```
mkarpman, line 40: This may also be a coincidence that specified lighting matches TABLE 9.5.1. Perhaps add it to the warning message.

mkarpman, line 40: To calculate the total wattage in a space, we need to sum up power_per_area x  space.floor_area for all interior_lighting objects and then divide by space area. This LPD would then be compared to value in Table 9.5.1.

docs/section6/Rule6-4.md
```diff
@@ -0,0 +1,46 @@
...
+          - Case 2:  Else if lighting status type in space is not-yet designed, and interior lighting power density does not match Table 9.5.1 : `else if ( space_lighting_status_type_dict[space.id] == "NOT-YET DESIGNED" ) and ( interior_lighting_p.power_per_area != data_lookup(table_9_5_1, lighting_building_area_type_p) ): FAIL and raise_warning "LIGHTING NEITHER EXISTS OR IS SUBMITTED WITH DESIGN DOCUMENTS. LIGHTING POWER DENSITY IN P_RMR DOES NOT MATCH TABLE 9.5.1."`  
+
```
mkarpman, line 42: I would update the message to say "Baseline lighting power was established using building area method, which is only allowed for spaces where lighting is not yet designed. However, lighting in the proposed design does not match the values in Table 9.5.1, as required for yet to be designed lighting."

docs/section6/Rule6-4.md
```diff
@@ -0,0 +1,46 @@
...
+          - Case 3: Else if lighting status type in space is as-designed or as-existing, and interior lighting power density in P_RMR matches U_RMR: `else if ( space_lighting_status_type_dict[space.id] == "AS-DESIGNED OR AS-EXISTING" ) and ( interior_lighting_p.power_per_area == match_data_element(U_RMR, InteriorLightings, interior_lighting_p.id).power_per_area ): PASS and raise_warning "LIGHTING EXISTS OR IS SUBMITTED WITH DESIGN DOCUMENTS. LIGHTING POWER DENSITY IN P_RMR MATCHES U_RMR"`  
+
```
mkarpman, line 44: The comparison should be based on the total space wattage assuming it may have multiple interiorlighting objects (i.e., same comment as above for row 40).

Why should there be a warning?

docs/section6/Rule6-4.md
```diff
@@ -0,0 +1,46 @@
...
+          - Case 4: Else if lighting status type in space is as-designed or as-existing, and interior lighting power density in P_RMR does not match U_RMR: `else if ( space_lighting_status_type_dict[space.id] == "AS-DESIGNED OR AS-EXISTING" ) and ( interior_lighting_p.power_per_area != match_data_element(U_RMR, InteriorLightings, interior_lighting_p.id).power_per_area ): CAUTION and raise_warning "LIGHTING EXISTS OR IS SUBMITTED WITH DESIGN DOCUMENTS. LIGHTING POWER DENSITY IN P_RMR DOES NOT MATCH U_RMR."`  
+
```
mkarpman, line 46: I think this should be "FAIL", please please remember to take into account that there may be multiple interior lighting objects for a given space in both P-RMR and U-RMR.

docs/section6/Rule6-4.md
xingcx: I'm thinking Case 2 I had might not be valid. If lighting status type is "not-yet designed", the LPD has to match Table 9.5.1. I updated the 4 cases please review.

docs/section6/Rule6-4.md
mkarpman: I think this should just be a PASS with no warning. Otherwise, as it stands now, RCT would be generating either caution or warning message in all cases, so there would be as many caution messages as spaces in the model just for this rule.

Perhaps we should not worry about the case of the specified LPD coincidently mathing the value in TABLE 9.5.1

docs/section6/Rule6-4.md
mkarpman: Case 2 seems right to me, except I would make the outcome FAIL

mkarpman: I don't think row 45 is correct. For yet-to-be-designed, the value in U-RMR doesn't matter, we just need to check that the total_space_LPD_p is equal to the value in Table 9.5.1. 

Ditto Case 4. Plust I would also change outcome for this case to FAIL.

docs/section6/Rule6-4.md
supriyagoel: @mkarpman I agree. Too many cautions/warnings dont provide useful information.

We have also been discussing having a section in the output report that summarizes the results for some 'Key Rules'. The lighting method used by the building could be one such key rule and could be highlighted in the report for the reviewer.

supriyagoel: @mkarpman i think summing up the W/ft2 will give the same result, right?

docs/section6/Rule6-4.md
supriyagoel: Case 1 and case 2 make sense. I am confused by Case 3 and case 4. 
@mkarpman if i understand your comment correctly, you're saying that for yet-to-be-designed spaces, we dont case that U-RMR = P-RMR? in this case, we only need to verify that P-RMR = 9.5.1?

supriyagoel (changes requested):

docs/section6/Rule6-4.md
xingcx: @mkarpman I have Case 3 and 4 for when the lighting is designed/existing but the LPD also matches Table 9.5.1. Let me know how we want to output in this case. If it is not-yet designed, the LPD will definitely match 9.5.1 as an outcome of the function, as you said.

docs/section6/Rule6-4.md
xingcx: Warning message for Case 1 is now deleted.

mkarpman (approved):

supriyagoel (approved):

docs/section6/Rule6-4.md
```diff
@@ -0,0 +1,46 @@
...
+          - Get total lighting power density in space in U_RMR: `total_space_LPD_u = sum(interior_lighting.power_per_area for interior_lighting in space_u.interior_lighting)`
+
+        **Rule Assertion:**  
+
+        - Case 1: If lighting status type in space is as-designed or as-existing, and interior lighting power density in P_RMR matches U_RMR: `if ( space_lighting_status_type_dict[space.id] == "AS-DESIGNED OR AS-EXISTING" ) and ( total_space_LPD_p == total_space_LPD_u ): PASS`
+
+        - Case 2: Else if lighting status type in space is as-designed or as-existing, and interior lighting power density in P_RMR does not match U_RMR: `else if ( space_lighting_status_type_dict[space.id] == "AS-DESIGNED OR AS-EXISTING" ) and ( total_space_LPD_p != total_space_LPD_u ): FAIL and raise_warning "LIGHTING EXISTS OR IS SUBMITTED WITH DESIGN DOCUMENTS. LIGHTING POWER DENSITY IN P_RMR DOES NOT MATCH U_RMR."`
+
+        - Case 3: Else, lighting status type in space is not-yet designed or is as-designed or as-existing but matches Table 9.5.1. If interior lighting power density in P_RMR matches U_RMR: `else if total_space_LPD_p == total_space_LPD_u: PASS"`
+
```
mkarpman, line 45: Case 3 is essentially an expanded version of Case 1.  Should we just replace Case 1 with Case 3, and reduce the number of cases?

docs/section6/Rule6-4.md
```diff
@@ -0,0 +1,46 @@
...
+        - Case 4: Else, interior lighting power density in P_RMR does not match U_RMR: `else: FAIL and raise_warning "LIGHTING IS NOT YET DESIGNED, OR LIGHTING IS AS-DESIGNED OR AS-EXISTING BUT MATCHES TABLE 9.5.1. LIGHTING POWER DENSITY IN P_RMR DOES NOT MATCH U_RMR."`
```
mkarpman, line 46: I recommend making it a "CAUTION" outcome.

xingcx: Updated.

docs/section6/Rule6-4.md
xingcx: Updated.

mkarpman (approved):

PR #83, opened 2021-06-28
Update _ruleset_functions.md

xingcx:
A few notes/questions I have:
**get_zone_conditioning_category**
1. It seems that the criteria for indirectly conditioned spaces is based on heated/cooled space vs outdoor/uncondiioned/semiheated spaces. So other indirectly conditioned spaces are not part of the consideration in the function right now. 

> indirectly conditioned space: an enclosed space within a building that is not a heated space or a cooled space, which is heated or cooled indirectly by being connected to adjacent spaces, provided: 
1.the product of the U-factors and surface areas of the space adjacent to connected spaces exceeds the combined sum of the product of the U-factors and surface areas of the space adjoining the outdoors, unconditioned spaces, and to or from semiheated spaces (e.g., corridors)

2. For the second type of indirectly conditioned spaces, i.e. transfer air from heated or cooled space > 3ACH, will the baseline have transfer air? If not, maybe we do not need to manually check this?

**get_surface_conditioning_category**
1. Can we keep the residential/non-residential determination in the actual RDS instead of putting it in the function? To have it in the function, we would need to check the actual envelope construction table (Table G3.4) and differentiate surface type (wall, ceiling, floor) if we want to raise warning when the residential construction is different from the non-res construction. And we also need to check subsurface, as sometimes the surface has same construction but the subsurface doesn't. This seems to make the function distracted. 

2. I'm still not clear if zone.surfaces only returns surfaces surrounding a zone (as is noted in the schema) or returns all surfaces that defines all spaces in the zone. If the former, and in cases where the zone has both res and non-res spaces and the residential and non-res baseline constructions are different, how can we do manual check (and do we need to raise warning)? It seems there's no indicator of whether the surface belongs to the residential or non-residential space.

### DIFF
--- a/docs/_ruleset_functions.md
+++ b/docs/_ruleset_functions.md
@@ -1,33 +1,135 @@
 ## get_zone_conditioning_category
-Description:  Determine the Space Conditioning Category for each zone.  This function would cycle through each space in an RMR and categorize it as ‘conditioned’, semi-heated’ or ‘unconditioned’.  If ‘conditioned’ it will also categorize the space as ‘residential’ or ‘non-residential’.  
+Description: Determine the Zone Conditioning Category for each zone. This function would cycle through each zone in an RMR and categorize it as ‘conditioned’, 'semi-heated’ or ‘unconditioned’.  
 
-Inputs:  
-- **Zone**: A Zone data element from the RMR.  
+Inputs:
+  - **RMR**: The RMR that needs to determine zone conditioning category.  
 
-Returns:  
-- **zone_conditioning_category**: The Zone Conditioning Category [conditioned, semi-heated, unconditioned].
+Returns:
+- **zone_conditioning_category**: The Zone Conditioning Category [conditioned, semi-heated, unconditioned].  
 
-Logic: [FILL IN LOGIC HERE]   
+
+Logic:  
+
+- Get building climate zone: ```climate_zone = RMR.weather.climate_zone```  
+
+  - Get heated space criteria: ```system_min_heating_output = data_lookup(table_3_2,climate_zone)```  
+
+- For each building segment in the RMR: ```for building_segment in RMR.building.building_segments:```  
+
+  - To determine zone conditioning category for zones served by an HVAC system, for each HVAC system in building segment: ```for hvac_system in building_segment.heating_ventilation_air_conditioning_systems:```  
+
+    - Check if the system meets the criteria for serving directly conditioned (heated or cooled) zones, save all zones served by the system as directly conditioned: ```if ( hvac_system.simulation_result_sensible_cool_capacity >= 3.4 ) OR ( hvac_system.simulation_result_heat_capacity >= system_min_heating_output ):  for zone in hvac_system.zones_served: zone_conditioning_category_dict[zone.id] = "DIRECTLY-CONDITIONED"```  
+
+    - Else check if the system meets the criteria for serving semi-heated zones, save all zones served by the system as semi-heated: ```else if hvac_system.simulation_result_heat_capacity >= 3.4: for zone in hvac_system.zones_served: zone_conditioning_category_dict[zone.id] = "SEMI-HEATED"```  
+
+    - Else, save all zones served by the system as unconditioned: ```else: for zone in hvac_system.zones_served: zone_conditioning_category_dict[zone.id] = "UNCONDITIONED"```  
+
+  - To determine zone conditioning category for zones not served by any HVAC system, for each thermal block in building segment: ```for thermal_block in building_segment.thermal_blocks:```  
+
+    - For each zone in thermal block: ```for zone in thermal_block.zones:```  
+
+      - If zone has not been classified: ```if zone.id not in zone_conditioning_category_dict:```  
+
+        - For each surface in zone: ```for surface in zone.surfaces:```  
+
+          - Check if surface is interior: ```if surface.adjacent_to == "INTERIOR":```  
+
+            - If the adjacent zone is directly conditioned (heated or cooled), add the product of the U-factor and surface area to the directly conditioned type: ```if zone_conditioning_category_dict[surface.adjacent_zone_id] == "DIRECTLY-CONDITIONED": directly_conditioned_product_sum += sum( ( fenestration.glazed_area + fenestration.opaque_area ) * fenestration.u_factor for fenestration in surface.fenestration_subsurfaces ) + ( surface.area - sum( ( fenestration.glazed_area + fenestration.opaque_area ) for fenestration in surface.fenestration_subsurfaces ) * surface.construction.u_factor```  
+
+            - Else, add the product of the U-factor and surface area to the other type (outdoor, semi-heated or unconditioned): ```else: other_product_sum += sum( ( fenestration.glazed_area + fenestration.opaque_area ) * fenestration.u_factor for fenestration in surface.fenestration_subsurfaces ) + ( surface.area - sum( ( fenestration.glazed_area + fenestration.opaque_area ) for fenestration in surface.fenestration_subsurfaces ) * surface.construction.u_factor```  
+
+          - Else check if surface is exterior, add the product of the U-factor and surface area to the other type (outdoor, semi-heated or unconditioned): ```else if surface.adjacent_to == "EXTERIOR": other_product_sum += sum( ( fenestration.glazed_area + fenestration.opaque_area ) * fenestration.u_factor for fenestration in surface.fenestration_subsurfaces ) + ( surface.area - sum( ( fenestration.glazed_area + fenestration.opaque_area ) for fenestration in surface.fenestration_subsurfaces ) * surface.construction.u_factor```  
+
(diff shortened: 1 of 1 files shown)


mcneillj:
Is the intent to have zone_conditioning_category be evaluated for an entire RMR or just return the value for a specific zone?

xingcx:
> @xingcx Is the intent to have zone_conditioning_category be evaluated for an entire RMR or just return the value for a specific zone?

@mcneillj The original input template for this function is to return the value for a specific zone. But my thought is it's more efficient and convenient to evaluate the entire RMR and return a dictionary instead of for a specific zone, for the following reasons:
1. We will need to pass the entire RMR to the function either way, as we need the climate zone, hvac system info, and conditioning category for some/all other zones (in some cases) to determine the conditioning category for a specific zone. 
2. We will only need to run the function once each RDS to get the information we need.

mkarpman:
@mcneillj <https://github.com/mcneillj> I agree



*From:* xingcx ***@***.***>
*Sent:* Wednesday, June 30, 2021 4:37 PM
*To:* pnnl/ruleset-checking-tool ***@***.***>
*Cc:* mkarpman ***@***.***>; Review requested <
***@***.***>
*Subject:* Re: [pnnl/ruleset-checking-tool] Update _ruleset_functions.md
(#83)



@xingcx <https://github.com/xingcx> Is the intent to have
zone_conditioning_category be evaluated for an entire RMR or just return
the value for a specific zone?

@mcneillj <https://github.com/mcneillj> The original input template for
this function is to return the value for a specific zone. But my thought is
it's more efficient and convenient to evaluate the entire RMR and return a
dictionary instead of for a specific zone, for the following reasons:

   1. We will need to pass the entire RMR to the function either way, as we
   need the climate zone, hvac system info, and conditioning category for
   some/all other zones (in some cases) to determine the conditioning category
   for a specific zone.
   2. We will only need to run the function once each RDS to get the
   information we need.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/pnnl/ruleset-checking-tool/pull/83#issuecomment-871710710>,
or unsubscribe

mcneillj:
> > @xingcx Is the intent to have zone_conditioning_category be evaluated for an entire RMR or just return the value for a specific zone?
> 
> @mcneillj The original input template for this function is to return the value for a specific zone. But my thought is it's more efficient and convenient to evaluate the entire RMR and return a dictionary instead of for a specific zone, for the following reasons:
> 
> 1. We will need to pass the entire RMR to the function either way, as we need the climate zone, hvac system info, and conditioning category for some/all other zones (in some cases) to determine the conditioning category for a specific zone.
> 2. We will only need to run the function once each RDS to get the information we need.

That's fine.  As this is just pseduo-code we really just need the representation of the general logic.  The level of detail of coding efficinecy will get sorted out during the Rule Definition coding process.  I just want to be clear for referencing the function in the RDS, so there isn't confusion.

docs/_ruleset_functions.md
```diff
@@ -1,33 +1,270 @@
...
+  - For each thermal_block in building segment: `for thermal_block in building_segment.thermal_blocks:`
+
+    - For each zone in thermal block: `zone in thermal_block.zones:`
+
+      - Check if zone is conditioned or semi-heated: `if zcc_dictionary[zone.id] in [CONDITIONED RESIDENTIAL, CONDITIONED NON-RESIDENTIAL, CONDITIONED MIXED, SEMI-HEATED]`
+
+        - For each surface in zone: `for surface in zone.surfaces:`
+
+          - If surface is above-grade wall and adjacent to the exterior: `if ((get_opaque_surface_type(surface) == "ABOVE-GRADE WALL") AND (surface.adjacent_to == "EXTERIOR")):`  
+
```
mkarpman, line 255: We have pending interpretation request about this. Based on discussions of the ECB subcommittee, it seems likely that we need to drop "adjacent_to=exterior condition". 
So for this function we'll need to get_surface_conditioning_category, and for all surfaces that are exterior res, exterior non res, exterior mixed and semi-exterior, and include all surfaces that have get_opaque_surface_type=wall.

docs/_ruleset_functions.md
```diff
@@ -1,33 +1,270 @@
...
+            - For each subsurface in surface: `for subsurface in surface.subsurfaces:`  
+
+              - If the glazed vision area is less than 50% of the total area, add glazed vision area to building segment total fenestration area: `if ( subsurface.glazed_area < ( subsurface.glazed_area + subsurface.opaque_area ) * 50% ): total_fenestration_area += subsurface.glazed_area`  
+
```
mkarpman, line 259: From 90.1 Section 3: door where one-half or less of the door area is glazed is considered an opaque door. 

So I think the logic should be to calculate (subsurface.glazed_area + subsurface.opaque_area) for all subsurfaces that have qualified parent surfaces (i.e., exterior or semi-exterior walls) AND (have subsurface.classification != "door" OR (subsurface.classification=door AND (subsurface.glazes_area=>0.5*(subsurface.glazed_area + subsurface.opaque_area))

mkarpman (changes requested):

docs/_ruleset_functions.md
```diff
@@ -1,33 +1,270 @@
...
-Inputs:  
-- **ADD_HERE**: Description of inputs.  
+          - Add to envelope total above-grade wall area: `total_envelope_wall_area += surface.area`
 
-Returns:  
-- **ADD_HERE**: Description of output.  
+- Calculate WWR of the building segment: `window_wall_ratio = total_fenestration_area / total_envelope_wall_area`
 
-Logic: [FILL IN LOGIC HERE]   
+**Returns** ```return window_wall_ratio```
```
mkarpman, line 270: I would suggest having a dedicated # for each function so that they can be individually approved.

docs/_ruleset_functions.md
xingcx: I was referencing G3.1-5(c) vertical fenestration...separates conditioned or semi-heated space from exterior. But in the current get_surface_conditioning_category, if a surface is semi-exterior, it could be between conditioned and unconditioned spaces, which is not exterior. And if a surface is exterior res or exterior non res, it could be between conditioned and unenclosed spaces.

docs/_ruleset_functions.md
```diff
@@ -1,33 +1,270 @@
...
+
+Returns:
+- **window_wall_ratio**: The window wall ratio of the building segment.  
+
+Function Call:
+
+- get_zone_conditioning_category()
+- get_surface_conditioning_category()
+- get_opaque_surface_type()
+
```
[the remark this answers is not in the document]
xingcx: get_zone_conditioning_category() is referenced on Line 250.

docs/_ruleset_functions.md
mkarpman: I submitted interpretation request to ASHRAE about this, and based on ECB subcommittee discussions it is all but certain that the interpretation will say that both exterior and semi-exterior surfaces in your example above count as adjacent to exterior because the unconditioned or unenclosed spaces are themselves pathway to exterior.

docs/_ruleset_functions.md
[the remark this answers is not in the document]
mkarpman: I might well be missing something, but I am not sure we need it there either because it's called from within get_surface_conditioning_category. So in WWR function we need to (1) call get_surface_conditioning_category to find all semi-heated and exterior surfaces, and then (2) call get_opaque_surface_type to identify all such surfaces that are walls.

docs/_ruleset_functions.md
```diff
@@ -1,33 +1,270 @@
...
+Logic:
+
+- Get zone conditioning category dictionary for RMR: `zcc_dictionary = get_zone_conditioning_category(RMR)`
+
```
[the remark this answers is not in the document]
supriyagoel: @xingcx and @mkarpman I find it easier to understand if it is explicitly mentioned. Though I'm ok if we feel it needs to be removed.

docs/_ruleset_functions.md
supriyagoel: @mkarpman Its been a while since we've discussed this. We have two WWR calculations, one for determining baseline U/SHGC- so this needs to be a single value for the entire building and one for determining the actual baseline WWR (this needs to be by building segment). Here I see we're doing the WWR calculation by building segment, so this is for purpose #2  correct?

docs/_ruleset_functions.md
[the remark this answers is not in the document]
mkarpman: I am good either way.

supriyagoel:
@xingcx would it make sense to have separate MD documents for each function? This document is now getting quite long.

docs/_ruleset_functions.md
mkarpman: Correct. The interpretation request that I submitted asks for clarification on both WWR calculations. #2 above has been discussed by the subcommittee, but we haven't talked about #1 yet.

xingcx: Surface.adjacent_to is now replaced with get_surface_conditioning_category outputs. @mkarpman

mcneillj:
Archive. Ruleset functions moved to individual files and PRs.

PR #84, opened 2021-06-30
Create Rule6-13.md

xingcx:
A few notes/questions I have:
1. Do we need to consider schedule.is_schedule_modified_for_workaround or is comparing EFLH sufficient for this rule?
2. Right now it seems the schema only has lighting.lighting_multiplier_schedule_name to find the schedule (string) from lighting object. I'm using match_data_element to locate the actual schedule, which means match_data_element will need to use name instead of id.  I remember there were some discussion on name vs id for the other rules, so just want to add a note here for future discussion.
3. We will need to add EFLH function to _function.md. I also put a note on the RDS tracker under the "request" tab.
4. The RDS spreadsheet currently has "Proposed building includes spaces with lighting systems with additional occupancy sensor control" as the rule trigger. Since it's part of the rule to loop through each space and check if space has lighting control, do we want to consider remove this trigger? Or we will have two almost identical loops. 
5. Do we need to add a trigger and apply this RDS only if the building does not use BAM?


supriyagoel:
Similar to the daylighting control test, we will use the function to compare hourly values of the lighting schedule between the B and P and report the vaariattion.
•	If the S-WG agrees to adding additional data elements, then we will ask about the type of occ control (partial on, manual on) and verify the hourly schedule variation based on the adjustment for the same (Rule 37-b)

JasonGlazer:
So do we need some new functions that compare two schedules?

mkarpman:
@JasonGlazer Yes, I agree.

docs/section6/Rule6-13.md
```diff
@@ -10,41 +10,69 @@
 - Table G3.7, Performance Rating Method Lighting Power Density Allowances and Occupancy Sensor Reductions Using the Space-by-Space Method  
 
 **Applicability:** All required data elements exist for P_RMR  
-**Applicability Checks:** None  
+**Applicability Checks:**  
+
+  1. Pass Rule 6-10.
+
```
supriyagoel, line 16: not sure why we need the applicability check for 6-10. 6-10 checks for occupancy controls in the baseline. occupancy controls in proposed are required irrespective of the outcome of 6-10. @mkarpman - any thoughts?

docs/section6/Rule6-13.md
```diff
@@ -10,41 +10,69 @@
...
+    - For each zone in thermal block: `zone_p in thermal_block_p.zones:`  
+
+      - For each space in zone: `space_p in zone_p.spaces:`  
+
+        - Get matching space from B_RMR: `space_b = match_data_element(B_RMR, Spaces, space_p.id)`  
+
+          - Get normalized space lighting schedule for B_RMR: `normalized_schedule_b = normalize_space_lighting_schedule(space_b)`  
+
+        - Get lighting space type: `lighting_space_type_p = space_p.lighting_space_type`  
+
```
supriyagoel, line 44: need normalized schedule at space level for proposed as well. Proposed could have several lighting objects with differant schedules and we should be calculating the normalized schedule at the space level for P as well.

supriyagoel (changes requested):
I believe this RDS could be quite straightforward. 

- we get the applicable occupancy reduction fraction for each space from table G3.7
- we use the function to calculate normalized schedule at space level for P and B
- we use the function to compare the normalized schedule at space level for P and B
- we can check if occ control = true. If yes, we display the outcome of the comparison result
- if false, P schedule should be = baseline schedule. Pass if true, caution is false with warning.

I think doing anything more will be challenging as the P and B could have any combination of shut off controls, programmable controls or daylighting controls. I think it would be best to simply do a comparison and report the outcome for the first round.

docs/section6/Rule6-13.md
mkarpman: Yes, I agree.

docs/section6/Rule6-13.md
```diff
@@ -0,0 +1,78 @@
...
+**Data Lookup:** Table G3.7  
+**Function Call:**  
+
+  - compare_schedules()
+  - normalize_space_lighting_schedule()
+
+## Rule Logic:  
+
+- **Applicability Check 1:** If pass Rule 6-10: `if rule_6_10 == TRUE:`  
+
```
mkarpman, line 28: This does not appear necessary per above

docs/section6/Rule6-13.md
mkarpman: Agreed

docs/section6/Rule6-13.md
```diff
@@ -0,0 +1,78 @@
...
+      - For each space in zone: `space_p in zone_p.spaces:`  
+
+        - Get matching space from B_RMR: `space_b = match_data_element(B_RMR, Spaces, space_p.id)`  
+
+          - Get normalized space lighting schedule for B_RMR: `normalized_schedule_b = normalize_space_lighting_schedule(space_b)`  
+
+        - Get lighting space type: `lighting_space_type_p = space_p.lighting_space_type`  
+
+          - If lighting space type is employee lunch and break rooms, conference/meeting rooms, or classrooms (not including shop classrooms, laboratory classrooms, and preschool through 12th-grade classrooms), get baseline lighting occupancy sensor reduction factor from Table G3.7: `if lighting_space_type_p in ["LOUNGE/BREAKROOM", "CONFERENCE/MEETING/MULTIPURPOSE ROOM", "CLASSROOM/LECTURE HALL/TRAINING ROOM - PENITENTIARY", "CLASSROOM/LECTURE HALL/TRAINING ROOM - ALL OTHER"]: reduction_factor_b = data_lookup(table_G3_7, lighting_space_type_p, "FULL-AUTO ON")`  
+
```
mkarpman, line 46: I think we can eliminate this condition because values in table G3.7 take care of these space types automatically.

docs/section6/Rule6-13.md
```diff
@@ -0,0 +1,78 @@
...
+          - Else, set baseline reduction factor to 0: `else: reduction_factor_b = 0`  
+
+        - For each lighting in space: `lighting_p in space_p.interior_lighting:`  
+
+          - Get lighting control type: `lighting_control_type_p = lighting_p.occupancy_control_type`  
+
+          - Get lighting occupancy sensor reduction factor from Table G3.7: `reduction_factor_p = data_lookup(table_G3_7, lighting_space_type_p, lighting_control_type_p)`  
+
```
mkarpman, line 54: I think here we need to calculate what the correct proposed normalized schedule should be (based on its difference from the baseline) and compare it to the modeled proposed normalized lighting schedule for each space.

I would establish the correct schedule based on (a) schedule reduction fraction in Table G3.7 and (b) whether this fraction should be increased based on the listed controls for each interiorlighting object as required in footnotes b (b. For manual-ON or partial-auto-ON occupancy sensors, the occupancy sensor reduction factor shall be multiplied by 1.25.) and c (c. For occupancy sensors controlling individual workstation lighting, occupancy sensor reduction factor shall be 30%.)

docs/section6/Rule6-13.md
```diff
@@ -0,0 +1,78 @@
...
+          - Calculate adjusted lighting occupancy sensor reduction factor based on B_RMR: `adjusted_reduction_factor_p = ( 1 - reduction_factor_p ) / ( 1 - reduction_factor_b )`  
+
+          - Get lighting schedule: `lighting_schedule_p = lighting_p.lighting_multiplier_schedule`  
+
```
mkarpman, line 58: The adjustment may be different for each interior_lighting object

docs/section6/Rule6-13.md
```diff
@@ -0,0 +1,78 @@
...
+          - Compare lighting schedules in P_RMR and B_RMR: `schedule_comparison_result = compare_schedules(lighting_schedule_p, normalized_schedule_b, building_open_schedule_p, adjusted_reduction_factor_p)`  
+
```
mkarpman, line 60: I think it should be compare_schedules(normalized_schedule_p, normalized_schedule_b, building_open_schedule_p, adjusted_reduction_factor_p).....`

docs/section6/Rule6-13.md
```diff
@@ -0,0 +1,78 @@
...
+            **Rule Assertion:**
+
+            - Case 1: For all hours, for each lighting, if lighting schedule in P_RMR is equal to lighting schedule in B_RMR times adjusted lighting occupancy sensor reduction factor: `if schedule_comparison_result == "MATCH": PASS`  
+
+            - Case 2: Else if lighting schedule in P_RMR is lower than or equal to lighting schedule in B_RMR times adjusted lighting occupancy sensor reduction factor: `if schedule_comparison_result == "EQUAL AND LESS": CAUTION`  
+
```
mkarpman, line 66: What does "equal and less" mean? I think if P_RMR schedule is lower than what it should be, the outcome should be FAIL (savings are exaggerated). And if it's greater than what it should be, the outcome may be caution (proposed lighting is modeled with higher runtime hours then we expect which is wrong but result in conservative compliance outcome.

mkarpman (changes requested):

docs/section6/Rule6-13.md
xingcx: Applicability check is now removed.

docs/section6/Rule6-13.md
xingcx: Applicability check is now removed.

docs/section6/Rule6-13.md
xingcx: Right now the RDS compared each lighting object schedule with baseline. This won't work when baseline has multiple lighting objects with different schedule and needs to be modified. 
I'm not sure if we need to normalize schedules in P_RMR (using the normalize schedule function we have right now). For example, if 
B_RMR:
- LPD1 = 1W/sf, schedule = 0.5
- LPD2 = 1W/sf, schedule = 0.8, (assuming B_RMR multiple LPD should be the same, probably with the exception of retail lights)
- Total LPD = 2W/sf, normalized schedule = 0.65

P_RMR if no controls:
- LPD1 = 0.5W/sf, schedule = 0.5
- LPD2 = 0.3W/sf, schedule = 0.8
- Total LPD = 0.8W/sf, normalized schedule = 0.6125
In this case, averaging all schedules then compare to the normalized baseline schedule will work, with the exception when multiple lighting objects in B_RMR have different LPDs.

P_RMR with occ sensor only:
- LPD1 = 0.5W/sf, schedule = 0.4 (assuming occ sensor reduction factor of 0.2)
- LPD2 = 0.3W/sf, schedule = 0.72 (assuming occ sensor reduction factor of 0.1)
- Total LPD = 0.8W/sf, normalized schedule = 0.52
In this case, we can calculate each lighting schedule without reduction factor and average all schedules then compare to the normalized baseline schedule. Again it won't seem to work if multiple lighting objects in B_RMR have different LPDs.

P_RMR with both occ sensor and other type of controls:
In this case maybe we just compare if the normalized schedule in P_RMR is less than B_RMR.

Or do I need to modify the normalize schedule function?

mkarpman: I think the baseline space lighting may include several interior lighting objects with different LPDs. For example, general and task lighting in a space may be modeled as separate interior lighting objects each with a different schedule. This would meet App G requirements as long as the total LPD equials the expected value in Table G3.7.  

To make this RDS work, it seems that we need to require that baseline either (a) has a single interior lighting object, or (b) has the same number of lighting objects as the proposed design. If (a), we'd compare each of the proposed schedules to the single baseline schedule individually, and confirm that they are either the same (no controls) or the difference is as expected. If (b), we'd expect pairs of interior lighting objects and compare schedules to establish whether they relate to each other as expected.

supriyagoel: @xingcx agree with MK. there is no requirement/guidance in App G for the B to have multiple lighting objects. In most cases, we would see the B RMR with a single lighting object with the prescribed LPD based on Table G3.7. The most conservative option would be to use the normalized schedules for both B and P. We can also discuss this tomorrow if you think otherwise.

docs/section6/Rule6-13.md
xingcx: @mkarpman As this compares all hours in 8760 schedule, "equal and less" means no hour is greater, just to cover cases when some hours are less and others are equal (or all hours are less). If P_RMR is lower than what's calculated in some hours, it might be because of other controls like daylighting. 
For Case 3, I'll change it from FAIL to CAUTION.

docs/section6/Rule6-13.md
xingcx: @mkarpman you are right, if B_RMR has the same allocation as P_RMR, normalizing LPD at the space level works. I'll update the rule.

docs/section6/Rule6-13.md
xingcx: Updated.

docs/section6/Rule6-13.md
xingcx: Control credit adjustment is now moved to normalize_space_schedules() function.

docs/section6/Rule6-13.md
xingcx: Removed.

docs/section6/Rule6-13.md
```diff
@@ -38,41 +34,26 @@
...
+        - Compare lighting schedules in P_RMR and B_RMR: `schedule_comparison_result = compare_schedules(normalized_schedule_b, normalized_schedule_b, building_open_schedule_p, adjusted_reduction_factor_p)`  
 
-          - Get lighting occupancy sensor reduction factor from Table G3.7: `reduction_factor_p = data_lookup(table_G3_7, lighting_space_type_p, lighting_control_type_p)`  
+          **Rule Assertion:**
 
-          - Calculate adjusted lighting occupancy sensor reduction factor based on B_RMR: `adjusted_reduction_factor_p = ( 1 - reduction_factor_p ) / ( 1 - reduction_factor_b )`  
+          - Case 1: For all hours, for each lighting, if lighting schedule in P_RMR is equal to lighting schedule in B_RMR times adjusted lighting occupancy sensor reduction factor: `if schedule_comparison_result == "MATCH": PASS`  
 
-          - Get lighting schedule: `lighting_schedule_p = lighting_p.lighting_multiplier_schedule`  
+          - Case 2: Else if lighting schedule in P_RMR is lower than or equal to lighting schedule in B_RMR times adjusted lighting occupancy sensor reduction factor: `if schedule_comparison_result == "EQUAL AND LESS": CAUTION`
```
mkarpman, line 48: I think this should be a FAIL, because P-RMR schedule fractions is less than B-RMR schedule fraction times adjustment. Or am I missing what this is doing?

mkarpman (changes requested):

docs/section6/Rule6-13.md
mkarpman: Why is this not "FAIL" as discussed below?

docs/section6/Rule6-13.md
```diff
@@ -0,0 +1,59 @@
...
+        - Compare lighting schedules in P_RMR and B_RMR: `schedule_comparison_result = compare_schedules(normalized_schedule_b, normalized_schedule_b, building_open_schedule_p, adjusted_reduction_factor_p)`  
+
+          **Rule Assertion:**
+
+          - Case 1: For all hours, for each lighting, if lighting schedule in P_RMR is equal to lighting schedule in B_RMR times adjusted lighting occupancy sensor reduction factor: `if schedule_comparison_result == "MATCH": PASS`  
+
+          - Case 2: Else if lighting schedule in P_RMR is lower than or equal to lighting schedule in B_RMR times adjusted lighting occupancy sensor reduction factor: `if schedule_comparison_result == "EQUAL AND LESS": CAUTION`  
+
+          - Case 3: Else, lighting schedule in P_RMR is higher than lighting schedule in B_RMR times adjusted lighting occupancy sensor reduction factor: `if schedule_comparison_result == "EQUAL AND MORE": CAUTION`  
+
```
mkarpman, line 50: Change outcome to "UNDERTERMINED"

mkarpman (changes requested):

docs/section6/Rule6-13.md
mkarpman: It still shows as "CAUTION" for me...

xingcx: @mkarpman Case 2 was updated to FAIL. Not sure why you still see CAUTION.

docs/section6/Rule6-13.md
xingcx: Updated with message also.

mkarpman (approved):

supriyagoel (approved):

PR #88, opened 2021-07-09
Lighting test JSON update

jugonzal07:
Made corrections for section, rule, and test ID values in the lighting test JSON


jugonzal07:
@charliepnnl I have merged in the updated ruletest engine code into this branch along with the lighting JSON changes. I think you should be good to continue your Rule development.

charliepnnl (approved):

PR #91, opened 2021-07-21
Create Rule6-5.md

xingcx:


docs/section6/Rule6-5.md
```diff
@@ -0,0 +1,30 @@
...
+    - For each zone in thermal block: ```zone_p in thermal_block_p.zones:```  
+
+      - For each space in zone: ```space_p in zone_p.spaces:```  
+
+        - Get interior lighting in U_RMR: ```interior_lighting_b = match_data_element(U_RMR, InteriorLightings, space_p.interior_lighting.id)```
+
+          **Rule Assertion:** 
+
+          - Case 1: Lighting power in the proposed RMR is as designed and cannot be verified by RCT: ```if space_p.interior_lighting == interior_lighting_b: CAUTION```
+
```
mkarpman, line 29: The RDS looks good to me except I am struggling with "CAUTION" outcome. RCT will be checking numerous aspects of P-RMR to confirm that they are equal to the corresponding aspects of U-RMR, so will generate numerous "CAUTIONS" for each project. For example, on a project with 100 spaces that all have the same LPD in the U-RMR and P-RMR (the best case outcome), RCT will generate 100 "CAUTION" flags. I think this will result in these flags being ignored by human reviewers working for AHJs / RAs. I recommend making it "PASS", and providing an umbrella statement in the RCT report to verify that LPD modeled in the proposed design reflects design documents.

supriyagoel: @mkarpman the RCT will generate two reports, one high level report which summarizes all rules that were pass/fail/caution and a second more detailed report where details for each space would be shown. At the high level report, this will show up as just one fail/caution. 
If we decide that 6-5 should be caution, then I think that 6-4 should be caution as well.

mkarpman: @supriyagoel we have RCT project checking report described in Section 7.4.3. There is not much in there currently, perhaps we should propose updates for this section and vet it with the committee? The section talks about RCT reporting schema. Do we need to be developing it also?

mcneillj:
Archiving. RDS is no longer active and PR is being closed.

PR #92, opened 2021-07-21
Create Rule6-6.md

xingcx:


docs/section6/Rule6-6.md
```diff
@@ -0,0 +1,32 @@
+
+# Lighting - Rule 6-6
+
+**Rule ID:** 6-6  
+**Rule Description:** Where lighting neither exists nor is submitted with design documents, lighting shall comply with but not exceed the requirements of Section 9. Lighting power shall be determined in accordance with the Building Area Method (Section 9.5.1)  
+**Appendix G Section:** Section G3.1-6(c) Modeling Requirements for the Proposed design  
+**Appendix G Section Reference:**  
+
+- Table G3.8, Lighting Power Density Allowance Using the Building Area Method  
+
```
mkarpman, line 10: The LPD should be based on Table 9.5.1, not G3.8

docs/section6/Rule6-6.md
```diff
@@ -0,0 +1,32 @@
...
+**Applicability:** All required data elements exist for P_RMR  
+**Applicability Checks:** None  
+**Manual Check:** None  
+**Evaluation Context:** Each Data Element  
+**Data Lookup:** Table G3.8
```
mkarpman, line 15: Should be Table 9.5.1 per above

docs/section6/Rule6-6.md
```diff
@@ -0,0 +1,32 @@
...
+## Rule Logic: 
+
+- For each building_segment in the Proposed Model: ```building_segment_p in P_RMR.building.building_segments```  
+
+  - Determine if the building segment uses Building Area Method: ```if building_segment_p.lighting_building_area_type != "None":```  
+
```
mkarpman, line 21: This rule is for projects where "lighting neither exists nor provided with design documents". I don't think inclusion of lighting_building_area_type into the RMR has anything to do with it. I think that in order to implement this check, we need to add is_lighting_specified element to Space object. @supriyagoel  and @JasonGlazer - do you agree? Once this is added, the RDS would check that for any space that has is_lighting_specified = False the LPD is equal to the value in Table 9.5.1 for building_segment_p.lighting_building_area_type

docs/section6/Rule6-6.md
```diff
@@ -0,0 +1,32 @@
...
+    - Get lighting power allowance: ```lighting_allowance_p = data_lookup(table_G3_8, building_segment_p.lighting_building_area_type)```  
+
```
mkarpman, line 23: Table 9.5.1

mkarpman (changes requested):

docs/section6/Rule6-6.md
supriyagoel: i think this is similar to the the discussion we had regarding 'is_lighting_exisiting' - as previously discussed, this is unlikely to be an input BPM tools will have. 
I think we simply need to check that U-RMR = P-RMR and write out a caution message for all Proposed building scenarios (existing, lighting designed, lighting not designed)
The RCT could also check IF P-RMR lighting = 9.5.1 and state that lighting is = 9.5.1. or not equal to 9.5.1. to facilitate manaul review.

docs/section6/Rule6-6.md
```diff
@@ -37,4 +37,4 @@
 
           - Case 1: If lighting building area type is not specified: ```if NOT building_segment_p.lighting_building_area_type: FAIL```  
 
-          - Case 2: If lighting power does not equal to the value prescribed in Table 9.5.1 or lighting power in P_RMR is not the same as U_RMR: ```if ( ( interior_lighting_p.power_per_area == lighting_allowance_p ) OR ( interior_lighting_p == interior_lighting_u ) ): FAIL```  
+          - Case 2: If lighting power does not equal to the value prescribed in Table 9.5.1 or lighting power in P_RMR is not the same as U_RMR: ```if ( ( interior_lighting_p.power_per_area != lighting_allowance_p ) OR ( interior_lighting_p != interior_lighting_u ) ): FAIL```
```
mkarpman, line 40: I think this RDS has the same issue as 6-4. Unless I am missing something, I don't see how RCT can tell that the lighting "neither exists nor is submitted with design documents". I don't think specifying or not specifying lighting_building_area_type means that. Also, on some projects lighting may be designed for some tenant spaces but not others, so we can't use the global building_segment_p.lighting_building_area_type attribute to determine where lighting is and isn't specified.  See my comment for 6-4 regarding the proposed solution.

JasonGlazer (approved):

docs/section6/Rule6-6.md
supriyagoel: this rule needs to be updated to use the new lighting function @xingcx has added to identify if lighting in proposed as as designed, as exisiting or not designed

supriyagoel (changes requested):

mcneillj:
Archiving. RDS has been merged into Rule 6-4 and PR is being closed.

PR #94, opened 2021-07-22
Create Rule5-6.md

xingcx:


docs/section5/Rule5-6.md
```diff
@@ -0,0 +1,32 @@
...
+
+      - For each surface in zone: ```for surface_p in zone_p.surfaces:```  
+
+        - Get matching surface from U-RMR: ```surface_u = match_data_element(U_RMR, surfaces, surface_p.id)```  
+
+          **Rule Assertion:**  
+
+          Case 1: Surface U-factor in P_RMR matches U_RMR: ```if surface_p.construction.u_factor == surface_u.construction.u_factor: PASS```  
+
+          Case 2: Else: ```else: FAIL```
```
mkarpman, line 32: @JasonGlazer and @supriyagoel, App G requires that "All components of the building envelope in the proposed design shall be modeled as shown on architectural drawings or as built for existing building envelopes." We currently have rules that verify alignment of the U/C/F-factors (as appropriate for the surface type) between U-RMR and P-RMR, and also a rule that verify alignment in area.  I think it's alright to stop there, but want to confirm that you agree as it affect multiple RDSs. 

Alternatively, we may go deeper and verify alignment of other relevant attributes of the Surface (e.g., tilt, azimuth, surface_optical_properties) and Construction (e.g. fraction_framing, primary_layers, r-_value etc.). This also affects the RDS that we discussed yesterday, verifying that unregulated surfaces are modeled the same in the baseline as in the proposed design, and is related to the general approach for verifying "sameness" of objects. Perhaps we should limit the sameness check to just the attributes that are required?

supriyagoel: @mkarpman i agree that this rule should only verify U/F/C. if we want RCT to verify other attributes like tilt, azimuth, fraction framing etc. then we should write a separate RDS for that. it would be one rule that verifies all other envelope attributes and similarly we could have another rule for all lighting attributes etc. which are not checked by any other rule. hope this sounds good. similar discussion on 5-17 
image

supriyagoel: discussed with @mkarpman the path outlined above is good

mkarpman (approved):

docs/section5/Rule5-6.md
xingcx: For this rule do we need to change it to just check U-factor for regulated roof, floor and above-grade walls? It seems 5-9 covers below-grade wall, 5-16 covers slabs and 5-17 covers unregulated surfaces. Right now I check U-factor for all surfaces. @mkarpman @supriyagoel

supriyagoel: @xingcx because this is P-RMR = u-RMR check, we should be checking all surfaces (i.e. regulated and unregulated). I think the way you have it is fine. For all surfaces that have a U factor defined, we can verify that they're the same between P and U RMRs. 5-9 would verify the C factor for below grade walls and 5-16 would verify the F factors for slab on grade floors.

supriyagoel: question is should these rules apply to both reg and unreg surfaces : YES
Question 2: should we check U/F/C for all surfaces or particular surface type? Recommendation to check U for roof/wall. slab on grade only F and below grade walls only check C.
MK will also update the TCDs (5-6 only) to say U factor of above grade surfaces are modeled as designed.
the description also needs to be updated to say 'above grade opaque surfaces'

docs/section5/Rule5-6.md
```diff
@@ -0,0 +1,32 @@
+
+# Envelope - Rule 5-6  
+
+**Rule ID:** 5-6  
+**Rule Description:** Building envelope U-factors must be modeled in proposed design as designed.  
+**Appendix G Section:** Section G3.1-5(a) Building Envelope Modeling Requirements for the Proposed building
```
supriyagoel, line 6: clarify that this is  'above grade opaque surfaces'

mkarpman (approved):

PR #95, opened 2021-07-22
Create Rule5-17.md

xingcx:


docs/section5/Rule5-17.md
```diff
@@ -0,0 +1,36 @@
...
+      - For each surface in zone: ```for surface_b in zone_b.surfaces:```  
+
+        - Check if surface is unregulated: ```if ( scc_dictionary_b[surface_b.id] == UNREGULATED ):```  
+
+          - Get matching surface from P_RMR: ```surface_p = match_data_element(P_RMR, surfaces, surface_b.id)```  
+
+            **Rule Assertion:**  
+
+            Case 1: Surface construction in B_RMR matches P_RMR: ```if surface_b.construction == surface_p.construction: PASS```  
+
```
mkarpman, line 35: Construction data object has multiple attributes. I think we can just check equivalency of U/C/F factors, surface_construction_input_option, and perhaps has_radiant_heating and has_radiant_cooling. @supriyagoel and @JasonGlazer - do you agree, or do you think we need to do a deep check covering all attributes? We don't need "Notes" field to match for example.

JasonGlazer: I thought we had discussed some general approach for ensuring that everything not modified by rules was the same.

mkarpman: I remember that also but I don't know what was decided. I think in order to minimize inconsequential RCT flags we may want to limit the sameness check to just the RMR attributes that are mandatory for App G. @supriyagoel should we check with James?

supriyagoel: @mkarpman I agree we only need to check u,f,c.
i dont think we need to check has radiant cooling/has radiant heating.
I dont remember a discussion on this topic but adding @mcneillj to confirm

mkarpman: @supriyagoel I wonder why you think we shouldn't check "has radiant heating/cooling". I don't have strong feelings, but I think we should come up with logic for what to check. Obviously, we want the "regulated" inputs to align (U/C/F factors and modeling method for surfaces, LPDs for lighting, etc.) Should we skip all other data points that may be included in the RMR?

mkarpman: As far as I understand, the RDS is currently checking for the full alignement of the surface.construction object between P-RMR and U-RMR for unregulated surfaces. Didn't we decide that we only check the regulated aspects, i.e. U/C/F-factors and optical properties for the roofs? I don't have strong feelings either way.

docs/section5/Rule5-17.md
```diff
@@ -0,0 +1,48 @@
...
+        - Check if surface is unregulated: `if ( scc_dictionary_b[surface_b.id] == UNREGULATED ):`  
+
+          - Get surface type: `surface_type_b = get_opaque_surface_type(surface_b)`
+
+          - Get matching surface from P_RMR: `surface_p = match_data_element(P_RMR, surfaces, surface_b.id)`  
+
+            **Rule Assertion:**  
+
+            - Case 1: If surface type is roof, floor or above-grade wall, and surface construction U-factor in B_RMR matches P_RMR: `if ( surface_type_b in ["ROOF", "FLOOR", "ABOVE-GRADE WALL"] ) AND ( surface_b.construction.u_factor == surface_p.construction.u_factor ): PASS`
+
```
supriyagoel, line 42: @xingcx this looks good. the only additional test we need for each case is the modeling method, which should be set to layer by layer input in both B and P RMR.

mkarpman: I don't think App G requires that for unregulated surfaces, so I think this may be handled as part of general equivalency check that the same modelnig approach is used in P-RMR and U-RMR.

mkarpman (approved):

docs/section5/Rule5-17.md
supriyagoel: @xingcx I agree, no changes required.

supriyagoel (approved):

PR #98, opened 2021-07-30
Create Rule6-9.md

xingcx:


docs/section6/Rule6-9.md
```diff
@@ -0,0 +1,32 @@
...
+## Rule Logic: 
+
+- For each building_segment in the Baseline model: ```for building_segment_b in B_RMR.building.building_segments:```  
+  - For each thermal_block in building segment: ```for thermal_block_b in building_segment.thermal_blocks:```  
+    - For each zone in thermal block: ```for zone_b in thermal_block_b.zones:```  
+      - For each space in zone: ```for space_b in zone.spaces:```  
+        - Add space floor area to building total floor area: ```building_total_area_b += space_b.floor_area```  
+- **Applicability Check 1:**```if building_total_area_b > 5000:```  
+
+- **Manual Check 1:** Manual review is required where building total area is more than 5,000sq.ft. to verify automatic shutoff controls are modeled correctly in the Baseline.
```
mkarpman, line 32: Since automatic lighting controls are also required in the proposed design, I think this rule has to confirm that during unoccupied hours the same lighting schedules are modeled in B-RMR and P-RMR. In addition, The hourly fractions used in these schedules should be 0 or a low number because only the emergency lighting would stay on. I'll discuss it with others and copy you.

from 90.1 2004:
9.4.1.1 Automatic Lighting Shutoff. Interior lighting
in buildings larger than 5000 ft2 shall be controlled with an
automatic control device to shut off building lighting in all
spaces.

supriyagoel: @mkarpman and @xingcx - I agree with this assessment. The RDS would review the occupancy schedule and the lighting schedule. for times where occupancy is <=10%, the lighting schedule should be very low, only accounting for emergency lighting. This RDS should verify 2 things (1) lighting schedule is very low (0 or close to 0) when occupancy schedule is <=10% in the B-RMR (2) P-RMR = B-RMR during unoccupied hrs (when occupancy is <=10%)

mkarpman: I think the trigger should be the schedule in building.building_open_schedule, not space.occupant_multiplier_schedule, because based on the quote from 90.1 2004 above the rule is for when the entire building is unoccupied. Agreed?

supriyagoel: @mkarpman - we couldn't hear you on the call. The approach you have summarized here looks good. we can use the building.building_open_schedule to identify unoccupied times (this would be when schedule =0)
during those times, we will verify that the lighting scheduled in B=P for each space.
This would need to use the lighting function which calculates a W/ft2 weighted average lighting schedule for each space.

docs/section6/Rule6-9.md
```diff
@@ -11,22 +11,62 @@
...
+
+      - For each space in zone: `space_b in zone_b.spaces:`  
+
+        - Get normalized space lighting schedule: `normalized_schedule_b = normalize_space_lighting_schedule(space_b)`  
+
+        - Get matching space in P_RMR: `space_p = match_data_element(P_RMR, Spaces, space_b.id)`  
+
+          - Get normalized space lighting schedule in P_RMR: `normalized_schedule_p = normalize_space_lighting_schedule(space_p)`
+
+        - Check if automatic shutoff control is modeled in space during unoccupied hours (i.e. if lighting schedule hourly value in B_RMR is equal to P_RMR during unoccupied hours): `schedule_comparison_result = compare_schedules(normalized_schedule_b, normalized_schedule_p, building_open_schedule_b, none)`
```
supriyagoel, line 54: we should clarify we mean by 'unoccupied hrs'. Hrs where building open schedule = 0 would be 'unoccupied hrs'.
we should also probably rename 'unoccupied' to 'open' or 'closed'. Occupancy schedule is differant from the building open schedule, so this should probably say 'check if automatic shutoff controls modeled during building closed hrs'

docs/section6/Rule6-9.md
```diff
@@ -11,22 +11,62 @@
...
+
+
+**Temporary Function note:**
+
+`compare_schedule_result = compare_schedules(Schedule 1, Schedule 2, Mask Schedule, reduction_factor)`
+
+(4 inputs, Schedule 1, Schedule 2, Mask Schedule, comparison factor)
+
+- Schedule 2 as the comparison basis, i.e. Schedule 1 = Schedule 2 * comparison factor
+- When Mask Schedule hourly value is 0, schedules need to be the same at that hour. If Mask Schedule hourly value is 1, Schedule 1 needs to be comparison factor times Schedule 2 at that hour. If Mask Schedule hourly value is 2, skip comparison.
```
supriyagoel, line 70: I'm not sure what 'Mask Schedule' means. Should this be 'Open Schedule'?

supriyagoel (changes requested):

docs/section6/Rule6-9.md
```diff
@@ -0,0 +1,72 @@
...
+    - For each zone in thermal block: `for zone_b in thermal_block_b.zones:`  
+
+      - For each space in zone: `for space_b in zone.spaces:`  
+
+        - Add space floor area to building total floor area: `building_total_area_b += space_b.floor_area`  
+
+- **Applicability Check 1:**`if building_total_area_b > 5000:`  
+
+- Get building open schedule in the baseline model: `building_open_schedule_b = B_RMR.building.building_open_schedule`  
+
```
mkarpman, line 39: I think we discussed adding BuildingSegment.open_schedule, because I believe lighting shut-off controls must be included for each occupancy type and not building as a whole.

docs/section6/Rule6-9.md
```diff
@@ -0,0 +1,72 @@
...
+  1. Building total area is more than 5,000sq.ft.  
+
+**Manual Check:** Yes  
+
+**Function Call:**  
+
+  - compare_schedules()
+  - normalize_space_lighting_schedule()
+
```
mkarpman, line 20: Was this function developed?

docs/section6/Rule6-9.md
mkarpman: Sounds like comparison factor should be a number instead of "none" for . E.g., some flag like - 101.

docs/section6/Rule6-9.md
mkarpman: @supriyagoel , the "mask schedule" for this RDS is building_open_schedule, as show in line 54. I think Mask Schedule is a generic name that will be used in the function so that it can apply to different scenarios.

docs/section6/Rule6-9.md
supriyagoel: @mkarpman not ye. Xing will work on it this week.

docs/section6/Rule6-9.md
supriyagoel: yes, this is a placeholder till that is discussed by the schema WG and added to the schema.
@xingcx maybe you could include a note towards the end on future revisions. This RDS should reference buildingsegment.openschedule once that is discussed, approved and added to the schema

docs/section6/Rule6-9.md
xingcx: Right, in this case we will not be using the comparison factor. The mask schedule needs to be 2 for all building open hours to skip the comparison. I'll put -111 for now and leave it up to the actual code.

docs/section6/Rule6-9.md
xingcx: Note added.

docs/section6/Rule6-9.md
xingcx: "Unoccupied hours" is now changed to "building closed hours".

docs/section6/Rule6-9.md
xingcx: Right, this is just trying to make the function more generic. We can adjust this later to suit our needs in other RDS as well.

mkarpman (approved):

PR #102, opened 2021-07-31
Create Rule5-15.md

xingcx:


docs/section5/Rule5-15.md
```diff
@@ -0,0 +1,57 @@
...
+              - Else: ```manual_review_flag = TRUE```  
+
+          **Rule Assertion:**  
+
+          Case 1: If zone has both residential and non-residential spaces and the construction requirements for slab-on-grade floor are different, request manual review: ```if manual_review_flag == TRUE: RAISE_WARNING```  
+
+          Case 2: Else if slab-on-grade floor F-factor matches Table G3.4: ```else if surface_construction_b.f_factor == target_f_factor: PASS```  
+
+          Case 3: Else: ```else: FAIL```  
+
```
mkarpman, line 56: Based on Section 3 definition, "slab-on-grade floor: that portion of a slab floor of the building envelope that is in contact
with the ground and that is either above grade or is less than or equal to 24 in. below the final elevation of the nearest exterior grade.". So if floor is more than 24" below grade it will be unregulated, but there is not way for RCT to determine that. So perhaps Case 3 outcome should be caution instead of fail?

supriyagoel (approved):

docs/section5/Rule5-15.md
xingcx: Modified.

docs/section5/Rule5-15.md
```diff
@@ -52,6 +52,6 @@
```
mkarpman, line 52: I don't think there should be a warning here?

docs/section5/Rule5-15.md
```diff
@@ -52,6 +52,6 @@
 
           Case 2: Else if slab-on-grade floor F-factor matches Table G3.4: ```else if surface_construction_b.f_factor == target_f_factor: PASS```  
 
-          Case 3: Else: ```else: FAIL```  
+          Case 3: Else: ```else: RAISE_WARNING: "SLABS BELOW 24" BELOW GRADE IS NOT REGULATED."```
```
mkarpman, line 56: I would change the warning message as follows: "Baseline slab F-factor is not as expected for slabs that are less than 24" below grade. Verify that the slab is more than 24" below grade and is unregulated." Or maybe do FAIL and include this message saying that it can actually be OK?

xingcx: Updated.

mkarpman: Following up our discussion during the bi-weekly calls, I suggest changing the outcome to FAIL and keepting the accompanying message.

mkarpman (changes requested):

supriyagoel:
the outcome for a rule should be PASS, FAIL or CAUTION. Each outcome can have a 'Message'. 
Some FAIL outcomes could clarify instances where the value might be correct (for example, if the building is an existing building etc.)

docs/section5/Rule5-15.md
```diff
@@ -0,0 +1,57 @@
...
+              - Else: `manual_review_flag = TRUE`  
+
+          **Rule Assertion:**  
+
+          Case 1: If zone has both residential and non-residential spaces and the construction requirements for slab-on-grade floor are different, request manual review: `if manual_review_flag == TRUE: CAUTION and raise_warning "ZONE HAS BOTH RESIDENTIAL AND NON-RESIDENTIAL SPACES AND THE CONSTRUCTION REQUIREMENTS FOR SLAB-ON-GRADE FLOOR ARE DIFFERENT. VERIFY CONSTRUCTION IS MODELED CORRECTLY."`  
+
+          Case 2: Else if slab-on-grade floor F-factor matches Table G3.4: `else if surface_construction_b.f_factor == target_f_factor: PASS`  
+
+          Case 3: Else: `else: FAIL and raise_warning: "BASELINE SLAB F-FACTOR IS NOT AS EXPECTED FOR SLABS THAT ARE LESS THAN 24" BELOW GRADE. VERIFY THAT THE SLAB IS MORE THAN 24" BELOW GRADE AND IS UNREGULATED."`  
+
```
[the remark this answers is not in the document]
mkarpman: Agreed.

mkarpman (approved):

PR #105, opened 2021-08-09
Update Rule6-2

xingcx:


docs/section6/Rule6-2.md
```diff
@@ -14,49 +14,47 @@
...
-- For each zone in thermal_block: ```zone_proposed in thermal_block_proposed.zones:```  
+  - For each thermal block in building segment: `thermal_block_p in building_segment_p.thermal_blocks:`  
+  
+    - For each zone in thermal block: `zone_p in thermal_block_p.zones:`  
 
-- For each space in zone: ```space_proposed in zone_proposed.spaces:```  
+      - For each space in zone: `space_p in zone_p.spaces:`  
 
-  - Get floor_area from space: ```floor_area_proposed = space_proposed.floor_area```  
+        - Add space lighting power to building segment total: `design_lighting_wattage += space_p.interior_lighting.power_per_area * space_p.floor_area`
```
supriyagoel, line 32: @xingcx if I understand correctly, i think this adds space lighting tower to the building segment lighting power?
If yes, I'm not sure if that is correct. i think we would calculate the lighting power allowance for each space based on G3.7 and the remaining area of the building segment based on G3.8. @mkarpman please confirm.

docs/section6/Rule6-2.md
```diff
@@ -14,49 +14,47 @@
...
+      - For each space in zone: `space_p in zone_p.spaces:`  
 
-  - Get floor_area from space: ```floor_area_proposed = space_proposed.floor_area```  
+        - Add space lighting power to building segment total: `design_lighting_wattage += space_p.interior_lighting.power_per_area * space_p.floor_area`  
 
-  - If bam_flag is TRUE then add floor_area to the total building segment floor area: ```total_building_segment_area_proposed += floor_area_proposed```  
+        - If building segment specifies lighting building area type , add space floor area to the total building segment floor area: `if allowable_LPD_BAM: total_building_segment_area_p += space_p.floor_area`  
 
-    - Get the allowable lighting power density for this building area type: ```if lighting_building_area_type_proposed in table_G3_8: allowable_LPD_proposed = data_lookup(table_G3_8, lighting_building_area_type_proposed) else raise_warning```  
+        - Check if any space does not specify lighting space type, flag for Building Area Method: `if NOT space_p.lighting_space_type in table_G3_7: BAM_flag = TRUE`
```
mkarpman, line 36: I think instead of BAM_flag=TRUE it should be SBS_flag=FALSE. Because missing space types means that we can't use space-by-space method. It doesn't automatically mean that we can use BAM because BAM area type may not be provided either.

docs/section6/Rule6-2.md
```diff
@@ -14,49 +14,47 @@
...
+          - Add to the total allowable lighting wattage for the building segment using Space-by-Space method: `allowable_lighting_wattage_SBS += allowable_LPD_space * space_p.floor_area`  
 
-    - If lighting_space_type_proposed exists in Table G3.7 then get the allowable lighting power density for this space type: ```if lighting_space_type_proposed in table_G3_7: allowable_LPD_proposed = data_lookup(table_G3_7, lighting_space_type_proposed) else raise_warning```  
+**Rule Assertion:**
 
-    - Calculate the total allowable lighting wattage for the building segment: ```building_segment_allowable_lighting_wattage_proposed += allowable_LPD_proposed * floor_area_proposed```  
+- Case 1: For each building segment, if both lighting building area type and lighting space type in all spaces are specified, and the total lighting power in P_RMR is less than or equal to the higher of the Building Area Method and Space-by-Space Method allowances: `if ( allowable_LPD_BAM ) and ( NOT BAM_flag ) and ( design_lighting_wattage <= max(allowable_LPD_BAM * total_building_segment_area_p, allowable_lighting_wattage_SBS) ): PASS`
```
mkarpman, line 44: I think we discussed the following logic:
a. Calculate allowable_LPD_BAM (per row 23 above); determine the total area of the 
b. Calculate the allowable wattage with space-by-space method in Table G3-7)this should say allowable_wattage_SAM=+= space_p.interior_lighting.power_per_area * space_p.floor_area. If lighting space types are not provided for ALL spaces, there should be a flag for that and only (a) would be checked.
c. Calculate the total specified lighting. Project passes if it's less than the MAX(a,b).

So I think the assertions should be as follows: 
Case 1: enough info is specified for either building area or space by space method, and proposed lighting is less than the maximum of the two, then pass. 
IF segment_designed_lighting_wattage!>MAX(allowable_lighting_wattage_SBS; allowable lighting_wattage_BAM) AND (BAM_flag OR SBS_flag):PASS

Case 2: enough info is specified for building area method but not space by space method, and specified lighting exceeds BAM allowance, then FAIL and show message that specified lighting exceeds allowance in Table G3.8, and not enough information is provided to establish alternative compliance via Table G3.7.

Case 3: enough info is specified for SBS method but not BAM, and specified lighting exceeds SBS allowance, then FAIL and show message that specified lighting exceeds allowance in Table G3.7, and not enough information is provided to establish alternative compliance via Table G3.8.

Case 4: BAM_flag=False AND SBS_flag=FALSE then FAIL and show message that not enough information is provided to establish compliance of the proposed lighting via either SBS (Tablle G3.7) or BAM (Table G3.8).

supriyagoel: @mkarpman i believe this is what is being done. Case 2 in your comment is Case 4 in the RDS and case 3 in your comment is case 6 in the RDS. Case 4 in your comment is case 7 in the RDS. Does anything need to change?

docs/section6/Rule6-2.md
xingcx: I will change the variable name to building_segment_design_lighting_wattage to avoid confusion. 
Right now this rule calculates:
1). building segment total designed lighting power from all spaces (Line 31, I'll change it to cover multiple lighting objects in space)
2). Table 3.7 allowance, based on each space's area and space type. If any space does not specify space.lighting_space_type, Table 3.7 path will not be used. 
That is I'm assuming we are not comparing Table 3.7 total allowance for spaces with space.lighting_space_type specified and comparing Table 3.8 total allowance for spaces with no space.lighting_space_type specified.
@supriyagoel was this your question? 
3). Table 3.8 allowance, based on building_segment total space area and building_segment.lighting_building_area_type.

docs/section6/Rule6-2.md
xingcx: @mkarpman it seems what you summarized is what's in the RDS.
Case 1 in your comment is Case 1 and 2 in the RDS, for when building_segment.lighting_building_area_type and all space.lighting_space_types are specified.
Case 2 in your comment is Case 5 and 6 in the RDS, for when building_segment.lighting_building_area_type is specified but not all space.lighting_space_types are specified.
Case 3 in your comment is Case 3 and 4 in the RDS, for when building_segment.lighting_building_area_type is not specified but all space.lighting_space_types are specified.
Case 4 in your comment is Case 7 in the RDS, for when building_segment.lighting_building_area_type is not specified and not all space.lighting_space_types are specified.
I split each each of yours into 2 to cover PASS and FAIL scenarios, with the exception of Case 4, which will be FAIL as the allowance could not be established.

docs/section6/Rule6-2.md
xingcx: This flag is just to mark space-by-space method should not be used and does not mean the rule will use BAM method (as the rule assertion below uses both the BAM_flag and the allowable_LPD_BAM flag to determine BAM method). I would prefer to have a TRUE flag instead of FALSE. I can change to name to avoid confusion.

docs/section6/Rule6-2.md
mkarpman: Agreed, except for #3 Table 3.8 path will not be used if building_segment.lighting_building_area_type is not specified. You are already doing it this way, so this is just to complete this summary of the logic.

docs/section6/Rule6-2.md
mkarpman: Changing the name would work, but then would it be No_SBS_flag=TRUE? Ultimately the names don't matter, but if they are not intuitive the logic may be screwed up during the implementation.

mkarpman (approved):

supriyagoel (approved):

PR #106, opened 2021-08-09
Create get_lighting_status_type.md

xingcx:


docs/ruleset_functions/get_lighting_status_type.md
```diff
@@ -0,0 +1,31 @@
...
+
+Inputs:
+  - **building_segment**: The building segment that needs to determine lighting status type.  
+
+Returns:
+- **space_lighting_status_type**: The Lighting Status Type Category [not-yet designed, as-designed or as-existing].    
+
+Data Lookup:
+- Table 9.5.1  
+
```
supriyagoel, line 14: We should add a clarification somewhere that this function ONLY applies to the proposed building.

supriyagoel (approved):

supriyagoel:
@xingcx this looks good to me.

docs/ruleset_functions/get_lighting_status_type.md
mkarpman: Agreed. One hole in our logic here is when lighting is existing or fully designed, but the LPD by chance is exactly the same as BAM LPD. Perhaps capture this scenario in the error message that the calling RDS would show.

mkarpman (approved):

docs/ruleset_functions/get_lighting_status_type.md
```diff
@@ -0,0 +1,31 @@
...
+- Get lighting power density allowance for building segment: `allowable_LPD = data_lookup(table_9_5_1, building_segment.lighting_building_area_type)`  
+
+- For each thermal block in building segment: `thermal_block in building_segment.thermal_blocks:`  
+
+  - For each zone in thermal block: `zone in thermal_block.zones:`  
+
+    - For each space in zone: `space in zone.spaces:`  
+
+      - If space lighting power density equals to allowance, save space interior lighting as not-yet designed: `if space.interior_lighting.power_per_area == allowable_LPD: space_lighting_status_type_dict[space.id] = "NOT-YET DESIGNED"`  
+
```
mkarpman, line 28: Need to account for possibility of multiple interior lighting objects within a space. So need to use  SUM(interior_lighting.power_per_area) in a given space for comparision.

mkarpman (changes requested):

docs/ruleset_functions/get_lighting_status_type.md
xingcx: Added this to the returned values.

docs/ruleset_functions/get_lighting_status_type.md
xingcx: Modified to account for multiple interior lighting objects.

mkarpman (approved):

PR #107, opened 2021-08-10
Create Rule6-10.md

xingcx:
Questions:
1). Is occupancy sensor control modeled through schedule adjustment only? e.g. not through LPD as Rule 6-8 requires baseline LPD matches Table G3.7.
2). Do we need to check the schedule values or just check if InteriorLighting.occupancy_control_type + InteriorLighting.are_schedules_used_for_modeling_occupancy_control are set up correctly?

### DIFF
--- a/docs/section6/Rule6-10.md
+++ b/docs/section6/Rule6-10.md
@@ -14,12 +14,15 @@
 **Function Call:**  
 
   - compare_schedules()
+  - normalize_space_lighting_schedule()
 
 ## Rule Logic: 
 
-- For each building_segment in the baseline model: `building_segment_b in B_RMR.building.building_segments:`  
+- Get building open schedule in the baseline model: `building_open_schedule_b = B_RMR.building.building_open_schedule`  
 
-  - For each thermal_block in building segment: `thermal_block_b in building_segment_b.thermal_blocks:`  
+- For each building segment in building: `building_segment_b in B_RMR.building.building_segments:`  
+
+  - For each thermal block in building segment: `thermal_block_b in building_segment_b.thermal_blocks:`  
 
     - For each zone in thermal block: `zone_b in thermal_block_b.zones:`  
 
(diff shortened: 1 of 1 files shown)


supriyagoel:
@xingcx as discussed yesterday, we will simply verify if the flag for occ control is true. If yes, PASS. if no, FAIL. In both cases, the lighting schedule function should be called. I am unclear on what the function will exactly do, but we can discuss that on the function RDS.
@mkarpman please weigh in.

docs/section6/Rule6-10.md
```diff
@@ -0,0 +1,65 @@
...
+
+          - Case 3: If space does not require occupancy sensor and interior lighting is modeled with full-auto on occupancy sensor control: `if ( NOT occ_sensor_flag ) and ( control_type_b == "FULL-AUTO ON" ): FAIL`  
+
+          - Case 4: If space does not require occupancy sensor and interior lighting is not modeled with full-auto on occupancy sensor control: `if ( NOT occ_sensor_flag ) and ( control_type_b != "FULL-AUTO ON" ): PASS and return schedule_comparison_result`  
+ 
+**Temporary Function note:**
+
+`compare_schedule_result = compare_schedules(lighting_schedule_p, normalized_schedule_b, building_open_schedule_p, reduction_factor_p)`
+
+(4 inputs, Schedule 1, Schedule 2, Mask Schedule, comparison factor)
```
supriyagoel, line 61: i think we need normalized schedule P and normalized schedule b. this would be the schedule normalized to the space level. I am not sure if we need the open schedule for this function. we could simply compare the normalized P schedule against the normalized b schedule. @mkarpman - any thoughts?

supriyagoel (changes requested):

docs/section6/Rule6-10.md
```diff
@@ -0,0 +1,65 @@
...
+        - Compare lighting schedule in B_RMR and P_RMR: `schedule_comparison_result = compare_schedules(normalized_schedule_b, normalized_schedule_p, building_open_schedule_b, 1)`  
+
+        - Get lighting space type: `lighting_space_type_b = space_b.lighting_space_type`  
+
+          - If lighting space types is employee lunch and break rooms, conference/meeting rooms, or classrooms (not including shop classrooms, laboratory classrooms, and preschool through 12th-grade classrooms), set occupancy sensor flag: `if lighting_space_type_b in ["LOUNGE/BREAKROOM", "CONFERENCE/MEETING/MULTIPURPOSE ROOM", "CLASSROOM/LECTURE HALL/TRAINING ROOM - PENITENTIARY", "CLASSROOM/LECTURE HALL/TRAINING ROOM - ALL OTHER"]: occ_sensor_flag = TRUE`  
+
+          **Rule Assertion:**  
+
+          - Case 1: If space requires occupancy sensor and interior lighting is modeled with full-auto on occupancy sensor control: `if ( occ_sensor_flag ) and ( control_type_b == "FULL-AUTO ON" ): PASS and return schedule_comparison_result`  
+
```
mkarpman, line 50: I think here we just need to establish that normalized schedules are the same for baseline and proposed 24/7. So I don't think we need to check for control_type. I would just check if normalized schedules are the same for the baseline and proposed, and have PASS if yes and FAIL if no.

We'll need to revisit once "schedule_compare" function is finalized so that we know what it returns.

docs/section6/Rule6-10.md
```diff
@@ -0,0 +1,65 @@
...
+          - Case 2: If space requires occupancy sensor and interior lighting is not modeled with full-auto on occupancy sensor control: `if ( occ_sensor_flag ) and ( control_type_b != "FULL-AUTO ON" ): FAIL`  
+
```
mkarpman, line 52: Per above, I think assertion should be strictly based on comparing the normalized schedules. Same comment for Caes 3 and Case 4 below.

docs/section6/Rule6-10.md
mkarpman: @supriyagoel what you propose would work in this case because the two schedules must align 24/7. But for other building types the new function would be necessary - e.g., because for buildings over 5,000 SF we'd need space schedules to differ by the fractions prescribed in Table G3.7 and based on the specified controls 24/7 for building less than 5000SF, and during unoccupied hours only for larger buildings.

mkarpman (changes requested):

docs/section6/Rule6-10.md
```diff
@@ -0,0 +1,65 @@
...
+# Lighting - Rule 6-10  
+
+**Rule ID:** 6-10  
+**Rule Description:** Baseline building is modeled with occupancy sensor controls in applicable space types, i.e. employee lunch and break rooms, conference/meeting rooms, and classrooms (not including shop classrooms, laboratory classrooms, and preschool through 12th-grade classrooms).  
+**Appendix G Section Reference:** Section G3.1-6 Modeling Requirements for the baseline building  
+
```
mkarpman, line 7: I think we can get rid of this rule because it's redundant to 6-13. 6-13 pulls occupancy sensor reductions from Table G3.7, and in that table occupancy sensor reduction (the third column) is set to none for these spaces.

mkarpman (changes requested):

docs/section6/Rule6-10.md
supriyagoel: @mkarpman i see your point, however, i strongly recommend keeping this as 2 separate rules mostly because it aligns with the way things are specific in App G. Also, the  outcome for 6-10 is specific for baseline and outcome for 6-13 would be specific to the proposed. I agree that it is redundant but it gives us alignment with the rules.

supriyagoel: as discussed, we will merge 6-10 and 6-13. It will verify that proposed models OC using G3.7. This will automatically check that the spaces required to ahve OC in the B-RMR have them (i.e. P RMR is same as B RMR for those space types) as G 3.7 is set to 'none' for those space types.

mcneillj:
Rule 6-10 is no longer needed.  This PR will be closed and the branch will be deleted.

PR #111, opened 2021-08-19
Create normalize_space_schedules.md

xingcx:


docs/ruleset_functions/normalize_space_schedules.md
```diff
@@ -0,0 +1,43 @@
...
+  - Get lighting power per area: `power_per_area = interior_lighting.power_per_area`
+
+  - Add lighting power per area to space total: `space_total_power_per_area += power_per_area`
+
+  - Get lighting schedule: `schedule = interior_lighting.lighting_multiplier_schedule`
+
+    - If schedule is hourly type, multiply hourly values with power per area and save to an 8760 hours array: `if schedule.schedule_sequence_type == "HOURLY": 8760_array.append(hourly_value * power_per_area for hourly_value in schedule.hourly_values)`
+
+    - Else, schedule is event type, convert schedule values to hourly: `else: converted_schedule_array = convert_schedule_to_hourly(schedule)`
+
```
mkarpman, line 34: I don't know what schedule_sequence_type=event means. Since we expect this function to be applicable to lighting and misc. equipment, should only "hourly" be allowed here? @supriyagoel if you agree perhaps this may be another general RCT check.

docs/ruleset_functions/normalize_space_schedules.md
```diff
@@ -0,0 +1,43 @@
...
+      - Multiply hourly values with power per area and save to an 8760 hours array: `8760_array.append(hourly_value * power_per_area for hourly_value in converted_schedule_array)`
+
```
mkarpman, line 36: I am a littly confused by the nomenclature. After this step we should be getting an 8760 array of hourly_use_per_area (kWh/SF) for each interior lighting object. Is that what this is goind? If so, perhaps change the name of the resulting array to hourly_use_per_area?

docs/ruleset_functions/normalize_space_schedules.md
```diff
@@ -0,0 +1,43 @@
...
+  - Add 8760 schedule array to space total: `space_total_8760_array = ( hourly_value_1 + hourly_value_2 for hourly_value_1, hourly_value_2 in zip(space_total_8760_array, 8760_array) )`
+
```
mkarpman, line 38: And here we should be summing up the hourly_use_per_area arrays for all interior_lighting objects in the space. Is that what it's doing?

docs/ruleset_functions/normalize_space_schedules.md
```diff
@@ -0,0 +1,43 @@
...
+- Divide space total 8760 hour array by space total power per area to get normalized schedule array: `space_normalized_schedule_array = (hourly_value / space_total_power_per_area for hourly_value in space_total_8760_array)`
+
```
mkarpman, line 40: Using the nomenclature suggested above, it would be hourly_use_per_area/pace_total_power_per_area.

docs/ruleset_functions/normalize_space_schedules.md
xingcx: It seems a schedule can be in either hourly or event format. I'm not sure if this event format is applicable to all type of schedules or just some special types.

docs/ruleset_functions/normalize_space_schedules.md
xingcx: Updated.

docs/ruleset_functions/normalize_space_schedules.md
mkarpman: Could you help me understand what event format means? Thanks!

docs/ruleset_functions/normalize_space_schedules.md
xingcx: That is right. Line 37 adds hourly values from both schedules. Sorry for the python syntax, couldn't think of a more generic way to write this.

docs/ruleset_functions/normalize_space_schedules.md
xingcx: It seems an event type schedule has two fields, event_time (when the schedule value changes) and event_values (schedule value). So instead of describing all 8760 hour values, it just reports a few time spot when the value changes (as a simplified way I assume?).

docs/ruleset_functions/normalize_space_schedules.md
xingcx: The hourly_value here is just local to whatever value used in the schedule in the for loop. Similar to Line 31 and Line 35. Can we keep the hourly_value here so the for loop is concise? I changed the schedule name to be more clear.

docs/ruleset_functions/normalize_space_schedules.md
mkarpman: My only outstanding comment is that convert_schedule_array function has to be developed to convert event schedule to hourly schedule.

mkarpman (approved):

supriyagoel (approved):

mkarpman (approved):

PR #115, opened 2021-08-21
Create Rule6-7.md

xingcx:


docs/section6/Rule6-7.md
```diff
@@ -0,0 +1,53 @@
...
+          - Get total lighting power density in space: `total_space_LPD_b = sum(interior_lighting.power_per_area for interior_lighting in space_b.interior_lighting)`
+
+          - Get lighting status type for space: `space_lighting_status_type = space_lighting_status_type_dict_p[match_data_element(P_RMR, Spaces, space_b.id).id]`
+
+            **Rule Assertion:**
+
+            - Case 1: If space lighting status type is as-designed or as-existing, and space total interior lighting power density in B_RMR matches Table G3.7: `if ( space_lighting_status_type == "AS-DESIGNED OR AS-EXISTING" ) AND ( total_space_LPD_b == data_lookup(table_G3_7, space_b.lighting_space_type) ): PASS`  
+
+            - Case 2: Else if space lighting status type is as-designed or as-existing, and space total interior lighting power density in B_RMR does not match Table G3.7 or if lighting space type is not specified: `else if ( space_lighting_status_type == "AS-DESIGNED OR AS-EXISTING" ) AND ( ( NOT space_b.lighting_space_type ) OR (total_space_LPD_b != data_lookup(table_G3_7, space_b.lighting_space_type)) ): FAIL`
+
```
mkarpman, line 47: I think we should differentiate between the two possible reasons for the FAIL outcome - (a) space type not provided and (b) space type is provided but the baseline LPD is not as expected, so I would split this into two "cases" - (a) would require a manual check, and (b) would be FAIL.

docs/section6/Rule6-7.md
```diff
@@ -0,0 +1,53 @@
...
+        - For each space in zone: `space_b in zone_b.spaces:`  
+
+          - Get total lighting power density in space: `total_space_LPD_b = sum(interior_lighting.power_per_area for interior_lighting in space_b.interior_lighting)`
+
+          - Get lighting status type for space: `space_lighting_status_type = space_lighting_status_type_dict_p[match_data_element(P_RMR, Spaces, space_b.id).id]`
+
+            **Rule Assertion:**
+
+            - Case 1: If space lighting status type is as-designed or as-existing, and space total interior lighting power density in B_RMR matches Table G3.7: `if ( space_lighting_status_type == "AS-DESIGNED OR AS-EXISTING" ) AND ( total_space_LPD_b == data_lookup(table_G3_7, space_b.lighting_space_type) ): PASS`  
+
```
mkarpman, line 45: The baseline must be based on Table G3.7 irrespective of the proposed lighting status type, so I think we don't need to check for "space_lighting_status_type == "AS-DESIGNED OR AS-EXISTING"" in Case 1, just that the baseline total space LPD is based on Table G3.7. So case 1 and case 3 should be merged.

docs/section6/Rule6-7.md
```diff
@@ -0,0 +1,53 @@
...
+            - Case 2: Else if space lighting status type is as-designed or as-existing, and space total interior lighting power density in B_RMR does not match Table G3.7 or if lighting space type is not specified: `else if ( space_lighting_status_type == "AS-DESIGNED OR AS-EXISTING" ) AND ( ( NOT space_b.lighting_space_type ) OR (total_space_LPD_b != data_lookup(table_G3_7, space_b.lighting_space_type)) ): FAIL`
+
+            - Case 3: Else if space lighting status type is not-yet designed or matches Table_9_5_1, and space total interior lighting power density in B_RMR matches Table G3.7: `else if ( space_lighting_status_type == "NOT-YET DESIGNED OR MATCH TABLE_9_5_1" ) AND ( total_space_LPD_b == data_lookup(table_G3_7, space_b.lighting_space_type) ): PASS`
+
```
mkarpman, line 49: I think this should be combined with Case 1, per above.

docs/section6/Rule6-7.md
```diff
@@ -0,0 +1,53 @@
...
+            - Case 4: Else if space lighting status type is not-yet designed or matches Table_9_5_1, and space total interior lighting power density in B_RMR matches Table G3.8: `else if ( space_lighting_status_type == "NOT-YET DESIGNED OR MATCH TABLE_9_5_1" ) AND ( total_space_LPD_b == data_lookup(table_G3_8, building_segment_b.lighting_building_area_type) ): PASS`
+
```
mkarpman, line 51: This rule was added in an addendum to 90.1 2019. Was it decided to included it in the RDS? Aside from that, I think this is the only scenario when a conincidental alignment between proposed LPDs and Table 9.5.1 may result in mistaken outcome, because the RDS sets baseline LPD based on building area method that does not apply in this case.

docs/section6/Rule6-7.md
supriyagoel: that makes sense

docs/section6/Rule6-7.md
supriyagoel: ok, merging case 1 and case 3 makes sense.

docs/section6/Rule6-7.md
supriyagoel: @mkarpman addendum AG clarifies this requirement and yes, it was decided to include this requirement for the RCT because we classified this as a 'clarification to 90.1 2019 requirements' and not a 'new requirement'.
we're trying to incorporate all addenda that clarify 2019 requirements. Let us know if you disagree. If yes, what should the check in this case be?

docs/section6/Rule6-7.md
supriyagoel: @xingcx - just one for comment. for space type = "sales area" we should add a outcome message for case 2 which requires manual check. This is because the proposed building might include some additional 'display lighting' which also increase the baseline lighting power allowance. If space type = Sales Area and the baseline LPD is > table G3.7, then a manual check is required.

docs/section6/Rule6-7.md
xingcx: @mkarpman Case 3 is for when lighting is design/existing but the LPD also matches Table 9.5.1 value, as the function right now cannot differentiate this case and not-yet designed. Let me know how we want to treat this case. Similar question for Rule 6-4.

docs/section6/Rule6-7.md
xingcx: Updated.

docs/section6/Rule6-7.md
xingcx: Split into Case 2, 3 and 4.

docs/section6/Rule6-7.md
mkarpman: The rule description in row #5 suggests that "Where lighting neither exists nor is submitted with design documents, baseline LPD shall be determined in accordance with Table G3-8.", but I don't see this language in 90.1 2019. This was introduced in a later addendum. So for 2019 version, the baseline is always based on Table G3-7.

docs/section6/Rule6-7.md
mkarpman: Following the discussion we had at the bi-weekly call, I think both (a) and (b) should be "FAIL", but with different messages.

docs/section6/Rule6-7.md
mkarpman: @supriyagoel I think we agreed that there should be no additional baseline allowance for retail lighting as it was added in an addendum to 90.1 2019.

mkarpman (changes requested):

docs/section6/Rule6-7.md
supriyagoel: @mkarpman maybe I misunderstood your email. I thought we would not check for the additional allowance but flag such spaces for manual review IF they do not align with G3.7. Do you not recommend doing that either?

docs/section6/Rule6-7.md
supriyagoel: @mkarpman i don't quite understand this requirement in 2019 App G. If space types are not known, how can the baseline LPD be set based on G3.7? there would be no lighting space type specified in such scenarios, correct?

docs/section6/Rule6-7.md
```diff
@@ -0,0 +1,60 @@
...
+# Lighting - Rule 6-7
+
+**Rule ID:** 6-7  
+**Rule Description:** Where a complete lighting system exists and where a lighting system has been designed and submitted with design documents, the baseline LPD is equal to expected value in Table G3.7. Where lighting neither exists nor is submitted with design documents, baseline LPD shall be determined in accordance with Table G3-7 for "Office-Open Plan" space type.
+
```
mkarpman, line 6: The second sentense should say "Where lighting neither exists nor is submitted with design documents and **lighting space type is unknown**, baseline LPD shall be determined in accordance with Table G3-7 for "Office-Open Plan" space type.

mkarpman (approved):

PR #126, opened 2021-09-03
table3.2 data lookup function and tests

vijayvardhan94:
Data lookup functions and tests for Table 3.2

### DIFF
--- a/rct229/data_fns/table_3_2_fns.py
+++ b/rct229/data_fns/table_3_2_fns.py
@@ -37,7 +37,8 @@ def table_3_2_lookup(climate_zone_enum_val):
         One of the ClimateZone2019ASHRAE901 enumeration values
     Returns
     -------
-    A dictionary of the form: { system_min_heating_ouput: float – the heating output for the given climate one given by Table 3.2 [Btu/h·ft^2] }
+    dict 
+        { system_min_heating_ouput: float – the heating output for the given climate one given by Table 3.2 [Btu/h·ft^2] }
     """
     climate_zone = climate_zone_enumeration_to_climate_zone_map[climate_zone_enum_val]
     osstd_entry = find_osstd_table_entry(
--- a/rct229/data_fns/table_3_2_fns.py
+++ b/rct229/data_fns/table_3_2_fns.py
@@ -39,8 +39,9 @@ climate_zone_enumeration_to_climate_zone_type_map = {
     "CZ6A": "6",
     "CZ6B": "6",
     "CZ7": "7",
-    "CZ8": "8"    
+    "CZ8": "8",
 }
+
 
 def table_3_2_lookup(climate_zone):
     """Returns the heating output for a climate zone as
@@ -54,9 +55,7 @@ def table_3_2_lookup(climate_zone):
     dict
         The Heating Output for a given Climate Zone given by Table 3.2 [Btu/h·ft^2]
     """
-    climate_zone_type = climate_zone_enumeration_to_climate_zone_type_map[
-        climate_zone
-    ]
+    climate_zone_type = climate_zone_enumeration_to_climate_zone_type_map[climate_zone]
     osstd_entry = find_osstd_table_entry(
         [("climate_zone", climate_zone_type)],
         osstd_table=data["ashrae_90_1_table_3_2"],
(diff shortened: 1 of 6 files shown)


rct229/data_fns/table_3_2_fns.py
```diff
@@ -0,0 +1,64 @@
...
+CZ3C -> "3C"
+CZ4A, CZ4B -> "4A,4B"
+CZ4C -> "4C"
+CZ5A, CZ5B, CZ5C -> "5"
+CZ6A, CZ6B -> "6"
+CZ7 -> "7"
+CZ8 -> "8"
+
+"""
+climate_zone_enumeration_to_climate_zone_type_map = {
```
charliepnnl, line 23: Since the column in Table 3.2 is called "Climate Zone" and you have rightly used the field name "climate_zone" in the json data file, I believe the name for this map should be `climate_zone_enumeration_to_climate_zone_map` (I've dropped the word "type").

rct229/data_fns/table_3_2_fns.py
```diff
@@ -0,0 +1,64 @@
...
+    "CZ5C": "5",
+    "CZ6A": "6",
+    "CZ6B": "6",
+    "CZ7": "7",
+    "CZ8": "8",
+}
+
+
+def table_3_2_lookup(climate_zone):
+    """Returns the heating output for a climate zone as
```
charliepnnl, line 47: I suggest inserting the word "minimum" as in "... the minimum heating output ...".

rct229/data_fns/table_3_2_fns.py
```diff
@@ -0,0 +1,64 @@
...
+    required by ASHRAE 90.1 Table 3.2
+    Parameters
+    ----------
+    climate_zone : str
+        One of the ClimateZone2019ASHRAE901 enumeration values
+    Returns
+    -------
+    dict
+        The Heating Output for a given Climate Zone given by Table 3.2 [Btu/h·ft^2]
```
charliepnnl, line 56: Better to say something like:

A dictionary of the form:
    {
        system_min_heating_ouput: float – the heating output for the given climate one given by  Table 3.2 [Btu/h·ft^2]
    }

rct229/data_fns/table_3_2_test.py
```diff
@@ -0,0 +1,61 @@
...
+def test__table_3_2_CZ7():
+    assert table_3_2_lookup("CZ7") == {"system_min_heating_output": 16}
+
+
+def test__table_3_2_CZ8():
+    assert table_3_2_lookup("CZ8") == {"system_min_heating_output": 19}
+
+
+# Testing climate_zone_enumeration_to_climate_zone_type_map ----------
+def test__lighting_space_enumeration_to_lpd_space_type_map():
```
charliepnnl, line 53: The name of this function should be changed to match the map being tested.

charliepnnl (commented):
My comments are primarily regarding documentation and just a few other picky things.

rct229/data_fns/table_3_2_fns.py
```diff
@@ -0,0 +1,64 @@
...
+    "CZ8": "8",
+}
+
+
+def table_3_2_lookup(climate_zone):
+    """Returns the heating output for a climate zone as
+    required by ASHRAE 90.1 Table 3.2
+    Parameters
+    ----------
+    climate_zone : str
```
charliepnnl, line 51: It might be helpful to us to call this function parameter `climate_zone_enum_val` so it is clearer that it is an enumeration value and to avoid a name collision with the table value with the same name. Then you can use the name `climate_zone` where you are using `climate_zone_type`.

rct229/data_fns/table_3_2_fns.py
charliepnnl: Oh sorry, I meant for you to keep this line: dict
Just replacing the explanation on the following line with "A dictiontary ...".

rct229/data_fns/table_3_2_fns.py
```diff
@@ -0,0 +1,48 @@
...
+def table_3_2_lookup(climate_zone_enum_val):
+    """Returns the minimum heating output for a climate zone as
+    required by ASHRAE 90.1 Table 3.2
+    Parameters
+    ----------
+    climate_zone_enum_val : str
+        One of the ClimateZone2019ASHRAE901 enumeration values
+    Returns
+    -------
+    A dictionary of the form: { system_min_heating_ouput: float – the heating output for the given climate one given by Table 3.2 [Btu/h·ft^2] }
```
charliepnnl, line 40: Oh sorry, I meant for you to keep the line showing the return type,  dict

charliepnnl (commented):
This is great!
Just the one miscommunication regarding the return type line.

charliepnnl (approved):

PR #129, opened 2021-09-04
Update and rename get_wwr to get_window_wall_areas

xingcx:
Changed function to return window wall areas for building segments instead of window wall ratio.
Changed function name.

### DIFF
--- a/docs/ruleset_functions/get_building_segment_window_wall_areas.md
+++ b/docs/ruleset_functions/get_building_segment_window_wall_areas.md
@@ -1,13 +1,14 @@
 
-## get_overall_building_segment_wwr
+## get_building_segment_window_wall_areas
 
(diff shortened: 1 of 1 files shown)


docs/ruleset_functions/get_building_segment_window_wall_areas.md
```diff
@@ -41,8 +42,8 @@ Logic:
```
mkarpman, line 42: for line 35, please change the note from "If surface is above-grade wall and adjacent to the exterior" to "If surface is above-grade wall and is part of exterior or semi-exterior building envelope"

xingcx: Updated.

supriyagoel (approved):

mkarpman (approved):

PR #130, opened 2021-09-04
Create Rule5-21.md

xingcx:


docs/section5/Rule5-21.md
```diff
@@ -0,0 +1,66 @@
...
+  - For each thermal block in building segment: `for thermal_block_b in building_segment_b.thermal_blocks:`
+
+    - For each zone in thermal block: `for zone_b in thermal_block_b.zones:`  
+
+      - For each surface in zone: `for surface_b in zone_b.surfaces:`  
+
+        - Check if surface is above-grade wall and is regulated: `if ( get_opaque_surface_type(surface_b.id) == "ABOVE-GRADE WALL" ) AND ( scc_dictionary_b[surface_b.id] != "UNREGULATED" ):`
+
+          - For each subsurface in surface: `for subsurface_b in surface_b.subsurfaces:`
+
```
mkarpman, line 47: In the following lines, you are calulating the total fenestration area on each regulated surface in the baselnie and proposed design. I don't think this is necessary. I think from line 46 you can jump right to the rule assertion to check that each subsurface accounts for the same percentage of overall fenestration area in the baseline as in the proposed design. You'd only exclude doors that are less than 50% glass from this check; all other regulated subsurfaces would need to be verified.

supriyagoel: area of each sub-surface should account for the same % of sub-surface area in the baseline as the proposed.
baseline, should have the same number of sub-surfaces as the proposed, i.e., there should be an alignment in subsurface IDs between B and P.

supriyagoel: this verification should happen at surface level and not zone level.

supriyagoel (changes requested):

docs/section5/Rule5-21.md
```diff
@@ -0,0 +1,66 @@
...
+# Envelope - Rule 5-21  
+
+**Rule ID:** 5-21  
+**Rule Description:** The vertical fenestration shall be distributed on each face of the building in the same proportion as in the proposed design.  
+**Rule Assertion:** B-RMR total (subsurface.glazed_area+subsurface.opaque_area) = expected value for each zone
```
mkarpman, line 6: Please change that it's the same value for each surface, not each zone.

docs/section5/Rule5-21.md
```diff
@@ -0,0 +1,66 @@
...
+              - Check if subsurface is door, and if the glazed vision area is more than 50% of the total area, add subsurface total area to surface total fenestration area: `if ( subsurface_p.classification == "DOOR" ) AND ( subsurface_p.glazed_area > ( subsurface_p.glazed_area + subsurface_p.opaque_area ) * 50% ): total_fenestration_area_surface_p += subsurface_p.glazed_area + subsurface_p.opaque_area`  
+
+              - Else if subsurface is not door, add total subsurface area to surface total fenestration area: `else if subsurface_p.classification != "DOOR" : total_fenestration_area_surface_p += surface_p.glazed_area + surface_p.opaque_area`
+
+        **Rule Assertion:**
+
+        - Case 1: For each surface, if total fenestration area in B_RMR is in the same proportion as in P_RMR: `if total_fenestration_area_surface_b / total_fenestration_area_b == total_fenestration_area_surface_p / total_fenestration_area_p: PASS`
+
+        - Case 2: Else: `else: FAIL and raise_warning "THE DISTRIBUTION OF VERTICAL FENESTRATION IN SURFACE IS NOT IN THE SAME PROPORTION AS IN THE PROPOSED DESIGN. VERIFY IF ENVELOPE IS EXISTING OR ALTERED AND CAN BE EXCLUDED FROM THIS CHECK."`
+
```
mkarpman, line 65: Please change "raise_warning" to "show message". Please make it say ""THE VERTICAL FENESTRATION IS NOT DISRIBUTED ACROSS BASELINE OPAQUE SURFACES IN THE SAME PROPORTION AS IN THE PROPOSED DESIGN. VERIFY IF ENVELOPE IS EXISTING OR ALTERED AND CAN BE EXCLUDED FROM THIS CHECK."

xingcx: Updated.

mkarpman (approved):

supriyagoel (approved):

PR #134, opened 2021-09-08
RS/CH/Rule5-2

charliepnnl:


charliepnnl:
I have refactored this rule to report at the building level with the list `failing_surface_ids` as a calculated value.

mcneillj (approved):

PR #135, opened 2021-09-08
RS/CH/Rule5-3

charliepnnl:


charliepnnl:
I have refactored this rule to report at the building level with a list `baseline_surfaces_casting_shade_ids` as a calculated value.

rct229/rules/section5.py
```diff
@@ -0,0 +1,58 @@
...
+from rct229.utils.jsonpath_utils import find_all
+
+# Rule Definitions for Section 5 of 90.1-2019 Appendix G
+
+
+# ------------------------
+
+
+class Section5Rule3(RuleDefinitionListIndexedBase):
+    """Rule 3 of ASHRAE 90.1-2019 Appendix G Section 6 (Lighting)"""
```
vijayvardhan94, line 15: I maybe wrong, the class name suggests Section 5 Rule 3, but the description/documentation mentions rule 6. Please ignore otherwise.

rct229/rules/section5.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self):
```
vijayvardhan94, line 17: I found descriptions of table functions really helpful. Would it make sense to include similar documentation for classes/functions and their return types within sections as well.?

rct229/rules/section5.py
charliepnnl: Good catch.  I will change that.

rct229/rules/section5.py
vijayvardhan94: The documentation can be found in rule_base.py related to sections.

mcneillj (approved):

rct229/rules/section5.py
```diff
@@ -0,0 +1,53 @@
...
+                required_fields={
+                    "$..surfaces[*]": ["adjacent_to"],
+                    "$..surfaces[?(@.adjacent_to='EXTERIOR')]": ["does_cast_shade"],
+                },
+            )
+
+        def get_calc_vals(self, context, data=None):
+            baseline_surfaces_casting_shade_ids = []
+            for surface in find_all(
+                "$..surfaces[?(@.adjacent_to='EXTERIOR')]", context.baseline
```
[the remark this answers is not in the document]
charliepnnl: Good idea. I have extracted this json path into a reusable constant.

PR #141, opened 2021-09-10
Create Rule5-35.md

xingcx:


mkarpman (approved):

supriyagoel:
Looks like we're simply checking is the B SRR = 3% but not the proportion of skylight area in all roof components. I think that is fine, its an accepted simplification.

supriyagoel (approved):

PR #146, opened 2021-09-15
Clarification on Rule5-2 RDS

mcneillj:
We are looking to have more clarification in the RDS about the evaluation context for the rules.  I discussed with @charliepnnl and we agree that it would be more useful for the "Evaluation Context" field to describe the data group that a rule is evaluated at.  We are proposing that this is updated to "Building" for this rule.  This would mean that the rule assertion is conducted one time for each building after collecting the list of all surfaces with non-matching orientations.  The list of surfaces is obtained by comparing each orientation for all surfaces in the building.

### DIFF
--- a/docs/section5/Rule5-2.md
+++ b/docs/section5/Rule5-2.md
@@ -10,7 +10,7 @@
 **Applicability Checks:** None  
 
 **Manual Check:** None  
-**Evaluation Context:** Each Data Element  
+**Evaluation Context:** Building Segment  
 **Data Lookup:** None  
 
 ## Rule Logic:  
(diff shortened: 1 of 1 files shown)


mkarpman:
As far as I understand, you suggest that the RCT cycles through all pairs of matching surfaces in U-RMR and P-RMR, and if there is at least one mismatch generate "FAIL" outcome with list the offending surfaces. Otherwise it will be PASS. I think this is a way better approach than generating a separate "FAIL" outcome for each mismatch, because it would make the RCT report much more readable by reducing the number of error messages from one per component (surface, subsurface, zone, etc.) to one per rule. Perhaps this logic should be applied to all rules, not just this one?

xingcx:
At building level is a better approach. I think this can also apply to most of the other rules. We can discuss exceptions when we want to evaluate each data element and when we want to apply this at building_segment level, etc.

PR #34, opened 2021-04-26
Update Rule12-6.md

xingcx:

PR #149, opened 2021-09-17
Create Rule5-47.md

xingcx:


docs/section5/Rule5-47.md
```diff
@@ -0,0 +1,53 @@
...
+    - For each zone in thermal block: `for zone_b in thermal_block_b.zones:`
+
+      - For each surface in zone: `for surface_b in zone_b.surfaces:`
+
+        - Check if surface is regulated, add zone total area of building envelope to building total: `if scc_dict_b[surface_b.id] != "UNREGULATED": building_total_envelope_area += sum(surface.area for surface in zone_b.surfaces)`
+
+      - Check if zone is conditioned or semi-heated, add zone air leakage rate to building total: `if zone_conditioning_category_dict_b[zone.id] in [CONDITIONED RESIDENTIAL, CONDITIONED NON-RESIDENTIAL, CONDITIONED MIXED, SEMI-HEATED]: building_total_air_leakage_rate += zone_b.infiltration.air_leakage_rate`
+
+- Calculate the required baseline air leakage rate at 75Pa: `target_air_leakage_rate_b = 1.0 * building_total_envelope_area * 0.00047194745`
+
```
supriyagoel, line 40: @xingcx I'm assuming the 0.00047 number is converting from in wg to Pa. Please add a note somewhere to state the same.

supriyagoel: @xingcx for additional clarity, we should probably state the unit of the calculated value - baseline infiltration airflow m3/s @ 75 Pa

docs/section5/Rule5-47.md
```diff
@@ -0,0 +1,53 @@
...
+**Rule Assertion:**  
+
+- Case 1: For B_RMR, if infiltration pressure difference is 75Pa and the total air leakage rate for conditioned and semi-heated zones is equal to the required baseline air leakage rate at 75Pa: `if ( B_RMR.ASHRAE229.infiltration_pressure_difference == 75 ) AND ( building_total_air_leakage_rate == target_air_leakage_rate_b ): PASS`
+
+- Case 2: Else if infiltration pressure difference is 75Pa for B_RMR and the total air leakage rate for conditioned and semi-heated zones is not equal to the required baseline air leakage rate at 75Pa: `if ( B_RMR.ASHRAE229.infiltration_pressure_difference == 75 ) AND ( building_total_air_leakage_rate != target_air_leakage_rate_b ): FAIL`
+
+- Case 3: Else if infiltration pressure difference is 0Pa for B_RMR and the total air leakage rate for conditioned and semi-heated zones is equal to the required baseline air leakage rate at 0Pa: `if ( B_RMR.ASHRAE229.infiltration_pressure_difference == 0 ) AND ( building_total_air_leakage_rate == target_air_leakage_rate_b * 0.112 ): PASS`
```
supriyagoel, line 47: I am not familiar with this rule for air leakage rate @ 0 Pa. What App G requirement does this refer to?

docs/section5/Rule5-47.md
mkarpman: @supriyagoel this is conversion from air leakage rate in CFM (1.0 CFM) to air leakage rate in cubic meters per second.

docs/section5/Rule5-47.md
```diff
@@ -0,0 +1,53 @@
...
+        - Check if surface is regulated, add zone total area of building envelope to building total: `if scc_dict_b[surface_b.id] != "UNREGULATED": building_total_envelope_area += sum(surface.area for surface in zone_b.surfaces)`
+
+      - Check if zone is conditioned or semi-heated, add zone air leakage rate to building total: `if zone_conditioning_category_dict_b[zone.id] in [CONDITIONED RESIDENTIAL, CONDITIONED NON-RESIDENTIAL, CONDITIONED MIXED, SEMI-HEATED]: building_total_air_leakage_rate += zone_b.infiltration.air_leakage_rate`
+
+- Calculate the required baseline air leakage rate at 75Pa: `target_air_leakage_rate_b = 1.0 * building_total_envelope_area * 0.00047194745`
+
+**Rule Assertion:**  
+
+- Case 1: For B_RMR, if infiltration pressure difference is 75Pa and the total air leakage rate for conditioned and semi-heated zones is equal to the required baseline air leakage rate at 75Pa: `if ( B_RMR.ASHRAE229.infiltration_pressure_difference == 75 ) AND ( building_total_air_leakage_rate == target_air_leakage_rate_b ): PASS`
+
```
mkarpman, line 44: Let's discuss at the RDS call if this should be pass or fail. (I.e., do we want to flag all RMRs that report air leakage at 75Pa as fail, since we don't know if they converted to leakage at wind pressure correctly.

docs/section5/Rule5-47.md
```diff
@@ -0,0 +1,53 @@
...
+- Case 2: Else if infiltration pressure difference is 75Pa for B_RMR and the total air leakage rate for conditioned and semi-heated zones is not equal to the required baseline air leakage rate at 75Pa: `if ( B_RMR.ASHRAE229.infiltration_pressure_difference == 75 ) AND ( building_total_air_leakage_rate != target_air_leakage_rate_b ): FAIL`
+
```
mkarpman, line 46: If we decide to not accept any infiltraiton input at 75Pa as correct, we can merge this with Case 1. Otherwise it looks good.

docs/section5/Rule5-47.md
mkarpman: @supriyagoel this is air leakage at wind pressure (no pressurization). Let's discuss at the next RDS call.

docs/section5/Rule5-47.md
supriyagoel: RCT team will be using a library for conversions. It will be based on actual units. Can instead say 'Pa_to_psi'. RCT team will eventually upload a file with the actual conversion factors.

supriyagoel: if there are value comparisons, then unit conversions don't need to be done in the RDS. However, if the unit conversion is a part of the RDS logic, then the conversion can be shown using terminology (like cubic feet per minute and cubic meter per second) but not values.

docs/section5/Rule5-47.md
supriyagoel: have two options- 1. a flag in the RMR which states that infiltration rate specified as wind pressure or 2. add a note in the RMR which states that a value of '0' indicates infiltration specified at wind pressure.

mkarpman: It seems that if we want to verify both (A) leakage at 75Pa, as prescribed in Table G3.1 #5 item (b) in the proposed column and item (h) in the baseline column, and (B) leakage at wind pressure as prescribed in G3.1.1.4, we'd need to rename the current infiltration.air_leakage_rate to something like infiltration.tested_air_leakage_rate, and add infiltration.air_leakage_at_wind_pressure. 

For the baseline, RCT will only verify infiltration.air_leakage_at_wind_pressure. For proposed, we would also verify infiltration.air_leakage_at_wind_pressure, and if the entered value is not equivalent to the default of 0.6 CFM/SF @75Pa converted to the rate at wind pressure per G3.1.1.4, will additionally check input for infiltration.tested_air_leakage_rate.

docs/section5/Rule5-47.md
mkarpman: Per last comment from Supriya, need to replace 0.00047 with (cubic feet per minute to cubic meter per second)

docs/section5/Rule5-47.md
mkarpman: Don't need to check that B_RMR.ASHRAE229.infiltration_pressure_difference == 75; just that the building_total_air_leakage_rate == target_air_leakage_rate_b

docs/section5/Rule5-47.md
```diff
@@ -0,0 +1,53 @@
...
+  - For each thermal block in building segment: `for thermal_block_b in building_segment_b.thermal_blocks:`
+
+    - For each zone in thermal block: `for zone_b in thermal_block_b.zones:`
+
+      - For each surface in zone: `for surface_b in zone_b.surfaces:`
+
+        - Check if surface is regulated, add zone total area of building envelope to building total: `if scc_dict_b[surface_b.id] != "UNREGULATED": building_total_envelope_area += sum(surface.area for surface in zone_b.surfaces)`
+
+      - Check if zone is conditioned or semi-heated, add zone air leakage rate to building total: `if zone_conditioning_category_dict_b[zone.id] in [CONDITIONED RESIDENTIAL, CONDITIONED NON-RESIDENTIAL, CONDITIONED MIXED, SEMI-HEATED]: building_total_air_leakage_rate += zone_b.infiltration.air_leakage_rate`
+
```
mkarpman, line 38: should be "building_total_air_leakage_rate += zone_b.infiltration.infiltration_flow_rate`

mkarpman (changes requested):

docs/section5/Rule5-47.md
xingcx: Updated.

docs/section5/Rule5-47.md
xingcx: Updated.

mkarpman (approved):

docs/section5/Rule5-47.md
mkarpman: @xingcx and @supriyagoel given the confirmed direction regarding the units, we should remove * CONVERSION("CFM", "M^3/S")` from line 35.

mkarpman (changes requested):

docs/section5/Rule5-47.md
supriyagoel: @mcneillj Its been a while since we discussed this. How should the units be mentioned here in the RDS. I believe it is ok to mention CONVERSION("CFM", "M^3/S")` - please clarify.

docs/section5/Rule5-47.md
```diff
@@ -0,0 +1,43 @@
...
+## Rule Logic:  
+
+- Get surface conditioning category dictionary for B_RMR: `scc_dict_b = get_surface_conditioning_category(B_RMR)`
+
+- Get zone conditioning category dictionary for B_RMR: `zone_conditioning_category_dict_b = get_zone_conditioning_category(B_RMR)`
+
+- For each zone in the Baseline model: `for zone_b in B_RMR...zones:`
+
+  - For each surface in zone: `for surface_b in zone_b.surfaces:`
+
```
supriyagoel, line 30: should this be for each exterior surface of the zone?

inkmug: The rule requirement (G3.1-5(h)) is "The baseline air leakage rate of the building **envelope** (I75Pa)...". And according to the definition, "**building envelope**: the exterior plus the semiexterior portions of a building." This rule currently checks if the surface is regulated, which includes all exterior (exterior res, exterior non res, exterior mixed) and semi-exterior.

xingcx: The rule requirement (G3.1-5(h)) is "The baseline air leakage rate of the building envelope (I75Pa)...". And according to the definition, "building envelope: the exterior plus the semiexterior portions of a building." This rule currently checks if the surface is regulated, which includes all exterior (exterior res, exterior non res, exterior mixed) and semi-exterior.

docs/section5/Rule5-47.md
xingcx: Updated. The unit conversion is removed.

docs/section5/Rule5-47.md
mkarpman: I agree with @xingcx

docs/section5/Rule5-47.md
```diff
@@ -0,0 +1,43 @@
...
+    - Check if surface is regulated, add zone total area of building envelope to building total: `if scc_dict_b[surface_b.id] != "UNREGULATED": building_total_envelope_area += sum(surface.area for surface in zone_b.surfaces)`
+
+  - Check if zone is conditioned or semi-heated, add zone infiltration flow rate to building total: `if zone_conditioning_category_dict_b[zone.id] in [CONDITIONED RESIDENTIAL, CONDITIONED NON-RESIDENTIAL, CONDITIONED MIXED, SEMI-HEATED]: building_total_air_leakage_rate_b += zone_b.infiltration.infiltration_flow_rate`
+
+- Calculate the required baseline infiltration flow rate in cfm: `target_infiltration_flow_rate_b = 1.0 * building_total_envelope_area`
+
+**Rule Assertion:**  
+
+- Case 1: For B_RMR, if the total zone infiltration rate for conditioned and semi-heated zones is equal to the required baseline infiltration rate at wind pressure: `if building_total_air_leakage_rate_b == target_infiltration_flow_rate_b: PASS`
```
mkarpman, line 39: The target flow rate calculated in line #35 is CFM @ 75PA. However, building_total_air_leakage_rate_b calculated in #34 is the sum of  zone.infiltration.infiltration_flow_rate, which is the rate at wind pressure. So I think the assertion should be 
if building_total_air_leakage_rate_b == target_infiltration_flow_rate_b / 0.112 : PASS

docs/section5/Rule5-47.md
supriyagoel: @xingcx line 35 says 'Calculate the required baseline infiltration flow rate in cfm: `target_infiltration_flow_rate_b = 1.0 * building_total_envelope_area`

i believe total envelope area will be in m2, and 1 is CFM/ft2. 
we should convert the total envelope area to ft2. Probably need to add a line that says building_total_envelope_area = building_total_envelope_area (m2 to ft2)
We don't want to include the actual conversion factors (since the RCT team is using Pint for that) but do want to identify when a conversion needs to take place.

docs/section5/Rule5-47.md
xingcx: Corrected.

mkarpman (approved):

docs/section5/Rule5-47.md
supriyagoel: as discussed at the last RDS call, there is a default list of units and the unit conversions will only be mentioned if a unit is something other than the units in that list.

supriyagoel (approved):

PR #150, opened 2021-09-17
Create Rule5-49.md

xingcx:


docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+
+- Case 4: Else if infiltration pressure difference is 0Pa for P_RMR and the total air leakage rate for conditioned and semi-heated zones is not equal to the required proposed design air leakage rate at 0Pa: `if ( P_RMR.ASHRAE229.infiltration_pressure_difference == 0 ) AND ( building_total_air_leakage_rate != target_air_leakage_rate_p * 0.112 ): FAIL`
+
+- Case 5: Else, infiltration pressure difference is neither 75Pa nor 0Pa for P_RMR: `else: FAIL and raise_warning "90.1 SECTION G3.1.1.4 PRESCRIBED PROPOSED DESIGN INFILTRATION RATE AT 75PA AND METHODOLOGY TO CONVERT AIR LEAKAGE RATE AT 75PA TO AIR LEAKAGE AT WIND PRESSURE. THE MODELED INFILTRATION IS AT PRESSURE DIFFERENTIAL OTHER THAN THE TWO OPTIONS PRESCRIBED BY 90.1. EXCEPTION APPLIES WHEN WHOLE-BUILDING AIR LEAKAGE TESTING IS SPECIFIED DURING DESIGN AND COMPLETED AFTER CONSTRUCTION. THE PROPOSED DESIGN AIR LEAKAGE RATE OF THE BUILDING ENVELOPE SHALL BE AS MEASURED. MANUAL CHECK IS REQUIRED TO VERIFY THAT PROPOSED DESIGN INFILTRATION WAS MODELED CORRECTLY."`
+
```
mkarpman, line 52: Everything looks good pending resolution of comments for Rule 47

docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+    - For each zone in thermal block: `for zone_p in thermal_block_p.zones:`
+
+      - For each surface in zone: `for surface_p in zone_p.surfaces:`
+
+        - Check if surface is regulated, add zone total area of building envelope to building total: `if scc_dict_p[surface_p.id] != "UNREGULATED": building_total_envelope_area += sum(surface.area for surface in zone_p.surfaces)`
+
+      - Check if zone is conditioned or semi-heated, add zone air leakage rate to building total: `if zone_conditioning_category_dict_p[zone.id] in [CONDITIONED RESIDENTIAL, CONDITIONED NON-RESIDENTIAL, CONDITIONED MIXED, SEMI-HEATED]: building_total_air_leakage_rate += zone_p.infiltration.air_leakage_rate`
+
+- Calculate the required proposed design air leakage rate at 75Pa: `target_air_leakage_rate_p = 0.6 * building_total_envelope_area * 0.00047194745`
+
```
supriyagoel, line 40: @xingcx need to remove the conversion metric from here. Based on input provided in 5-47 , this RDS should be updated as well.

docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+- For each building segment in the Proposed model: `for building_segment_p in P_RMR.building.building_segments:`
+
+  - For each thermal block in building segment: `for thermal_block_p in building_segment_p.thermal_blocks:`
+
+    - For each zone in thermal block: `for zone_p in thermal_block_p.zones:`
+
+      - For each surface in zone: `for surface_p in zone_p.surfaces:`
+
+        - Check if surface is regulated, add zone total area of building envelope to building total: `if scc_dict_p[surface_p.id] != "UNREGULATED": building_total_envelope_area += sum(surface.area for surface in zone_p.surfaces)`
+
```
supriyagoel, line 36: same questions as before, this is all surfaces or only exterior surfaces?

docs/section5/Rule5-49.md
supriyagoel: @xingcx similar to #149 this needs to remove the actual conversion factor and use the cfm to m3-s explanation instead

docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+**Rule Assertion:**  
+
+- Case 1: For P_RMR, if infiltration pressure difference is 75Pa and the total air leakage rate for conditioned and semi-heated zones is equal to the required proposed design air leakage rate at 75Pa: `if ( P_RMR.ASHRAE229.infiltration_pressure_difference == 75 ) AND ( building_total_air_leakage_rate == target_air_leakage_rate_p ): PASS`
+
+- Case 2: Else if infiltration pressure difference is 75Pa for P_RMR and the total air leakage rate for conditioned and semi-heated zones is not equal to the required proposed design air leakage rate at 75Pa: `if ( P_RMR.ASHRAE229.infiltration_pressure_difference == 75 ) AND ( building_total_air_leakage_rate != target_air_leakage_rate_p ): FAIL`
+
+- Case 3: Else if infiltration pressure difference is 0Pa for P_RMR and the total air leakage rate for conditioned and semi-heated zones is equal to the required proposed design air leakage rate at 0Pa: `if ( P_RMR.ASHRAE229.infiltration_pressure_difference == 0 ) AND ( building_total_air_leakage_rate == target_air_leakage_rate_p * 0.112 ): PASS`
+
```
supriyagoel, line 48: should add some explanation of this *0.112 - maybe a reference to the App G section. External reviewers might be puzzled by this

docs/section5/Rule5-49.md
inkmug: The rule requirement (G3.1-5(h)) is "The baseline air leakage rate of the building envelope (I75Pa)...". And according to the definition, "building envelope: the exterior plus the semiexterior portions of a building." This rule currently checks if the surface is regulated, which includes all exterior (exterior res, exterior non res, exterior mixed) and semi-exterior.

docs/section5/Rule5-49.md
inkmug: Reference is now added.

docs/section5/Rule5-49.md
xingcx: The rule requirement (G3.1-5(h)) is "The baseline air leakage rate of the building envelope (I75Pa)...". And according to the definition, "building envelope: the exterior plus the semiexterior portions of a building." This rule currently checks if the surface is regulated, which includes all exterior (exterior res, exterior non res, exterior mixed) and semi-exterior.

docs/section5/Rule5-49.md
xingcx: Reference is now added.

docs/section5/Rule5-49.md
xingcx: Updated. The unit conversion is now updated.

docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+- Get surface conditioning category dictionary for P_RMR: `scc_dict_p = get_surface_conditioning_category(P_RMR)`
+
+- Get zone conditioning category dictionary for P_RMR: `zone_conditioning_category_dict_p = get_zone_conditioning_category(P_RMR)`
+
+- For each building segment in the Proposed model: `for building_segment_p in P_RMR.building.building_segments:`
+
+  - For each thermal block in building segment: `for thermal_block_p in building_segment_p.thermal_blocks:`
+
+    - For each zone in thermal block: `for zone_p in thermal_block_p.zones:`
+
```
mkarpman, line 32: Please remove "thermal block" nomenclature, so that the RDS starts with looping through zones similar to 5-47

docs/section5/Rule5-49.md
mkarpman: Agree with @xingcx

docs/section5/Rule5-49.md
mkarpman: I still see the conversion. Should be `target_air_leakage_rate_p = 0.6 * building_total_envelope_area`, analogous to 5-47

docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+      - For each surface in zone: `for surface_p in zone_p.surfaces:`
+
+        - Check if surface is regulated, add zone total area of building envelope to building total: `if scc_dict_p[surface_p.id] != "UNREGULATED": building_total_envelope_area += sum(surface.area for surface in zone_p.surfaces)`
+
+      - Check if zone is conditioned or semi-heated, add zone air leakage rate to building total: `if zone_conditioning_category_dict_p[zone.id] in [CONDITIONED RESIDENTIAL, CONDITIONED NON-RESIDENTIAL, CONDITIONED MIXED, SEMI-HEATED]: building_total_air_leakage_rate += zone_p.infiltration.air_leakage_rate`
```
mkarpman, line 37: The name of the data attribute was changed to infiltration.measured_air_leakage_rate; I would also change the name of the building level rate accordingly, to building_total_measured_air_leakage_rate.

I this we also need to calculate building_total_infiltration_flow_rate, as done in 5-47. Unless exception to the rule applies, air leakage test was not performed so measured rate would not be entered so we would use the modeled rate at wind pressure for this check.

docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+
+- Calculate the required proposed design air leakage rate at 75Pa: `target_air_leakage_rate_p = 0.6 * building_total_envelope_area * 0.00047194745`
+
+**Rule Assertion:**  
+
+- Case 1: For P_RMR, if infiltration pressure difference is 75Pa and the total air leakage rate for conditioned and semi-heated zones is equal to the required proposed design air leakage rate at 75Pa: `if ( P_RMR.ASHRAE229.infiltration_pressure_difference == 75 ) AND ( building_total_air_leakage_rate == target_air_leakage_rate_p ): PASS`
+
```
mkarpman, line 44: This case does not check the part of the rule about converting leakage @ 75PA to wind pressure. Since 0.6 CFM/SF only applies when air leakage was not tested, I recommend replacing Case 1 with the following check: building_total_infiltration_flow_rate == target_infiltration_flow_rate_p*0.112

docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+- Case 2: Else if infiltration pressure difference is 75Pa for P_RMR and the total air leakage rate for conditioned and semi-heated zones is not equal to the required proposed design air leakage rate at 75Pa: `if ( P_RMR.ASHRAE229.infiltration_pressure_difference == 75 ) AND ( building_total_air_leakage_rate != target_air_leakage_rate_p ): FAIL`
+
```
mkarpman, line 46: This case would flag as FAIL situations when measured air leakage rate was not entered, however this is an expected outcome for cases where proposed design did not have air leakage tested. Recommend removing Case 2

docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+- Case 4: Else if infiltration pressure difference is 0Pa for P_RMR and the total air leakage rate for conditioned and semi-heated zones is not equal to the required proposed design air leakage rate at 0Pa (0.112 is referenced in G3.1.1.4 Modeling Building Envelope Infiltration): `if ( P_RMR.ASHRAE229.infiltration_pressure_difference == 0 ) AND ( building_total_air_leakage_rate != target_air_leakage_rate_p * 0.112 ): FAIL`
+
+- Case 5: Else, infiltration pressure difference is neither 75Pa nor 0Pa for P_RMR: `else: FAIL and raise_warning "90.1 SECTION G3.1.1.4 PRESCRIBED PROPOSED DESIGN INFILTRATION RATE AT 75PA AND METHODOLOGY TO CONVERT AIR LEAKAGE RATE AT 75PA TO AIR LEAKAGE AT WIND PRESSURE. THE MODELED INFILTRATION IS AT PRESSURE DIFFERENTIAL OTHER THAN THE TWO OPTIONS PRESCRIBED BY 90.1. EXCEPTION APPLIES WHEN WHOLE-BUILDING AIR LEAKAGE TESTING IS SPECIFIED DURING DESIGN AND COMPLETED AFTER CONSTRUCTION. THE PROPOSED DESIGN AIR LEAKAGE RATE OF THE BUILDING ENVELOPE SHALL BE AS MEASURED. MANUAL CHECK IS REQUIRED TO VERIFY THAT PROPOSED DESIGN INFILTRATION WAS MODELED CORRECTLY."`
+
```
mkarpman, line 52: To summarize, I think we need the following Cases:
1. PASS when total infiltration_flow_rate equal to target infiltration_flow_rate (no air leakage test was performed). 
2. FAIL when total infiltration_flow_rate not equal to target infiltration_flow_rate but measured air leakage was not entered.
3. PASS or UNDERTERMINED when modeled infiltration is not equal to 0.6 CFM/SF but reflects a measured value correctly converted to infiltration_flow_rate. (UNDERTERMINED outcome would flag for reviewer that completion of air leakage measurements needs to be verified.)
4. FAIL when modeled infiltration is not equal to 0.6 CFM/SF and reflects a measured value incorrectly converted to infiltration_flow_rate. This is a variation of Case 2, so perhaps then can be combined.

docs/section5/Rule5-49.md
xingcx: Updated.

docs/section5/Rule5-49.md
xingcx: Changed to infiltration.infiltration_flow_rate.

docs/section5/Rule5-49.md
xingcx: Updated.

docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+    - For each surface in zone: `for surface_p in zone_p.surfaces:`
+
+      - Check if surface is regulated, add zone total area of building envelope to building total: `if scc_dict_p[surface_p.id] != "UNREGULATED": building_total_envelope_area += sum(surface.area for surface in zone_p.surfaces)`
+
+    - Check if zone is conditioned or semi-heated, add zone air leakage rate to building total: `if zone_conditioning_category_dict_p[zone.id] in [CONDITIONED RESIDENTIAL, CONDITIONED NON-RESIDENTIAL, CONDITIONED MIXED, SEMI-HEATED]: building_total_air_leakage_rate += zone_p.infiltration.infiltration_flow_rate`
+
+      - Check if measured air leakage rate is not entered for zone, raise empty measured air leakage rate flag: `if NOT zone_p.infiltration.measured_air_leakage_rate: empty_measured_air_leakage_rate_flow_flag = TRUE`
+
+      - Else, measured air leakage rate is entered for zone, add measured air leakage rate to building total: `else: building_total_measured_air_leakage_rate += zone_p.infiltration.measured_air_leakage_rate`
+
```
mkarpman, line 40: What if measured air leakage rate is entered only for some of the zones? Do we treat empty field as 0 or as RMD error? Treating it as 0 seems to make sense, but I wonder what others think.

docs/section5/Rule5-49.md
```diff
@@ -0,0 +1,53 @@
...
+- Case 3: else if 1). the building total air leakage rate for conditioned and semi-heated zones is not equal to the required proposed design air leakage rate at 75Pa with a conversion factor of 0.112 as per Section G3.1.1.4, and 2). the measured air leakage rate is entered for all conditioned and semi-heated zones, and 3). the building total air leakage rate is equal to the measured air leakage rate with a conversion factor of 0.112 as per Section G3.1.1.4: `else if ( building_total_air_leakage_rate != target_air_leakage_rate_75pa_p * 0.112 ) AND ( empty_measured_air_leakage_rate_flow_flag == FALSE ) AND ( building_total_air_leakage_rate == building_total_measured_air_leakage_rate * 0.112 ): PASS`
+
```
mkarpman, line 50: Suggest also confirming that ReulsetModelInstance.measured_infiltration_pressure_difference=75

I also don't think we need RulesetModelInstance.is_measured_infiltration_based_on_test. If others agree, I'll post it as RMD issue.

supriyagoel: yes, I dont think we need that boolean. Not sure if that was added for another ruleset

docs/section5/Rule5-49.md
supriyagoel: @mkarpman i guess we can assume zero. If a zone is incorrectly zero, then it would fail case 3 and be flagged as fail.
correct?

supriyagoel: discussed on RDS call - decided to leave the flag as is.

mkarpman (approved):

supriyagoel (approved):

PR #157, opened 2021-09-27
Update schema 2021-09-27

mcneillj:
Updated the schema.json files to the most recent.


charliepnnl (approved):
I have modified the code related to the Rules to conform to the new Schema. Namely, I modified the example RMRs, the Rules and needed, and just a few RDS.

WARNING: To get the pytests passing, I also commented out Juan's pytests in `rct229/ruletest_engine/tests/test_ruletest_json_schemas.py.  I figured that Juan is going to have to make many other changes to the ruletests because of this schema change and those changes should probably be made in a separate PR.

PR #158, opened 2021-09-27
Update _rule_template.md

xingcx:
Add standard output types.


weilixu:
Close this PR because it was updated at #229

PR #162, opened 2021-10-03
Create Rule21-1.md

xingcx:


docs/section21/Rule21-1.md
```diff
@@ -0,0 +1,45 @@
...
+## Rule Logic:  
+
+- For each conditioning component in P_RMR: `for conditioning_component_p in P_RMR.ASHRAE229.conditioning_components:`
+
+  - Check if conditioning component in P_RMR is purchased hot water or steam: `if ( conditioning_component_p IS ExternalFluidSource ) AND ( conditioning_component_p.type in ["HOT_WATER", "STEAM] ):`
+
+    - Set applicability flag: `rule_applicability_check = TRUE`
+
+    - Get matching external fluid source from B_RMR: `external_fluid_source_b = match_data_element(B_RMR, ExternalFluidSources, conditioning_component_p.id)`
+
```
mkarpman, line 29: 1. ServiceWaterHeatingSystem RMR is still in flux, be we may potentially have externalFluidSource apply to it as well. In that case, this RDS would need to differentiate between space heating and SWH.
2. App G is vague regarding handling of projects that have a mixture of purchased heat and on-site plants in the proposed design. So I am not sure we need to match data elements here. One potential interpretation is that if some systems in the proposed desing use purchased steam/hot water, all baseline systems must use purchased steam / hot water. For example, if proposed design has space heating via purchased steam + DOAS that uses on-site furnace, the baseline may be modeled with 100% purchased heat.

docs/section21/Rule21-1.md
```diff
@@ -0,0 +1,45 @@
...
+**Manual Check:** None  
+**Evaluation Context:** Building  
+**Data Lookup:** None  
+**Function Call:** None  
+
+## Rule Logic:  
+
+- For each conditioning component in P_RMR: `for conditioning_component_p in P_RMR.ASHRAE229.conditioning_components:`
+
```
supriyagoel, line 23: @xingcx ASHRAE.229 data group does not have a data element for conditioning_components. We could instead check if ASHRAE229.external_fluid_source is not NULL

docs/section21/Rule21-1.md
supriyagoel: @mkarpman i believe SWH is now approved and i dont believe there are any connects between SHW and externalfluidsource. Regarding (2) we can move all non-mandatory RDS to a separate branch and not discuss them right now. @mcneillj is that a good solution?

xingcx:
This rule is now changed from non-mandatory to mandatory.

docs/section21/Rule21-1.md
xingcx: conditioned_component was an old element when this RDS was create. This rule has been updated.

mkarpman (approved):

docs/section21/Rule21-1.md
```diff
@@ -0,0 +1,47 @@
...
+purchased hot water or steam in both the proposed design and baseline building design. If any system in the proposed design uses purchased hot water or steam, all baseline systems with hot water coils shall use the same type of purchased hot water or steam.  
+**Rule Assertion:** B-RMR = expected value  
+**Appendix G Section:** Section 21 Boiler  
+**Appendix G Section Reference:** Section G3.1.1.1 & G3.1.1.3.1 Building System-Specific Modeling Requirements for the Baseline model  
+
+**Applicability:** All required data elements exist for B_RMR  
+**Applicability Checks:**  
+
+1. B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11, or 12.
+
```
supriyagoel, line 15: 1. @mkarpman I am unsure about what this RDS is doing. I've missed some calls so maybe this was discussed. If P-RMR is modeled with purchased hot water, then  the baseline is not necessarily system type 1,5,7,11 or 12 if the cooling is also purchased chilled water. In that case system 1 is replaced with FPFCU etc.
2. This rule isn't checking anything for the B-RMR, only the P-RMR. Is that not the intent of the RDS?

docs/section21/Rule21-1.md
```diff
@@ -0,0 +1,47 @@
...
+    - Set applicability flag: `rule_applicability_check = TRUE`
+
+**Rule Assertion:**
+
+- Case 1: If P-RMR is modeled with purchased hot water or steam: `if rule_applicability_check: UNDETERMINED and raise_message "P-RMR IS MODELED WITH PURCHASED HOT WATER OR STEAM. VERIFY B-RMR HEATING SOURCE IS MODELED CORRECTLY."`
+
+**Applicability Check 2:**
+
+1. Rule is applicable if P-RMR is modeled with purchased hot water or steam: `if rule_applicability_check: is_applicable = TRUE`
+
```
mkarpman, line 40: should there be else is_applicable = FALSE, or is it assumed without saying?

docs/section21/Rule21-1.md
mkarpman: @supriyagoel I agree with your first comment - we should remove this applicability check.
Aside from that, we agreed to limit implementation of this rule to just checking if it applies to the project, which is determined based on P-RMR. Notice that the rule doesn't have PASS or FAIL outcome. The only options are either (A) "UNDERTERMINED" (meaning that the manual review is required) + is_applicable=TRUE, or (B) no outcome except is_applicable=FLASE

supriyagoel: makes sense. @xingcx please remove this applicability check

docs/section21/Rule21-1.md
```diff
@@ -0,0 +1,45 @@
...
+1. check_purchased_chw_hhw()
+
+## Rule Logic:  
+
+- Check if P-RMR is modeled with purchased cooling or purchased hot water/steam: `purchased_chw_hhw_status_dict = check_purchased_chw_hhw(P_RMR)`
+
+  - If P-RMR is not modeled with purchased hot water or steam: `if NOT purchased_chw_hhw_status_dict["PURCHASED_HEATING"]:`
+
+    - Check if B-RMR is modeled with any external fluid source that is hot water or steam type: `if external_fluid_source_b.type in ["HOT_WATER", "STEAM"] for external_fluid_source_b in B_RMR.ASHRAE229.external_fluid_source:`
+
```
mkarpman, line 28: I think this should be checking P_RMR.

mkarpman: I think this RDS should have two outcomes:
1. If statement above is true (when applied to P-RMR), the outcome should be UNDERTERMINED.
2. If statement above is false (when applied to P-RMR), the outcome should be NOT APPLICABLE.

I don't think the rule assertions below reflect this.

mkarpman (changes requested):

docs/section21/Rule21-1.md
xingcx: Updated.

supriyagoel (approved):

mkarpman (approved):

docs/section21/Rule21-1.md
```diff
@@ -0,0 +1,42 @@
...
+  - Else, P-RMR is modeled with purchased hot water/steam, continue to rule logic: `else: CHECK_RULE_LOGIC`
+
+## Rule Logic:  
+
+- Check if B-RMR is modeled with any air-side system that is Type-1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 1a, 3a, 7a, 8a, 11a, 12a, 13a, i.e. any air-side system using heating source other than purchased hot water or steam, set check flag to True: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-1", "SYS-2", "SYS-3", "SYS-4", "SYS-5", "SYS-6", "SYS-7", "SYS-8", "SYS-9", "SYS-10", "SYS-11", "SYS-12", "SYS-13", "SYS-1A", "SYS-3A", "SYS-7A", "SYS-8A", "SYS-11A", "SYS-12A", "SYS-13A"]): check_flag = TRUE`
+
+**Rule Assertion:**
+
+- Case 1: If check flag is True: `if check_flag: FAIL`
```
supriyagoel, line 38: @xingcx everything makes sense and looks good except for this statement 'check_flag: FAIL'. Not sure wwwhat this means.

mkarpman: I agree with Supriya

mkarpman (approved):

mkarpman (approved):

docs/section21/Rule21-1.md
supriyagoel: @xingcx - this is the last remaining fix for this RDS. Any chance you could update this?
Yopu could simply rename check_flag to 'not modeled with purchased HW' and set that to TRUE in line 34.

supriyagoel (approved):

PR #33, opened 2021-04-26
Update Rule12-4.md

xingcx:

PR #164, opened 2021-10-03
Create Rule21-3.md

xingcx:


docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,44 @@
...
+
+**Manual Check:** None  
+**Evaluation Context:** Building  
+**Data Lookup:** None  
+**Function Call:** None  
+
+## Rule Logic:  
+
+- For each fluid loop in B_RMR: `for fluid_loop_b in B_RMR.ASHRAE229.fluid_loops:`
+
```
mkarpman, line 23: Seems like we also need to have a check that there is no more than 1 HW loop in the baseline. I don't think there is a rule like that now.

mkarpman (changes requested):

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,43 @@
...
+# Boiler - Rule 21-3  
+
+**Rule ID:** 21-3  
+**Rule Description:** Plant capacities shall be based on coincident loads.
```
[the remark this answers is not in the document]
xingcx: Rule 22-35 is now added to check CHW plant capacity is modeled based on coincident loads.

docs/section21/Rule21-3.md
xingcx: Rule is now added to check that no more than 1 HW loop is modeled in the baseline, Rule 21-18.

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,43 @@
...
+
+      - Case 1: For each heating fluid loop, if heating design and control component is sized using coincident load: `if heating_design_and_control_b.is_sized_using_coincident_load: PASS`
+
+      - Case 2: Else: `else: FAIL`
+
+**[Back](../_toc.md)**
+
+**Note:**
+
+1. Do we need to expand the logic to cover all cases, e.g. if logic cannot find any FluidLoop in B-RMR, or any FluidLoop that is heating type? This needs to be covered in either the logic or in applicability check.
```
mkarpman, line 42: The applicability check is that B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11, or 12 (line 13 above). It is my understanding that this will actually involve referring to the appropriate air-side HVAC rules (or a function) which will involve verifying that HVAC systems in b-RMR are confitured as required in order to be classified as systems 1,5,7,11 or 12

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,43 @@
...
+2. Dependency, B-RMR air-side system that is Type-1, 5, 7, 11, or 12 is modeled correctly with heating hot water plant.
```
mkarpman, line 43: RMR does not have data element tagging a system as Type 1, 5, 7, etc. So it would be up to the RDS (of function) mentioned above to look at the configuration of HVAC systems in the baseline and map them to App G baseline system types.

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,43 @@
...
+
+**Manual Check:** None  
+**Evaluation Context:** Building  
+**Data Lookup:** None  
+**Function Call:** None  
+
+**Applicability Check:**
+
+1. Rule is applicable if B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11, or 12: `PLACEHOLDER`
+
```
mkarpman, line 23: Another applicability check is that proposed design doesn't have purchased heating

docs/section21/Rule21-3.md
xingcx: We might need to set the boundary for what we want to check under water-side and what we want to check under air-side and make sure we cover all cases.

docs/section21/Rule21-3.md
xingcx: Added.

docs/section21/Rule21-3.md
xingcx: Right I added the applicability check but did not delete this note. This note is now deleted.

docs/section21/Rule21-3.md
mkarpman: I don't think we need the first check (that there is at least one air-side system of Type 1, 5, etc.) since this is a prerequisite for Rule 21-1,

mkarpman (approved):

docs/section21/Rule21-3.md
mkarpman: Also, I don't think we have "NOT APPLICABLE" outcome... should be "UNDERTERMINED"

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,50 @@
...
+**Manual Check:** None  
+**Evaluation Context:** Building  
+**Data Lookup:** None  
+**Function Call:** None  
+
+**Applicability Check:**
+
+1. Rule is applicable if B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11, or 12: `PLACEHOLDER`
+2. B-RMR is not modeled with purchased heating: `if RULE-21-1 == "NOT APPLICABLE":`
+
```
supriyagoel, line 25: 21-1 checke the P-RMR not the B-RMR. Correct?

supriyagoel (approved):

docs/section21/Rule21-3.md
xingcx: 21-1 is updated.

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,53 @@
...
+**Evaluation Context:** Building  
+**Data Lookup:** None  
+**Function Call:**  
+
+1. check_purchased_chw_hhw()
+
+**Applicability Check:**
+
+1. Pass Rule 21-1: `if rule-21-1 == "PASS":`
+
```
mkarpman, line 26: This should be calling ceck_purchase_chw_hw function instead of referring to the rule.

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,53 @@
...
+2. B-RMR is modeled with at least one air-side system that is Type-1a, 7, 11, 12 or that is Type-1, 5, 7, 11, 12:
+
+  - Check if P-RMR is modeled with purchased cooling or purchased hot water/steam: `purchased_chw_hhw_status_dict = check_purchased_chw_hhw(P_RMR)`
+
```
mkarpman, line 30: I think this function call should be moved to Line 25.

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,53 @@
...
+  - Check if fluid loop type is heating: `if fluid_loop_b.type == "HEATING":`
+
+    - Get heating design and control component for fluid loop: `heating_design_and_control_b = fluid_loop_b.heating_design_and_control`
+
+      **Rule Assertion - Component:**
+
+      - Case 1: For each heating fluid loop, if heating design and control component is sized using coincident load: `if heating_design_and_control_b.is_sized_using_coincident_load: PASS`
+
+      - Case 2: Else: `else: FAIL`
+
```
mkarpman, line 52: There should also be Case 3: if purchased_chw_hhw_status_dict = check_purchased_chw_hhw(P_RMR) indicates that project has purchased HW, the outcome would be NOT APPLICABLE

mkarpman (changes requested):

docs/section21/Rule21-3.md
xingcx: Updated.

docs/section21/Rule21-3.md
xingcx: Updated.

docs/section21/Rule21-3.md
xingcx: This is already included in the applicability check. If P-RMR has purchased HW, this rule is not applicable.

docs/section21/Rule21-3.md
xingcx: As per discussion, each component will have 4 outcome, Pass, Fail, Undetermined and Not Applicable.

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,59 @@
...
+**Applicability Check:**
+
+1. Check if P-RMR is modeled with purchased cooling or purchased hot water/steam: `purchased_chw_hhw_status_dict = check_purchased_chw_hhw(P_RMR)`
+
+  - If P-RMR is modeled with purchased hot water/steam, rule is not applicable: `if purchased_chw_hhw_status_dict["PURCHASED_HEATING"]: rule_applicability_flag = FALSE`
+
+  - Else, P-RMR is not modeled with purchased hot water/steam, continue to next applicability check: `if NOT purchased_chw_hhw_status_dict["PURCHASED_HEATING"]:`
+
+2. B-RMR is modeled with at least one air-side system that is Type-1a, 7, 11, 12 or that is Type-1, 5, 7, 11, 12:
+
```
supriyagoel, line 32: @xing not sure I understand why we have listed the systems twice. Is one set for the modified baseline purchased H system? If yes, maybe we can come up with a different terminology for them maybe 'G3.1.1.3.3. baseline systems for PCHW and PHW'

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,59 @@
...
+**Rule ID:** 21-3  
+**Rule Description:** Heating hot water plant capacity shall be based on coincident loads.  
+**Rule Assertion:** B-RMR FluidLoop.heating_design_and_control: is_sized_using_coincident_load = True  
+**Appendix G Section:** Section 21 Boiler  
+**Appendix G Section Reference:** Section G3.1.2.2 Building System-Specific Modeling Requirements for the Baseline model  
+
+**Applicability:** All required data elements exist for B-RMR  
+**Applicability Checks:**  
+
+1. P-RMR is modeled with purchased heating.
```
mkarpman, line 13: This should be "not" modeled with purchased heating. 
I also think it should apply to U-RMR, but P-RMR here and on the following lines.

docs/section21/Rule21-3.md
```diff
@@ -0,0 +1,59 @@
...
+2. B-RMR is modeled with at least one air-side system that is Type-1a, 7, 11, 12 or that is Type-1, 5, 7, 11, 12.
+
```
mkarpman, line 15: Should this just be 1a, 1, 5, 7, 11 and 12? Why are there two separate lists?

docs/section21/Rule21-3.md
mkarpman: I agree. I would just use one list here 1a, a, 5,7,11,12

docs/section21/Rule21-3.md
mkarpman: I think Lines #33 - 43 can be replaces with 
 - Check if B-RMR is modeled with at least one air-side system that is Type-1a, 1, 5, 7, 11 or 12, continue to rule logic: `PLACEHOLDER`

    - Else, rule is not applicable: `else: rule_applicability_flag = FALSE`

docs/section21/Rule21-3.md
mkarpman: I agree that it's part of applicability check, but I think we agreed a few calls ago to provide explicit "NOT APPLICABLE" outcome in such cases? (Prior to that, the RDS would produce no outcome at all in such cases.) Let's discuss on the call and, if agreed, add the following Case 3: If rule_applicability_flag = FALSE: "NOT APPLICALBE"

docs/section21/Rule21-3.md
supriyagoel: as discussed on the RDS call now we do not need to check if the P RMR is modeled with P cooling, we can simply reference the function and check if the baseline system type is 1a, 7a, etc. etc.

docs/section21/Rule21-3.md
supriyagoel: The applicability check looks good now!

docs/section21/Rule21-3.md
supriyagoel: good decision to add the CZ  check for System 11.

supriyagoel (approved):

docs/section21/Rule21-3.md
mkarpman: I think we need to include Not Applicable outcome?

mkarpman (approved):

PR #165, opened 2021-10-03
Create Rule21-4.md

xingcx:


docs/section21/Rule21-4.md
```diff
@@ -0,0 +1,45 @@
...
+
+**Applicability Check:**
+
+1. Rule is applicable if B-RMR is modeled with at least one boiler: `if rule_applicability_check: is_applicable = TRUE`
+
+**[Back](../_toc.md)**
+
+**Notes:**
+
+1. Do we need to check if baseline system is 1,5,7,11,12, or does this rule apply to all boilers in B-RMR?
```
mkarpman, line 44: It allpies to all baseline systems that have boilers

docs/section21/Rule21-4.md
```diff
@@ -0,0 +1,45 @@
...
+
+**Manual Check:** None  
+**Evaluation Context:** Building  
+**Data Lookup:** None  
+**Function Call:** None  
+
+## Rule Logic:  
+
+- For each conditioning component in B_RMR: `for conditioning_component_b in B_RMR.ASHRAE229.conditioning_components:`
+
```
supriyagoel, line 23: @xingcx i dont quite follow what 'conditioning_components' is. 


@mkarpman  - i am not sure how this test would work. My suggestion is that to verify that the baseline has a boiler, we would have to reference a function we haven't written yet to identify the baseline system type(s) in the B-RMR.
The function would be able to go through each zone and identify the baseline system type for that zone.
so if the outcome of that function identifies system types 1,5,7,11 or 12 for any zone, we need to verify that the heating source is a natural draft boiler for that zone.

docs/section21/Rule21-4.md
```diff
@@ -0,0 +1,45 @@
...
+  - Check if conditioning component is boiler: `if conditioning_component_b IS Boiler:`
+
+    - Check if boiler serves hot water loop: `if conditioning_component_b.loop.type == "HEATING":`
+
+      - Set applicability flag: `rule_applicability_check = TRUE`
+
+        **Rule Assertion:**
+
+        - Case 1: For each boiler, if it is modeled as natural draft type: `if conditioning_component_b.draft_type == "NATURAL": PASS`
+
```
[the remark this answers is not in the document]
supriyagoel: @mkarpman i understand what you're proposing but i am not sure how we would do that.

docs/section21/Rule21-4.md
xingcx: @supriyagoel The "conditioning_component" was an old schema element before it has separate boilers, chillers, pumps, etc. This rule is not mandatory so was not updated since I first created the PR.

docs/section21/Rule21-4.md
```diff
@@ -0,0 +1,45 @@
...
+**Rule ID:** 21-4  
+**Rule Description:** When baseline building does not use purchased heat, baseline systems 1,5,7,11,12 shall be modeled with natural draft boilers.  
+**Rule Assertion:** B-RMR  
+**Appendix G Section:** Section 21 Boiler  
+**Appendix G Section Reference:** Section G3.1.3.2 Building System-Specific Modeling Requirements for the Baseline model  
+
+**Applicability:** All required data elements exist for B_RMR  
+**Applicability Checks:**  
+
```
supriyagoel, line 12: do we also need the baseline system type = 1, 5, 7, 11 (except CZ 0-3a), 12, 1a, 7a, 11a (expect cz 0-3a) check over here?

mkarpman: I agree - it should be system type 1,5,7,11,12, 1a, 7a, 11a and 12a. 

Also, I would consider assigning different letters to Systems 11 and 11a in CZ 0-3a vs all other CZ. For example, 11.1 
 would be for System 11 w/electric resistance (CZ 0-3a) and 11.2 would be System 11 w/boiler, for all other CZs.

docs/section21/Rule21-4.md
mkarpman: This is non-mandatory rule so we don't need to implement the logic, just determine whether it is applicable. 
That said, since projects often have boilers in the baseline, I think it may be worthwhile to make this rule mandatory. 
The RDS currently refers to "conditioning_component" that I don't think exists in the current schema. So, if we decide to implement it, the logic would be to check that each boiler in the B-RMR that references a fluid loop of type "heating" is natural draft.

mkarpman (approved):

supriyagoel (approved):

PR #26, opened 2021-04-19
Update Rule12-5.md

xingcx:

PR #166, opened 2021-10-03
Create Rule21-5.md

xingcx:


docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,65 @@
...
+**Rule Description:** The baseline building design boiler plant shall be modeled as having a single boiler if the baseline building design plant serves a conditioned floor area of 15,000sq.ft. or less, and as having two equally sized boilers for plants serving more than 15,000sq.ft.  
+**Rule Assertion:** B-RMR  
+**Appendix G Section:** Section 21 Boiler  
+**Appendix G Section Reference:** Section G3.1.3.2 Building System-Specific Modeling Requirements for the Baseline model  
+
+**Applicability:** All required data elements exist for B_RMR  
+**Applicability Checks:**  
+
+1. Baseline number of HW Loops > 0 and number of boilers in HW loop > 0.  
+
```
mkarpman, line 14: Few thoughts:
1. The rule should have rule dependency that baseline HVAC system type was established correctly (i.e., there should indeed be a boiler in the baseline). 
2. I think having more than one hot water loop in the baseline should be flagged as an error.

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,65 @@
...
+  - Set applicability flag: `rule_applicability_check = TRUE`
+
+  - Get all boilers on the loop: `boilers_array = loop_boiler_dict[hhw_loop]`
+
+    **Rule Assertion:**
+
+    - Case 1: If total conditioned area under the loop is 15,000sq.ft. or less and if only one boiler in on the loop: ``if ( loop_area_dict[hhw_loop] <= 15000 ) AND ( boilers_array.size == 1 ): PASS`
+
+    - Case 2: Else if total conditioned area under the loop is more than 15,000sq.ft. and two boilers are on the loop and the two boilers are sized equally: `else if ( loop_area_dict[hhw_loop] > 15000 ) AND ( boilers_array.size == 2 ) AND ( boilers_array[0].design_capacity == boilers_array[1].design_capacity ): PASS`
+
```
mkarpman, line 53: Should we use design_capacity or rated_capacity attribute?

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,65 @@
...
+**Applicability Check:**
+
+1. Rule is applicable if B-RMR is modeled with at least one hot water loop: `if rule_applicability_check: is_applicable = TRUE`
+
+**[Back](../_toc.md)**
+
+**Notes:**
+
+1. Indirectly conditioned zones does not have HVAC system, and may be adjacent to multiple directly conditioned zones served by different hvac systems. how to assign indirectly conditioned zone area?
```
mkarpman, line 64: Indirectly conditioned zones would be classified as conditioned, so I think we are good with that.

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,65 @@
...
+2. Do we need to also check boiler.rated_capacity?
```
mkarpman, line 65: I am not clear on this either.

docs/section21/Rule21-5.md
supriyagoel: as discussed on the RDS call - need to have a separate RDS to verify that the baseline HVAC is established correctly. SG will upload the TCDs for baseline HVAC checks and Xing will update the applicability check to verify the baseline system typee.

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,70 @@
...
+
+  - Check if zone has HVAC terminals, get zone terminals: `if zone_b.terminals: zone_terminals_b = zone_b.terminals`
+
+    - For each zone terminal: `for zone_terminal_b in zone_terminals_b:`
+
+      - Check if zone area has not been added to any heating hot water loop: `if NOT zone_flag:`
+
+        - Get HVAC system serving the terminal: `hvac_b = zone_b.terminals.served_by_heating_ventilation_air_conditioning_systems`
+
+          - Get heating system from HVAC system: `heating_system_b = hvac_b.heating_system`
```
supriyagoel, line 42: @xingcx as mentioned on another RDS, the zone > terminal connection has not been finalized and this might have to be changed in the future.

mkarpman: So I think the logic should be as follows:
1. Identified all zones that are conditioned based on get_zone_conditioning_category(B_RMR) (List 1)_
2. Identify all zones that are EITHER (a) connect to HW fluid loop through Terminal or (b) through HeatingVentilationAirConditioningSystem -> HeatingSystem. (List 2) I don't think you do (a)?
3. Calculate the total area of all zones that are included in both List 1 and List 2

Once the terminal connections are finalized, we should revisit to confirm the logic.

docs/section21/Rule21-5.md
mkarpman: I think we may need design capacity for the baseline, unless we say that since the baseline is autosized design and rated capacity should be equal. It doesn't matter here which one we check, but would metter for RDS that establishes baseline boiler efficiency which will depend on capacity.

docs/section21/Rule21-5.md
xingcx: I had the original question under the assumption that baseline may have more than one HHW loop, e.g. one for purchased HHW and one for boiler loop. We have since clarified that baseline shall have either purchased loop or boiler loop. So I think this question is irrelevant now.

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,70 @@
...
+
+**Applicability Check:**
+
+1. B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11, or 12: `if PLACEHOLDER_AIRSIDE_SYSTEM_RULE-X-XX:`
+2. B-RMR is not modeled with purchased heating: `if NOT Rule-21-1:`
+
+## Rule Logic:  
+
+- Get zone conditioning category dictionary for B_RMR: `zcc_b = get_zone_conditioning_category(B_RMR)`
+
```
supriyagoel, line 31: Based on MK's comment below, we need to get terminals only of the conditioned zones.

docs/section21/Rule21-5.md
supriyagoel: other than (1), i believe the RDS is doing everything else on the list.

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,70 @@
...
+        - Get HVAC system serving the terminal: `hvac_b = zone_b.terminals.served_by_heating_ventilation_air_conditioning_systems`
+
+          - Get heating system from HVAC system: `heating_system_b = hvac_b.heating_system`
+
+            - Check if heating system has heating hot water loop, get heating hot water loop: `if heating_system_b.hot_water_loop: hhw_loop_b = heating_system_b.hot_water_loop`
+
+              - Save zone area to heating hot water loop: `loop_area_dict[hhw_loop_b] += sum(space.floor_area for space in zone_b.spaces)`
+
+              - Set zone flag as True to avoid double counting the zone area: `zone_flag == TRUE`
+
```
supriyagoel, line 49: @xingcx need a step before here to find the boiler on the hot water loop.

docs/section21/Rule21-5.md
mkarpman: @supriyagoel I agree, don't see where zcc_b is used - the RDS just seems to be doing (2).
Another issue is that my logic above does not account for indirectly-conditioned spaces which may be included in List 1 but not List 2 - e.g., if an indirectly conditioned space is not by any heating system. I think area of such zones should also be counted.

supriyagoel: @mkarpman correct, I think Xing pointed that out as well somewhere. Not sure how that would work though - we need a data element to identify if the indirectly conditioned zone is being conditioned by a zone that falls under (2). Zones have a transfer air flow rate but no data element that would identify where the transfer air is coming from. Any thoughts?

mkarpman: Few thoughts:
1. I think we agreed to have a function that calculates conditioned floor area. (This is also mentioned in 21-10.) So I would include a call to this future function here and in 21-10, and move the logic for determining the conditioned floor area that we started developing here to that function. To support the functionality that we need, perhaps the function should also or alternatively return the list of conditioned zones (List 1 per my comment from Nov 16 above).
2. To your question about indirectly conditioned zones, I would include area of zones that do not have terminals but that are included in the list of conditioned zones produced by the function discussed in #1. I think this will be correct in most cases and will be a conservative assumption as will result in more stringent requirements. The RDS may generate a message with the list of zones that have no terminals that are included in the conditioned area total, and the modeler can then explain to reviewer if the calculation was off. (If it is off.)

supriyagoel: @mkarpman that makes sense. I think having the list of conditioned zones which do not have terminals will be a good indication of an indirectly conditioned space.

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,67 @@
...
+**Evaluation Context:** Each Data Element  
+**Data Lookup:** None  
+**Function Call:**
+
+  1. get_zone_conditioning_category()
+  2. get_loop_zone_list_w_area()
+
+**Applicability Check:**
+
+1. Pass Rule 21-1: `if rule-21-1 == "PASS":`
```
mkarpman, line 26: This should be calling the function, not the rule.

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,67 @@
...
+2. B-RMR is modeled with at least one air-side system that is Type-1a, 7, 11, 12 or that is Type-1, 5, 7, 11, 12:
+
+  - Check if P-RMR is modeled with purchased cooling or purchased hot water/steam: `purchased_chw_hhw_status_dict = check_purchased_chw_hhw(P_RMR)`
+
+    - If P-RMR is not modeled with purchased cooling (Note: Passing Rule 21-1 means P-RMR is not modeled with purchased heating.): `if NOT purchased_chw_hhw_status_dict["PURCHASED_COOLING"]:`
+
```
mkarpman, line 32: This should be moved to Line #26

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,67 @@
...
+      - Check if B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11 or 12: `PLACEHOLDER`
+
+    - Else, P-RMR is modeled with purchased cooling: `else:`
+
+      - Check if B-RMR is modeled with at least one air-side system that is Type-1a, 7, 11 or 12: `PLACEHOLDER`
+
```
mkarpman, line 38: I forgot why we need to check this. This will only be the case if project has purchased heating, but then the applicability check will not be met and the outcome will be Not Applicable?

docs/section21/Rule21-5.md
xingcx: Updated.

docs/section21/Rule21-5.md
xingcx: Updated.

docs/section21/Rule21-5.md
xingcx: If baseline does not have system type that is with HHW loop then this rule is not applicable.

docs/section21/Rule21-5.md
xingcx: Updated.

docs/section21/Rule21-5.md
supriyagoel: same question as 21-3, 'B-RMR is modeled with at least one air-side system that is Type-1a, 7, 11, 12 or that is Type-1, 5, 7, 11, 12.' is a little confusing. I assume Type-1, 5, 7, 11, 12 are the typical baseline systems and 'Type-1a, 7, 11, 12' are the modified baseline systems which have PHW or PCHW in proposed? If yes, we should instead use terminology 'Baseline System Type based on G3.1.1.3.3.' @mkarpman to confirm.

mkarpman: Yes, I agree

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,70 @@
...
+**Data Lookup:** None  
+**Function Call:**
+
+  1. get_zone_conditioning_category()
+
+**Applicability Check:**
+
+1. B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11, or 12: `if PLACEHOLDER_AIRSIDE_SYSTEM_RULE-X-XX:`
+2. B-RMR is not modeled with purchased heating: `if NOT Rule-21-1:`
+
```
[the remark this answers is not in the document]
mkarpman: Same as for 21-3, I think this check should be applied to U-RMR

docs/section21/Rule21-5.md
mkarpman: I think rows 35-43 can be merged to say
Check if B-RMR is modeled with at least one air-side system that is Type-1, 1a, 5, 7, 11 or 12, continue to rule logic: PLACEHOLDER
Else, rule is not applicable: else: rule_applicability_flag = FALSE

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,77 @@
...
+    - Classify zone as indirectly conditioned and add zone total area to total area served by heating loop: `heating_loop_conditioned_zone_area += SUM(space.floor_area for space in zone.spaces)` (Note XC, in this logic zone might be adjacent to directly conditioned zones that are cooling only, but its area will still be added to HHW loop)
+
+**Rule Assertion:**
+
+- Case 1: If baseline building design plant serves a conditioned floor area of 15,000sq.ft. or less, and if only one boiler is modeled in B_RMR: `if ( heating_loop_conditioned_zone_area <= 15000 ) AND ( LEN(B_RMR.ASHRAE229.boilers) == 1 ): PASS`
+
+- Case 2: Else if baseline building design plant serves a conditioned floor area more than 15,000sq.ft. and two boilers are modeled in B_RMR and the two boilers are sized equally: `else if ( heating_loop_conditioned_zone_area > 15000 ) AND ( LEN(B_RMR.ASHRAE229.boilers) == 2 ) AND ( B_RMR.ASHRAE229.boilers[0].rated_capacity == B_RMR.ASHRAE229.boilers[1].rated_capacity ): PASS`
+
+- Case 3: Else: `else: FAIL`
+
```
mkarpman, line 72: Should we add Case 4 with NOT APPLICALBE outcome?

supriyagoel (approved):

docs/section21/Rule21-5.md
```diff
@@ -0,0 +1,70 @@
...
+  3. get_loop_zone_list_w_area()
+
+**Applicability Check:**
+
+- Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`
+
+  - Check if B-RMR is modeled with any air-side system that is Type-11 or 11a: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-11", "SYS-11A"]):`
+
+    - Check if B-RMR is in climate zones other than 0 through 3A, continue to rule logic: `if NOT B_RMR.ASHRAE229.weather.climate_zone in ["0A", "0B", "1A", "1B", "2A", "2B", "3A"]: CHECK_RULE_LOGIC`
+
```
mkarpman, line 31: I think the CZ check for System 11 should be part of the function - see my comment on the function. Then applicability check in row 36 would include Systems 11.2 and 11.2a

docs/section21/Rule21-5.md
xingcx: As discussed on call, we will skip NOT APPLICABLE outcome in RDS.

docs/section21/Rule21-5.md
xingcx: Updated.

mkarpman (approved):

supriyagoel (approved):

PR #167, opened 2021-10-06
data table look up function and tests for Table G3.6

vijayvardhan94:
Includes Table look up functions and tests for Table G3.6.

### DIFF
--- a/rct229/data_fns/table_G3_6_fns.py
+++ b/rct229/data_fns/table_G3_6_fns.py
@@ -1,4 +1,3 @@
-import rct229
 from rct229.data import data
 from rct229.data_fns.table_utils import find_osstd_table_entry
 from rct229.schema.config import ureg
@@ -34,7 +33,8 @@ def table_G3_6_lookup(building_exterior_type_enum_val):
     Returns
     -------
     dict
-        { lpd: Quantity - The lighting power density in watt per square foot given by Table G3.6, linear_lpd: Quantity - The lighting power density in watt per linear foot given by Table G3.6 }
+        { lpd: Quantity - The lighting power density in watt per square foot given by Table G3.6,
+        linear_lpd: Quantity - The lighting power density in watt per linear foot given by Table G3.6 }
 
     """
     building_exterior_type = building_exterior_enumeration_to_lpd_space_type_map[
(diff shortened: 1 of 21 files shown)


rct229/data_fns/table_G3_6_fns.py
```diff
@@ -0,0 +1,60 @@
+import rct229
```
charliepnnl, line 1: I don't think line 1 is needed.

rct229/data_fns/table_G3_6_fns.py
```diff
@@ -0,0 +1,60 @@
...
+    required by ASHRAE 90.1 Table G3.6
+    Parameters
+    ----------
+    building_exterior_type : str
+        One of the ExteriorLightingAreas2019ASHRAE901TableG36 enumeration values
+
+    Returns
+    -------
+    dict
+        { lpd: Quantity - The lighting power density in watt per square foot given by Table G3.6, linear_lpd: Quantity - The lighting power density in watt per linear foot given by Table G3.6 }
```
charliepnnl, line 37: Black won't reformat inside a comment, so you should break this into multiple lines.

rct229/data_fns/table_G3_6_fns.py
```diff
@@ -0,0 +1,60 @@
...
+        building_exterior_type_enum_val
+    ]
+
+    osstd_entry = find_osstd_table_entry(
+        [("building_exterior_type", building_exterior_type)],
+        osstd_table=data["ashrae_90_1_table_3_6"],
+    )
+    watts_per_ft2 = osstd_entry["w/ft^2"]
+    watts_per_linear_ft = osstd_entry["w/ft"]
+    if watts_per_linear_ft is None:
```
charliepnnl, line 50: Worth considering as a alternate method of variable assignment:
```
lpd = watts_per_ft2 * ureg("watt / foot**2") if watts_per_linear_ft is not None else None
linear_lpd = watts_per_linear_ft * ureg("watt / foot") if watts_per_linear_ft is not None else None
```
The important thing is for the logic to be as clear as possible to anyone reading the code. The fact that lines 51 and57, for example, are the same is a tip off that the code could be simplified.

rct229/data_fns/table_G3_6_fns_test.py
```diff
@@ -0,0 +1,113 @@
...
+    building_exterior_enumeration_to_lpd_space_type_map,
+    table_G3_6_lookup,
+)
+from rct229.data_fns.table_utils import (
+    check_enumeration_to_osstd_match_field_value_map,
+    find_osstd_table_entry,
+)
+from rct229.schema.config import ureg
+
+watts_per_ft2 = ureg("watt / foot**2")
```
charliepnnl, line 15: It is customary in python to use ALL_CAPS for constants such as this.

rct229/data_fns/table_G3_6_fns.py
vijayvardhan94: Did you mean - 

lpd = watts_per_ft2 * ureg("watt / foot**2") if watts_per_ft2 is not None else None
linear_lpd = watts_per_linear_ft * ureg("watt / foot") if watts_per_linear_ft is not None else None

PR #177, opened 2021-10-21
Add unit convention documentation scripts.

jugonzal07:
This PR adds the capability to generate unit_conventions.json files to keep track of TCD vs RMR unit conventions. It also includes updated documentation in the README for the RCT spreadsheet to account for these updates and removes some deprecated functionality.


jugonzal07:
forgot to do the black formatting. Will do soon!

PR #184, opened 2021-11-08
RS/CH/get_zone_conditioning_category

charliepnnl:
Please ignore the name of the branch. This PR is primarily for the implementation of the get_zone_conditioning_category ruleset function.

This also includes an implementation of  get_opaque_surface_type() and a complete reworking of the pint units registry for the project.


charliepnnl:
The RDS markdown for get_zone_conditioning_category needs to be updated because of substantial changes to the schema.  The plan is for @xingcx review the python code to check the logic of the function. Once there is agreement on that, we can update the RDS to match the implementation.

jugonzal07:
Hi Charlie,

Quick question about the pint changes.

- I saw C = [Celsius ](https://github.com/pnnl/ruleset-checking-tool/pull/184/files#diff-ff0d7c8c6b214d9a0da0c891d8b4033665b78590b4de12feace7d28d1ecab23cR141)but also [coulomb](https://github.com/pnnl/ruleset-checking-tool/pull/184/files#diff-ff0d7c8c6b214d9a0da0c891d8b4033665b78590b4de12feace7d28d1ecab23cR228). Does this cause any problems to have the same representation for each?

charliepnnl:
Good catch!

I will delete the C for coulomb. It is only used in the definition of abcoulomb, which can also be deleted.

Many more units can, and probably should, be deleted.  I would like to discuss this in the team meeting this week.

jugonzal07 (approved):
Looks good! I didn't realize how tricky schema updates must be for you with all the enumeration changes. Hopefully they settle. Wonder if there's an alternative way to store them all as constants somewhere so you just have to change them in one place as they update. Might be more trouble than it's worth though.

charliepnnl:
@jugonzal07  Actually, enumerations changes aren't bad. The hard part here was all the changes to the structure of the schema.

rct229/ruleset_functions/get_zone_conditioning_category_dict.py
```diff
@@ -0,0 +1,228 @@
...
+    # semi-heated zones
+    directly_conditioned_zone_ids = []
+    semiheated_zone_ids = []
+    for zone in building_zones:
+        zone_id = zone["id"]
+
+        zone_hvac_systems = [
+            building_hvac_systems_map[hvac_system_id]
+            for hvac_system_id in zone[
+                "served_by_heating_ventilation_air_conditioning_systems"
```
xingcx, line 56: Possible change in the schema from zone.served_by_heating_ventilation_air_conditioning_systems to zone.terminal.served_by_heating_ventilation_air_conditioning_systems but has not been finalized yet.

PR #187, opened 2021-11-21
Create Rule22-3.md

xingcx:


supriyagoel (approved):
Looks good

docs/section22/Rule22-3.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Rule Logic:  
+
+- For each chiller in B_RMR: `for chiller_b in B_RMR.ASHRAE229.chillers:`
+
+  - Get cooling loop that chiller servers: `chw_loop_b = chiller_b.cooling_loop`
+
+    - If cooling loop has not been saved in CHW loop array: `if chw_loop_b NOT in chw_loop_array:`
+
+      - Save cooling loop to CHW loop array: `chw_loop_array.append(chw_loop_b)`
```
supriyagoel, line 34: the RDS looks good. I am not sure if we need this chiller check after we have done the applicability check for system 7,8,11,12,13. i.e. once we know that the CHW loop is associated with one of those systems, then we dont need the chiller check. 
we could instead start with identifying zones served by system 7,8,11,12 or 13 and then find the associated fluid loop serving those HVAC systems and then check the reset control. i do not know if one approach is better than tthe other or unnecessarily complicated. would help to know what @mcneillj or @mkarpman think.

docs/section22/Rule22-3.md
```diff
@@ -0,0 +1,46 @@
...
+**Rule ID:** 22-3  
+**Rule Description:** For Baseline chilled water loop that is not purchased cooling, chilled-water supply temperature shall be reset based on outdoor dry-bulb temperature.  
+**Rule Assertion:** B-RMR = expected value  
+**Appendix G Section:** Section 22 CHW&CW Loop  
+**Appendix G Section Reference:** Section G3.1.3.9 Chilled-water supply temperature reset (System 7, 8, 11, 12 and 13)  
+
+**Applicability:** All required data elements exist for B_RMR  
+**Applicability Checks:**  
+
+1. B-RMR is modeled with at least one air-side system that is Type-7, 8, 11, 12, or 13
```
mkarpman, line 13: Please remove system 11 from the list because it is subject to Exception 1 to G3.1.3.9 which we are covering in Rule 22-5

docs/section22/Rule22-3.md
```diff
@@ -0,0 +1,46 @@
...
+
+- Case 1: For each fluid loop that chiller serves, if supply temperature is reset based on outdoor dry-bulb temperature: `if ( loop.cooling_or_condensing_design_and_control.temperature_reset_type == "OUTSIDE_AIR_RESET" for loop in chw_loop_array): PASS`
+
+- Case 2: Else: `else: FAIL`
+
+**[Back](../_toc.md)**
+
+**Notes:**
+
+1. Need to check secondary loop?
```
mkarpman, line 46: I beleive this should just be applied to primary.

docs/section22/Rule22-3.md
```diff
@@ -0,0 +1,46 @@
...
+2. B-RMR is not modeled with purchased chilled water.
+
+**Manual Check:** None  
+**Evaluation Context:** Building  
+**Data Lookup:** None  
+**Function Call:** None  
+
+**Applicability Checks:**  
+
+1. B-RMR is modeled with at least one air-side system that is Type-7, 8, 11, 12, or 13: `PLACEHOLDER:`
```
mkarpman, line 23: Per above, please remove System 11

mkarpman, line 23: I think the applicability check should be that there is at least on air-side system of Type 7,8,12 or 13 AND that there is no system type 11. Because if project includes System 11 then Exception to G3.1.3.9 would apply and reset will be base on loads and not outdoor temperature

xingcx: Removed System 11 in the applicability check.

xingcx:
The rule is updated with new applicability checks.

mkarpman (approved):

supriyagoel (approved):

PR #189, opened 2021-11-21
Create Rule22-5.md

xingcx:


docs/section22/Rule22-5.md
```diff
@@ -0,0 +1,59 @@
...
+**Manual Check:** None  
+**Evaluation Context:** Building  
+**Data Lookup:** None  
+**Function Call:** None  
+
+**Applicability Checks:**  
+
+1. B-RMR is modeled with at least one air-side system serves computer room HVAC system (System Type-11):
+ `LEN(hvac.does_serve_computer_room == TRUE for hvac in B-RMR...heating_ventilation_air_conditioning_systems) > 0`
+
```
supriyagoel, line 26: similar comment as the one made in 22-4, need to  find zones associated with HVAC system, spaces associated with the zone, space type of each space and then identify if it is a computer room.

supriyagoel (changes requested):

docs/section22/Rule22-5.md
```diff
@@ -0,0 +1,59 @@
...
+**Appendix G Section:** Section 22 CHW&CW Loop  
+**Appendix G Section Reference:** Section G3.1.3.9 Chilled-water supply temperature reset (System 7, 8, 11, 12 and 13)  
+
+**Applicability:** All required data elements exist for B_RMR  
+**Applicability Checks:**  
+
+1. B-RMR is modeled with at least one air-side system serves computer room HVAC system (System Type-11)
+2. B-RMR is modeled with at least one chilled water system that does not use purchased chilled water.
+3. Pass Rule 22-3.
+
```
mkarpman, line 16: I don't think #3 belongs here. My understanding of the rules is that if loops serves at least one system 11 its supply temperature should be reset based on load instead of outdoor air temperature.

docs/section22/Rule22-5.md
mkarpman: I think this is already determined by the RDS that confirmed that there is a System 11 on the project. I don't think we need to check it again.

docs/section22/Rule22-5.md
```diff
@@ -0,0 +1,59 @@
...
+**Applicability Checks:**  
+
+1. B-RMR is modeled with at least one air-side system serves computer room HVAC system (System Type-11):
+ `LEN(hvac.does_serve_computer_room == TRUE for hvac in B-RMR...heating_ventilation_air_conditioning_systems) > 0`
+
+2. B-RMR is modeled with at least one chilled water system that does not use purchased chilled water: `if B_RMR.ASHRAE229.chillers:`
+
+3. Pass Rule 22-3: `if Rule-22-3 == "PASS"`
+
```
mkarpman, line 30: Don't think this applies, per above. Also, why is this repeated?

docs/section22/Rule22-5.md
```diff
@@ -0,0 +1,59 @@
...
+## Rule Logic:  
+
+- For each HVAC system in B-RMR: `for hvac_b in B-RMR...heating_ventilation_air_conditioning_systems:`
+
```
mkarpman, line 34: I think lines 33 - 37 should be replaced by just checking that the loop serves System 11. (It doesn't matter if it serves additional systems.)

docs/section22/Rule22-5.md
```diff
@@ -0,0 +1,59 @@
...
+  - Check if HVAC system serves computer room: `if hvac_b.does_serve_computer_room:`
+
+    - Save CHW loop that serves the cooling system of HVAC system serving computer room to an array: `chw_loop_for_computer_room_array.append(hvac_b.cooling_system.chilled_water_loop)`
+
+- For each chiller in B_RMR: `for chiller_b in B_RMR.ASHRAE229.chillers:`
+
+  - Get cooling loop that chiller servers: `chw_loop_b = chiller_b.cooling_loop`
+
+    - If cooling loop has not been saved in CHW loop array and serves computer room HVAC system: `if ( chw_loop_b NOT in chw_loop_reset_check_array ) AND ( chw_loop_b in chw_loop_for_computer_room_array ):`
+
```
mkarpman, line 44: This would need to be updated per above, to remove computer room criteria

docs/section22/Rule22-5.md
```diff
@@ -0,0 +1,59 @@
...
+  - Case 1: If CHW loop is modeled with load reset: `if ( loop_b.cooling_or_condensing_design_and_control.temperature_reset_type == "LOAD_RESET" ): PASS`
+
+  - Case 2: Else: `else: FAIL`
+
+**[Back](../_toc.md)**
+
+**Notes:**
+
+1. Need to check secondary loop for reset?
```
mkarpman, line 59: I don't think so.

docs/section22/Rule22-5.md
xingcx: Right I will remove this.

docs/section22/Rule22-5.md
xingcx: This Pass Rule 22-3 applicability check is removed. One applicability check is descriptive, one is with pseudo code. We can just keep one if preferred.

docs/section22/Rule22-5.md
xingcx: I'm not sure what needs to be changed. Line 33-37 gets all the CHW loops that serve computer room system, primary or secondary.

docs/section22/Rule22-5.md
xingcx: Not sure why we need to remove computer room criteria here. Is it because we assume there can only be one primary CHW loop in Baseline (so that the computer room system can only be served by this loop)? If so, to summarize the logic:
1). check baseline does not use purchased CHW, as part of applicability check
2). check if there is Sys-11 in baseline, as part of applicability check
3). if so, get CHW primary loop and check if temperature reset is "LOAD_RESET" type

docs/section22/Rule22-5.md
supriyagoel: @mkarpman sure, then this applicability check should be changed to 'does the B RMR has a HVAC system type 11.'

docs/section22/Rule22-5.md
```diff
@@ -0,0 +1,56 @@
...
+ `LEN(hvac.does_serve_computer_room == TRUE for hvac in B-RMR...heating_ventilation_air_conditioning_systems) > 0`
+
+2. B-RMR is modeled with at least one chilled water system that does not use purchased chilled water: `if B_RMR.ASHRAE229.chillers:`
+
+## Rule Logic:  
+
+- For each HVAC system in B-RMR: `for hvac_b in B-RMR...heating_ventilation_air_conditioning_systems:`
+
+  - Check if HVAC system serves computer room: `if hvac_b.does_serve_computer_room:`
+
```
supriyagoel, line 33: @xingcx i guess i dont undersatnd what 'if hvac_b.does_serve_computer_room'' iss doing. I previously assumed you were looking as a data element in the HVAC data group (which does not exist). is that correct?

supriyagoel: based on what MK said, we could simply find the CHW loop associated with System 11 and run the verification on those.

docs/section22/Rule22-5.md
supriyagoel: i believe your approach and the checks you have summarized are correct.

mkarpman (approved):

supriyagoel (approved):

supriyagoel (approved):

PR #190, opened 2021-11-21
Create Rule22-7.md

xingcx:


docs/section22/Rule22-7.md
```diff
@@ -0,0 +1,48 @@
...
+
+    - If cooling loop has not been saved in CHW loop array: `if chw_loop_b NOT in chw_loop_check_array:`
+
+      - Save cooling loop to CHW loop array that needs to comply with the requirement: `chw_loop_check_array.append(chw_loop_b)`
+
+- For each CHW loop that chiller serves: `for loop_b in chw_loop_check_array:`
+
+  **Rule Assertion:**
+
+  - Case 1: If CHW loop is modeled with secondary loop with pump: `if ( loop_b.child_loops ) AND ( loop in loop_with_pump_array for loop in loop_b.child_loops ): PASS`
```
supriyagoel, line 44: @xingcx I'm not sure I follow the logic of this RDS. Are we checking that a CHW loop has 2 fluid loops, one which has a chiller and one which does not? And then assuming that the secondary loop is the one without the chiller?

supriyagoel (changes requested):
I didnt quite understand this RDS, would help to discuss.

docs/section22/Rule22-7.md
```diff
@@ -0,0 +1,48 @@
+
+# CHW&CW - Rule 22-7  
+
+**Rule ID:** 22-7  
+**Rule Description:** Baseline chilled water system that does not use purchased chilled water shall be modeled as primary/secondary systems.
```
mkarpman, line 5: There is a mismatch in the way the rule is formulated here (as a check that CHW system is modeled with primary/secondary loops) and the RDS logic in lines #26 and down. I would re-formulate the rule in the spreadsheet to focus in verifying primary/secondary arrangement and update the rule logic below accordingly.

docs/section22/Rule22-7.md
```diff
@@ -0,0 +1,48 @@
...
+**Data Lookup:** None  
+**Function Call:** None  
+
+**Applicability Checks:**  
+
+1. B-RMR is modeled with at least one air-side system that is Type-7, 8, 11, 12, or 13: `PLACEHOLDER`
+2. B-RMR is modeled with at least one chilled water system that does not use purchased chilled water: `if B_RMR.ASHRAE229.chillers:`
+
+## Rule Logic:  
+
```
mkarpman, line 27: I would update the rule logic to confirm that 
(a) there is a loop referenced by a chiller (this would be the primary loop), 
(b) that this primary loop references a child loop (this would be the secondary loop), 
(c) that the secondary loop is not referenced by another chiller or boiler and does not have a child loop of its own (tetriary loop)
(d) that there are no loops of type "cooling" other than these two.

xingcx: Do we also need to check that all HVAC cooling coils are on the secondary loop and not on primary loop.

docs/section22/Rule22-7.md
xingcx: The original TCD rule was "CHW Loop has a secondary pump", and I changed it to the current description. But the RDS was more written in a way to check that the secondary loop has pump as I'm a bit confused about the focus of this rule. It will be helpful to discuss.

docs/section22/Rule22-7.md
```diff
@@ -0,0 +1,56 @@
...
+**Manual Check:** None  
+**Evaluation Context:** Building  
+**Data Lookup:** None  
+**Function Call:** None  
+
+**Applicability Checks:**  
+
+1. B-RMR is not modeled with purchased chilled water: `if Rule-18-8 == "NOT APPLICABLE":`
+
```
mkarpman, line 23: Change NOT APPLICABLE to UNDERTERMINED

docs/section22/Rule22-7.md
```diff
@@ -0,0 +1,56 @@
...
+## Rule Logic:  
+
+- For each building segment in B_RMR: `for building_segment_b in B_RMR...building_segments:`
+
+  - For each HVAC system in building segment: `for hvac_b in building_segment_b.heating_ventilation_air_conditioning_systems`
+
+    - Check if HVAC system has chilled water coil: `if hvac_b.cooling_system.chilled_water_loop:`
+
```
mkarpman, line 31: I think we also need to check that this loop is not referenced in any of the chller.cooling_loop and is referenced in FluidLoop.child_loops

docs/section22/Rule22-7.md
```diff
@@ -0,0 +1,56 @@
...
+- For each fluid loop in B_RMR: `for fluid_loop_b in B_RMR.ASHRAE229.fluid_loops:`
+
+  - Check if fluid loop is connected to chiller(s): `if fluid_loop_b in loop_chiller_dict.keys()`
+
+    - Check if fluid loop serves any cooling coils, set check coil flag to True: `if fluid_loop_b in secondary_chw_loop_array: check_coil_flag = TRUE`
+
+    - Check if fluid loop is not modeled with one secondary loop, set check number flag to True: `if LEN(fluid_loop_b.child_loops) != 1: check_number_flag = TRUE`
+
+    - Check if secondary CHW loop array are not the same as child loops of primary CHW loop, set check secondary loop flag to True: `if secondary_chw_loop_array == fluid_loop_b.child_loops: check_secondary_loop_flag = TRUE`
+
```
mkarpman, line 45: This is a little confusing but looks correct.

mkarpman (approved):

docs/section22/Rule22-7.md
supriyagoel: small correct to line 30, cooling system would be a "fluid loop" and there needs to be an additional check that verifies that the fluid loop type = "cooling"

docs/section22/Rule22-7.md
supriyagoel: @xingcx I dont quite understand whats going on here.

1. Looks like lines 28-32 identify the secondary loops by looking at HVAC systems which are connected to fluid loops
2. 34-38 identify primary loops by looking at CHW loops connected to chillers.
3. Line 40 identifies if any of the loops are in the secondary CHW loop array. There is an issue with this logic: lines 28-32 might also end up including primary only loops and line 40 will assume those to be secondary loops. There need to be a check to confirm that these loops dont have chiller on them.

This logic is slightly confusing, so might be helpful to add a narrative somewhere.

xingcx: Line 40 will flag the rule to FAIL if the loop is connected to both chiller and cooling coils.

docs/section22/Rule22-7.md
```diff
@@ -0,0 +1,40 @@
...
+- Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`
+
+  - Check if B-RMR is modeled with at least one air-side system that is Type-7, 8, 11.1, 11.2, 12, 13, 7b, 8b, 11b, 12b, 13b, i.e. with air-side system served by chiller(s), continue to rule logic: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-7", "SYS-8", "SYS-11.1", "SYS-11.2", "SYS-12", "SYS-13", "SYS-7B", "SYS-8B", "SYS-11B", "SYS-12B", "SYS-13B"]): CHECK_RULE_LOGIC`
+
+  - Else, rule is not applicable to B-RMR: `else: RULE_NOT_APPLICABLE`
+
+## Rule Logic:  
+
+- Get primary and secondary loops for B-RMR: `primary_secondary_loop_dictionary = get_primary_secondary_loops(B_RMR, "COOLING")`
```
mkarpman, line 33: I included comments on the function. I would consider just using the logic right within this RDS:
a. Identify loops referenced by chillers (LOOP List A)
b. Identify loops that are referenced by cooling systems or terminals (LOOP List B).
c. Baseline CHW primary/secondary is modeled correctly if (i) Lists A and B each have only one loop (ii) loops in List A and List B are different ones (iii) loop in list A references loop B as it's child loop.

docs/section22/Rule22-7.md
```diff
@@ -0,0 +1,40 @@
+
+# CHW&CW - Rule 22-7  
+
+**Rule ID:** 22-7  
+**Rule Description:** Baseline chilled water system that does not use purchased chilled water shall be modeled as primary/secondary systems.
```
supriyagoel, line 5: the rule description should be updated to remove 'does not use purchased chilled water' and instead say ' all systems using chilled water..'

docs/section22/Rule22-7.md
supriyagoel: The function is also used by 22-34 would make sense to keep it

supriyagoel (approved):

PR #191, opened 2021-11-21
Create Rule22-8.md

xingcx:


docs/section22/Rule22-8.md
```diff
@@ -0,0 +1,54 @@
...
+- For each chiller in B_RMR: `for chiller_b in B_RMR.ASHRAE229.chillers:`
+
+  - Get cooling loop that chiller servers: `chw_loop_b = chiller_b.cooling_loop`
+
+    - Add chiller rated capacity to CHW loop total capacity: `chw_loop_capacity_dict[chw_loop_b] += chiller_b.rated_capacity`
+
+- For each CHW loop that chiller serves: `for loop_b in chw_loop_capacity_dict.keys():`
+
+  - Check if CHW loop total cooling capacity is 300 tons or more: `if chw_loop_capacity_dict[loop_b] >= 300 * CONVERSION(TON_TO_BTUH):`
+
```
supriyagoel, line 43: rated capacity is in W in the schema, conversion should be from W to tons

docs/section22/Rule22-8.md
```diff
@@ -0,0 +1,54 @@
...
+    - For each secondary loops served by CHW loop: `for secondary_loop_b in loop_b.child_loops:`
+
```
supriyagoel, line 45: @xingcx Similar to RDS 22-7, i'm unclear on how the secondary loop is being identified. would be helpful to have a narrative of what is happening out here as well.

docs/section22/Rule22-8.md
```diff
@@ -0,0 +1,54 @@
...
+      - For each secondary pump associated with secondary loop: `for secondary_pump in loop_pump_dictionary[secondary_loop_b]:`
+
```
supriyagoel, line 47: @JasonGlazer I have created an issue for this in the RMR repo as well- we have two data elements in pump data group 'is_variable_speed' and 'PumpSppednControlOptions' for 'fixed_speed' and 'variable_speed'. we should remove one of them.

docs/section22/Rule22-8.md
```diff
@@ -0,0 +1,54 @@
...
+  - Save loop served by pump and pump to a dictionary: `loop_pump_dictionary[pump_b.loop_or_piping].append(pump)`
+
+- For each chiller in B_RMR: `for chiller_b in B_RMR.ASHRAE229.chillers:`
+
+  - Get cooling loop that chiller servers: `chw_loop_b = chiller_b.cooling_loop`
+
+    - Add chiller rated capacity to CHW loop total capacity: `chw_loop_capacity_dict[chw_loop_b] += chiller_b.rated_capacity`
+
+- For each CHW loop that chiller serves: `for loop_b in chw_loop_capacity_dict.keys():`
+
```
mkarpman, line 41: If fule 22-7 passes there would be only one CHW loop connected to all chillers.

docs/section22/Rule22-8.md
mkarpman: Didn't we agree that unit conversions will happen at the time RMRs are parsed and we don't need to include it here? Perhaps we should just include a note that chiller capacity should be converted from W to Tons and not Btu/hr.

docs/section22/Rule22-8.md
mkarpman: Agreed. The seconary loop is the one referenced from the primary loop as a child loop (chw_loop_b).

docs/section22/Rule22-8.md
mkarpman: Agreed. I also noted it under one of the HW RDSs.

docs/section22/Rule22-8.md
xingcx: This conversion is for the 90.1, which has a unit of Ton. I would assume the equipment capacity unit would be reported either in W (for SI)or Btuh/kBtuh (for IP). This is not for for SI / IP conversion.

docs/section22/Rule22-8.md
xingcx: If Rule 22-7 does not pass, do we still need to check this rule?

supriyagoel: if 22-7 doesn't pass then 22-8 would not be applicable and should not be tested. 
@mkarpman - do you agree?

docs/section22/Rule22-8.md
supriyagoel: If i remember the discussion on this during the RDS call, this can just be the conversion from the unit in the RMR (i.e. W) to the value needed by 90.1 (i.e. Tons).

supriyagoel (approved):

supriyagoel (approved):

mkarpman (approved):

docs/section22/Rule22-8.md
```diff
@@ -0,0 +1,66 @@
...
+# CHW&CW - Rule 22-8  
+
+**Rule ID:** 22-8  
+**Rule Description:** For Baseline chilled water system with cooling capacity of 300 tons or more, the secondary pump shall be modeled with variable-speed drives.  
+**Rule Assertion:** B-RMR = expected value  
+**Appendix G Section:** Section 22 CHW&CW Loop  
+**Appendix G Section Reference:** Section G3.1.3.10 Chilled-water supply temperature reset (System 7, 8, 11, 12 and 13)  
+
```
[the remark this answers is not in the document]
xingcx: Corrected.

supriyagoel:
PR 22-8 #212 has been combined with PR 22-10 #341 this PR can be archived.

supriyagoel (approved):

PR #196, opened 2021-12-06
Create Rule23-1.md

xingcx:


docs/section23/Rule23-1.md
```diff
@@ -0,0 +1,48 @@
...
+## Rule Logic:  
+
+- For each building segment in B_RMR: `for building_segment_b in B_RMR...building_segments:`
+
+  - Get HVAC system: `hvac_b = building_segment_b.heating_ventilation_air_conditioning_systems`
+
+    - Check if HVAC system is type 5, 6, 7, 8 or 11: `if baseline_hvac_type(hvac_b) in ["SYS-5", "SYS-6", "SYS-7", "SYS-8", "SYS-11"]:`
+
+      - Set applicability flag: `rule_applicability_check = TRUE`
+
```
supriyagoel, line 31: this should be for each system. i.e. for each system, get fan.

docs/section23/Rule23-1.md
```diff
@@ -0,0 +1,48 @@
...
+1. baseline_hvac_type()  
+
+## Rule Logic:  
+
+- For each building segment in B_RMR: `for building_segment_b in B_RMR...building_segments:`
+
+  - Get HVAC system: `hvac_b = building_segment_b.heating_ventilation_air_conditioning_systems`
+
+    - Check if HVAC system is type 5, 6, 7, 8 or 11: `if baseline_hvac_type(hvac_b) in ["SYS-5", "SYS-6", "SYS-7", "SYS-8", "SYS-11"]:`
+
```
supriyagoel, line 29: @xingcx since we don't have an enumeration for the system type in the schema, we would have functions which determine the system type for each system in thee baseline. This could instead  call the placeholder function to determine the baseline HVAC type for each system.

mkarpman: @supriyagoel I believe this is the call for a future function

docs/section23/Rule23-1.md
mkarpman: Agreed. So line 26 should be for each HVAC system and FanSystem

xingcx: Corrected.

docs/section23/Rule23-1.md
```diff
@@ -0,0 +1,46 @@
...
+# CHW&CW - Rule 23-1  
+
+**Rule ID:** 23-1  
+**Rule Description:** For baseline systems 5-8 and 11, the SAT is reset higher by 5F under minimum cooling load conditions.  
+**Rule Assertion:** B-RMR = expected value
```
supriyagoel, line 6: @mkarpman it seems strange that App G has this SAT reset requirement for system 11. System 11 is a single zone system, the SAT will automatically be based on the heating or cooling demand of that zone, correct? Does SAT reset even apply to such a system in reality?

supriyagoel (approved):

docs/section23/Rule23-1.md
mkarpman: I think the intent is that it applies to all VAV systems since they can adjust amount of extracted heat by adjusting either flow rate or SAT. So it is my understanding that this requirement say that SAT can go up from 55F to 60F, and after that the further reduction in heat extraction would be due to reduced flow rate.

mkarpman (approved):

docs/section23/Rule23-1.md
supriyagoel: got it, this is helpful.

PR #197, opened 2021-12-06
Create Rule23-5.md

xingcx:


docs/section23/Rule23-5.md
```diff
@@ -0,0 +1,48 @@
...
+## Rule Logic:  
+
+- For each zone in B_RMR: `for zone_b in B_RMR...zones:`
+
+  - For each terminal in zone: `for terminal_b in zone_b.terminals:`
+
+    - Get HVAC system serving terminal: `hvac_b = terminal_b.served_by_heating_ventilation_air_conditioning_systems`
+  
+      - Check if HVAC system is type 6 or 8: `if baseline_hvac_type(hvac_b) in ["SYS-6", "SYS-8"]:`
+
```
supriyagoel, line 31: based on MK's response to 23-1, my understanding is that line 23 is referencing a function to be written in the future, which will identify the baseline system type for each HVAC system in the B-RMR.

docs/section23/Rule23-5.md
```diff
@@ -0,0 +1,48 @@
...
+        - Set applicability flag: `rule_applicability_check = TRUE`
+
+          **Rule Assertion:**
+
+          - Case 1: For each terminal that is served by HVAC system that is Type-6 or 8, if fan in parallel VAV-powered box is sized for 50% of the peak design primary air (from the VAV air-handling unit) flow rate (CFM) and is modeled with 0.35W/cfm fan power: `if ( terminal_b.secondary_airflow == terminal_b.primary_airflow * 0.5) AND ( terminal_b.fan.design_airflow == terminal_b.primary_airflow * 0.5 ) AND ( terminal_b.fan.design_electric_power == 0.35 * terminal_b.fan.design_airflow ): PASS`
+
```
supriyagoel, line 37: I am confused about line 36.

1. VAV fan-powered boxes shall be sized for 50% of the peak design primary air should be terminal_b.primary_airflow == terminal_b.fan.design_airflow * 0.5 (here the fan should be the central supply fan)
2. I dont think the seconrady airflow needs to be 50% of the primary airflow, it just needs to be less than the primary airflow
3. I think the 0.35W/CFM should 0.35 * terminal_b.primary_airflow

mkarpman: This will probably need a discussion, but here is my take on it:
1. Terminal.primary_airflow attribute is the airflow rate that zone needs to recieve at design conditions. This attribute is independent on whether zone terminal is fan powered.
2. If terminal has a fan, than its design_airflow needs to be 50% of the terminal.primary_airflow.
3. I think we should remove secondary_airflow attribute because it's redundant to fan.design_airflow. 
4. I agree with Xing's fan power condition. So Case 1 would then be
`if ( terminal_b.fan.design_airflow == terminal_b.primary_airflow * 0.5 ) AND ( terminal_b.fan.design_electric_power == 0.35 * terminal_b.fan.design_airflow ): PASS`

mkarpman (changes requested):

docs/section23/Rule23-5.md
xingcx: Updated.

docs/section23/Rule23-5.md
```diff
@@ -0,0 +1,50 @@
...
+  - For each terminal in zone: `for terminal_b in zone_b.terminals:`
+
+    - Get HVAC system serving terminal: `hvac_b = terminal_b.served_by_heating_ventilation_air_conditioning_systems`
+  
+      - Check if HVAC system is type 6, 8, 8a, 6b, 8b: `if any(is_baseline_system_type(hvac_b, sys_type) == TRUE for sys_type in ["SYS-6", "SYS-8", "SYS-8A", "SYS-6B", "SYS-8B"]):`
+
+        **Rule Assertion:**
+
+        - Case 1: For each terminal that is served by HVAC system that is Type-6, 8, 8a, 6b, 8b, if fan in parallel VAV-powered box is sized for 50% of the peak design primary air (from the VAV air-handling unit) flow rate (CFM) and is modeled with 0.35W/cfm fan power: `if ( terminal_b.fan.design_airflow == terminal_b.primary_airflow * 0.5 ) AND ( terminal_b.fan.design_electric_power == 0.35 * terminal_b.fan.design_airflow ): PASS`
+
```
supriyagoel, line 43: need to discuss this during the RDS call. Need to confirm that design airflow = 50% of primary airflow.

docs/section23/Rule23-5.md
mkarpman: Either way, I will mark it as not used for App G - agreed?

mkarpman (approved):

supriyagoel (approved):

PR #200, opened 2021-12-06
Create Rule23-10.md

xingcx:


docs/section23/Rule23-10.md
```diff
@@ -0,0 +1,53 @@
...
+      - Set applicability flag: `rule_applicability_check = TRUE`
+
+      - Get fan system serving HVAC system: `fan_system_b = hvac_b.fan_systems[0]` (See Note#1)
+
+        - For fan system supply, return and relief fan: `for fan_b in [fan_system_b.supply_fan, fan_system_b.return_fan, fan_system_b.relief_fan]:`
+
+          **Rule Assertion:**
+
+          - Case 1: For supply, return and relief fan in each fan system serving Baseline system type 5, 6, 7, 8 or 11, if fan part-load VAV is modeled as per Table G3.1.3.15: `if fan_b.output_validation_points[i] == data_lookup(table_G_3_1_3_15, i) for i in [0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0]: PASS`
+
```
supriyagoel, line 39: @xingcx we might need some guidance here on how to convert airflow to 0, 0.01, 0.02 etc data points.
Also need some guidance on what needs to be verified.
the RDS should look up the full load power (fan.design_electric_power) and multiply by the fraction in the table to show the expected wattage. Not sure how to best write it but a little more guidance on the required verification would be helpful.

mkarpman: 1. This RDS should apply to supply fan only.
2. It is my understanding that in the output_validation-points array we expect 11 sets of value pairs:
a. The first value in the pair would be the product of Fan Part-Load Ratio (first column in G3.1.3.15) by  Fan.design_airflow
b. The second value in the pair would be the product of Fraction of Full-Load Power (second column in G3.1.3.15) by Fan.design_electric_power,

docs/section23/Rule23-10.md
```diff
@@ -0,0 +1,53 @@
...
+**Applicability Checks:**  
+
+1. B-RMR is modeled with at least one air-side system that is Type-5, 6, 7, 8 or 11: `if rule_applicability_check: is_applicable = TRUE`
+
+**[Back](../_toc.md)**
+
+**Notes:**
+
+1. For baseline hvac.fan_systems, assuming only one fan system is modeled. Might need a separate check for this. Note from schema: "Normally one fan system is used but second fan systems may be used when a direct outdoor air system is used."
+
```
mkarpman, line 51: We can handle it either way but make consistent across RDSs, as it will come up on many of them.

docs/section23/Rule23-10.md
```diff
@@ -0,0 +1,53 @@
...
+2. RDS assumes exhaust fan in HVAC system could be constant. Please confirm this approach is valid.
+
```
mkarpman, line 53: This RDS only applies to supply fan: "VAV system supply fans shall have variable-speed drives, and their part-load performance characteristics shall be modeled using either Method 1 or Method 2 specified in Table G3.1.3.15."

docs/section23/Rule23-10.md
```diff
@@ -0,0 +1,53 @@
...
+# Airside System - Rule 23-10  
+
+**Rule ID:** 23-10  
+**Rule Description:** System 5-8 and 11 - part load VAV fan power shall be modeled using either method 1 or 2 in Table G3.1.3.15. This rule will only validate data points from Method-1 Part-load Fan Power Data.
```
mkarpman, line 5: This effectively verifies both methods, because it is my understanding that the difference between the two methods is just in a way variation in input power is expressed, not in what it should be.

mkarpman (changes requested):

docs/section23/Rule23-10.md
supriyagoel: this looks very good now!

supriyagoel (approved):

docs/section23/Rule23-10.md
xingcx: Updated.

docs/section23/Rule23-10.md
mkarpman: I believe this the second value in each pair should be based of fan.design_electric_power, not fan.input_power. Input power is "Power delivered to the fan’s shaft and does not include the mechanical drive losses. Equivalent to fan brake
horsepower for inchpound units. Only used when specification_method is set to Detailed.

docs/section23/Rule23-10.md
mkarpman: I would add to the rule description "however, both methods are equivalent. When modeling inputs are based on Method 2, values should be converted to Method 1 when writing to RMD

mkarpman (approved):

docs/section23/Rule23-10.md
xingcx: Rule description is updated.

docs/section23/Rule23-10.md
```diff
@@ -0,0 +1,62 @@
...
+
+    - Check if HVAC system is type 5, 6, 7, 8, 11.1, 11.2, 7a, 8a, 11.1a, 11.2a, 5b, 6b, 7b, 8b, 11b, 7c, 11c: ``if any(is_baseline_system_type(hvac_b, sys_type) == TRUE for sys_type in ["SYS-5", "SYS-6", "SYS-7", "SYS-8", "SYS-11.1", "SYS-11.2", "SYS-7A", "SYS-8A", "SYS-11.1A", "SYS-11.2A", "SYS-5B", "SYS-6B", "SYS-7B", "SYS-8B", "SYS-11B", "SYS-7C", "SYS-11C"]):`
+
+      - Get fan system serving HVAC system: `fan_system_b = hvac_b.fan_system`
+
+        - For each supply fan in fan system: `for supply_fan_b in fan_system_b.supply_fans:`
+
+          **Rule Assertion:**
+
+          - Case 1: For supply fan in fan system, if fan part-load VAV is modeled as per Table G3.1.3.15: `if supply_fan.output_validation_points == {0: 0,
```
mkarpman, line 45: Based on schema note, "Airflow is input to each validation point and energy output is the result." So, in order to verify the match between Table G3.1.3.15 and fan validation point in the RMD, we need to divide the first value in fan.validation_point pair by fan.design_airflow, and the second value in each pair by fan.design_electric_power.

mkarpman: I have also submitted schema issue for this.

mkarpman (changes requested):

docs/section23/Rule23-10.md
supriyagoel: Link to Schema issue https://github.com/open229/ruleset-model-description-schema/issues/109

mkarpman (approved):

PR #201, opened 2021-12-06
Create Rule23-19.md

xingcx:


docs/section23/Rule23-19.md
```diff
@@ -0,0 +1,52 @@
...
+    - Get HVAC system serving terminal: `hvac_b = terminal_b.served_by_heating_ventilation_air_conditioning_systems`
+  
+      - Check if HVAC system is type 5 through 8: `if baseline_hvac_type(hvac_b) in ["SYS-5", "SYS-6", "SYS-5", "SYS-8"]:`
+
+        - Set applicability flag: `rule_applicability_check = TRUE`
+
+        - Get zone design heating temperature setpoint: `zone_heating_setpoint_b = zone_b.design_thermostat_heating_setpoint`
+
+        - Save HVAC system and zone design heating temperature setpoint if temperature is higher than current value saved: `if zone_heating_setpoint_b > hvac_max_zone_setpoint_dict[hvac_b]: hvac_max_zone_setpoint_dict[hvac_b] = zone_heating_setpoint_b`
+
```
supriyagoel, line 37: this needs to be done for each HVAC system. 
So we need to find all the zones being served by each HVAC system and then find the max heating setpoint for all of those zones. Then the corresponding preheat coil for that HVAC system needs to be identified and verified.

I believe this is already being done int he RDS but line 36 suggests that the max across all zones for all HVAC systems is being identified.

docs/section23/Rule23-19.md
```diff
@@ -0,0 +1,52 @@
...
+- For each zone in B_RMR: `for zone_b in B_RMR...zones:`
+
+  - For each terminal in zone: `for terminal_b in zone_b.terminals:`
+
+    - Get HVAC system serving terminal: `hvac_b = terminal_b.served_by_heating_ventilation_air_conditioning_systems`
+  
+      - Check if HVAC system is type 5 through 8: `if baseline_hvac_type(hvac_b) in ["SYS-5", "SYS-6", "SYS-5", "SYS-8"]:`
+
+        - Set applicability flag: `rule_applicability_check = TRUE`
+
```
mkarpman, line 33: Do Building->BuildingSegment->HVAC system instead, similar to 23-10?

docs/section23/Rule23-19.md
mkarpman: App G doesn't say how to handle cases where design room heating temperature setpoint differs between zones. Perhaps it should be clarified in App G whether this should be based on the maximum setpoint? I started a separate exchange on this.

docs/section23/Rule23-19.md
```diff
@@ -0,0 +1,52 @@
...
+        - Save HVAC system and zone design heating temperature setpoint if temperature is higher than current value saved: `if zone_heating_setpoint_b > hvac_max_zone_setpoint_dict[hvac_b]: hvac_max_zone_setpoint_dict[hvac_b] = zone_heating_setpoint_b`
+
+- For each HVAC system that is type 5 through 8: `for hvac_b in hvac_max_zone_setpoint_dict.keys():`
+
+  - Get preheat system of HVAC system: `preheat_b = hvac_b.preheat_system`
+
+    **Rule Assertion:**
+
+    - Case 1: For each Baseline system type 5, 6, 7, or 8, if preheat coils is controlled to a fixed set point 20F less than the design room heating temperature setpoint: `if preheat_b.heating_coil_setpoint == hvac_max_zone_setpoint_dict[hvac_b] - 20: PASS`
+
```
mkarpman, line 45: This RDS should also verify that preheat system is a coil, so I suggest to add ".... AND preheat_b.heating_system_type="FLUID LOOP"". 
I think we also need to verify that it references the right fluid loop.

mkarpman (changes requested):

docs/section23/Rule23-19.md
xingcx: It is confirmed that the reset should be based on max setpoint.

docs/section23/Rule23-19.md
xingcx: For this rule, we also need to get zone setpoint so it starts with zone.terminal.

docs/section23/Rule23-19.md
xingcx: "AND preheat_b.heating_system_type="FLUID LOOP" is added. Also added FluidLoop.type == "Heating". Rule 21-18 will verify there is only one heating fluid loop.

docs/section23/Rule23-19.md
```diff
@@ -0,0 +1,54 @@
...
+# Airside System - Rule 23-19  
+
+**Rule ID:** 23-19  
+**Rule Description:** Systems 5 - 8, the baseline system shall be modeled with preheat coils controlled to a fixed set point 20F less than the design room heating temperature setpoint.
```
supriyagoel, line 5: @mkarpman If i remember correctly, previously App G used to require preheat in baseline ONLY when the proposed had preheat. Now it looks like all baseline systems 5-8 need to have preheat, irrespective of the proposed. Is that correct?

docs/section23/Rule23-19.md
supriyagoel: I do not understand what 'AND ( preheat_b.hot_water_loop )' is doing

supriyagoel: here preheat_b.heating_coil_setpoint == hvac_max_zone_setpoint_dict[hvac_b] - 20 ): PASS we assume that heating coil setpoint is in F. We should probably provide a unit conversion tip here, maybe heating_coil_setpoint[C to F]  or whatever convention we have previously followed.

mkarpman: As Supriya noted above, I do not exactly follow what is "AND ( preheat_b.hot_water_loop)". Perhaps this is meant to check that the referenced loop  has FluidLoop.type == "Heating"?

xingcx: Correct. This is modified to check that it is connected to a heating hot water loop.

mkarpman (approved):

supriyagoel (approved):

Issue #10, opened 2021-03-19
RDDS Rules 4-4 /4-5


mcneillj:
•	4-4: Count of assigned weeks in an Annual Sch is the same in Base RMR and User RMR
•	4-5: Week names are same in Baseline RMR and in User RMR

PR #204, opened 2021-12-07
Create 4-3.md

claperle:
Here is my first RDS for review. A couple of notes

Do we also need to check if the fan operation schedule is consistent with the occupancy schedules of the zones served by the fan system. That is complicated because the zone occupancy schedule is a multiplier schedule on the occupant density so we would have to decide above which fraction represents occupied. This may be different across buildings and space types. I heard talk of using the Building.building_open_schedule for this type of comparison but building occupancy schedules can differ significantly from zone schedules so this could led erroneous fails. Any thoughts?


docs/section4/4-3.md
```diff
@@ -0,0 +1,46 @@
...
+**Appendix G Section:** Section G3.1-4 Schedules for the proposed building  
+**Appendix G Section Reference:**  
+
+**Applicability:** All required data elements exist for P_RMR  
+**Applicability Checks:** 
+
+1. Not applicable for for heating or cooling system is being simulated only to meet the requirements described in Table G3.1.
+
+**Manual Check:** None  
+**Evaluation Context:** Evaluate each data element
```
mkarpman, line 16: I think this should be each HeatingVentilationAirAconditioningSystem

docs/section4/4-3.md
```diff
@@ -0,0 +1,46 @@
...
+					- Get minimum OA: `minimum_outdoor_airflow_p = fan_systems_p. minimum_outdoor_airflow`
+					- Get operation during occupied periods: `operation_during_occupied_p = fans_systems_p.operation_during_occupied`
+
+**Rule Assertion:**
+- Case 1: if minimum OA is greater than 0 and fan system operation during occupied periods is continuous in the P_RMR: `if minimum_outdoor_airflow_p >0 and operation_during_occupied_p == “CONTINUOUS”: PASS`
+
+- Case 2: Else if minimum OA is greater than 0 and fan system operation during occupied periods is not set to CONTINUOUS P_RMR: `else if minimum_outdoor_airflow_p > 0 and operation_during_occupied_p == “KEEP_OFF”: FAIL and raise_warning “FAN SYSTEM IS OFF DURING OCCUPIED PERIODS WHEN THE UNIT IS MODELED AS PROVIDING MINUMUM OA CFM IN THE P_RMR.”`
+
+- Case 3: Else if minimum OA is equal to 0 and fan system operation during occupied periods is equal to continuous in the P_RMR: `else if minimum_outdoor_airflow_p == 0 and operation_during_occupied_p == “CONTINUOUS”: FAIL and raise_warning “FAN SYSTEM OPERATES CONTINUOUSLY DURING OCCUPIED PERIODS WHEN IT DOES NOT PROVIDE OA CFM IN THE P_RMR. ”`
+
```
mkarpman, line 35: There are no rules in App G for fan operation in case OA=0, so think this Case should be removed. Same applies to all other cases down below.

docs/section4/4-3.md
```diff
@@ -0,0 +1,46 @@
+# ASHVAC – Rule 4-3
+
+**Rule ID:** 4-3
+ 
+**Rule Description:** HVAC Fan Schedules. Schedules for HVAC fans that provide outdoor air for ventilation shall run continuously whenever spaces are occupied.
```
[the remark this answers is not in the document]
claperle: To be discussed on call. We decided to separate to limit the number of TCD combiations.

docs/section4/4-3.md
```diff
@@ -0,0 +1,46 @@
+# ASHVAC – Rule 4-3
+
+**Rule ID:** 4-3
+ 
+**Rule Description:** HVAC Fan Schedules. Schedules for HVAC fans that provide outdoor air for ventilation shall run continuously whenever spaces are occupied.  
+**Rule Assertion:** P-RMR = expected value                                           
+**Appendix G Section:** Section G3.1-4 Schedules for the proposed building
```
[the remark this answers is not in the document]
claperle: Which line are you suggesting that this be added to?

docs/section4/4-3.md
```diff
@@ -0,0 +1,46 @@
...
+**Rule Description:** HVAC Fan Schedules. Schedules for HVAC fans that provide outdoor air for ventilation shall run continuously whenever spaces are occupied.  
+**Rule Assertion:** P-RMR = expected value                                           
+**Appendix G Section:** Section G3.1-4 Schedules for the proposed building  
+**Appendix G Section Reference:**  
+
+**Applicability:** All required data elements exist for P_RMR  
+**Applicability Checks:** 
+
+1. Not applicable for for heating or cooling system is being simulated only to meet the requirements described in Table G3.1.
+
```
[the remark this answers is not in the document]
claperle: This is TBD, I think, because right now this RDS only applies to occupied hours so those other exceptions would not be relevant during occ hours. Correct?

docs/section4/4-3.md
claperle: Okay, change made.

docs/section4/4-3.md
```diff
@@ -0,0 +1,46 @@
...
+**Manual Check:** None  
+**Evaluation Context:** Evaluate each data element  
+**Data Lookup:** None  
+**Function Call:** None
+
+## Rule Logic:
+- For each HVAC system in P_RMR: `for hvac_p in P_RMR...heating_ventilation_air_conditioning_systems:`
+    - Get HVAC system status: `status_type_p = hvac_p.status_type:`
+        - Check if HVAC system status is new or existing: `if status_type_p in [NEW,EXISTING]:`
```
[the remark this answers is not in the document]
claperle: I need an explanation on this from the group/yuo on this. I am having trouble understanding how the logic knows this. I see that Xing has included code to determine the applicability so that is what I did. I think exception #1 applies to the status because if its a yet to be designed system then #1 would come into play.

docs/section4/4-3.md
```diff
@@ -0,0 +1,46 @@
...
+            - Set applicability flag: `rule_applicability_check = TRUE`
+				- For each fan system in each HVAC system that meets the applicabliity check: `fan_systems_p in heating_ventilation_air_conditioning_systems_p.fan_systems:`
+					- Get minimum OA: `minimum_outdoor_airflow_p = fan_systems_p. minimum_outdoor_airflow`
+					- Get operation during occupied periods: `operation_during_occupied_p = fans_systems_p.operation_during_occupied`
+
+**Rule Assertion:**
+- Case 1: if minimum OA is greater than 0 and fan system operation during occupied periods is continuous in the P_RMR: `if minimum_outdoor_airflow_p >0 and operation_during_occupied_p == “CONTINUOUS”: PASS`
+
+- Case 2: Else if minimum OA is greater than 0 and fan system operation during occupied periods is not set to CONTINUOUS P_RMR: `else if minimum_outdoor_airflow_p > 0 and operation_during_occupied_p == “KEEP_OFF”: FAIL and raise_warning “FAN SYSTEM IS OFF DURING OCCUPIED PERIODS WHEN THE UNIT IS MODELED AS PROVIDING MINUMUM OA CFM IN THE P_RMR.”`
+
```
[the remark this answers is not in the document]
claperle: I think if the unit does not provide minimum OA then it can cycle with heating and cooling loads even during occupied periods. Right? That is how I interpret the language. This would apply to a unit heater serving stairs for example.

docs/section4/4-3.md
claperle: See above. The lanaguage specifically states that it applies to HVAC fans that "provide OA for ventilation" so I am not sure I agree.

claperle:
Unoccupied is covered in a separate rule.

*Christina LaPerle* PE, BEMP, CEM, CBCP
________________
[image: image.png]
Phone 860.266.6220

docs/section4/4-3.md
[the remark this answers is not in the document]
supriyagoel: @mkarpman and @claperle I'm not sure what we have the applicability checks (2) and (3)
2. P-RMR airside HVAC system are been modeled correctly per Table G3.1 #10 proposed c: `if rule 10-10 == "PASS"`
3. P-RMR airside HVAC system are been modeled correctly per Table G3.1 #10 proposed d: `if rule 10-11 == "PASS"`

10-10 states - Where no heating system exists or no heating system has been submitted with design documents, the proposed building system type shall be the same system as modeled in the baseline building design and shall comply with but not exceed the requirements of Section 6

10-11 Except for spaces with baseline system 9 or 10, if no cooling system exists or no cooling system has been submitted with design documents, the proposed building cooling system type shall be the same as modeled in the baseline building design and shall comply with the requirements of Section 6.

Not sure what 10-10 and 10-11 need to pass as an applicability check for this rule.

docs/section4/4-3.md
supriyagoel: @claperle I dont see the change. I'm assuming the intent was to change 'Evaluate each data element ' to 'Evaluate each HeatingVentilationAirAconditioningSystem'. Maybe the changes didn't save?

docs/section4/4-3.md
```diff
@@ -0,0 +1,104 @@
...
+    - Reset heating system exists boolean variable equal to false each time it loops to a different zone: `heating_system_exists_u = FALSE`
+    - Reset cooling system exists boolean variable equal to false each time it loops to a different zone: `cooling_system_exists_u = FALSE`
+    - Reset heating system exists boolean variable equal to false each time it loops to a different zone: `heating_system_exists_p = FALSE`
+    - Reset cooling system exists boolean variable equal to false each time it loops to a different zone: `cooling_system_exists_p = FALSE`
+    - Check if there are terminal units associated with the zone, if not skip to proposed: `if len(zone_u.terminals) != 0:`
+        - For each terminal unit associated with each zone: `for terminal_u in zone_u:`
+            - if reheat_source is not null then set heating exists boolean variable equal to true: `if terminal_u.reheat_source != "None":`
+                - Set heating system exists boolean variable equal to true: `heating_system_exists_u = TRUE`
+            - if heat_capacity is not zero then set heating exists boolean variable equal to true: `if terminal_u.heat_capacity != 0:`
```
supriyagoel, line 34: i'm assuming here we're checking if a heating ssystem exisits or has been designed. 2 questions/suggestions
1. Change the term heating_system_exists_u to heating_system_exists_is-designed_u . the term exists implies that we're system is modeled as existing. If thats what the intent is, then please ignore.
2. we're checking both the reheat source and the heat capacity, is that to account for both reheat systems and single zone systems?

docs/section4/4-3.md
```diff
@@ -0,0 +1,104 @@
...
+                    - For each heating system in hvac_p: `for heating_system_p in hvac_p.heating_system:`
+                        - if heating_system_type is not "None": `if heating_system_p.heating_system_type != "None":`
+                            - Set heating system exists boolean variable equal to true: `heating_system_exists_p = TRUE`
+                    - For each cooling system in hvac_p: `for cooling_system_p in hvac_p.cooling_system:`
+                        - if cooling_system_type is not "None": `if cooling_system_p.cooling_system_type != "None":`
+                            - Set cooling system exists boolean variable equal to true: `cooling_system_exists_p = TRUE`
+                    - For each preheat system in hvac_p: `for preheat_system_p in hvac_p.preheat_system:`
+                        - if preheat_system_type is not "None": `if preheat_system_p.heating_system_type != "None":`
+                            - Set heating system exists boolean variable equal to true: `heating_system_exists_p = TRUE`  
+
```
supriyagoel, line 70: @claperle and @mkarpman i now understand why the applicability checks for 10-10 and 10-11 have been added. this is quite detailed and very well written- thank you!

Does it make sense to convert the check for if system is existing/has been designed to a function? There might be more rules that need this check.

docs/section4/4-3.md
supriyagoel: @claperle since the logic is so convoluted and has so many steps, it might help to have a statement right under "Rule Logic" that states "Determine if a HVAC system has been designed or modeled as exisiting in U-RMR" and then right before line 50
. add a statement that says "Determine if the same  zone has been analyzed with a HVAC system in P-RMR" or something similar to clarify the intent.

docs/section4/4-3.md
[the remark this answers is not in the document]
claperle: We decided to create separate rules for unoccupied and occupied as it helps with creating the TCDs

docs/section4/4-3.md
claperle: @supriyagoel I just did not quite under, this has been updated

docs/section4/4-3.md
claperle: All of these have been addressed. Good suggestions! Thanks.

docs/section4/4-3.md
claperle: Yup, done!

docs/section4/4-3.md
claperle: Thinking more about this I think I am going to include that OA is greater than 0 because we would not want a fail outcome for a scenario where the terminal unit is VRF and cycles with load during occupied hours with a DOAS operating contunuously providing OA. I will make the scenario where OA = 0 and it cycles undetermined. PLease see edits.

docs/section4/4-3.md
```diff
@@ -0,0 +1,39 @@
...
+**Manual Check:** None  
+
+**Rule Logic:**
+**Applicability Check 1:** 
+- Call function to get list of inapplicable HVAC systems: `inapplicable_HVAC_systems_list_p = get_HVAC_systems_applicable_10cd(U_RMR,P_RMR)`
+- For each HeatingVentilationAirAconditioningSystem in the P_RMR check if this check is applicable: `for each hvac_p in P_RMR...hvac_systems:`
+    - Check if hvac_p is not in the list of inapplicable systems, if it is not found then set applicability flag to true: `if hvac_p not in inapplicable_HVAC_systems_list_p: rule_applicability_flag = TRUE`   
+    - If rule_applicability_flag = TRUE for hvac_p: `if rule_applicability_flag == TRUE:`
+        - For each fan system in hvac_p: `fan_systems_p in hvac_p.fan_systems:`
```
mkarpman, line 28: I submitted RMR issue recommending to require that a single fan_system is referenced by HeatingVentilationAirConditioningSystem data group. If this is accepted, line 28 will need to be updated to point to this system.

docs/section4/4-3.md
```diff
@@ -0,0 +1,39 @@
...
+            - Get operation during occupied periods: `operation_during_occupied_p = fans_systems_p.operation_during_occupied`
+            - Get minimum_outdoor_airflow: `minimum_outdoor_airflow_p = fans_systems_p.minimum_outdoor_airflow`
+
+            **Rule Assertion:**
+            - Case 1: For each applicable fan system, if fan system operation during occupied periods is continuous and the minimum OA is greater than 0 in the P_RMR: `if operation_during_occupied_p == “CONTINUOUS” and minimum_outdoor_airflow_p >0 : PASS`
+            - Case 2: else if, fan system operation during occupied periods is cycling and the minimum OA equals 0 in the P_RMR: `else if operation_during_occupied_p == “CYCLING” and minimum_outdoor_airflow_p == 0 : UNDETERMINED`
+
```
mkarpman, line 35: I think this should be "PASS".

docs/section4/4-3.md
```diff
@@ -0,0 +1,39 @@
...
+- Call function to get list of inapplicable HVAC systems: `inapplicable_HVAC_systems_list_p = get_HVAC_systems_applicable_10cd(U_RMR,P_RMR)`
+- For each HeatingVentilationAirAconditioningSystem in the P_RMR check if this check is applicable: `for each hvac_p in P_RMR...hvac_systems:`
+    - Check if hvac_p is not in the list of inapplicable systems, if it is not found then set applicability flag to true: `if hvac_p not in inapplicable_HVAC_systems_list_p: rule_applicability_flag = TRUE`   
+    - If rule_applicability_flag = TRUE for hvac_p: `if rule_applicability_flag == TRUE:`
+        - For each fan system in hvac_p: `fan_systems_p in hvac_p.fan_systems:`
+            - Get operation during occupied periods: `operation_during_occupied_p = fans_systems_p.operation_during_occupied`
+            - Get minimum_outdoor_airflow: `minimum_outdoor_airflow_p = fans_systems_p.minimum_outdoor_airflow`
+
+            **Rule Assertion:**
+            - Case 1: For each applicable fan system, if fan system operation during occupied periods is continuous and the minimum OA is greater than 0 in the P_RMR: `if operation_during_occupied_p == “CONTINUOUS” and minimum_outdoor_airflow_p >0 : PASS`
```
mkarpman, line 33: Does this align with the approach that we agreed on for providing outcome on individual component level vs. rule level?

claperle: Yes

docs/section4/4-3.md
```diff
@@ -0,0 +1,39 @@
...
+ 
+**Function Call:** 
+
+1. get_HVAC_systems_applicable_10cd
+
+**Manual Check:** None  
+
+**Rule Logic:**
+**Applicability Check 1:** 
+- Call function to get list of inapplicable HVAC systems: `inapplicable_HVAC_systems_list_p = get_HVAC_systems_applicable_10cd(U_RMR,P_RMR)`
```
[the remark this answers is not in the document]
claperle: Okay, updated.

docs/section4/4-3.md
```diff
@@ -0,0 +1,39 @@
...
+- For each HeatingVentilationAirAconditioningSystem in the P_RMR check if this check is applicable: `for each hvac_p in P_RMR...hvac_systems:`
+    - Check if hvac_p is not in the list of inapplicable systems, if it is not found then set applicability flag to true: `if hvac_p not in inapplicable_HVAC_systems_list_p: rule_applicability_flag = TRUE`
```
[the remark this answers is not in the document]
claperle: Maria to email PNNL or someone about how this should be handled.

docs/section4/4-3.md
claperle: Updated as this appears that it was accepted.

docs/section4/4-3.md
claperle: OK, changed.

docs/section4/4-3.md
```diff
@@ -0,0 +1,41 @@
+# Airside System – Rule 4-3
```
mkarpman, line 1: I see bunch of text above this line - should it be removed?

docs/section4/4-3.md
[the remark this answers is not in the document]
mkarpman: Christina, I emailed the question. I wonder if the current App G rule can even be implemented as written in the simulation tools, as it seems to require in some cases having two different HVAC system types serving the same zone. Consider a proposed design involving a 50,000 SF elementary school with unit ventilators (UVs) serving each classroom and no cooling specified. Based on G3.1 #10 (d), the proposed design should be modeled with the same cooling system type as in the baseline, which gives us System 5. So in this case, it seems that the proposed design should be modeled with UVs running continuously during heating season, and with cycling System 5 providing cooling but no OA during cooling season.

docs/section4/4-3.md
supriyagoel: i believe that text should be removed.

docs/section4/4-3.md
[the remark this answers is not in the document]
supriyagoel: @mkarpman in the scenario you've described above - there would need to be 2 HVAC systems in the RMR, right? So this should not be a problem. But in the scenario where the there is a single system with as design H and baseline C, it would have the same fan and be represented as one system in the schema. Correct?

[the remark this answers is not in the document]
mkarpman: Correct - if a separate HVAC system is created, it's not a problem from the RMR structure perspective but not sure how it would actually be modeled, as there would be effictively two separate HVAC systems, one operating in cooling mode only and another in heating mode only. So during swing months, they would need to alternate based on weather?

If a single HVAC system is modeled for the proposed, then it becomes a problem.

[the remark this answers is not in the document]
supriyagoel: @mkarpman i assume in the UV + cycling system 5 - the rule would be flagged as fail since the UV will likely have min OA as 0. I think that is the right approach- anything I'm missing here?

supriyagoel (approved):

docs/_functions.md
```diff
@@ -28,4 +28,14 @@ Inputs:
 Returns:  
 - **data_element**: A RMR Data Element object.
 
+## match_data_element_exist
+Description: Returns a true or false to indicate whether there is a data element with the matching `name` in either a User, Proposed or Baseline RMR. True means that it exists and false means that it does not exist.
```
mkarpman, line 32: I believe the match is based on ID and not the "name"

mkarpman (approved):

PR #212, opened 2021-12-21
Create Rule18-8.md

xingcx:


docs/section18/Rule18-8.md
```diff
@@ -0,0 +1,41 @@
...
+  - Else, P-RMR is modeled with purchased chilled water, continue to rule logic: `else: CHECK_RULE_LOGIC`
+
+## Rule Logic:  
+
+- Check if B-RMR is modeled with any air-side system that is Type-1, 2, 3, 4, 5, 6, 7, 8, 9, 11, 12, 13, 1b, 3b, 5b, 6b, 7b, 8b, 11b, 12b, 13b, i.e. any air-side system using cooling source other than purchased chilled water, set check flag to True: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-1", "SYS-2", "SYS-3", "SYS-4", "SYS-5", "SYS-6", "SYS-7", "SYS-8", "SYS-9", "SYS-11", "SYS-12", "SYS-13", "SYS-1B", "SYS-3B", "SYS-5B", "SYS-6B", "SYS-7B", "SYS-8B", "SYS-11B", "SYS-12B", "SYS-13B"]): check_flag = TRUE`
+
+**Rule Assertion:**
+
+- Case 1: If check flag is True: `if check_flag: FAIL`
+
```
supriyagoel, line 38: @xingcx I had the same question for 21-1 - not sure what this statement means If check flag is True: `if check_flag: FAIL`

mkarpman (approved):

docs/section18/Rule18-8.md
supriyagoel: @xingcx the same comment above is still confusing. Now I understand that check_flag = true means that the baseline has been modeled with a system that is now PCHW- hence it fails. Could we change 'Check_flag' to something more intuitive like 'non-pchw-system'
then it would be if non_P-CHW_system = TRUE, then FAIL
if non_P-CHW_systeem = FALSE, then PASS

supriyagoel: Updates required, the check is still confusing. I will approve, the RCT team should correct this during implementation.

supriyagoel (approved):

mkarpman (approved):

PR #215, opened 2022-01-10
Create check_purchased_chw_hhw.md

xingcx:


docs/ruleset_functions/check_purchased_chw_hhw.md
```diff
@@ -0,0 +1,75 @@
...
+      - For each zone in building segment: `for zone in building_segment.zones:`
+
+        - For each terminal in zone: `for terminal in zone.terminals:`
+
+          - Check if zonal reheat is connected to any loop in purchased heating loop array: `if terminal.reheat_from_loop in purchased_hhw_loop_array`
+
+            - Set purchased heating flag as True and break logic loop #2: `purchased_chw_hhw_status_dict["PURCHASED_HEATING"] = TRUE, BREAK`
+
+**Returns** `return purchased_chw_hhw_status_dict`  
+
```
supriyagoel, line 70: @xingcx I am slightly confused about the output of this function. Will this simply write out 'TRUE, TRUE' to say that PCHW=TRUE and PHW = TRUE? or will it also provide a list of zones or HVAC systems  that are served by PCHW and PCHW

docs/ruleset_functions/check_purchased_chw_hhw.md
```diff
@@ -0,0 +1,75 @@
...
+**Notes:**
+
+1. For purchased heating, another option is to check if SHW uses purchased heating. If not and RMR has external fluid source, then assume purchased heating is used for space heating.
+
```
supriyagoel, line 74: i think the way it has been done - by checking the HVAC system and zones is more correct and looks good.

docs/ruleset_functions/check_purchased_chw_hhw.md
```diff
@@ -0,0 +1,75 @@
...
+2. Pending question on baseboard, radiant system, active beams in zone, whether these use reheat_from_loop or is part of an HVAC system, and whether terminal has cooling output.
```
supriyagoel, line 75: Terminal has an input for 'heating_from_loop', since baseboards would be modeled as terminals, my guess is they would use 'heating_from_loop' parameter. I do not see a data element for 'reheat_from_loop'

supriyagoel (approved):

docs/ruleset_functions/check_purchased_chw_hhw.md
```diff
@@ -0,0 +1,75 @@
+
+## check_purchased_chw_hhw
+
+Description: Check if RMR is modeled with purchased chilled water as space cooling source or purchased hot water/steam as space heating source. If any system in RMR uses purchased chilled water, function shall return True for purchased chilled water as space cooling source. Similarly, if any system in RMR uses purchased hot water or steam, function shall return True for purchased hot water/steam as space heating source.
+
+Inputs:  
+- **RMR**: The RMR that needs to be checked.
```
mkarpman, line 7: Let's use RMD in all new RDSs.

docs/ruleset_functions/check_purchased_chw_hhw.md
```diff
@@ -0,0 +1,75 @@
...
+
+Returns: 
+- **purchased_chw_hhw_status_dictionary**: A dictionary that saves whether RMR is modeled with purchased chilled water as space cooling source or purchased hot water/steam as space heating source, i.e. {"PURCHASED_COOLING": TRUE, "PURCHASED_HEATING": FALSE}.
+
+Function Call:
+- GET_COMPONENT_BY_ID()
+
+Logic:  
+
+- Check if RMR is not modeled with external fluid source: `if NOT RMR.ASHRAE229.external_fluid_source:`
```
mkarpman, line 17: Don't we need to return the dictionary in this case as well? Suggesting moving line 19 above line 17.

docs/ruleset_functions/check_purchased_chw_hhw.md
mkarpman: I agree. There may also be external fluid source for process load, and the above approach avoids incorrect flags.

docs/ruleset_functions/check_purchased_chw_hhw.md
```diff
@@ -0,0 +1,75 @@
...
+
+          - Check if heating system is connected to any loop in purchased heating loop array: `if preheat_system.hot_water_loop in purchased_hhw_loop_array`
+
+            - Set purchased heating flag as True and break logic loop #2: `purchased_chw_hhw_status_dict["PURCHASED_HEATING"] = TRUE, BREAK`
+
+      - For each zone in building segment: `for zone in building_segment.zones:`
+
+        - For each terminal in zone: `for terminal in zone.terminals:`
+
+          - Check if zonal reheat is connected to any loop in purchased heating loop array: `if terminal.reheat_from_loop in purchased_hhw_loop_array`
```
mkarpman, line 65: I think this should be terminal.heating_from_loop

docs/ruleset_functions/check_purchased_chw_hhw.md
```diff
@@ -0,0 +1,75 @@
...
+
+    - For each building segment in RMR (logic loop #1): `for building_segment in RMR...building_segments:`
+
+      - For each HVAC system in building segment: `for hvac_sys in building_segment.heating_ventilation_air_conditioning_systems:`
+
+        - For each cooling system in HVAC system: `for cooling_system in hvac_sys.cooling_system:`
+
+          - Check if cooling system is connected to any loop in purchased cooling loop array: `if cooling_system.chilled_water_loop in purchased_chw_loop_array`
+
+            - Set purchased cooling flag as True and break logic loop #1: `purchased_chw_hhw_status_dict["PURCHASED_COOLING"] = TRUE, BREAK` (Note XC, under the assumption that there is no orphan hvac system not serving any zones.)
```
mkarpman, line 37: I think we also need to check if purchased cooling is referenced at the terminal level, similar to what is done for heating in lines 61-65 but for terminal.cooling_from_loop

docs/ruleset_functions/check_purchased_chw_hhw.md
xingcx: Updated.

docs/ruleset_functions/check_purchased_chw_hhw.md
xingcx: Added.

PR #216, opened 2022-01-10
RS/CH/Rule5_5

charliepnnl:
This code has not been tested.  Will need to return to it for testing after get_zone_conditioning_category() is finished.


rct229/rules/section5.py
```diff
@@ -110,6 +116,114 @@ def rule_check(self, context, calc_vals, data=None):
...
+class Section5Rule5(RuleDefinitionListIndexedBase):
+    """Rule 5 of ASHRAE 90.1-2019 Appendix G Section 5 (Envelope)"""
+
+    def __init__(self):
+        super(Section5Rule5, self).__init__(
+            rmrs_used=UserBaselineProposedVals(False, True, False),
+            required_fields={
+                "weather": ["climate_zone"],
+            },
+            each_rule=Section5Rule3.BuildingRule(),
```
weilixu, line 128: @charliepnnl  Revise it to Section5Rule5.BuildingRule()

rct229/rules/section5.py
```diff
@@ -110,6 +116,114 @@ def rule_check(self, context, calc_vals, data=None):
...
+
+        class RoofRule(RuleDefinitionBase):
+            def __init__(self):
+                super(Section5Rule5.BuildingRule.RoofRule, self).__init__(
+                    rmrs_used=UserBaselineProposedVals(False, True, False),
+                    required_fields={},
+                )
+
+            def get_calc_vals(self, context, data=None):
+                climate_zone: str = data["climate_zone"]
```
weilixu, line 175: I like the way it defines the variable types and wondering whether we should enforce the style in this project? Will it cause a lot overheads?

rct229/rules/section5.py
```diff
@@ -110,6 +116,114 @@ def rule_check(self, context, calc_vals, data=None):
...
+                    )["u_value"]
+                    if target_u_factor_res == target_u_factor_nonres:
+                        target_u_factor = target_u_factor_res
+
+                return {
+                    "roof_u_factor": roof_u_factor,
+                    "target_u_factor": target_u_factor,
+                    "target_u_factor_res": target_u_factor_res,
+                    "target_u_factor_nonres": target_u_factor_nonres,
+                }
```
weilixu, line 207: Shall we add the ID of the surface in the calc_val?

rct229/rules/section5.py
charliepnnl: Done

rct229/rules/section5.py
```diff
@@ -97,3 +109,113 @@ def rule_check(self, context, calc_vals, data=None):
...
+        def create_data(context, data=None):
+            rmr_baseline = context.baseline
+            return {"climate_zone": rmr_baseline["weather"]["climate_zone"]}
+
+    class BuildingRule(RuleDefinitionListIndexedBase):
+        def __init__(self):
+            super(Section5Rule5.BuildingRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                required_fields={},
+                each_rule=Section5Rule3.BuildingRule.RoofRule(),
```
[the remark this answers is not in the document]
charliepnnl: Done

rct229/rules/section5.py
charliepnnl: Yes, I have started to do this with new code – at least for the few variables that are primitive types. I need to learn more about how to handle more complex types.

rct229/rules/section5.py
charliepnnl: Each surface (roof) will be a node in the output which will include its id, the calc_vals, etc.  So I don't think we need to also add it to the calc_vals.

PR #219, opened 2022-01-18
RS/CH/get_hvac_zone_list_w_area

charliepnnl:


rct229/ruleset_functions/get_hvac_zone_list_w_area_dict.py
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+    """
+    hvac_zone_list_w_area_dict = {}
+
+    for zone in find_all("$..zones[*]", building):
+        terminals = zone.get("terminals")
+        # Note: None and [] are falsey
+        if terminals:
+            zone_area = pint_sum(find_all("spaces[*].floor_area", zone))
```
weilixu, line 32: I am a bit confused by the sum part
It could be because I do not fully understand the zone-space relationship in the schema.

With the current data structure in the sample Json file, it seems to me that:
1. All terminals found in a zone will serve all spaces in the same zone (no terminal-space relationship)
2. We sum the floor area of all spaces under this zone and assign the values to the all connected hvac systems.

Question to clarify:
I am thinking in EnergyPlus, how do we define zone-space-terminal relationship. Typically, one ZoneHVAC:EquipmentList maps one or multiple terminals to one zone, and the zone is a thermal zone, with no spaces under this zone.

With the current setup, one scenario is what if there is a terminal A and terminal B connects to HVAC Sys A and HVAC Sys B and they serve space A and space B correspondently. Both space A and space B is in zone A. Then this line of code will double count the floor area for HVAC sys A and HVAC sys B. 

Feel free to let me know if the zone-space relationship in the schema is different than I interpreted. Thanks!

weilixu (approved):

PR #221, opened 2022-01-21
Rs/wx/rule5 44/46

weilixu:
Rule 5-44 and 5-46

### DIFF
--- a/rct229/rules/section5.py
+++ b/rct229/rules/section5.py
@@ -12,9 +12,9 @@ from rct229.ruleset_functions.get_opaque_surface_type import (
 from rct229.ruleset_functions.get_surface_conditioning_category_dict import (
     get_surface_conditioning_category_dict,
 )
+from rct229.schema.config import ureg
 from rct229.utils.jsonpath_utils import find_all
 from rct229.utils.match_lists import match_lists_exactly_by_id
-from rct229.schema.config import ureg
 
 # Rule Definitions for Section 5 of 90.1-2019 Appendix G
 CONSTANT = schema_enums["InfiltrationMethodType"].CONSTANT.name
@@ -108,6 +108,7 @@ class Section5Rule3(RuleDefinitionListIndexedBase):
 
 
 # ------------------------
+
 
 class Section5Rule8(RuleDefinitionListIndexedBase):
     """Rule 8 of ASHRAE 90.1-2019 Appendix G Section 5 (Envelope)"""
@@ -213,6 +214,7 @@ class Section5Rule8(RuleDefinitionListIndexedBase):
 
 # ------------------------
 
+
 class Section5Rule44(RuleDefinitionListIndexedBase):
     """Rule 44 of ASHRAE 90.1-2019 Appendix G Section 5 (Envelope)"""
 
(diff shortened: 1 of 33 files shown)


rct229/rules/section5.py
```diff
@@ -97,3 +99,88 @@ def rule_check(self, context, calc_vals, data=None):
...
+                rmrs_used=UserBaselineProposedVals(False, True, True)
+            )
+
+        def get_calc_vals(self, context, data=None):
+            failing_infiltration_zone_ids = []
+            baseline_zones = find_all("$..zones[*]", context.baseline)
+            proposed_zones = find_all("$..zones[*]", context.proposed)
+
+            # This assumes that the surfaces all match
+            matched_baseline_zones = match_lists_exactly_by_id(
```
charliepnnl, line 165: I'm thinking that it might be better to use `match_lists_by_id()` here and treat any missing baseline matches as failures. Otherwise, any missing baseline matches will cause this code to throw an exception.

rct229/rules/section5.py
```diff
@@ -97,3 +99,88 @@ def rule_check(self, context, calc_vals, data=None):
...
+
+    class BuildingRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section5Rule44.BuildingRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                required_fields={"$..infiltration[*]": ["modeling_method"]},
+            )
+
+        def get_calc_vals(self, context, data=None):
+            baseline_infiltration = find_all("$..infiltration[*]", context.baseline)
```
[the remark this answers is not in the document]
weilixu: Make sense, I add the fix in the next commit to this PR.

rct229/rules/section5.py
weilixu: In this case, I can add a failure check after this line:

if len (matched_baseline_zones) != len(proposed_zones):
   # Search for the missing zones in the proposed_zone list
   # Add the missing zones to the failed list for report
   return {"failed_zones" : failed_zone_list}

proposed_baseline_zone_pairs = zip(proposed_zones, matched_baseline_zones)
.... continue the code.

rct229/rules/section5.py
```diff
@@ -97,3 +99,88 @@ def rule_check(self, context, calc_vals, data=None):
...
+            index_rmr="proposed",
+            id="5-46",
+            description="The infiltration shall be modeled using the same methodology and adjustments for weather and building operation in both the proposed design and the baseline building design.",
+            rmr_context="buildings",
+        )
+
+    class BuildingRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section5Rule46.BuildingRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, True)
```
charliepnnl, line 156: Should there be required fields? 
1) all zones have an `infiltration` field
2) all infiltration objects have `algorithm_name` and `model_method fields`

charliepnnl (commented):
The two RDS need a little adjustment because thermal_blocks no longer exist in the new schema version.  Those modifications should go into this PR as well.

rct229/rules/section5.py
charliepnnl: I think you will still need to check for b_zone is None inside your for loop.

Here is an alternate idea that avoids match_lists() all together. Starting after line 162:
```
b_zone_dict = {b_zone["id"]: b_zone for b_zone in baseline_zones}. # Hurray for dict comprehensions!

for p_zone in proposed_zones:
    # Get matching baseline zone
    b_zone = b_zone_dict.get(p_zone["id"]) # returns None if the id is not a key in b_zone_dict
    
    if b_zone is None:
      failing_infiltration_zone_ids.append(p_zone["id"])
    else:
       ...
```

weilixu: Done

rct229/rules/section5.py
weilixu: Add the algorithm name and model method to required fields

charliepnnl (approved):

PR #225, opened 2022-01-28
Rt/vt/envelope tcd

vijayvardhan94:
Making a pull request to merge the following jsons to develop: 5-2 to 5-16, 5-33, 5-44, 5-46.


charliepnnl (approved):

jugonzal07:
I'm happy to add this for now just to have the files in the repo for testers to use. Validating these against the TCDs will be a time consuming task that we should think through. I'm wondering if running the rules against them can serve to test them (that's what we did on transformers). This might require further discussion.

jugonzal07:
Actually-- seeing other issues introduced here Vijay. We should reduce what you're merging to just the JSONs.

weilixu:
> I'm happy to add this for now just to have the files in the repo for testers to use. Validating these against the TCDs will be a time consuming task that we should think through. I'm wondering if running the rules against them can serve to test them (that's what we did on transformers). This might require further discussion.

I will vote for smaller PRs with just the JSONs.

I am going to merge this PR for now. In the meantime, I will add this to our next standup for discussion

weilixu (approved):

PR #228, opened 2022-02-02
RCT/CH/set_black_version

charliepnnl:
This is intended to fix the issue with a failing github action. The problem is that a new version of black has been release which the github action was using. But our code was formatted with an older version of black that has slightly different formatting. The simplest solutions seemed to be to explicitly set the version used by the github action to match our own code version.

I also set the version of black in the PipFile so will not inadvertently get the new version if we reinstall on our local machines.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/setup-python@v2
         # This will halt the action if formatting fails
         # Note: This will also fail if there are syntax errors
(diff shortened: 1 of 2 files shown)


mcneillj (approved):

mcneillj:
@charliepnnl Github Actions still are failing when loading black.

charliepnnl:
@mcneillj   My syntax was a little off in specifying the black version for the github action. It all works now and I've merged it into develop.

PR #231, opened 2022-02-04
Create get_HVAC_systems_applicable_10cd.md

claperle:


docs/ruleset_functions/get_HVAC_systems_applicable_10cd.md
```diff
@@ -0,0 +1,82 @@
...
+
+**Returns:**
+- **applicable_hvac_systems_10cd_list_p**: A list that saves all the HVAC systems in which Appendix G Table G3.1 #10 c and d are applicable.
+ 
+**Function Call:** 
+
+1. match_data_element()
+
+**Logic:**
+- # Determine if a HVAC system has been designed or modeled as existing in U-RMR for each zone**
```
mkarpman, line 17: I imagined this function would compare HVAC systems in the U-RMR and P-RMR and identify those that have cooling or heating component in U but not P. Perhaps it also needs to check if there are heating/cooling associated with terminals (via the existing heat_capacity field and cool_capacity field to be added). But it seems to be doing much more than that. Let's discuss the intended logic on the call.

docs/ruleset_functions/get_HVAC_systems_applicable_10cd.md
```diff
@@ -0,0 +1,82 @@
+## get_HVAC_systems_applicable_10cd
+
+**Description:** Get the list of HeatingVentilationAirAconditioningSystem in which Appendix G Table G3.1 #10 c and d are applicable.
+
```
supriyagoel, line 4: @claperle to clarify, we can add the following descriptive text tto the rule description
Where no heating system exists or no heating system has
been submitted with design documents, the system type
shall be the same system as modeled in the baseline building
design and shall comply with but not exceed the requirements
of Section 6.
d. Where no cooling system exists or no cooling system has
been submitted with design documents, the cooling system
type shall be the same as modeled in the baseline building
design and shall comply with the requirements of Section 6

claperle: @supriyagoel I do not think this function is used anywhere. Please let me know if I am mistaken. I have not looked at it since early Feb.

weilixu:
Do we need to close this PR? @claperle @mkarpman

claperle:
> Do we need to close this PR? @claperle @mkarpman

claperle:
Yes, I think we can close this. I do not think it was used anywhere.

*Christina LaPerle (she/her)* PE, BEMP, CEM, CBCP
________________
[image: image.png]
Phone 860.266.6220

weilixu:
Close because the function is no longer needed.

PR #233, opened 2022-02-04
Create get_zones_computer_rooms.md

claperle:


docs/ruleset_functions/get_zones_computer_rooms.md
```diff
@@ -0,0 +1,27 @@
...
+
+**Returns:**
+- **applicable_computer_room_list_p**: A list that saves all zones that have computer rooms associated with them.
+ 
+**Function Call:** 
+
+
+**Logic:**
+- For each zone in P_RMR: `for zone_p in P_RMR...zones:`
+    - Reset applicability flag: `rule_applicability_check = FALSE`
```
supriyagoel, line 16: small comment to rename rule_applicability_check to something like 'has_computer_room'

supriyagoel (approved):
small suggestion to rename the applicability check, else it looks good

docs/ruleset_functions/get_zones_computer_rooms.md
claperle: Good point, modified.

docs/ruleset_functions/get_zones_computer_rooms.md
```diff
@@ -0,0 +1,24 @@
+# get_zones_computer_rooms  
+
+**Description:** Get the list of zones that have at least one computer room space associated with them in the P-RMR.  
+
+**Inputs:**  
+- **P-RMR**: To determine if any of the zones have computer rooms associated with them in the P-RMR.  
+
+**Returns:**  
+- **zones_with_computer_room_list_p**: A list that saves all zones that have at least one computer room associated with them.  
+
```
mkarpman, line 10: The associated RDS is for systems that primarily serve computer rooms. I suggest assuming that "primarily"= computer rooms accounting for more than 50% of floor area of spaces served by the HVAC systems. So I would update this function accordingly, or just implement the criteria within the RDS without using a function.

supriyagoel: @mkarpman here we're simply checking if a zone includes spaces of type computer room. So maybe this function should return 2 values- List of zones with space types = computer room and Area of space type = computer room for each zone.
The calculation of area served by HVAC system would be done within the function, correct?

supriyagoel (changes requested):

docs/ruleset_functions/get_zones_computer_rooms.md
claperle: Implemented a hybrid of your two comments. The function will return the following:
Returns a dictionary with the zones that have at least one computer room space associated with them in the P-RMR as the keys. The values associated with each key are in a list form. The list associated with each key contains the computer room floor area as the first item in the list and the total zone floor area as the second item in the list.

docs/ruleset_functions/get_zones_computer_rooms.md
```diff
@@ -0,0 +1,33 @@
...
+## Logic:   
+- Create object from the RMR input into the function: `X_RMR = RMR`
+- Create empty dictionary: `zones_with_computer_room_dict_x = dict()`
+- For each zone in X_RMR: `for zone_x in X_RMR...Zone:`
+    - Reset zone has computer room boolean variable: `zone_has_computer_room_check = FALSE` 
+    - Reset zone_computer_room_floor_area variable: `zone_computer_room_floor_area = 0`
+    - Reset total_zone_floor_area variable: `total_zone_floor_area = 0`
+    - For each space in zone: `for space_x in zone_x.Spaces:`        
+        - Get space floor_area: `space_floor_area = space_x.floor_area * CONVERSION(m2_TO_ft2)`
```
mkarpman, line 21: I believe we don't need to include conversion tag, as it is done by RCT as part of parsing the RMD.

claperle: That is correct! Revised.

supriyagoel (approved):

PR #235, opened 2022-02-04
Rds/cml/4 11

claperle:


docs/section4/4-11.md
```diff
@@ -0,0 +1,33 @@
...
+**Manual Check:** Check that fan schedules have been modeled appropriately in the baseline and proposed when fan schedules are allowed to differ between the baseline and proposed per Appendix G Section G3.1.1(c).
+
+**Applicability Check 1:** 
+- Get dict of zone id keys and hvac id values applicable to this check: `section_G3.1.1c_applies_zones_hvac_dict_b = get_zones_G3.1.1c_40_EFLHs(B_RMR)`
+- Create list of baseline hvac systems in which Appendix G Section G3.1.1(c) applies: `hvac_systems_section_G3.1.1c_applies_list_b = list(section_G3.1.1c_applies_zones_hvac_dict_b.values())`
+- For each hvac_b in the list in which Section G3.1.1c applies: `for hvac_b in hvac_systems_section_G3.1.1c_applies_list_b:`    
+    **Rule Assertion:**
+    - Case 1: For each applicable hvac system, outcome is undetermined: `UNDETERMINED`
+    - Case 2: Else, for each non applicable zone: `Else: IN_APPLICABLE`
+
```
mkarpman, line 31: Perhaps the following logic should be used for the RDS:
1. Applicability check: verify that the baseline has (a) systems 5-9 (there are additional systems listed in G3.1.1(c) but it was a mistake that was corrected in addendum ab) AND (b) that there are baseline systems 3-4 on the same floor (i.e., zones served by these systems have the same zone.floor_name).   
2. Identify baseline HVAC systems associated with zones 1(b).
3. Identify proposed HVAC systems associated with zones 1(b). For each such system,
    if proposed HVAC system is multizone, and the baseline FanSystem.operating_schedule != TBD then pass, else fail.
4. The question is how to determine the correct schedule that should be modeled based on occupancy schedule of the associated spaces.

supriyagoel: @mkarpman I think i understand this logic. 
Though not sure why in (3) the proposed needs to be MZ. Even if the proposed is SZ, the baseline would need to be SZ system 3 or 4 depending on EFLH, correct? Do we need (3) at all?

supriyagoel:
Comment from MK: MK: Commented. It looks good except I think it should flag scenarios when schedules are modeled the same but should be different.  
I think this logic misses scenarios when schedules erroneously match between baseline and proposed. I think PASS should only apply in the following cases:
schedules match AND (baseline & proposed both served by single-zone systems OR baseline & proposed both served by multi-zone systems)
Since we ask for manual verification, I think the outcome should be UNDETERMINED.
I think we should have Case 2a that would also have an undetermined outcome if schedule_mismatch ==FALSE AND hvac_type_check == TRUE AND proposed_served_by_multizone == TRUE because schedules may be erroneously modeled as the same where they should differ.

docs/section4/4-11.md
```diff
@@ -0,0 +1,65 @@
...
+5. get_list_hvac_systems_associated_with_zone()
+6. get_hvac_zone_list_w_area()
+
+## Rule Logic:  
+**Applicability Check 1:**  
+- Get dictionary for finding baseline system types for hvac systems: `baseline_hvac_sys_type_ids_dict_b = get_baseline_system_types(B_RMR)`
+- Get dictionary with list of zones associated with each HVAC system: `dict_hvac_sys_zones_served_p = get_hvac_zone_list_w_area(P_RMR)`
+- For each zone in the P_RMR: `for zone_p in P_RMR...Zone:`
+    - Reset schedule mismatch variable which is used to track if the schedules mismatches: `schedule_mismatch = FALSE`
+    - Get hvac system with the longest fan operating schedule in the proposed to compare to the baseline fan schedule: `hvac_p = get_hvac_system_with_longest_fan_schedule_p(zone_p.id, P_RMR)`
```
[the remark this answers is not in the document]
supriyagoel: @mkarpman i'm wondering if instead off the longest scehdule, we should be getting the merged schedule. In your example of a zone having a fan schedule from 0-12, and another zone having 13-24, the baseline should have a fan schedule of 0-24, correct? so should this be a merged schedule instead of the longest schedule?

docs/section4/4-11.md
```diff
@@ -0,0 +1,65 @@
...
+                - Loop through each hvac system and check if any serve multiple zones: `for hvac_pr in list_hvac_systems_p:`
+                    - Check if the hvac system serves more than one zone: `if len(list(dict_hvac_sys_zones_served_p[hvac_pr.id]["Zone_List"].values())) > 1:`
+                        - Loop through terminal units associated with the zone to determine which is/are served by this HVAC system and check if it has reheat: `for terminal_p in zone_p.terminals:`
+                            - Check if the terminal unit is served by the HVAC system: `if terminal_p.served_by_heating_ventilation_air_conditioning_system == hvac_pr.id:`
+                                - Check if it has reheat, if it does then assume its a multizone unit and set the proposed served by multizone boolean variable to TRUE: `if terminal_p.heating_capacity >0:proposed_served_by_multizone = TRUE`   
+                                
+    **Rule Assertion:**
+    - Case 1: For each applicable zone, if no schedule mismatch is observed then pass: `if schedule_mismatch == FALSE: PASS`
```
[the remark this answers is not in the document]
claperle: @mkarpman There can be scenarios though where the baseline is multizone and the proposed is single zone and they match. I am thinking of a model we did recently for a school where the baseline was system 5 and then the proposed was all single RTUs. Matching schedules was appropriate in that case so I don't think we would want to fail in that scenario. Maybe if they are matching in the scenarios you describe we can make it undetermined?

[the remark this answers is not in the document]
claperle: I don't think it should fail if there is single zone in proposed and MZ in baseline and the schedules match. I have modeled scenarios where this is the case and the schedules should be modeled the same. A recent school comes to mind. 

I think your recommendation below addresses this.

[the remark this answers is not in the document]
claperle: Also, what if the single zone is being modeled due to different internal loads rather than different schedules? Then there may not be a schedule mismatch.

docs/section4/4-11.md
```diff
@@ -0,0 +1,65 @@
...
+    - Case 2: Else if, there is a fan schedule mismatch but Section G3.1.1(c) appears to apply then Pass: `Elif schedule_mismatch == TRUE AND hvac_type_check == TRUE AND proposed_served_by_multizone == TRUE: PASS and raise_message "THERE IS A SCHEDULE MISMATCH BETWEEN THE BASELINE AND PROPOSED BUT SECTION G3.1.1(c) APPEARS APPLICABLE. VERIFY THAT THE FAN OPERATING SCHEDULE IN THE BASELINE IS IN ALIGNMENT WITH THE OCCUPANCY SCHEDULES."`
```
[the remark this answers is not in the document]
claperle: Also, what if the single zone is being modeled due to different internal loads rather than different schedules? Then there may not be a schedule mismatch.

docs/section4/4-11.md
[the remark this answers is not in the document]
supriyagoel: @mkarpman's reply over email - 	Right. I think modeling the same schedules would also be correct for your #1. The problem is that we cannot definitely establish when #2 applies, and it may apply concurrently with #1, 3 and 4. That said, agree with your scenario above about a school where a zone is served by System 5 in the baseline but a single zone system in the proposed. To accommodate that, I suggest the following change to the logic: PASS should only apply when schedules match AND (baseline & proposed both served by single-zone systems OR baseline & proposed both served by multi-zone systems)

docs/section4/4-11.md
supriyagoel: I believe this RDS is getting quite complicated, it will be difficult to implement and difficult to verify. My suggestion would be to only check for Case 1 - hvac schedules match and the proposed and baseline are both served by either SZ  or MZ systems. All other cases could be 'UNDETERMINED'.

Not sure if there is a clear fail case as well, if yes, we could add that. I think each fail case could have an exception which  could be causing it so maybe there is no value in testing all those additional scenarios as well.

supriyagoel: Other discussions:

image

mkarpman: Exhaust fan schedule is still an open issue https://github.com/open229/ruleset-model-description-schema/issues/93 .

docs/section4/4-11.md
```diff
@@ -0,0 +1,67 @@
...
+    - Case 4: Else if, there is a schedule mismatch between the baseline and proposed then undetermined: `elif schedule_mismatch == TRUE: outcome == "UNDETERMINED" and raise_message "THERE IS A FAN SCHEDULE MISMATCH BETWEEN THE BASELINE AND PROPOSED RMRs FOR THE HVAC SYSTEM(S) SERVING THIS ZONE. FAIL UNLESS TABLE G3.1 SECTION 4 BASELINE COLUMN EXCEPTIONS #1, #2 or #3 IS APPLICABLE"` 
+    - Case 3: Else (schedules match between the baseline and proposed), undertermined: `Else: outcome == "UNDETERMINED" and raise_message "FAN SCHEDULES MATCH BETWEEN THE BASELINE AND PROPOSED RMRs FOR THE HVAC SYSTEM(S) SERVING THIS ZONE. VERIFY THAT MATCHING SCHEDULES ARE APPROPRIATE IN THAT NONE OF THE SECTION 4 BASELINE COLUMN EXCEPTIONS #1, #2 or #3 ARE APPLICABLE"`  
+
```
supriyagoel, line 66: This RDS looks very good to me.
However, I still recommend just checking for case 1 and for all all other scenarios, providing an undetermined outcome with a generic message which could identify the presence of any of these 3 exceptions.
This is adding a lot of checking and work for the RCT and i don't see value of doing all of this because we still get an undetermined outcome for each scenario.

claperle: No problem. Do we all agree then that we do not need lines 53 to 59? These lines are only to see if exception #3 likely applies.

supriyagoel: @claperle - my apologies. we had discussed this at the last RDS  call and I forgot about that conversation. I think it is ok to leave lines 53-59, most of the work is being done before and this is simply grouping zones by floor name and then HVAC systems by floor. This looks good as is and I'll approve it

supriyagoel (approved):

docs/section4/4-11.md
claperle: Issue closed. Exhaust will be checked via an umbrella check on the zone level.

mkarpman (approved):

PR #236, opened 2022-02-04
Log method development

weilixu:
@mcneillj @charliepnnl 

This PR is meant for review (not ready for merge)

This PR implemented the python native logging module to log the msg.
The framework is completed and an example of using the logger is given in Section 5 Rule 2.
Also a sample log file is generated for review /schema/resources/rct.log

If this is good approach, I can start adding the logger to functions and rulesets. Then, for missing keys such as *tilt*, we can handle at code level instead of schema and report them in logs.


weilixu:
@mcneillj @charliepnnl 
Somehow I mixed Rule 44 and 46 in this PR...

For review, you can select the latest three commits to see the logging system...

PR #237, opened 2022-02-04
Log dev

weilixu:
@mcneillj @charliepnnl 

This PR is meant for review (not ready for merge)

This PR implemented the python native logging module to log the msg.
The framework is completed and an example of using the logger is given in Section 5 Rule 2.
Also a sample log file is generated for review /schema/resources/rct.log

If this is good approach, I can start adding the logger to functions and rulesets. Then, for missing keys such as *tilt*, we can handle at code level instead of schema and report them in logs.


rct229/utils/json_utils.py
```diff
@@ -1,11 +1,36 @@
...
-
+import logging
+import sys
 
 def load_json(json_file_path):
     with open(json_file_path) as f:
         return json.load(f)
 
 
+def get_value_from_key(key, context, func_name=None):
```
weilixu, line 10: Added this function to handle the key reading from a dict.

The get function returns None if key is not exist.

Use response to return the error message to parent routine call.

weilixu (commented):
Added KeyError handling.
@charliepnnl @mcneillj

PR #238, opened 2022-02-06
Create Rule21-19.md

xingcx:


mkarpman (approved):

docs/section21/Rule21-19.md
```diff
@@ -0,0 +1,44 @@
...
+  - Else, rule is not applicable to B-RMR: `else: RULE_NOT_APPLICABLE`
+
+## Rule Logic:  
+
+- For each boiler in B-RMR: `for boiler_b in B_RMR.RulesetModelInstance.boilers:`
+
+  **Rule Assertion - Component:**
+
+  - Case 1: For each boiler, if its rated capacity is less than 300,000Btuh, and its efficiency metric is AFUE, and its efficiency is 80%: `if ( boiler_b.rated_capacity < 300000 ) AND ( boiler_b.efficiency_metric == "ANNUAL_FUEL_UTILIZATION" ) AND ( boiler_b.efficiency == 0.8 ): PASS`
+
```
supriyagoel, line 37: @mkarpman do we need to convert the minimum eff in a specific metric to other metrics and verify the value? For example, E+ only takes in Et, the RMR is likely to include Et on all 3 cases. 
The challenge with this is using standard conversion methodology, one option would be to follow the PRM RM and document the same out here.

supriyagoel (changes requested):
question on supporting other metrics for eff

docs/section21/Rule21-19.md
supriyagoel: as discussed on the RDS call, 
image

supriyagoel (approved):

PR #240, opened 2022-02-06
Create get_primary_secondary_loops.md

xingcx:


docs/ruleset_functions/get_primary_secondary_loops.md
```diff
@@ -0,0 +1,61 @@
...
+
+Description: Get the list of primary and secondary loops for HHW or CHW for a RMR.
+
+Inputs:  
+- **RMR**: The RMR that needs to get the list of primary and secondary loops.
+- **loop_type**: "COOLING" (or "HEATING", if needed in future)
+
+Returns: 
+- **primary_secondary_loop_dictionary**: A dictionary that saves the list of primary and secondary loops for the selected loop type, i.e. "HEATING" or "COOLING", e.g. {"PRIMARY": [loop_1.id, loop_2.id], "SECONDARY": [loop3.id]}. If primary loop or secondary loop is not modeled correctly in RMR, return an empty list for both.
+
```
mkarpman, line 12: I find this function a bit confusing. Few thoughts:
1. I don't think producing a list of primary and secondary loops would work because it doesn't specify relationship. E.g., in your example above, it's not clear whether loop3 is associated with loop1 or loop2. 
2. For this function, why not just check the following:
a. Loops referenced by chillers are primary loops (LOOP List A)
b. Loops that are referenced by cooling systems (LOOP List B).
c. Baseline CHW primary/secondary is modeled correctly if the loop in List A (is NOT also in List B) AND (references the loop in List B as its child loop).

xingcx: Rule 22-8 might also need this. "For Baseline chilled water system with cooling capacity of 300 tons or more, the secondary pump shall be modeled with variable-speed drives."

xingcx: The current intent of the function is to return primary and secondary loops in a RMR, and not just judging if primary/secondary loops are modeled correctly. The reason for that is for rule 22-8, we will also need to know which loop is secondary to find the secondary pump. 
For Item#1 above, we can know the relationship by calling FluidLoop.child_loop for all primary loops. 
I think the current logic already checks 2a, 2b, 2c(ii) and 2c(iii), but it does not check 2c(i).

docs/ruleset_functions/get_primary_secondary_loops.md
```diff
@@ -0,0 +1,61 @@
...
+**Returns** `return primary_secondary_loop_dictionary`  
+
+**Notes:**
+
+1. Primary secondary configuration is correct if:
+1). a cooling loop is either connected to chiller or cooling coil, but not both
+2). if a cooling loop is connected to chiller, it shall have child loop(s)
+3). if a cooling loop is connected to cooling coil, it shall not have any child loops
+4). the collection of child loops of all of 2) shall be the same as all of 3)
```
supriyagoel, line 61: @xingcx as discussed, maybe we can add some information here to help future reviewers
1. Primary Only Loop: A cooling loop which is connected to chiller and cooling coil.
2. Primary Secondary Loop: Primary loop would be connected to chiller but not to cooling coil. The loop connected to the chiller would have child loop(s). The secondary loop would be connected to the cooling coil and not have child loop.

docs/ruleset_functions/get_primary_secondary_loops.md
```diff
@@ -0,0 +1,71 @@
+
+## get_primary_secondary_loops
+
+Description: Get the list of primary and secondary loops for HHW or CHW for a RMR.
+
+Inputs:  
+- **RMR**: The RMR that needs to get the list of primary and secondary loops.
+- **loop_type**: "COOLING" (or "HEATING", if needed in future)
+
```
mkarpman, line 9: As it stands now, the function only applies to cooling (e.g., it looks at chillers but not boilers) so I think we either need to remove references to heating here (and perhaps update name of the function), or expand the function to also cover heating loops.

docs/ruleset_functions/get_primary_secondary_loops.md
```diff
@@ -0,0 +1,61 @@
...
+  - For each HVAC system in building segment: `for hvac in building_segment.heating_ventilation_air_conditioning_systems`
+
+    - Check if HVAC system has chilled water coil: `if hvac.cooling_system.chilled_water_loop:`
+
+      - Save chilled water loop that serves cooling systems to CHW coil loop array to get all loops connected to cooling coils: `chw_coil_loop_array.append(hvac.cooling_system.chilled_water_loop)` (Note XC, need to include zonal cooling coil as well once the schema is updated)
+
+- For each fluid loop in RMR: `for fluid_loop in RMR.RulesetModelInstance.fluid_loops:`
+
+  - Check if loop type is the same as input loop_type: `if fluid_loop.type == loop_type:`
+
```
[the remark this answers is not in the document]
mkarpman: As it stands now, the function only supports cooling so suggest taking out line 35 and updating function name accordingly.

docs/ruleset_functions/get_primary_secondary_loops.md
```diff
@@ -0,0 +1,66 @@
...
+    - Check if HVAC system is baseline system Type-7, 8, 11.1, 11.2, 12, 13, 7b, 8b, 11b, 12b, 13b: `if any(hvac.id in baseline_hvac_system_dict[sys_type] for sys_type in ["SYS-7", "SYS-8", "SYS-11.1", "SYS-11.2", "SYS-12", "SYS-13", "SYS-7B", "SYS-8B", "SYS-11B", "SYS-12B", "SYS-13B"]):`
+
+      - Save chilled water loop that serves cooling systems to non-process CHW coil loop array (array for all loops connected to cooling coils): `chw_coil_loop_array.append(hvac.cooling_system.chilled_water_loop)`
+
+- For each fluid loop in B-RMR: `for fluid_loop in B-RMR.RulesetModelInstance.fluid_loops:`
+
+  - Check if loop type is cooling: `if fluid_loop.type == "COOLING":`
+
+    - Check if loop is connected to both chiller(s) and cooling coil(s), break logic and return an empty dictionary: `if ( fluid_loop.id in chiller_loop_array ) AND ( fluid_loop in chw_coil_loop_array ): BREAK and return primary_secondary_loop_dictionary = {}`
```
mkarpman, line 30: So this terminates the function if there is at least one loop that is modeled as primary only. 
a. What if this is a process loop? 
b. Also, even if it's an incorrectly modeled baseline CHW loop, I wonder if instead it should return primary loop with no secondary? This way, we could still use the function to verify rules that apply just to primary loops.

docs/ruleset_functions/get_primary_secondary_loops.md
```diff
@@ -0,0 +1,66 @@
...
+
+      - Check if all child loops of loop serve baseline system Type-7, 8, 11.1, 11.2, 12, 13, 7b, 8b, 11b, 12b, 13b only (to exclude CHW loop served by process chiller(s)): `if child_loop_id in chw_coil_loop_array for child_loop_id in fluid_loop.child_loops:`
+
+        - Save loop to primary loop array: `primary_loop_array.append(fluid_loop.id)`
+
+        - Save all child loops to child loop array: `child_loop_array.append(child_loop_id for child_loop_id in fluid_loop.child_loops)`
+
+    - Else if loop is connected to baseline system cooling coil(s) only: `else if fluid_loop in chw_coil_loop_array:`
+
+      - Check if loop has any child loop, break logic and return an empty dictionary: `if fluid_loop.child_loops != NULL: BREAK and return primary_secondary_loop_dictionary = {}`
```
mkarpman, line 42: Maybe not terminate the function in this case but just not count this loop as secondary loop in the dictionary returned by the function?

docs/ruleset_functions/get_primary_secondary_loops.md
```diff
@@ -0,0 +1,66 @@
...
+
+      - Else, save loop to secondary loop array: `else: secondary_loop_array.append(fluid_loop.id)`
+
+- Check if child loop array and secondary loop array is not the same, break logic and return an empty dictionary: `if child_loop_array.sort() != secondary_loop_array.sort(): BREAK and return primary_secondary_loop_dictionary = {}`
```
mkarpman, line 46: I am not quite sure what this is doing?

docs/ruleset_functions/get_primary_secondary_loops.md
xingcx: a. The logic will skip process loops. Line 22-24 only save loops that serves baseline system type. The description on this line is updated to make it more clear. Line 34 only saves primary loops connected to baseline systems, which excludes process systems (assuming a process system will be the same as P-RMR and will not be a baseline system type).

b. The logic is currently set to determine whether a baseline CHW system is modeled with primary-secondary configuration. So if any CHW loop in the baseline does not meet the primary-secondary requirement, the function will return an empty dictionary (i.e. baseline is not modeled with primary-secondary configuration). We can discuss in the meeting if we want to return primary loop with no secondary loop.

docs/ruleset_functions/get_primary_secondary_loops.md
xingcx: This checks if all loops connected to baseline system cooling coils are the same as the child loops of primary loop(s). For example, loops connected to baseline system cooling coils could be tertiary loops instead of secondary loops.

docs/ruleset_functions/get_primary_secondary_loops.md
supriyagoel: as discussed on 2022.04.26, we discussed to modify the function to return the list of primary loops and the list of secondary rules

supriyagoel: As discussed, 

- 22-25, would verify the primary loop is continuous constant flow
- 22-36, would verify the secondary loop is variable
- Need one more RDS to verify that baseline loop configuration is primary secondary

xingcx: Pump rules:
22-25 would verify the primary pump is constant speed.
22-10 would verify the secondary pump is constant speed (<300ton) or with VFD (>=300ton).
Loop rules:
22-7 would verify that baseline has been modeled with P S configuration.
22-36 would verify continuous constant flow primary loop and variable-flow secondary loop.
22-34 would verify baseline has been modeled with only one CHW plant (i.e. one primary and one secondary loop).

mkarpman (approved):

supriyagoel (approved):

PR #241, opened 2022-02-06
Create Rule22-6.md

xingcx:


docs/section22/Rule22-6.md
```diff
@@ -0,0 +1,44 @@
+
+# CHW&CW - Rule 22-6  
+
+**Rule ID:** 22-6  
+**Rule Description:** For Baseline chilled water loop that is not purchased chilled water and serves computer room HVAC systems (System Type-11), The maximum reset chilled-water supply temperature shall be 54F.
```
mkarpman, line 5: Re-reading 3.1.3.9, I think the CHW reset temperatures are the same for all CHW systems including System 11 (computer rooms). The only difference is that for system 11 the reset is based on coil with the maximum cooling load (rule 22-5) and for all other systems it is based on OA temperature (rule 22-3). So we should either update 22-4 RDS so that it applies to System 11 (which I think is a better option) OR update this RDS so that it's exactly the same as 22-4 but applies to system 11 only.

mkarpman (approved):

supriyagoel (approved):

supriyagoel:
@weilixu this is now ready to merge

PR #243, opened 2022-02-06
Create Rule22-20.md

xingcx:


docs/section22/Rule22-20.md
```diff
@@ -0,0 +1,40 @@
+
+# CHW&CW - Rule 22-20  
+
+**Rule ID:** 22-20  
+**Rule Description:** The baseline heat rejection Loop temperature reset = Load Reset.
```
mkarpman, line 5: G3.1.3.11 that this RDS is meant to implement has requirements quoted below. I believe the requirements represent reset based on wet bulb temperature, not based on load. Let's discuss and perhaps consider adding wet-bulb temperature reset option to the schema. 

Condenser-water design supply temperature shall be calculated using the cooling tower approach to the 0.4% evaporation design wet-bulb temperature as generated by the formula below, with a design temperature rise of 10°F:
Approach10°F Range = 25.72 – (0.24 × WB) 
where WB is the 0.4% evaporation design wet-bulb temperature (°F); valid for wet bulbs from 55°F to 90°F.

supriyagoel: @mkarpman I might be misunderstanding this requirement, but I don't read this as a wet bulb reset either. we should discuss this

mkarpman: I quoted a wrong section. I think the relevant rule is below. I don't think this is LOAD_RESET, but no other TemperatureResetTypeOptions seem to fit. Perhaps "Other"? 

The tower shall be controlled to maintain a leaving water temperature, where
weather permits, per Table G3.1.3.11, floating up to the design leaving water temperature
for the cooling tower.

docs/section22/Rule22-20.md
```diff
@@ -0,0 +1,40 @@
...
+  - Else, rule is not applicable to B-RMR: `else: RULE_NOT_APPLICABLE`
+
+## Rule Logic:  
+
+- For each heat rejection device in B_RMR: `for tower_b in B_RMR.RulesetModelInstance.heat_rejections:`
+
+  **Rule Assertion:**
+
+  - Case 1: If heat rejection temperature reset type is load reset: `if tower_b.reset_type == "LOAD_RESET": PASS`
+
```
supriyagoel, line 37: FluidLoopDesignAndControl has TemperatureResetOptions which can be 'Other'. However, I think should be more specific like 'SystemSetpoint' or 'LeavingWaterTemperature' and 'EnteringWaterTemperature' and for the baseline we would verify that it is 'LeavingWaterTemperature'

docs/section22/Rule22-20.md
xingcx: Updated.

supriyagoel (approved):

docs/section22/Rule22-20.md
```diff
@@ -0,0 +1,43 @@
...
+# CHW&CW - Rule 22-20  
+
+**Schema Version:** 0.0.16  
+**Mandatory Rule:** True
+**Rule ID:** 22-20  
+**Rule Description:** The tower shall be controlled to maintain a leaving water temperature, where weather permits, per Table G3.1.3.11.
```
mkarpman, line 7: Please change this to say "The tower shall be controlled to maintain a leaving water temperature, where weather permits." Compliance with the table G3.1.3.11 is verified by Rule 22-21. It's my mistake - sorry about that!

mkarpman (approved):

PR #246, opened 2022-02-06
Create Rule22-23.md

xingcx:


docs/section22/Rule22-23.md
```diff
@@ -0,0 +1,52 @@
...
+    - Else if chiller rated capacity is 300Ton or more and is less than 600Ton, set size category to >= 300TONS and < 600TONS": `else if rated_capacity_b < 600 * CONVERSION(TON_TO_BTUH): size_category = ">= 300TONS AND < 600TONS"`
+
+    - Else, set size category to > 600TONS: `else: size_category = "> 600TONS"`
+
+    - Get baseline chiller efficiency requirements: `kW_ton_full_load_b = data_lookup(table_G3_5_3, chiller_b.compressor_type, size_category, "FL"), kW_ton_part_load_b = data_lookup(table_G3_5_3, chiller_b.compressor_type, size_category, "IPLV")`
+
+      **Rule Assertion - Component:**
+
+      - Case 1: For each chiller, if chiller full-load and part-load efficiencies are modeled at the minimum efficiency levels in accordance with Tables G3.5.3 `if ( chiller_b.full_load_efficiency == kW_ton_full_load_b ) AND ( chiller_b.full_load_efficiency == part_load_efficiency ) AND ( chiller_b.part_load_efficiency_metric == "INTEGRATED_PART_LOAD_VALUE" ): PASS`
+
```
mkarpman, line 49: 1. chiller.full_load_efficiency is expressed in COP, while the values in table G3.5.3 are in kW/ton, so please add conversion to chiller_b.full_load_efficiency * CONVERSION (COP to kW_ton)==kW_ton_full_load_b
2. after AND should say (chiller_b.part_load_efficiency==...) (instead of full load)

xingcx: Updated.

supriyagoel: @mkarpman we had tacked the same discussion on efficient conversion when we evaluated boiler efficient rule and decided that is the RMR uses a value other than AFUE then the rule outcome would be fail. The logic there has been that the conversion from AFUE to Et isnt as straightforward. However, here we are doing a kW per ton to COP conversion - i assume thats because there is a standard protocol for this conversion? Please confirm

mkarpman: Correct. AFUE reflects annualized average efficiency while Et is steady state thermal efficiency, so there is no physics-based conversion that is correct for any model of boiler. On the other hand, chiller.full_load_efficiency is expressed as COP (W/W) while ASRHAE tables have it as kW/Ton, so we need to apply conversion.

mkarpman: Taking a fresh look at this RDS I think we should separate part load efficiency into a separate RDS that would be based on chiller.power_validation_points. IPLV is typically not entered / used by simulation tools, which would mean that majority of models would fail the RDS

supriyagoel: @mkarpman I believe the decision on the IPLV verification was to verify that the baseline chiller validation points align with the part load performance curves that are being added to App G 2022, is that correct?

mkarpman: I thought the plan was to check the IPLV field via RDS and have an RMR test to verify that the tool converts either performance curve or validation points to the IPLV correctly.

docs/section22/Rule22-23.md
```diff
@@ -0,0 +1,50 @@
...
+# CHW&CW - Rule 22-23  
+
+**Schema Version:** 0.0.10 **Mandatory Rule:** True
+**Rule ID:** 22-23  
+**Rule Description:** The baseline chiller efficiencies shall be modeled at the minimum efficiency levels, both part load and full load, in accordance with Tables G3.5.3.
```
mkarpman, line 6: I suggest that we split this RDS to separately verify full load and part load efficiency. All tools can output the full load, but some may not be able to do part load.

docs/section22/Rule22-23.md
supriyagoel: @mkarpman recommends splitting this RDS into two RDS. the first RDS will verify full load and part load efficiency. 

Regarding IPLV test:  We will check the absolute value of the IPLV through RDS and had curve or validation points to IPLV conversion checked via RMR test. 
Jeremy Lerond  will provide curves and table. For each set of curves and tables, there will be a corresponding IPLV and a RMD test to verify that simulation tests do this correctly and RDS will only verify IPLV.

xingcx: 22-23 is updated to only check full load efficiency. Add new RDS Rule 22-37 for part load efficiency check.

supriyagoel (approved):

mkarpman (approved):

PR #248, opened 2022-02-07
Rds/cml/10 1

claperle:


docs/section10/10-1.md
```diff
@@ -0,0 +1,42 @@
...
+            - Get the served_by_heating_ventilation_air_conditioning_systems for each terminal: `heating_ventilation_air_conditioning_systems_b = terminal_b.served_by_heating_ventilation_air_conditioning_systems`
+            - Add to list of heating_ventilation_air_conditioning_systems_list_b as the code loops through the terminal units: `heating_ventilation_air_conditioning_systems_list_b = heating_ventilation_air_conditioning_systems_list_b.append(heating_ventilation_air_conditioning_systems_b)`                
+- Convert the list of heating_ventilation_air_conditioning_systems_list_b associated with the applicable zones to a set and the back to a list to eliminate duplicates after looping through all zones: `heating_ventilation_air_conditioning_systems_list_b = list(set(heating_ventilation_air_conditioning_systems_list_b))`
+- For each hvac_b in heating_ventilation_air_conditioning_systems_list_b: `For in heating_ventilation_air_conditioning_systems_list_b:` 
+    **Rule Assertion:**
+    - Case 1: For each applicable hvac system, outcome is undetermined, manual check needs to be performed: `UNDETERMINED`
+
```
supriyagoel, line 39: @claperle and @mkarpman we could check the humidification type for the baseline HVAC system and identify the ones which are adiabatic and non adiabatic. we could then identify the corresponding proposed HVAC systems and identify the ones which need to comply with the prescriptive requirements and the ones that dont. That might help provide more information. We're already getting all the prerequisite data points here.

supriyagoel: @mcneillj we should probably add a category up to which identifies a non-mandatory rule. I didn't realize this one was a non-mandatory rule and instead assumed the undetermined outcome was due to other limitations (which is also the case in this one)

claperle: I'll wait for a discussion on this with James and the group.

docs/section10/10-1.md
```diff
@@ -0,0 +1,41 @@
...
+**Evaluation Context:** Evaluate each HeatingVentilationAirAconditioningSystem   
+
+**Applicability Checks:** 
+
+1. The zone in the P-RMR includes humidification.
+
+**Function Call:** 
+
+1. match_data_element()
+2. get_zones_hvac_system_humid_p()
```
mkarpman, line 21: Not sure we need this function - see below.

docs/section10/10-1.md
```diff
@@ -0,0 +1,41 @@
...
+- For each zone in P_RMR: `zone_p in P_RMR...Zone:`
+    - Check if the zone has humidification in the P_RMR, if so then proceed with logic: `if applicable_zones_hvac_sys_humid_dict_p.has_key(zone_p.id):`
+        - Get analogous zone ID in the B_RMR: `zone_b = match_data_element(B_RMR,Zones,zone_p.id)`
+        - Get list of hvac systems associated with the zone (should only be one): `heating_ventilation_air_conditioning_systems_b = get_list_hvac_systems_associated_with_zone(B_RMR, zone_b.id)`
+        - Add to list of heating_ventilation_air_conditioning_systems_list_b as the code loops through the zones: `heating_ventilation_air_conditioning_systems_list_b = heating_ventilation_air_conditioning_systems_list_b.append(heating_ventilation_air_conditioning_systems_b)`                
+- Convert the list of heating_ventilation_air_conditioning_systems_list_b associated with the applicable zones to a set and the back to a list to eliminate duplicates after looping through all zones: `heating_ventilation_air_conditioning_systems_list_b = list(set(heating_ventilation_air_conditioning_systems_list_b))`
```
mkarpman, line 34: For the applicability check, I would do the following:
1. Cycle through all HVAC systems in the proposed design to see if any of them have heatingsystem.humidity_type != None OR Null? If yes, the RDS applies and we need to identify the corresponding baseline systems as described in the following steps. 
2. Identify zones served by these proposed systems (I think we already have a function for that?), and then use another (I think existing) function to identify HVAC systems in the baseline that serve at least one of these zones.  HVAC systems in the baseline that serve

docs/section10/10-1.md
```diff
@@ -0,0 +1,41 @@
+# Airside systems – Rule 10-1  
+**Schema Version:** 0.0.12  
+**Mandatory Rule:** False    
+**Rule ID:** 10-1  
+ 
+**Rule Description:** The proposed design includes humidification (which does not comply with 90.1-2019 Section 6.5.2.4) and the baseline building design has been modeled with adiabatic humidification, as expected.  
+
```
mkarpman, line 7: I suggest that we aggregate 10-1 and 10-2, to avoid having to differentiate between systems in the proposed design that do and do not comply with Section 6.5.2.4.

docs/section10/10-1.md
supriyagoel: @claperle and @mkarpman i believe we would use #365 to get list of all HVAC systems and then in this RDS identify the HVAC systems in that list where humidification type is adiabatic or other (if null or NONE then no humidification): `if humidification_type_p in ["ADIABATIC", "OTHER"]:`

If yes, then that makes sense to me.

dont think we need #258 then, correct?

docs/section10/10-1.md
```diff
@@ -0,0 +1,41 @@
...
+1. match_data_element()
+2. get_zones_hvac_system_humid_p()
+3. get_list_hvac_systems_associated_with_zone()
+
+**Manual Check:** Check that the applicable baseline HVAC systems were modeled per the rules of G3.1-10 HVAC System baseline column which states that if the proposed design includes humidification then the baseline building design shall use adiabatic humidification (if the proposed humidification system does not comply with 90.1-2019 Section 6.5.2.4). For each applicable zone (i.e. zones with humidification in the proposed) check if the humidification system associated with the zone complies with 90.1-2019 Section 6.5.2.4. If it does NOT then check that the baseline system serving the zone was modeled with adiabatic humidification. 
+ 
+**Rule Logic:**  
+**Applicability Check 1:** 
+- Call function to get dictionary with applicable zones (key) and hvac systems (values) in P_RMR: `applicable_zones_hvac_sys_humid_dict_p = get_zones_hvac_system_humid_p(P_RMR)`
```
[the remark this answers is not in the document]
supriyagoel: Yup, I agree that we dont need a new function. We should try and be efficient with the use of functions as much as we can and re-use them. Each function is a significant amount of work for the RCT team.

docs/section10/10-1.md
supriyagoel: as @mkarpman summarized in the email, #248  could be merged with #250

docs/section10/10-1.md
```diff
@@ -0,0 +1,41 @@
...
+
+1. The zone in the P-RMR includes humidification.
+
+**Function Call:** 
+
+1. match_data_element()
+2. get_zones_hvac_system_humid_p()
+3. get_list_hvac_systems_associated_with_zone()
+
+**Manual Check:** Check that the applicable baseline HVAC systems were modeled per the rules of G3.1-10 HVAC System baseline column which states that if the proposed design includes humidification then the baseline building design shall use adiabatic humidification (if the proposed humidification system does not comply with 90.1-2019 Section 6.5.2.4). For each applicable zone (i.e. zones with humidification in the proposed) check if the humidification system associated with the zone complies with 90.1-2019 Section 6.5.2.4. If it does NOT then check that the baseline system serving the zone was modeled with adiabatic humidification.
```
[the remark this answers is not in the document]
claperle: We are doing this RDS based on the HVAC system an dnot the zone so I don't think offering the zone woudl work unless we wanted to change it around

docs/section10/10-1.md
claperle: Done

docs/section10/10-1.md
claperle: All comments implemented

docs/section10/10-1.md
```diff
@@ -0,0 +1,65 @@
...
+
+1. match_data_element()
+2. get_baseline_system_types() 
+3. get_list_hvac_systems_associated_with_zone()
+4. get_hvac_zone_list_w_area()
+
+**Manual Check:** For zones modeled with humidification in the baseline and proposed,check that the baseline system serving this zone is modeled with adiabatic humidification if the specified humidification system complies with 90.1 - 2019 Section 6.5.2.4, and that the baseline system serving this zone is modeled with non-adiabatic humidification if the specified system does not comply with Section 6.5.2.4. 
+ 
+**Rule Logic:**  
+**Applicability Check 1:**
```
supriyagoel, line 28: @claperle i believe you've done this in Section 4 RDS, you've added a statement explaining what each chunk of code is doing. For example, before line 31, we would add 'Get all systems in PRM with humidification specified in heating system or preheat system and get corresponding list of zones served by those systems'

You're already doing this in line 45, which makes that part clear.

supriyagoel (approved):

docs/section10/10-1.md
claperle: Added

mkarpman (approved):

PR #251, opened 2022-02-07
Rds/cml/10 3

claperle:


docs/section10/10-3.md
```diff
@@ -0,0 +1,40 @@
...
+1. get_zones_computer_rooms
+
+**Manual Check:** Check that the applicable baseline HVAC systems (hvac systems that serve comuter rooms) were modeled per the rules of G3.1-10 HVAC System baseline column which states that for systems serving computer rooms, the baseline building design shall not have reheat for the purpose of dehumidification.  
+ 
+
+**Rule Logic:**
+**Applicability Check 1:** 
+- Call function to get list of applicable zones in P_RMR: `applicable_zones_comp_room_p = get_zones_computer_rooms(P_RMR)`
+- For each zone in P_RMR: `zone_p in P_RMR...zones:`
+    - Check if the zone includes a computer room, if so then applicability flag is true: `if zone_p.id in applicable_zones_comp_room_p: rule_applicability_flag = TRUE`
```
supriyagoel, line 27: As commented in the function 'get_zones_computer_rooms', having a more intutive name than 'rule_applicability_flag' would be helpful. This could be something like 'zone list computer room'

docs/section10/10-3.md
```diff
@@ -0,0 +1,40 @@
...
+            - Get the served_by_heating_ventilation_air_conditioning_systems for each terminal: `heating_ventilation_air_conditioning_systems_b = terminal_b.served_by_heating_ventilation_air_conditioning_systems`
+            - Add to list of heating_ventilation_air_conditioning_systems_list_b as the code loops through the terminal units: `heating_ventilation_air_conditioning_systems_list_b = heating_ventilation_air_conditioning_systems_list_b.append(heating_ventilation_air_conditioning_systems_b)`                
+- Convert the list of heating_ventilation_air_conditioning_systems_list_b associated with the applicable zones to a set and the back to a list to eliminate duplicates after looping through all zones: `heating_ventilation_air_conditioning_systems_list_b = list(set(heating_ventilation_air_conditioning_systems_list_b))`
+- For each hvac_b in heating_ventilation_air_conditioning_systems_list_b: `For in heating_ventilation_air_conditioning_systems_list_b:` 
+    **Rule Assertion:**
+    - Case 1: For each applicable hvac system, outcome is undetermined, manual check needs to be performed: `UNDETERMINED`
+
```
supriyagoel, line 37: This looks good, we could potentially add an outcome message which provides a list of HVAC systems which could not be verified for rule ID 10-3. Something like '_HVAC System ID [X] serves a computer room and should not use reheat for dehumidification. This needs to be manually verified'_

claperle: Addressed

docs/section10/10-3.md
claperle: Addressed

docs/section10/10-3.md
```diff
@@ -0,0 +1,40 @@
...
+1. get_zones_computer_rooms()  
+2. get_list_hvac_systems_associated_with_zone()  
+
+**Manual Check:** Check that the applicable baseline HVAC systems (hvac systems that serve comuter rooms) were modeled per the rules of G3.1-10 HVAC System baseline column which states that for systems serving computer rooms, the baseline building design shall not have reheat for the purpose of dehumidification.   
+
+**Rule Logic:**  
+**Applicability Check 1:**  
+- Call function to get dictonary of zones that include computer rooms and associated square footage in P_RMR: `zones_that_include_comp_rooms_dict_p = = get_zones_computer_rooms(P_RMR)`
+- Get list of the zones that include computer rooms: `zones_that_include_comp_rooms_list_p = zones_that_include_comp_rooms_dict_p.keys()`
+- For each zone in zones_that_include_comp_rooms_list_p: `zone_p in zones_that_include_comp_rooms_list_p:`
```
mkarpman, line 29: I think for each such zone, we should check the following:
(1) If zone.maximum_humidity_setpoint_schedule is Null OR have all values equal to 1 (100% relative humidity), then pass; otherwise
(2) Find all terminals associated with each identified zone; if for each of them Terminal.heating_capacity=0 then Pass, otherwise Fail

docs/section10/10-3.md
mkarpman: It seems that we could fairly easily verify it as suggested above. So I would consider making this rule mandatory.

docs/section10/10-3.md
supriyagoel: @mkarpman not sure why we need 1, a zone might not even have a humidity setpoint schedule, correct?
for terminal heating capacity =0, wouldnt a better check be terminal.heating_source = 'None'?

docs/section10/10-3.md
```diff
@@ -0,0 +1,40 @@
...
+
+1. get_zones_computer_rooms()  
+2. get_list_hvac_systems_associated_with_zone()  
+
+**Manual Check:** Check that the applicable baseline HVAC systems (hvac systems that serve comuter rooms) were modeled per the rules of G3.1-10 HVAC System baseline column which states that for systems serving computer rooms, the baseline building design shall not have reheat for the purpose of dehumidification.   
+
+**Rule Logic:**  
+**Applicability Check 1:**  
+- Call function to get dictonary of zones that include computer rooms and associated square footage in P_RMR: `zones_that_include_comp_rooms_dict_p = = get_zones_computer_rooms(P_RMR)`
+- Get list of the zones that include computer rooms: `zones_that_include_comp_rooms_list_p = zones_that_include_comp_rooms_dict_p.keys()`
```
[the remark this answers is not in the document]
supriyagoel: @mkarpman I agree with you on #1, though i wonder if that is an oversight on the 90.1 committee's side. essentially we're saying that if even a single zone on a floor is a computer room, then the baseline cannot have reheat (correct?)

docs/section10/10-3.md
supriyagoel: @mkarpman we need to discuss the solution you've proposed, would be good to make this rule mandatory

docs/section10/10-3.md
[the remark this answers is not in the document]
claperle: Implemented suggestions

docs/section10/10-3.md
claperle: Implemented suggestions (both MK and SG)

docs/section10/10-3.md
```diff
@@ -0,0 +1,47 @@
...
+            - Check if the value does not equal one, if it does not then set boolean variable to TRUE: `if zone_b.maximum_humidity_setpoint_schedule.hourly_values[x] != 1: does_zone_have_dehumification = TRUE`
+    - Loop through the terminal units associated with the zone and check if any have heating: `For terminal_b in zone_b.terminals:`
+        - Check if the heating source is Null: `if terminal_b.heating_source == Null: has_terminal_heating = FALSE`
+        - Else if, check if the heating source equals None: `elif terminal_b.heating_source == "None": has_terminal_heating = FALSE`
+        - Else if, check if heating source does not equal none but heating capacity is 0: `elif terminal_b.heating_source != "None" and terminal_b.heating_capacity == 0 : has_terminal_heating = FALSE`
+        - Else, it has terminal heating: `Else: has_terminal_heating = TRUE`
+
+    **Rule Assertion:**
+    - Case 1: For each applicable hvac system (hvac_b above) if the zone served does not have dehumidification then pass: `if does_zone_have_dehumification ==  FALSE: outcome == "PASS"`
```
supriyagoel, line 42: @mkarpman I am slightly confused here, the rule states that the zone shall not have reheat for dehumidification, it doesn't say that it shall not have dehumidification at all. Is there any scenario where the baseline might have dehumidification?

docs/section10/10-3.md
```diff
@@ -0,0 +1,47 @@
...
+- For each zone_b in zones_that_include_comp_rooms_list_b: `For zone_b in zones_that_include_comp_rooms_list_b:`
+    - Reset does_zone_have_dehumification boolean variable: `does_zone_have_dehumification = FALSE`
+    - Reset has_terminal_heating boolean variable: `has_terminal_heating = FALSE`
+    - Get the hvac system associated with the zone (should only be one, this RDS assumes one HVAC system): `hvac_b = get_list_hvac_systems_associated_with_zone(B_RMR, zone_b.id)`
+    - Check if the zone's maximum_humidity_setpoint_schedule equals Null: `if zone_b.maximum_humidity_setpoint_schedule == Null: does_zone_have_dehumification = FALSE`
+    - Else (i.e. if the zone's maximum_humidity_setpoint_schedule does NOT equal Null): `Else:`
+        - Loop though all elements in the schedule and check if the value equals 1 (convert to an 8760 schedule with a function if needed): `For x in Range(len(zone_b.maximum_humidity_setpoint_schedule.hourly_values))`
+            - Check if the value does not equal one, if it does not then set boolean variable to TRUE: `if zone_b.maximum_humidity_setpoint_schedule.hourly_values[x] != 1: does_zone_have_dehumification = TRUE`
```
supriyagoel, line 34: @claperle and @mkarpman I am unclear on what 'maximum humidity setpoint schedule' is meant to contain. Attached screenshot. The description says its meant to contain the min or max relative humidity setpoint and then the notes say it would be a fraction between 0 and 1. Which one is correct?
If it is a fraction, do we have a place where we specifying the absolute humidity setpoint which is getting adjusted by this fraction schedule. 


image

mkarpman: @supriyagoel I think there is no conflict because relative humidity is a percentage needed for saturation at the same temperature.

docs/section10/10-3.md
mkarpman: Baseline would have dehumidificaiton if there is dehumidification in the proposed. However, it cannot be achieved through reheat. Here is a relevant section from 90.1 User's manual: "Computer rooms also typically have an upper 
humidity limit. This requires dehumidification. Due to the high sensible load in computer rooms, dehumidification is typically achieved without reheat. Due to this, reheating with computer-room air conditioners is not allowed in the baseline building regardless of the space relative humidity. "

docs/section10/10-3.md
```diff
@@ -0,0 +1,47 @@
...
+    - Loop through the terminal units associated with the zone and check if any have heating: `For terminal_b in zone_b.terminals:`
+        - Check if the heating source is Null: `if terminal_b.heating_source == Null: has_terminal_heating = FALSE`
+        - Else if, check if the heating source equals None: `elif terminal_b.heating_source == "None": has_terminal_heating = FALSE`
+        - Else if, check if heating source does not equal none but heating capacity is 0: `elif terminal_b.heating_source != "None" and terminal_b.heating_capacity == 0 : has_terminal_heating = FALSE`
+        - Else, it has terminal heating: `Else: has_terminal_heating = TRUE`
+
+    **Rule Assertion:**
+    - Case 1: For each applicable hvac system (hvac_b above) if the zone served does not have dehumidification then pass: `if does_zone_have_dehumification ==  FALSE: outcome == "PASS"`
+    - Case 2: For each applicable hvac system (hvac_b above) if none of the terminal units have heating then pass:`elif has_terminal_heating == FALSE: outcome == "PASS"`  
+    - Case 3: Else (i.e. HVAC system has dehumification and the system serves computer room space): `Else: outcome == "FAIL"`
```
mkarpman, line 44: Case 3 text should be (HVAC system that serves computer room space has dehumidification AND reheat

mkarpman (approved):

docs/section10/10-3.md
supriyagoel: discussed on RDS call, baseline should not have dehumidification + reheat, it can have dehumidification or it can have reheat but not both.

docs/section10/10-3.md
claperle: Modified

mkarpman (approved):

docs/section10/10-3.md
supriyagoel: @mkarpman, makes sense. then I guess we just need to add units as 'unitless' and clarify that this is a %. I'll create a RMD issue for @JasonGlazer

supriyagoel (approved):

PR #24, opened 2021-04-18
Update Rule6-3.md

xingcx:

PR #252, opened 2022-02-07
Rds/cml/function get heating fuels p

claperle:


docs/ruleset_functions/get_heating_fuels_p.md
```diff
@@ -0,0 +1,75 @@
+# get_heating_fuels_p
+
+**Description:** Get a list of the heating fuels used in the P-RMR.
```
mkarpman, line 3: I would make it agnostic to the type of the RMR, so that it works with either U, P or B-RMR.

docs/ruleset_functions/get_heating_fuels_p.md
```diff
@@ -0,0 +1,75 @@
...
+Set boolean variables to FALSE. Default is fuel is not used for space heating in the P_RMR. These boolean variable will be set to true as this function goes through the heating sources in the P_RMR and determines if each specific fuel has been modeled.
+- Set boolean variable to FALSE: `fuels_propane_p = FALSE`
+- Set boolean variable to FALSE: `fuels_natural_gas_p = FALSE`
+- Set boolean variable to FALSE: `fuels_electricity_p = FALSE`
+- Set boolean variable to FALSE: `fuels_oil_p = FALSE`
+- Set boolean variable to FALSE: `fuels_other_p = FALSE`
+
+Loop through boilers to determine the fuels used by the boilers in the P_RMR. Boolean variables are set to TRUE accordingly.
+- For each rulesetModelInstance_p in the P_RMR: `for rulesetModelInstance_p in P_RMR:`
+    - For each boiler_p in rulesetModelInstance_p.boilers: `for boiler_p in rulesetModelInstance_p.boilers:`
```
mkarpman, line 24: Some boilers may be serving process loads. Do we want to include them also? What about fuels used for ServiceWaterHeatingEquipment.heater_fuel_type? If we are only interested in space heating, then #24 would need to be expanded.

docs/ruleset_functions/get_heating_fuels_p.md
```diff
@@ -0,0 +1,75 @@
...
+        - Else, set other boolean variable to true: `Else:fuels_other_p = TRUE`  
+    Loop through all terminal units and check the heat source associated with each. Boolean variables are set to TRUE accordingly.
+    - For each zone_p in each building_segment_p: `for zone_p in building_segment_p.zones:`
+        - For each terminal_p in zone_p: `for terminal_p in zone_p.terminals:`
+            - Reset energy source variable: `energy_source_type_p = ""`
+            - Get terminal heat source: `energy_source_type_p = terminal_p.heat_source`
+            - Check if the energy source is electric, if so then set boolean variable to true: `if energy_source_type_p in ["ELECTRIC"]:fuels_electricity_p = TRUE`
+            - Check else if the energy source is other, if so then set boolean variable to true (else, do nothing): `elif energy_source_type_p in ["OTHER"]:fuels_other_p = TRUE`  
+This section creates a list of the applicable fuel types that provide heating in the proposed design model based on the outcome of the boolean variables above.
```
mkarpman, line 60: Per above, we need to be clear whether the list includes only space heating fuels or all fuels.

docs/ruleset_functions/get_heating_fuels_p.md
supriyagoel: maybe we start with the function which gets all HVAC systems in a building based on all zones in a building and then identify all the heating fuel types for all HVAC systems.

claperle: I think we need to include all boilers whether serving process or SWH because if natural gas issued for anything in the proposed then it is likely available at the building. Or maybe this is not a safe assumption?

docs/ruleset_functions/get_heating_fuels_p.md
claperle: Good point

docs/ruleset_functions/get_heating_fuels_p.md
```diff
@@ -0,0 +1,75 @@
+# get_heating_fuels_p
+
```
claperle, line 2: Function replaced by a function that does not solely apply to proposed and that expanded the checks.

weilixu:
Closed because the function is replaced.

PR #254, opened 2022-02-08
Rds/cml/10 5

claperle:


docs/section10/10-5.md
```diff
@@ -0,0 +1,36 @@
...
+
+**Applicability Checks:**  
+
+1. HVAC with mechanical cooling.  
+
+**Function Call:** None  
+
+**Manual Check:**  
+
+ 1. Check that the mechanical equipment cooling efficiency has been adjusted from actual design conditions to standard rating conditions.
```
mkarpman, line 22: My interpretation of this rule is that efficiency inputs may be at either rated or design conditions as long as the inputs correctly reflect what type of efficiency is entered. If so, I think it becomes the type of check where we are verifying that user model reflects design documents. I think that this and other similar types of checks are all manual but we do not provide a separate flag for each. Instead, I think the plan is to just produce a report showing what was modeled for the user model, to allow reviewer verify it.

supriyagoel: @mkarpman and @claperle do we even need an RDS for this then?

claperle: This is meant to provide the reviewer with the efficiencies so they can check them. Updated accordingly.

docs/section10/10-5.md
```diff
@@ -0,0 +1,35 @@
+# Airside systems – Rule 10-5  
+**Schema Version:** 0.0.12  
+**Mandatory Rule:** False   
+**Rule ID:** 10-5  
+ 
+**Rule Description:** For HVAC systems designed, mechanical equipment cooling efficiencies shall be adjusted from actual design conditions to the standard rating conditions.
```
mkarpman, line 6: The corresponding rule in appendix G says that it should only be done "... if required by the simulation model". Please add this to the description because I think this is an important detail. I also updated the language in the Rules spreadsheet accordingly.

docs/section10/10-5.md
mkarpman: Let's discuss on the call what to do with this and other similar rules that verify that U-RMR reflects design documents.

supriyagoel: as  discussed on RDS call
- should apply to U-RMR, and then have umbrella check for U RMR = P RMR.
- should treat this in the same way as we treat all rules that apply to user models / verify alignment with design documents. 
- maybe instead  have a list of parameters of user model reported so that reviewer can verify it complies with design documents.
Have 2 options: 
- Option 1: Have one single RDS that says U RMR say it reflects design documents, with the exception of XYZ and has outcome of undetermined
- Option 2: Have one RDS that says U RMR needs to reflect design documents and here are all elements - need to provide aggregate of data (wall area by exposure, WWR by  orientation etc.)
-

docs/section10/10-5.md
claperle: Got it. Done.

docs/section10/10-5.md
claperle: ARCHIVED THE RULE

docs/section10/10-5.md
```diff
@@ -0,0 +1,35 @@
+# Airside systems – Rule 10-5
```
claperle, line 1: ARCHIVED per discussion regarding rules that compared U_RMR to design documents and then compare U_RMR to P_RMR. This rule has been moved to the General Rules Workbook to be addressed by a master rule.

weilixu:
Closed because:
ARCHIVED per discussion regarding rules that compared U_RMR to design documents and then compare U_RMR to P_RMR. This rule has been moved to the General Rules Workbook to be addressed by a master rule.

PR #255, opened 2022-02-08
Rds/cml/10 6

claperle:


docs/section10/10-6.md
```diff
@@ -0,0 +1,35 @@
...
+**Applicability Checks:** 
+
+1. HVAC with mechanical cooling.
+
+**Function Call:** None
+
+**Manual Check:** 
+
+ 1. Check that the mechanical equipment cooling efficiency has been adjusted to remove the supply fan energy from the efficiency rating. 
+
```
mkarpman, line 22: I don't think this RDS, as it stands now, is doing much. I posted RMR issues about adding COPnfcooling and COPnfheating to the schema. Then, this RDS could check that these data elements are included for each packaged HVAC system in the proposed design. We could perhaps also include a check to confirm that COPnfcooling is not just a unit convertion for EER.

supriyagoel: @mkarpman i do not know how this would work. E+ only takes in the COPnfcooling and heating values, there is no way to enter rated EER etc.
Even if we add COPnfcooling/COPnfheating, we would need to assume that values for EER etc might not be provided and this check cannot be performed. How does eQuest handle it?

I had a similar comment for verifying boiler efficiency, it is unlikely that the RMR would include AFUE, Et, Ec for models generated using E+, E+ only takes in Et. Again, not sure how eQuest handles this.

docs/section10/10-6.md
```diff
@@ -0,0 +1,34 @@
...
+## Rule Logic:  
+**Applicability Check 1:** 
+- For each hvac system in the P_RMR: `for hvac_p in P_RMR...HeatingVentilationAirConditioningSystem:`
+    - For each cooling system in the HVAC system: `for cooling_system_p in hvac_p:`
+        - Check that cooling system type does not equal None:: `if cooling_system_p.cooling_system_type != "None":`
+            - Get the full load cooling efficiency: `full_load_efficiency_p = cooling_system_p.full_load_efficiency`
```
mkarpman, line 28: The RDS may just be checking that CoolingSystem.full_load_efficiency_no_fan is entered. Also need to check that HeatingSystem.full_load_efficiency_no_fan is provided. There is currently no such field in the RMD, so I added an issue on github

claperle: @mkarpman  There are separate checks for heating and cooling. See 10-13

claperle: Updated per your suggestion (please check if I got the outcomes correct). Need to update 10-13 as well.

docs/section10/10-6.md
supriyagoel: @claperle and @mkarpman not sure where we landed with MK's first comment. Looks like we're now checking if full load eff and non fan eff is entered but are not doing a unit conversion check. I think a check just on no fan eff is sufficient but want to amke sure we're not missing anything

PR #257, opened 2022-02-08
Rds/cml/10 9

claperle:


docs/section10/10-9.md
```diff
@@ -0,0 +1,53 @@
+# Airside systems – Rule 10-9  
+**Schema Version:** 0.0.12  
+**Mandatory Rule:** True     
+**Rule ID:** 10-9  
+ 
+**Rule Description:** Where multiple HVAC zones or residential spaces are combined into a single thermal block, baseline HVAC System Types 5 or 6 efficiencies shall be based on the cooling equipment capacity of a single floor.  
+
```
mkarpman, line 7: I believe this RDS is related to possible use of multipliers. We discussed it with Jason a while back, and his posibition was that even if multipliers were used in the simulation, the RMR should include explicit listing of all zones, systems, etc. @JasonGlazer please chime in. I believe one of the outcomes of this discussion is the Zone.ID note that suggests that "No multipliers or floor multipliers are used with the Zone data group so each zone should be individually identified." I think question of multipliers was brought by one of the vendors, so I suggest revisiting it at the next RDS call and perhaps also the next schema WG call. 

If that's the case, we should tag any baseline HVAC systems of Type 5-8 that span multiple floors as an error, so the check wouldn't apply

claperle: As discussed with Mike R per G3.1.1#4 "Floors with identical thermal blocks can be grouped for modeling purposes.", which means that a single unit could literally serve multiple floors (which I think is different than a multiplier situation).

docs/section10/10-9.md
```diff
@@ -0,0 +1,51 @@
...
+**Manual Check:** None.  
+ 
+## Rule Logic:  
+**Applicability Check 1:**  
+- Get baseline hvac system type dictionary: `baseline_hvac_sys_type_ids_dict_b = get_baseline_system_types(B_RMR)`
+- Get dictionary with hvac system ids serving more than one floor as keys and the number of floors served as values: `hvac_sys_serve_more_than_one_flr_dict_b = get_hvac_systems_5_6_serving_multiple_floors_b(B_RMR)`
+- Get list of hvac systems of type 5 and 6 that serve more than one floor: `hvac_sys_serve_more_than_one_flr_list_b = hvac_sys_serve_more_than_one_flr_dict_b.keys()`
+- For each hvac_b in hvac_sys_serve_more_than_one_flr_list_b: `for hvac_b in hvac_sys_serve_more_than_one_flr_list_b:`
+    - Get the hvac system type for hvac_b: `hvac_sys_type_b = baseline_hvac_sys_type_ids_dict_b.keys()[list(baseline_hvac_sys_type_ids_dict_b.values()).index(hvac_b.id)]`
+    - Check if HVAC system is type 5, 5b, 6, 6b, if it is then carry on: `if get_baseline_system_types(hvac_b.id) in ["SYS-5", "SYS-5b", "SYS-6", "SYS-6b"]:`
```
mkarpman, line 32: Lines 31 and 32 seems redundant because function only returns those system types

docs/section10/10-9.md
```diff
@@ -0,0 +1,51 @@
...
+        - Get the number of floors that the hvac system serves: `num_floors_hvac_b = hvac_sys_serve_more_than_one_flr_dict_b[hvac_b.id]`
+        Loop through each and get the attributes of the cooling system(s) associated with this HVAC (e.g. efficiency, capacity, etc.)
+        - For each cooling_system_b in hvac_b: `for cooling_system_b in hvac_b.cooling_system:`
+            - Get the total_cool_capacity: `total_cool_capacity_b = cooling_system_b.design_total_cool_capacity`
```
mkarpman, line 36: Should this be design or rated cool capacity? (Same question as for 10-7.)

docs/section10/10-9.md
claperle: Revised

docs/section10/10-9.md
claperle: Revised per discussion on the call

docs/section10/10-9.md
```diff
@@ -0,0 +1,51 @@
...
+
+1. get_baseline_system_types()
+2. get_hvac_systems_5_6_serving_multiple_floors_b()
+
+**Manual Check:** None.  
+ 
+## Rule Logic:  
+**Applicability Check 1:**  
+- Get baseline hvac system type dictionary: `baseline_hvac_sys_type_ids_dict_b = get_baseline_system_types(B_RMR)`
+- Get dictionary with hvac system ids serving more than one floor as keys and the number of floors served as values: `hvac_sys_serve_more_than_one_flr_dict_b = get_hvac_systems_5_6_serving_multiple_floors_b(B_RMR)`
```
supriyagoel, line 28: just a small comment here. This rule only applies to System 5 or 6, so after line 27, if baseline system type is not equal to 5 or 6, then we should have a not applicable outcome and not do the rest.
same for line 28, if the baseline system is 5 or 6 but it is not serving multiple floors, then the outcome should be NA

claperle: @supriyagoel , I modified it accordingly but based on previous conversations we are not explicitly creating NA outcomes.

mkarpman (approved):

docs/section10/10-9.md
```diff
@@ -0,0 +1,52 @@
...
+        - Get the number of floors that the hvac system serves: `num_floors_hvac_b = hvac_sys_serve_more_than_one_flr_dict_b[hvac_b.id]`
+        Loop through each and get the attributes of the cooling system(s) associated with this HVAC (e.g. efficiency, capacity, etc.)
+        - For each cooling_system_b in hvac_b: `for cooling_system_b in hvac_b.cooling_system:`
+            - Check if rated cooling capacity does not equal Null: `if cooling_system_b.rated_total_cool_capacity <> Null:`
+                - Set cooling capacity equal to the rated cooling capacity: `total_cool_capacity_b = cooling_system_b.rated_total_cool_capacity`
+            - Else, rated equals Null then set the cooling capacity equal to the design cooling capacity:`total_cool_capacity_b = cooling_system_b.design_total_cool_capacity`
+            - Get the full load no fan cooling efficiency: `full_load_efficiency_b = cooling_system_b.full_load_efficiency_no_fan`
+            - Determine the capacity (Btuh) that should be used to determine the modeled baseline COPnfcooling efficiency based upon the number of floors that the hvac system serves: `cool_cap_hvac_b = total_cool_capacity_b/num_floors_hvac_b`
```
mkarpman, line 39: Looks good, but please update based on the new RMD efficiency nomenclature (same as for 10-7).

mkarpman (approved):

docs/section10/10-9.md
```diff
@@ -0,0 +1,57 @@
...
+        - Get the number of floors that the hvac system serves: `num_floors_hvac_b = hvac_sys_serve_more_than_one_flr_dict_b[hvac_b.id]`
+        - Reset the efficiency found variable to false, if efficiency is not reported in the correct units for defining the baseline then the outcome of this RDS is fail: `was_efficiency_found = FALSE`
+        Loop through each and get the attributes of the cooling system(s) associated with this HVAC (e.g. efficiency, capacity, etc.)
+        - For each cooling_system_b in hvac_b: `for cooling_system_b in hvac_b.cooling_system:`
+            - Check if rated cooling capacity does not equal Null: `if cooling_system_b.rated_total_cool_capacity <> Null:`
+                - Set cooling capacity equal to the rated cooling capacity: `total_cool_capacity_b = cooling_system_b.rated_total_cool_capacity`
+            - Else, rated equals Null then set the cooling capacity equal to the design cooling capacity:`total_cool_capacity_b = cooling_system_b.design_total_cool_capacity`
```
mkarpman, line 38: Same comment as for 10-7. If neither capacity is entered but efficiency meets requirements for the most stringent bracked, suggest to PASS

mkarpman (approved):

supriyagoel (approved):

docs/section10/10-9.md
```diff
@@ -0,0 +1,65 @@
...
+**Evaluation Context:** Evaluate each HeatingVentilationAirAconditioningSystem   
+
+**Applicability Checks:**  
+
+1. Applicable to baseline HVAC system types 5 or 6 that include more than one floor.  
+
+**Function Call:**  
+
+1. get_baseline_system_types()
+2. get_hvac_systems_5_6_serving_multiple_floors_b()
```
mkarpman, line 21: please add compare_standard_val function here along with its summary description

docs/section10/10-9.md
```diff
@@ -0,0 +1,65 @@
...
+            - Check else if capacity is <760,000 Btuh and greater or equal to 240,000 Btuh: `elif cool_cap_hvac_b < 760000 AND cool_cap_hvac_b >= 240000 : size_category_b = 240_760Btuh`
+            - Else capacity is >=760,000 Btuh: `Else: size_category_b = 760Btuh`
+            - Set the COPnf_full_load_expected_b variable using data_lookup: `COPnf_full_load_expected_b = data_lookup(table_G3_5_1,size_category_b)`  
+            **Rule Assertion:**
+            - Case 1: For each applicable hvac system, outcome is pass if the modeled efficiency is as expected (compare_standard_val func shall return True or False, "greater than" means that if the modeled_efficiency_b is greater than or equal to the COPnf_full_load_expected_b it would be accepted if Permit_More_Stringent_Value equals TRUE ) and was_efficiency_found equals True and the capacity was defined: `if compare_standard_val(Permit_More_Stringent_Value, COPnf_full_load_expected_b, modeled_efficiency_b, "greater than") == TRUE and was_efficiency_found == TRUE and was_capacity_not_defined == FALSE: PASS`
```
mkarpman, line 59: It seems that we don't need was_efficiency_found==TRUE because if this is false than the compare-standard-val will not return TRUE.

docs/section10/10-9.md
```diff
@@ -0,0 +1,65 @@
...
+ 
+## Rule Logic:  
+**Applicability Check 1:**  
+- Get baseline hvac system type dictionary: `baseline_hvac_sys_type_ids_dict_b = get_baseline_system_types(B_RMR)`
+- Get dictionary with hvac system ids serving more than one floor as keys and the number of floors served as values: `hvac_sys_serve_more_than_one_flr_dict_b = get_hvac_systems_5_6_serving_multiple_floors_b(B_RMR)`
+- Get list of hvac systems of type 5 and 6 that serve more than one floor: `hvac_sys_serve_more_than_one_flr_list_b = hvac_sys_serve_more_than_one_flr_dict_b.keys()`
+- For each hvac_b in the B_RMR: `for hvac_b in the B_RMR...HeatingVentilationAirConditioningSystem:`     
+    - Check if hvac_b is in hvac_sys_serve_more_than_one_flr_list_b (outcome is Not Applicable for all HVAC systems not in this list): `if hvac_b in hvac_sys_serve_more_than_one_flr_list_b:`  
+        - Get the number of floors that the hvac system serves: `num_floors_hvac_b = hvac_sys_serve_more_than_one_flr_dict_b[hvac_b.id]`
+        - Reset the efficiency found variable to false, if efficiency is not reported in the correct units for defining the baseline then the outcome of this RDS is fail: `was_efficiency_found = FALSE`
```
mkarpman, line 33: I wonder if we need this variable. If it is deleted, then we can also remove line 45 and also take it out of Case 1 assertion.

docs/section10/10-9.md
claperle: It shouldn't because then modeled_efficiency_b will equal Null so I will remove.

PR #262, opened 2022-02-13
Rds/cml/10 8

claperle:


docs/section10/10-8.md
```diff
@@ -0,0 +1,39 @@
+# Airside systems – Rule 10-8  
+**Schema Version:** 0.0.12  
+**Mandatory Rule:** False   
+**Rule ID:** 10-8  
+ 
+**Rule Description:** Where multiple HVAC zones or residential spaces are combined into a single thermal block the cooling efficiencies (for baseline HVAC System Types 3 and 4) shall be based on the  equipment capacity of the thermal block divided by the number of HVAC zones or residential spaces.
```
mkarpman, line 6: I think this check is basically to verify that model reflects design documents. So it should be handled the same way as 10-5. Let's discuss on the call.

claperle: This is about the baseline efficiency; I think this RDS should be absorbed by 10-7 as it is covered there.

docs/section10/10-8.md
```diff
@@ -0,0 +1,39 @@
+# Airside systems – Rule 10-8
```
claperle, line 1: ARCHIVED because this is captured in 10-7

weilixu:
Closed because:
ARCHIVED because this is captured in 10-7

PR #265, opened 2022-02-14
Rds/cml/function get hvac systems 5 6 serving multiple floors b

claperle:


docs/ruleset_functions/get_hvac_systems_5_6_serving_multiple_floors_b.md
```diff
@@ -0,0 +1,40 @@
+# get_hvac_systems_5_6_serving_multiple_floors_b
+
+**Description:** Get a dictionary of the system 5, 5a, 5b, 6, 6a, 6b hvac system IDs that are modeled as serving more than one floor in the baseline design model.  The dictionary consists of the hvac system ids as the key and the number of floors served as the value associated with the key.  
+
```
supriyagoel, line 4: @claperle is system 5 or 6 are serving more than one floor in the baseline, then they would be incorrect, right?

mkarpman: @supriyagoel we agreed that this is allowed by G3.1.1 a4: ".....Floors with identical thermal blocks can be grouped for modeling purposes."

mkarpman (approved):

supriyagoel (approved):

PR #266, opened 2022-02-14
Rds/cml/10 14

claperle:


docs/section10/10-14.md
```diff
@@ -0,0 +1,74 @@
...
+1. Applicable to baseline HVAC system types 2, 3, 3a, 4, and 9.  
+
+**Function Call:**  
+
+1. get_baseline_system_types()  
+
+## Rule Logic:   
+**Applicability Check 1:**  
+- Get baseline hvac system type dictionary: `baseline_hvac_sys_type_ids_dict_b = get_baseline_system_types(B_RMR)`
+- For each hvac system in the B_RMR: `for hvac_b in B_RMR...HeatingVentilationAirConditioningSystem:`
```
supriyagoel, line 25: similar to #371 #375 and #376 need a case with NA outcome.

docs/section10/10-14.md
```diff
@@ -0,0 +1,74 @@
...
+    - Reset boolean variable that indicates if the hvac system is type 3 or 4 (used in rule assertion): `sys_type_3_or_4 = FALSE`
+    - Reset boolean variable that indicates if the hvac system is in the smallest capacity bracket (used in rule assertion): `smallcapBracket = FALSE`
+    - Reset boolean variable that indicates if the hvac system has 1 cooling and 1 heating system (used in rule assertion): `num_cool_heat_system_equal_1 = TRUE`  
+    Check if the hvac system is applicable to this check based on system type  
+    - Get the hvac system type for hvac_b: `hvac_sys_type_b = baseline_hvac_sys_type_ids_dict_b.keys()[list(baseline_hvac_sys_type_ids_dict_b.values()).index(hvac_b.id)]`
+    - Check if HVAC system is type 2, 3, 3a, 4, and 9: `if hvac_sys_type_b in ["SYS-2", "SYS-3", "SYS-3a", "SYS-4","SYS-9"]:`  
+        Get the attributes of the hvac system relvant to this check   
+        - Get number of heating systems, there should only be one heating system - if there is more than one then the outcome of this RDS is UNDETERMINED: `num_heating_sys = len(hvac_b.heating_system)`
```
supriyagoel, line 33: @claperle not sure I understand why there can only one heating system

supriyagoel: also, should add a case outcome of undetermined based on this logic.

docs/section10/10-14.md
supriyagoel: as discussed during the RDS  call, the RCT applies the applicability check before the 'rule outcome' step and the NA outcome is not required in the RDS for that check to happen in the RCT

docs/section10/10-14.md
```diff
@@ -0,0 +1,189 @@
...
+**Evaluation Context:** Evaluate each HeatingVentilationAirAconditioningSystem   
+
+**Applicability Checks:**  
+
+1. Applicable to baseline HVAC system types 2, 3, 3a, 4, and 9.  
+
+**Function Call:**  
+
+1. get_baseline_system_types()  
+2. get_hvac_zone_list_w_area()
```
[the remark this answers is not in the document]
claperle: Done

docs/section10/10-14.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Rule Logic:   
+**Applicability Check 1:**  
+- Get baseline hvac system type dictionary: `baseline_hvac_sys_type_ids_dict_b = get_baseline_system_types(B_RMR)`
+- For each hvac system in the B_RMR: `for hvac_b in B_RMR...HeatingVentilationAirConditioningSystem:`  
+    Check if the hvac system is applicable to this check based on system type  
+    - Get the hvac system type for hvac_b: `hvac_sys_type_b = baseline_hvac_sys_type_ids_dict_b.keys()[list(baseline_hvac_sys_type_ids_dict_b.values()).index(hvac_b.id)]`
+    - Check if HVAC system is type 2, 3, 3a, 4, and 9 (else then the check is not applicable): `if hvac_sys_type_b in ["SYS-2", "SYS-3", "SYS-3a", "SYS-4","SYS-9"]:`  
+        - Reset the efficiency_matches_acceptable_value boolean variable to FALSE: `efficiency_matches_acceptable_value = FALSE`
+        - Reset the efficiency found variable to false, if efficiency is not reported in the correct units for defining the baseline then the outcome of this RDS is fail: `was_efficiency_found = FALSE`
```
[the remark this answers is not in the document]
claperle: I can extract it. Its used throughout the code but I agree not needed.

docs/section10/10-14.md
```diff
@@ -0,0 +1,189 @@
...
+                        - Set the efficiency found boolean variable to TRUE: `was_efficiency_found = TRUE`
+                    - Else, do nothing and move on to the next metric: `Else:`   
+                - Check if was_efficiency_found equals TRUE: `if was_efficiency_found == TRUE:`  
+                    - Set expected_baseline_eff_b equal to 78% AFUE: `expected_baseline_eff_b = 0.78`
+                    - Check if the modeled efficiency meets the expected baseline efficiency by using the RCT compare_standard_val function, Permit_More_Stringent_Value is defined globally and is TRUE OR FALSE: `efficiency_matches_acceptable_value = compare_standard_val(Permit_More_Stringent_Value, expected_baseline_eff_b, modeled_efficiency_b, "greater than")`  func shall return True or False, "greater than" means that if the modeled_efficiency_b is greater than or equal to the expected_baseline_eff_b it would be accepted if Permit_More_Stringent_Value equals TRUE                  
+                - Reset was_efficiency_found equal to FALSE: `was_efficiency_found = FALSE`
```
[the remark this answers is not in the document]
claperle: @mkarpman It can be one or the other but if both are included they should both be correct based on this logic (I think).

docs/section10/10-14.md
```diff
@@ -0,0 +1,189 @@
...
+                            - Set the efficiency found boolean variable to TRUE: `was_efficiency_found = TRUE`
+                        - Else, do nothing and move on to the next metric: `Else:`   
+                    - Check if was_efficiency_found equals TRUE: `if was_efficiency_found == TRUE:`  
+                        - Check if the modeled efficiency meets the expected baseline efficiency by using the RCT compare_standard_val function, Permit_More_Stringent_Value is defined globally and is TRUE OR FALSE: `efficiency_matches_acceptable_value = compare_standard_val(Permit_More_Stringent_Value, expected_baseline_eff_b, modeled_efficiency_b, "greater than")`  func shall return True or False, "greater than" means that if the modeled_efficiency_b is greater than or equal to the expected_baseline_eff_b it would be accepted if Permit_More_Stringent_Value equals TRUE                   
+                    - Check if efficiency_matches_acceptable_value equals False, if it equals True then skip below: `if efficiency_matches_acceptable_value == False:`  
+                        - Reset was_efficiency_found equal to FALSE: `was_efficiency_found = FALSE`
```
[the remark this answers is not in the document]
claperle: Discuss on call.

docs/section10/10-14.md
```diff
@@ -0,0 +1,189 @@
...
+            - Check if cooling capacity is >=65,000 Btuh and <135000 Btuh OR the zone aggregation factor was defined and the capacity is >=65,000 Btuh and <135000 Btuh OR the capacity was not defined (this is one of the most stringent) AND the efficiency matches acceptable value is equal to FALSE: `if [(zone_list_b[0].aggregation_factor != Null AND total_heating_capacity_b/zone_list_b[0].aggregation_factor >= 65000 AND total_heating_capacity_b/zone_list_b[0].aggregation_factor < 135000) OR (total_cooling_capacity_b >= 65000 AND total_cooling_capacity_b < 135000) OR was_cool_capacity_not_defined == FALSE OR (zone_list_b[0].aggregation_factor == Null AND total_heating_capacity_b >= 135000)] AND efficiency_matches_acceptable_value == FALSE:`
```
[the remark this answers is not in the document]
claperle: Agreed with respect to the was_cool_capacity_not_defined variable.

The reason for efficiency_matches_acceptable_value == FALSE is because if the previous if statement was triggered and the efficiency was found to meet an acceptable value then there is no reason to run this next set of logic.

docs/section10/10-14.md
```diff
@@ -0,0 +1,189 @@
...
+        **Rule Assertion:**  
+        - Case 1: if the modeled efficiency matches the expected baseline efficiency based on the size category or the modeled efficiency matches the most conservative efficiency across its size category and below and the efficiency was defined in the correct units then pass: `if efficiency_matches_acceptable_value == TRUE AND was_capacity_not_defined == FALSE: PASS`    
+        - Case 2: Elif capacity was not defined but the modeled efficiency meets the highest (most conservative) efficiency in the appropriate table based on system type then pass: `elif was_capacity_not_defined == TRUE AND efficiency_matches_acceptable_value == TRUE: PASS`
```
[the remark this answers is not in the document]
claperle: Yes these can be merged. I only separated them so we could understand the scenarios better.

docs/section10/10-14.md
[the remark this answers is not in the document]
claperle: From call if both are defined and one is correct then accept

docs/section10/10-14.md
```diff
@@ -0,0 +1,189 @@
...
+                    - Check if the modeled efficiency meets the expected baseline efficiency by using the RCT compare_standard_val function, Permit_More_Stringent_Value is defined globally and is TRUE OR FALSE: `efficiency_matches_acceptable_value = compare_standard_val(Permit_More_Stringent_Value, expected_baseline_eff_b, modeled_efficiency_b, "greater than")`  func shall return True or False, "greater than" means that if the modeled_efficiency_b is greater than or equal to the expected_baseline_eff_b it would be accepted if Permit_More_Stringent_Value equals TRUE                   
+            - Else, capacity is greater than or equal to 225,000 Btuh: `Else:`
+                - Loop through each of the efficiency_metric_types associated with the heating system until COMBUSTION_EFFICIENCY is found. Get the associated value of the efficiency from the list of efficiency_metric_values associated with the heating_system_b : `For x in range(len(heating_system_b.efficiency_metric_types)):`
+                    - Check if the efficiency metric equals THERMAL_EFFICIENCY: `if heating_system_b.efficiency_metric_types[x] == "COMBUSTION_EFFICIENCY":`
```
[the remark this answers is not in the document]
claperle: Send SG note about this: 78% per PNNL PRM RM Et whether or not Permit_More_Stringent_Value=True or FALSE

docs/section10/10-14.md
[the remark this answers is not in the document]
claperle: Implemented the following: . For furnaces <225000 Btuh where the baseline efficiency can be modeled as 78% AFUE or 80% Et, if both are defined under HeatingSystem since efficiency can be defined in multiple units, we pass it if the project meets at least one of the 2 values in App G.

docs/section10/10-14.md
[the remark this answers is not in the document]
claperle: Implemented but do we want to implement this for any other units. For example, in eQuest I do not believe you define COP at 47 and 17 so all eQuest projects would fail unless the vendors transcribed/mapped the EIR to those entries based ont he performance curves.

docs/section10/10-14.md
```diff
@@ -0,0 +1,177 @@
...
+        - Else, system type is 4:`Else:` 
+            - Get dictionary list of baseline zones and the associated HVAC systems: `hvac_zone_list_w_area_dict_b = get_hvac_zone_list_w_area (B_RMR)`   
+            - Get list of zones that the HVAC system serves (this is to get the zone associated with sys 3 or 4): `zone_list_b = hvac_zone_list_w_area_dict_b[hvac_b.id]["ZONE_LIST"]`  
+            Get the cooling capacity (needed for determining heating efficiency for system 4)
+            - Check if rated cooling capacity does not equal null, if it does not equal null then set the cooling capacity equal to this data element's value: `if cooling_system_b.rated_total_cool_capacity != Null:`
+                - Set cooling capacity equal to the rated cooling capacity: `total_cool_capacity_b = cooling_system_b.rated_total_cool_capacity`
+            - Else if, check that design cooling capacity does not equal null:`Elif cooling_system_b.design_total_cool_capacity != Null:`  
+                - Set cooling capacity equal to the design cooling capacity: `total_cool_capacity_b = cooling_system_b.design_total_cool_capacity`
```
claperle, line 115: We allow either design or rated. Are we still okay with this now that we have adopted a stricter approach generally speaking?

docs/section10/10-14.md
```diff
@@ -0,0 +1,177 @@
...
+                            - Set the modeled efficiency value: `modeled_efficiency_b = heating_system_b.efficiency_metric_values[x]`  
+                        - Else, do nothing and move on to the next metric: `Else:`   
+                    - Check that modeled_efficiency_b does not equal Null: `if modeled_efficiency_b != Null:`
+                        - Set expected_baseline_eff_b equal to 2.1 COPnf17: `expected_baseline_eff_b = 2.1`
+                        - Check if the modeled efficiency meets the expected baseline efficiency: `efficiency_matches_acceptable_value = expected_baseline_eff_b == modeled_efficiency_b`                         
+        **Rule Assertion:**  
+        - Case 1: if the modeled efficiency matches the expected baseline efficiency based on the size category or the modeled efficiency matches the most conservative efficiency across its size category and below and the efficiency was defined in the correct units then pass or capacity was not defined but the modeled efficiency meets the highest (most conservative) efficiency in the appropriate table based on system type then pass: `if efficiency_matches_acceptable_value == TRUE: PASS`
```
claperle, line 172: Are we okay with the approach where if we do not have all of the needed information we let it pass if it modeled the most stringent efficiency? I think we should when the aggregation factor is not defined because I think this will rarely be defined in simulation tools.

docs/section10/10-14.md
```diff
@@ -0,0 +1,177 @@
...
+2. get_hvac_zone_list_w_area()  
+3. baseline_system_type_compare()  
+
+## Rule Logic:   
+**Applicability Check 1:**  
+- Get baseline hvac system type dictionary: `baseline_hvac_sys_type_ids_dict_b = get_baseline_system_types(B_RMR)`
+- Create a list of the target system types: `target_system_types = [HVAC_SYS.SYS_2, HVAC_SYS.SYS_3, HVAC_SYS.SYS_3a, HVAC_SYS.SYS_4, HVAC_SYS.SYS_9]`   
+- For each hvac system in the B_RMR: `for hvac_b in B_RMR...HeatingVentilationAirConditioningSystem:`  
+    Check if the hvac system is applicable to this check based on system type  
+    - Get the hvac system type for hvac_b: `hvac_sys_type_b = baseline_hvac_sys_type_ids_dict_b.keys()[list(baseline_hvac_sys_type_ids_dict_b.values()).index(hvac_b.id)]`
```
JacksonJ-KC, line 30: `hvac_sys_type_b = next((sys_type for sys_type, hvac_id_list in baseline_hvac_sys_type_ids_dict_b.items() if hvac_b.id in hvac_id_list), None)`

docs/section10/10-14.md
```diff
@@ -0,0 +1,177 @@
...
+**Function Call:**  
+
+1. get_baseline_system_types()  
+2. get_hvac_zone_list_w_area()  
+3. baseline_system_type_compare()  
+
+## Rule Logic:   
+**Applicability Check 1:**  
+- Get baseline hvac system type dictionary: `baseline_hvac_sys_type_ids_dict_b = get_baseline_system_types(B_RMR)`
+- Create a list of the target system types: `target_system_types = [HVAC_SYS.SYS_2, HVAC_SYS.SYS_3, HVAC_SYS.SYS_3a, HVAC_SYS.SYS_4, HVAC_SYS.SYS_9]`
```
JacksonJ-KC, line 27: HVAC_SYS.SYS_3a is called out separately from  HVAC_SYS.SYS_3, does this mean we do not want to consider the other variations b or c for SYS_3?

If we only want 3A to be applicable, then I believe HVAC_SYS.SYS_3 needs to be removed from the list because baseline_system_type_compare() will accept all variations of SYS_3.

docs/section10/10-14.md
```diff
@@ -0,0 +1,177 @@
...
+1. Applicable to baseline HVAC system types 2, 3, 3a, 4, and 9.  
+
+**Function Call:**  
+
+1. get_baseline_system_types()  
+2. get_hvac_zone_list_w_area()  
+3. baseline_system_type_compare()  
+
+## Rule Logic:   
+**Applicability Check 1:**
```
JacksonJ-KC, line 25: The RCT will likely have all of the expected efficiency values stored separately in the path `RCT\rct229\rulesets\ashrae9012019\data` 
Should we structure the RMD similarly so that all expected efficiency values are organized together? There are more changes we could make to the RMD to have it be more concise and less repetitive but I'm not sure it's worth improving the RMD since the RCT dev team will likely write the logic differently anyway.

Example chiller efficiency data is stored as shown below.
{
  "prm_water_chiller_min_efficiency": [
    {
      "equipment_type": "displacement",
      "inclusive_minimum_capacity_tons": 0.0,
      "exclusive_maximum_capacity_tons": 150.0,
      "minimum_full_load_efficiency_kw_per_ton": 0.790,
      "minimum_integrated_part_load_kw_per_ton": 0.676
    },
    {
      "equipment_type": "displacement",
      "inclusive_minimum_capacity_tons": 150.0,
      "exclusive_maximum_capacity_tons": 300.0,
      "minimum_full_load_efficiency_kw_per_ton": 0.718,
      "minimum_integrated_part_load_kw_per_ton": 0.629
    },
    {
      "equipment_type": "displacement",
      "inclusive_minimum_capacity_tons": 300.0,
      "exclusive_maximum_capacity_tons": 9999.99,
      "minimum_full_load_efficiency_kw_per_ton": 0.639,
      "minimum_integrated_part_load_kw_per_ton": 0.572
    },

docs/section10/10-14.md
JacksonJ-KC: I think in this case the logic should allow any of the efficiency values from system capacity ranges equal to or less than the defined capacity. The aggregated units don't necessarily need to be within the smallest capacity range.

docs/section10/10-14.md
JacksonJ-KC: I think it is good that the logic prioritizes rated capacity, but uses design capacity if the rated capacity does not exist.

docs/section10/10-14.md
claperle: Smallest is most conservative that is why we selected it. Otherwise we could be passing for a more lenient scenario.

docs/section10/10-14.md
claperle: Agree, both should be fail.

claperle: It is not clear what you are suggesting? That I rewrite it?

docs/section10/10-14.md
JacksonJ-KC: Disregard my comment I agree with the logic that if aggregation factor is not specified then the efficiency needs to match the highest efficiency available in the data tables for that system type.

docs/section10/10-14.md
```diff
@@ -0,0 +1,167 @@
...
+                            - Set the modeled efficiency value: `modeled_efficiency_b = heating_system_b.efficiency_metric_values[x]`  
+                        - Else, do nothing and move on to the next metric: `Else:`   
+                    - Check that modeled_efficiency_b does not equal Null: `if modeled_efficiency_b != Null:`    
+                        - Check if the modeled efficiency meets the expected baseline efficiency : `efficiency_matches_acceptable_value = expected_baseline_eff_b == modeled_efficiency_b`                
+                    - Check if efficiency_matches_acceptable_value equals False, if it equals True then skip below: `if efficiency_matches_acceptable_value == False:`  
+                        - Reset modeled_efficiency_b to equal Null: `modeled_efficiency_b = Null`  
+                        - Loop through each of the efficiency_metric_types associated with the heating system until FULL_LOAD_EFFICIENCY_NO_FAN is found. Get the associated value of the efficiency from the list of efficiency_metric_values associated with the heating_system_b : `For x in range(len(heating_system_b.efficiency_metric_types)):`
```
JacksonJ-KC, line 115: `FULL_LOAD_EFFICIENCY_NO_FAN` should be `FULL_LOAD_COEFFICIENT_OF_PERFORMANCE_NO_FAN`

docs/section10/10-14.md
```diff
@@ -0,0 +1,167 @@
...
+        - Reset the cooling capacity not defined variable to false: `was_cool_capacity_not_defined = FALSE`
+        - Reset the heating capacity not defined variable to false: `was_heat_capacity_not_defined = FALSE`
+        Get the attributes of the hvac system relevant to this check   
+        - Assign the heating system to an object (there should only be 1 heating system): `heating_system_b = hvac_b.heating_system` 
+        - Assign the cooling system to an object (there should only be 1 cooling system): `cooling_system_b = hvac_b.cooling_system` 
+
+        Determine the expected modeled efficiency based on the system type and capacity and compare to modeled value:
+        - SYSTEM TYPE 2  
+        - Check if the system is system type 2: `if hvac_sys_type_b = "SYS-2":`
+            - Loop through each of the efficiency_metric_types associated with the heating system until FULL_LOAD_EFFICIENCY_NO_FAN is found. Get the associated value of the efficiency from the list of efficiency_metric_values associated with the heating_system_b : `For x in range(len(heating_system_b.efficiency_metric_types)):`
```
JacksonJ-KC, line 42: `FULL_LOAD_EFFICIENCY_NO_FAN` should be `FULL_LOAD_COEFFICIENT_OF_PERFORMANCE_NO_FAN`

docs/section10/10-14.md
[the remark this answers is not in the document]
claperle: Going to implement a special fail for if a project gets the high temperature efficiency correct and the low temperature efficiency is undefined.

docs/section10/10-14.md
claperle: Christina and Jackson talked and are on the same page with the structure. Leave as is.

docs/section10/10-14.md
claperle: Change to HEAT_PUMP_COEFFICIENT_OF_PERFORMANCE_HIGH_TEMPERATURE_NO_FAN since in AHRI 310 the rating condition is 47F

docs/section10/10-14.md
```diff
@@ -0,0 +1,177 @@
...
+                - Set the expected baseline efficiency for <65000 (if the modeled part OR full load efficiency values equals 3.4 then it passes; this is captured in the subsequent logic): `expected_baseline_eff_b = 3.4`  
+                    - Loop through each of the efficiency_metric_types associated with the heating system until PART_LOAD_EFFICIENCY_NO_FAN  is found. Get the associated value of the efficiency from the list of efficiency_metric_values associated with the heating_system_b. : `For x in range(len(heating_system_b.efficiency_metric_types)):`
+                        - Check if the efficiency metric equals HEAT_PUMP_COEFFICIENT_OF_PERFORMANCE_HIGH_TEMPERATURE_NO_FAN : `if heating_system_b.efficiency_metric_types[x] == "HEAT_PUMP_COEFFICIENT_OF_PERFORMANCE_HIGH_TEMPERATURE_NO_FAN":` 
+                            - Set the modeled efficiency value: `modeled_efficiency_b = heating_system_b.efficiency_metric_values[x]`
```
claperle, line 111: Would like to discuss this. Should we assume the rating condition for the <65,000 Btuh category is 47F db? It is not clear since the resting standard is AHRI 210/240 which performances seasonal tests. I assume the intent of the 90.1 committee is that the rating condition be 47F but I do not see anything definitive.

JacksonJ-KC (approved):

docs/section10/10-14.md
```diff
@@ -0,0 +1,189 @@
...
+                - Check if zone aggregation_factor equals Null and efficiency_matches_acceptable_value does not equal TRUE, Else go to Rule Assertion: `if zone_list_b[0].aggregation_factor == Null AND efficiency_matches_acceptable_value == FALSE:` 
+                    - Reset was_efficiency_found equal to FALSE: `was_efficiency_found = FALSE`
+                    - Loop through each of the efficiency_metric_types associated with the heating system until AFUE is found. Get the associated value of the efficiency from the list of efficiency_metric_values associated with the heating_system_b : `For x in range(len(heating_system_b.efficiency_metric_types)):`
```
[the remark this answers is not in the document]
claperle: Remove any leniency

docs/section10/10-14.md
claperle: Yes, per group use 47F.

docs/section10/10-14.md
```diff
@@ -0,0 +1,142 @@
...
+
+        Determine the expected modeled efficiency based on the system type and capacity and compare to modeled value:
+        - SYSTEM TYPE 2  
+        - Check if the system is system type 2: `if baseline_system_type_compare(hvac_sys_type_b, HVAC_SYS.SYS_2, false):`    
+            - Loop through each of the efficiency_metric_types associated with the heating system until HEAT_PUMP_COEFFICIENT_OF_PERFORMANCE_HIGH_TEMPERATURE_NO_FAN is found. Get the associated value of the efficiency from the list of efficiency_metric_values associated with the heating_system_b : `For x in range(len(heating_system_b.efficiency_metric_types)):`
+                    - Check if the efficiency metric equals HEAT_PUMP_COEFFICIENT_OF_PERFORMANCE_HIGH_TEMPERATURE_NO_FAN (rating condition in 310 if 47F): `if heating_system_b.efficiency_metric_types[x] == "HEAT_PUMP_COEFFICIENT_OF_PERFORMANCE_HIGH_TEMPERATURE_NO_FAN":`  
+                        - Set the modeled efficiency value: `modeled_efficiency_b = heating_system_b.efficiency_metric_values[x]`
+                    - Else, do nothing and move on to the next metric: `Else:`
```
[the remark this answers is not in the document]
claperle: Done!

docs/section10/10-14.md
```diff
@@ -0,0 +1,142 @@
...
+                - Set heating capacity equal to the rated heating capacity: `total_heat_capacity_b = heating_system_b.rated_capacity`
+            - Else if, check that design heating capacity does not equal null:`Elif heating_system_b.design_capacity != Null:`  
+                - Set heating capacity equal to the design heating capacity: `total_heat_capacity_b = heating_system_b.design_capacity`
+            - Else, set the was_heat_capacity_not_defined boolean variable equal to TRUE: `Else: was_heat_capacity_not_defined = TRUE`
+            - Check if capacity is <225,000 Btuh (below checks if the modeled value is equal to the most stringent size category which is <225,000 for when capacity is not defined) or (if the aggregation factor was defined (i.e., does not equal Null) and the resultant capacity is less than 225000) or if the aggregation factor = Null : `if total_heat_capacity_b < 225000 OR was_heat_capacity_not_defined boolean variable == TRUE OR (zone_list_b[0].aggregation_factor != Null AND total_heating_capacity_b/zone_list_b[0].aggregation_factor < 225000) OR zone_list_b[0].aggregation_factor == Null:`
```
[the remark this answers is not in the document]
claperle: This made me realized that we should be producing an UNDETERMINED outcome when the aggregation factor is not defined AND when it is needed (sometimes it not needed to figure out the expected efficiency). I revised the code accordingly. It is not defaulted to one. Thanks for the suggestion but due to moving this along I left the code as is since it accomplishes what we need it to (please check that it does though!).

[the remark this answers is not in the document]
JacksonJ-KC: > just so I'm clear on the intent of how zone.aggregation_factor is used - looks like you are assuming that if it does not exist, it is essentially equal to 1. Is this correct?
> 
> the math / logic in this rule could be simplified by creating a dictionary that mapped a number for zone aggregation factor to a number (either 1 for null entries, or the number, if it exists). This would look like this:
> 
> ```
> for zone in zone_list
>    zone_aggregation_dict = {}
>    if zone.aggregation_factor == Null
>       zone_aggregation_dict[zone] = 1
>    else
>       zone_aggregation_dict[zone] = zone.aggregation_factor
> ```
> 
> & would be done at the start of the rule
> 
> the above line is then simplified to:
> 
> `if (total_heat_capacity_b/zone_aggregation_dict[zone_list_b[0].aggregation_factor]) < 225000 OR was_heat_capacity_not_defined boolean variable == TRUE OR zone_list_b[0].aggregation_factor == Null:`
> 
> similar simplifications could be made throughout the rule

@KarenWGard I had the same idea of defaulting the aggregation factor to 1 in the implementation of this rule!

KarenWGard (approved):

docs/section10/10-14.md
```diff
@@ -0,0 +1,159 @@
...
+                - Set cooling capacity equal to the rated cooling capacity: `total_cool_capacity_b = cooling_system_b.rated_total_cool_capacity`  
+            - Else if, check that design cooling capacity does not equal null:`Elif cooling_system_b.design_total_cool_capacity != Null:`   
+                - Set cooling capacity equal to the design cooling capacity: `total_cool_capacity_b = cooling_system_b.design_total_cool_capacity`  
+            - Else, set the was_capacity_not_defined boolean variable equal to TRUE: `Else: was_cool_capacity_not_defined = TRUE`   
+          
+            - SYSTEM 4 <65000   
+            - Check if cooling capacity is <65,000 Btuh OR the zone aggregation factor was defined and the capacity is less than 65000 Btuh: `if (zone_list_b[0].aggregation_factor != Null AND total_heating_capacity_b/zone_list_b[0].aggregation_factor < 65000) OR total_cooling_capacity_b < 65000 AND was_cool_capacity_not_defined == FALSE:`   
+                - Check if the capacity is greater than or equal to 65000 and zone aggregation not defined: `if total_cooling_capacity_b >= 65000 AND zone_list_b[0].aggregation_factor == Null :`
```
JacksonJ-KC, line 110: @claperle Does this need to be moved into the SYSTEM 4  >= 135,000 Btu/h block instead?

docs/section10/10-14.md
```diff
@@ -0,0 +1,159 @@
...
+                        - Set the modeled efficiency value: `modeled_efficiency_b = heating_system_b.efficiency_metric_values[x]`  
+                        - Break out of loop if efficiency is found: `break`  
+                - Check that modeled_efficiency_b does not equal Null: `if modeled_efficiency_b != Null:`     
+                    - Set expected_baseline_eff_b equal to 78% AFUE: `expected_baseline_eff_b = 0.78`   
+                - Check if the modeled efficiency does not equal Null, if false (i.e., does not equal Null) then go to the Rule Assertion, if it equals Null perform logic below: `if modeled_efficiency_b == Null:`   
+                    - Loop through each of the efficiency_metric_types associated with the heating system until THERMAL_EFFICIENCY is found. Get the associated value of the efficiency from the list of efficiency_metric_values associated with the heating_system_b : `For x in range(len(heating_system_b.efficiency_metric_types)):`  
+                        - Check if the efficiency metric equals THERMAL_EFFICIENCY: `if heating_system_b.efficiency_metric_types[x] == "THERMAL_EFFICIENCY":`
```
JacksonJ-KC, line 85: What if a project reports both `ANNUAL_FUEL_UTILIZATION_EFFICIENCY` AND `THERMAL_EFFICIENCY`, `AFUE=0.77` and `THERMAL_EFFICIENCY=0.81`?

docs/section10/10-14.md
```diff
@@ -0,0 +1,159 @@
...
+            - Check if rated cooling capacity does not equal null, if it does not equal null then set the cooling capacity equal to this data element's value: `if cooling_system_b.rated_total_cool_capacity != Null:`  
+                - Set cooling capacity equal to the rated cooling capacity: `total_cool_capacity_b = cooling_system_b.rated_total_cool_capacity`  
+            - Else if, check that design cooling capacity does not equal null:`Elif cooling_system_b.design_total_cool_capacity != Null:`   
+                - Set cooling capacity equal to the design cooling capacity: `total_cool_capacity_b = cooling_system_b.design_total_cool_capacity`  
+            - Else, set the was_capacity_not_defined boolean variable equal to TRUE: `Else: was_cool_capacity_not_defined = TRUE`   
+          
+            - SYSTEM 4 <65000   
+            - Check if cooling capacity is <65,000 Btuh OR the zone aggregation factor was defined and the capacity is less than 65000 Btuh: `if (zone_list_b[0].aggregation_factor != Null AND total_heating_capacity_b/zone_list_b[0].aggregation_factor < 65000) OR total_cooling_capacity_b < 65000 AND was_cool_capacity_not_defined == FALSE:`
```
JacksonJ-KC, line 109: inconsistent use of total_heating_capacity and total_cooling_capacity in this line

docs/section10/10-14.md
claperle: I don't know. I guess we figured that would not happen. Are you suggesting it should check both and then fail if either one is incorrect?

docs/section10/10-14.md
claperle: OK, changed throughout to cooling and heating instead of cool and heat.

docs/section10/10-14.md
claperle: I modified to make the zone_aggregation_factor_undefined_and_needed always TRUE for this condition. Reading through it I think it is all good across the other conditions as well. No need to make adjustments to those. Let me know if I am missing something.

docs/section10/10-14.md
JacksonJ-KC: Yes I think it should at least be one of these options you mention, instead of the current method of prioritizing AFUE

claperle: Done

PR #267, opened 2022-02-14
update json name while validating schema

vijayvardhan94:
Updating the json name in run_json_schema_validation.py without having to hard code file name.


jugonzal07:
I would rather not add this to the repo. The schema validation should be separate from the JSON generation and mixing the two introduces more coupling between the two than I would like.

I think it's fine to have this locally if it makes your work quicker! But this feels more like something you should set in a stash rather than commit to the repo.

When it comes to excel_to_json and run_json_schema_validation, it's probably best to remove the inputs from them and instead keep them as their own functions long term. Functions could be called from outside the repo and not cause diff issues (I assume that's the main motivation here but I'm not sure).

vijayvardhan94:
Closing this pull request based on @jugonzal07 comments above.

PR #340, opened 2022-03-07
Create Rule22-9.md

### LINKED ISSUE #301: Create Rule22-9.md
Secondary pump min flow setpoint is 25% when rule 8

xingcx:
resolves #301


docs/section22/Rule22-9.md
```diff
@@ -0,0 +1,65 @@
...
+    - Get primary loop: `primary_loop_b = get_component_by_id(loop_b_id)`
+
+      - For each secondary loop served by primary CHW loop: `for secondary_loop_id_b in primary_loop_b.child_loops:`
+
+        - For each secondary pump associated with secondary loop: `for secondary_pump in loop_pump_dictionary[secondary_loop_id_b]:`
+
+          **Rule Assertion - Component:**
+
+          - Case 1: If secondary pump is modeled with a minimum flow of 25% of the design flow rate: `if secondary_pump.minium_flow == 0.25: PASS`
+
```
mkarpman, line 62: I think this check should apply to FuildLoopDesignAndControl.minimum_flow_fraction

docs/section22/Rule22-9.md
```diff
@@ -0,0 +1,65 @@
...
+- **Check 1:** Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`
+
+  - Check if B-RMR is modeled with at least one air-side system that is Type-7, 8, 11.1, 11.2, 12, 13, 7b, 8b, 11b, 12b, 13b, i.e. with air-side system served by chiller(s), continue to the next applicability check: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-7", "SYS-8", "SYS-11.1", "SYS-11.2", "SYS-12", "SYS-13", "SYS-7B", "SYS-8B", "SYS-11B", "SYS-12B", "SYS-13B"]): NEXT_APPLICABILITY_CHECK`
+
+  - Else, rule is not applicable to B-RMR: `else: RULE_NOT_APPLICABLE`
+
+- **Check 2:** Get primary and secondary loops for B-RMR: `primary_secondary_loop_dictionary = get_primary_secondary_loops(B_RMR, "COOLING")`
+
+  - Check if primary and secondary loops are modeled correctly, continue to rule logic: `if primary_secondary_loop_dictionary["PRIMARY"] AND primary_secondary_loop_dictionary["SECONDARY"]: CHECK_RULE_LOGIC`
```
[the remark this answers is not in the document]
mkarpman: I think this RDS should do the following:
1. Use primary_secondary_loop_dictionaryIdentify to identify pairs of primary/secondary loop. (There should be just one such pair, but this is checked by 22-34, so it would flag any mistakes related to this.)
2. For each pair, (a) check that primary loop is served by chillers with capacity over 300 ton; if true then (b) check that secondary loop has min flow = 25%. If true, then pass.

supriyagoel (approved):

docs/section22/Rule22-9.md
```diff
@@ -0,0 +1,65 @@
...
+- For each chiller in B_RMR: `for chiller_b in B_RMR.RulesetModelInstance.chillers:`
+
+  - Get cooling loop that chiller serves: `chw_loop_b_id = chiller_b.cooling_loop`
+
+    - Add chiller rated capacity to primary CHW loop total capacity: `chw_loop_capacity_dict[chw_loop_b_id] += chiller_b.rated_capacity`
+
+- For each primary CHW loop: `for loop_b_id in primary_secondary_loop_dictionary["PRIMARY"]:`
+
+  - Check if CHW primary loop total cooling capacity is 300 tons or more: `if chw_loop_capacity_dict[loop_b_id] >= 300 * CONVERSION(TON_TO_BTUH):`
+
```
supriyagoel, line 54: @xingcx and @weilixu the chiller rated capacity is in W in the RMR. Should the conversion here say  conversion (W_to_TON) and the check be >300 Tons?

xingcx: The 300ton is from 90.1 table, and it is converted to Btuh as  we assumes the chiller rated capacity in IP unit is Btuh.

docs/section22/Rule22-9.md
```diff
@@ -0,0 +1,65 @@
...
+  - Check if B-RMR is modeled with primary-secondary configuration, continue to rule logic: `if LEN(primary_secondary_loop_dictionary) != 0: CHECK_RULE_LOGIC`
+
+  - Else, rule is not applicable to B-RMR: `else: RULE_NOT_APPLICABLE`
+
+## Rule Logic:  
+
+- For each pump in B_RMR: `for pump_b in B_RMR.RulesetModelInstance.pumps:`
+
+  - Save loop served by pump and pump to a dictionary: `loop_pump_dictionary[pump_b.loop_or_piping].append(pump)`
+
```
mkarpman, line 44: Why do we need to go through pumps? Can't we skip it and for each secondary loop returned by get_primary_secondary_loops(B_RMR, "COOLING") check that FluidLoopDesignAndControl.minimum_flow_fraction=25%?

docs/section22/Rule22-9.md
```diff
@@ -0,0 +1,65 @@
...
+- For each chiller in B_RMR: `for chiller_b in B_RMR.RulesetModelInstance.chillers:`
+
```
mkarpman, line 46: Project may have process chillers. So perhaps instead only consider chillers that have chiller.cooling_loop equal to the primary loop ID returned by get_primary_secondary_loops(B_RMR, "COOLING"). That is also not entirely full proof because process chiller may also have primary/secondary arrangement, but I don't think it's common so this method is less likely to misfire I think.

docs/section22/Rule22-9.md
supriyagoel: yup, makes sense to skip pumps. This might be leftover from when we were checking the pump min flow fraction instead of loop min flow fraction.

docs/section22/Rule22-9.md
supriyagoel: I don't think we should assume the IP units for RDS. The RDS is reading the RMR which is in SI units. The RDS should identify the required conversion from the units listed in the schema to the units required for the RDS verification- i.e. the units in Standard 90.1

mkarpman: @supriyagoel I believe we agreed that when RMR files are parsed by RCT, all metric values will be converted to IP. However for capacity, this default conversion would be from W to Btu/hr, so @xingcx included conversion from 300 ton to Btu/hr for apples to apples comparison.

docs/section22/Rule22-9.md
supriyagoel: as discussed during the RDS call on 2022.04.12 - the 'get primary secondary loops function' should be updated to exclude chilled water loops with process loads.

docs/section22/Rule22-9.md
xingcx: Updated.

docs/section22/Rule22-9.md
xingcx: get_primary_secondary_loops() function is now updated. 22-9 updated.

mkarpman (approved):

PR #343, opened 2022-03-07
Create Rule22-28.md

xingcx:
resolves #[309]


docs/section22/Rule22-28.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Rule Logic:
+
+- For each pump in B_RMR: `for pump_b in B_RMR.RulesetModelInstance.pumps:`
+
+  - Save loop served by pump and pump to a dictionary: `loop_pump_dictionary[pump_b.loop_or_piping].append(pump)`
+
+- For each chiller in B_RMR: `for chiller_b in B_RMR.RulesetModelInstance.chillers:`
+
+  - Check if condensing loop is connected to other chiller, raise fail flag: `if chiller_b.condensing_loop in condensing_loop_array: fail_flag = TRUE`
```
mkarpman, line 36: I am confused with this check. Does it suggest that there should be as many CW loops as chillers? I am not sure how this would be modeled.

supriyagoel: Maybe this is verifying that there is only one condenser loop serving a chiller, else it results in fail.  Should discuss.

supriyagoel (approved):
RDS looks good to me but should discuss MK's comment

docs/section22/Rule22-28.md
mkarpman: I am not sure. Seems like it's verifying that there is no more than one chiller on each condenser water loop, but don't we want one condenser water loop per project, just like we want one primary and once secondary CHW loop per project?

xingcx: Updated.

docs/section22/Rule22-28.md
```diff
@@ -0,0 +1,39 @@
...
+  - Else, rule is not applicable to baseline ruleset model instance: `else: RULE_NOT_APPLICABLE`
+
+## Rule Logic:
+
+- For each chiller in baseline ruleset model instance: `for chiller_b in ASHRAE229.RulesetModelInstance['BASELINE_0'].chillers:`
+
+  **Rule Assertion:**
+
+  - Case 1. If chiller is interlocked with condenser water pump: `if chiller_b.is_condenser_water_pump_interlocked: PASS`
+
```
supriyagoel, line 36: small correction required here - should check if if chiller_b.is_condenser_water_pump_interlocked = TRUE, then PASS

supriyagoel (approved):

PR #353, opened 2022-03-10
Rs/wx/envelope test updates

weilixu:


weilixu:
The updates fixes the running tests with test engine on the new rule structure

jugonzal07 (approved):

PR #363, opened 2022-03-18
Rs/wx/area type wwr

weilixu:
Complete the implementation of area type WWR function with test file (0.0.5)


weilixu:
@xingcx I made some comments in the RDS for the area type WWR function, could you take a look and confirm the changes?
If needed, we can discuss in the next RDS review meeting

rct229/ruleset_functions/get_surface_conditioning_category_dict.py
```diff
@@ -1,102 +1,104 @@
 import pandas as pd
 
 from rct229.ruleset_functions.get_zone_conditioning_category_dict import (
-    CONDITIONED_MIXED,
-    CONDITIONED_NON_RESIDENTIAL,
-    CONDITIONED_RESIDENTIAL,
-    SEMI_HEATED,
-    UNCONDITIONED,
-    UNENCLOSED,
+    ZoneConditioningCategory,
```
charliepnnl, line 4: Do you want to abbreviate this as you did for SurfaceConditioningCategory?
```
ZoneConditioningCategory as ZCC
```

rct229/ruleset_functions/get_surface_conditioning_category_dict.py
```diff
@@ -1,102 +1,104 @@
...
-EXTERIOR_RESIDENTIAL = "EXTERIOR RESIDENTIAL"
-SEMI_EXTERIOR = "SEMI-EXTERIOR"
-UNREGULATED = "UNREGULATED"
+
+
+# Intended for export and internal use
+class SurfaceConditioningCategory:
+    """Enumeration class for zone conditioning categories"""
+
+    EXTERIOR: str = "EXTERIOR"
```
charliepnnl, line 17: I suggest not including EXTERIOR, GROUND, INTERIOR here because these are not surface conditioning categories.  Ideally, we would get these values from the schema enumerations.  In the mean time, they can be constants in this file outside of the SurfaceConditioningCategory class.

charliepnnl (approved):
I have made several comments that I hope you will address before merging. But I am approving now to save time.

rct229/ruleset_functions/get_area_type_window_wall_area.py
```diff
@@ -0,0 +1,77 @@
...
+                scc_dictionary[surface["id"]]
+                in [
+                    SCC.EXTERIOR_RESIDENTIAL,
+                    SCC.EXTERIOR_NON_RESIDENTIAL,
+                    SCC.EXTERIOR_MIXED,
+                    SCC.SEMI_EXTERIOR,
+                ]
+            ):
+                assert_("area" in surface, f"surface:{surface['id']} has no area")
+                window_wall_areas_dictionary[area_type]["TOTAL_WALL_AREA"] += surface[
```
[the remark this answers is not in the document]
weilixu: this is in the RDS, I wonder if we can put lower case? I would also prefer lower case

rct229/ruleset_functions/get_surface_conditioning_category_dict.py
weilixu: To import the enum, will this looks good?
`SurfaceAdjacentTo = schema_enums["SurfaceAdjacentTo"]`
`                if surface_adjacent_to == SurfaceAdjacentTo.INTERIOR.name
                else surface_adjacent_to,`

rct229/ruleset_functions/get_area_type_window_wall_area.py
```diff
@@ -0,0 +1,77 @@
+from rct229.data.schema_enums import schema_enums
+from rct229.ruleset_functions.get_opaque_surface_type import (
+    ABOVE_GRADE_WALL,
+    get_opaque_surface_type,
+)
+from rct229.ruleset_functions.get_surface_conditioning_category_dict import (
+    SurfaceConditioningCategory as SCC,
+)
+from rct229.ruleset_functions.get_surface_conditioning_category_dict import (
```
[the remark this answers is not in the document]
weilixu: Black forced the separate imports.
do it 
`from rct229.ruleset_functions.get_surface_conditioning_category_dict import (
    get_surface_conditioning_category_dict,
    SurfaceConditioningCategory as SCC,
)`
will fail the black format..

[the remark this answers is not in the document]
charliepnnl: Interesting.  I just tried it and it works fine for me.

rct229/ruleset_functions/get_surface_conditioning_category_dict.py
charliepnnl: That works.  At some point, we should try to find a more elegant syntax for handling the enumerations. The `.name` part is just ugly. Probably, instead of using Python's Enumeration type, we should just create a class for each enumeration. Then the syntax could simply be `SurfaceAdjacentTo.INTERIOR` without the `.name`.

I think all we will need to do is to replace line 38 of rct229/data/schema_enums.py with:
```
schema_enums = {
    class_name: type(class_name, (object, ),  enum_dict) 
    for class_name, enum_dict in _enum_dicts.items()}
```
The `type(class_name, (object, ),  enum_dict)` part creates a new type (class) with the given class name, that inherits from `object`, and has class variables given by `enum_dict`.

Let me know if you'd like me to make this change and update wherever schema_enums is used.

charliepnnl (approved):

rct229/ruleset_functions/get_surface_conditioning_category_dict.py
weilixu: Yea, could you update the method? I agree the .name is ugly and the functioin in schma_enums should takes care of the .name. Thanks!

PR #366, opened 2022-03-20
Create Rule22-7.md

xingcx:


docs/section22/Rule22-7.md
```diff
@@ -31,9 +29,10 @@
 ## Rule Logic:  
 
 - Get primary and secondary loops for B-RMR: `primary_secondary_loop_dictionary = get_primary_secondary_loops(B_RMR, "COOLING")`
+
 **Rule Assertion - Component:**
 
-- Case 1: If primary and secondary loops are modeled correctly: `if primary_secondary_loop_dictionary["PRIMARY"] AND primary_secondary_loop_dictionary["SECONDARY"]: PASS`
+- Case 1: If B-RMR is modeled with primary-secondary configuration: `if LEN(primary_secondary_loop_dictionary) != 0: PASS`
```
mkarpman, line 35: I think here we need to check that there is only one primary and only one secondary loop. As it stands now, having multiple primary and secondary loops would pass.

mkarpman (approved):

supriyagoel (approved):

PR #374, opened 2022-03-21
Rds/cml/10 12 b

claperle:


docs/section10/10-12B.md
```diff
@@ -0,0 +1,36 @@
+# Airside systems – Rule 10-12B
```
claperle, line 1: ARCHIVED per discussion regarding rules that compared U_RMR to design documents and then compare U_RMR to P_RMR. This rule has been moved to the General Rules Workbook to be addressed by a master rule.

PR #385, opened 2022-03-24
Added Tests for envelope: 5-2,3,5, and 8

jugonzal07:
Added 5-2,3,5, and 8 in 0.0.12.


weilixu (approved):

charliepnnl (approved):

jugonzal07:
As just an FYI, this does not address the pint unit conversion issue with U-factor-- I will investigate that and create a separate PR.

PR #387, opened 2022-03-28
Fixed issue related to u-factors having incorrect values from TCDs.

jugonzal07:
Fixed issues related to u-factors having incorrect values. It appears the TCD spreadsheet had incorrect values. The pint calculations were being done correctly.

### DIFF
--- a/rct229/ruletest_engine/ruletest_jsons/envelope_tests.json
+++ b/rct229/ruletest_engine/ruletest_jsons/envelope_tests.json
@@ -215,7 +215,7 @@
                                         "id": "Building Segment 1",
                                         "zones": [
                                             {
-                                                "id": "Zone 2",
+                                                "id": "Zone 1",
                                                 "thermostat_cooling_setpoint_schedule": "Cooling Schedule 1",
                                                 "thermostat_heating_setpoint_schedule": "Heating Schedule 1",
                                                 "surfaces": [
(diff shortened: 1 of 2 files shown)


rct229/ruletest_engine/ruletest_jsons/envelope_tests.json
```diff
@@ -390,7 +390,7 @@
                                                         "tilt": 0,
                                                         "construction": {
                                                             "id": "Construction 1",
-                                                            "u_factor": 2.0328185600815645
+                                                            "u_factor": 0.06305330288229763
```
weilixu, line 393: The strict equality test will fail in this case because the U-factor in 90.1 table 3-4 is 0.064 for zone 4 (A,B,C). Could you check again to see if there is any conversion errors or if the starting value is incorrect?

weilixu (changes requested):

rct229/ruletest_engine/ruletest_jsons/envelope_tests.json
weilixu: There are a number of updates required to be addressed for Rule 5-5 and Rule 5-8 test cases.
1. All test cases for rule 5-5 and 5-8 are missing HVAC system
2. All spaces data group is missing floor_area data (required for zone conditioning category function).
3. rule-5-5-f, rule-5-5-g are missing spaces.

jugonzal07: Hi Weili,

I believe I fixed the issues you asked. 

- 5-5f and 5-5g were missing spaces in the TCDs for some reason so I added those back in
- HVAC systems were missing due to a formula issue in the spreadsheet. Fixed this.
- Floor area of 250 ft2 was added for all spaces

weilixu: Awesome, thank you Juan. I will merge this PR.

weilixu (approved):

PR #392, opened 2022-03-29
Rds/cml/function get dict of zones hvac sys serving specific floor

claperle:


rct229/ruleset_functions/get_dict_of_zones_hvac_sys_serving_specific_floor.md
```diff
@@ -0,0 +1,24 @@
...
+**Returns:**
+- **dict_of_zones_hvac_sys_serving_specific_floor**: a dictionary with zone ids as the keys and the associated HVAC system as the values for zones serving a specific floor the applicable RMR i.e. {zone_1.id: [hvac_1.id, hvac_2.id, hvac_3.id], zone_2.id: [hvac_1.id, hvac_2.id, hvac_3.id]}.  
+ 
+**Function Call:** 
+
+1. get_list_hvac_systems_associated_with_zone()  
+
+**Logic:**  
+- For each zone in the RMR (whichever RMR (U, P, or B) that was input to the function): `For zone_x in RMR...Zone:`
+    - If the floor name associated with the zone is equal to floor_name (i.e. the function input floor name) then add it to the dictionary with list of hvac systems: `if zone_x.floor_name = floor_name:`
```
mkarpman, line 18: I wonder if we need to have Floor data object so that matching can happen via unique ID as opposed to a string "Name" field. The problem though is that a floor may not be a model input.

supriyagoel: right, I'm not sure how this would work. RMR has floor_name, which will typically not be a value in energy models and as a string, cannot be used for matching. need to discuss.

claperle: @supriyagoel @mkarpman The model will have a floor number, can it be floor number in the schema and can the schema be changed from string to numeric for this data element? Below is a screenshot of the current schema.
image

supriyagoel: as discussed during the RDS call on 04.05.2022, we will use the floor name (will not add a separate data group for floors to get an ID) and the baseline HVAC RDS will check if floor name = NULL and through out an error.

supriyagoel (approved):

PR #393, opened 2022-03-29
RCT/CH/equality-comparison

charliepnnl:
This provides an equality comparison function `std_equal`. The name is intended to indicate equality according to whatever standard is being used (90.1 for us).

There are several design decisions for which I encourage feedback:

- I've placed this utility in the `ruleset_functions` folder. It feels like a standards-specific function to me.
- I've called the function `std_equal`. I was looking for something fairly short that indicates its purpose.
- The signature is `std_equal(std_val, val, units, ndigits = 0)`.  `units` can be either a string or a pint Quantity. `ndigits` is the parameter name that the built-in `round` function uses. I've defaulted it to 0, which would round to the nearest integer value just as the bult-in `round` function does.

I have purposefully not modified any code files where this new function would be actually used. I wanted to get feedback on this much and get it merged in before tackling its uses.

A typical use would like this:
```
return std_equal(standard_value, calc_vals["some_value"], "ft2", 1)
```
Or you could use keyed arguments:
```
 return std_equal(standard_value, calc_vals["some_value"], units="ft2", ndigits=1)
```


yunjoonjung-PNNL (approved):

charliepnnl:
I have rewritten this function in light of the feedback Weili received from the RDS team. (See email from 3/29/22.)

It turned out to be easy to allow the operands to be numbers or pint Quantities.

weilixu:
Thank you Charlie, below is the copy of the request for documentation purpose.

weilixu (approved):

PR #396, opened 2022-03-30
Rds/cml/4 14

claperle:


docs/section4/4-14.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Rule Logic:  
+**Applicability Check 1:**  
+- For each space in the P_RMR: `For space_p in P_RMR...Space:`
+    - Check if the lighting space type classification equals computer room: `if space_p.lighting_space_type == "COMPUTER_ROOM":`
+        - Reset the total space wattage variable to zero: `total_space_wattage = 0`
+        - For each miscellaneous equipment id in the space: `For misc_p in space_p.miscellaneous_equipment:`
+            - Check if it is IT equipment: `if misc_p.miscellaneous_equipment_type == "INFORMATION_TECHNOLOGY_EQUIPMENT":`
+                - Add Wattage to the total space Wattage: `total_space_wattage = total_space_wattage + misc_p.power`
+        - Get the square footage of the space: `floor_area = space_p.floor_area * CONVERSION(M2 to FT2)`
```
mkarpman, line 29: I don't think we need "conversion" statement here as conversion of all RMD values will happen at the time when the RCT will parse the RMD.

docs/section4/4-14.md
```diff
@@ -0,0 +1,38 @@
...
+**Function Calls:**  None  
+
+
+## Rule Logic:  
+**Applicability Check 1:**  
+- For each space in the P_RMR: `For space_p in P_RMR...Space:`
+    - Check if the lighting space type classification equals computer room: `if space_p.lighting_space_type == "COMPUTER_ROOM":`
+        - Reset the total space wattage variable to zero: `total_space_wattage = 0`
+        - For each miscellaneous equipment id in the space: `For misc_p in space_p.miscellaneous_equipment:`
+            - Check if it is IT equipment: `if misc_p.miscellaneous_equipment_type == "INFORMATION_TECHNOLOGY_EQUIPMENT":`
```
mkarpman, line 27: I wonder if we should use this descriptor or just sum up all misc. equipment power entered for the space.

docs/section4/4-14.md
```diff
@@ -0,0 +1,38 @@
...
+
+1. Applicable to spaces with the space.lighting_space_type equal to "COMPUTER_ROOM".
+
+**Function Calls:**  None  
+
+
+## Rule Logic:  
+**Applicability Check 1:**  
+- For each space in the P_RMR: `For space_p in P_RMR...Space:`
+    - Check if the lighting space type classification equals computer room: `if space_p.lighting_space_type == "COMPUTER_ROOM":`
```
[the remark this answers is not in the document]
supriyagoel: image

docs/section4/4-14.md
supriyagoel: I agree that we should sum up all - BEM tools don't have a way to characterize electric equipment and this value might not be filled out consistently.

docs/section4/4-14.md
supriyagoel: @mkarpman I agree that we don't need the conversion metric (i.e. 10.76) but we do need the conversion(m2 to ft2) to inform the RCT  team on what metrics to use.. The way CLP has it here looks good to me.

docs/section4/4-14.md
```diff
@@ -0,0 +1,38 @@
...
+        - Reset the total space wattage variable to zero: `total_space_wattage = 0`
+        - For each miscellaneous equipment id in the space: `For misc_p in space_p.miscellaneous_equipment:`
+            - Check if it is IT equipment: `if misc_p.miscellaneous_equipment_type == "INFORMATION_TECHNOLOGY_EQUIPMENT":`
+                - Add Wattage to the total space Wattage: `total_space_wattage = total_space_wattage + misc_p.power`
```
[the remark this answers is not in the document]
claperle: I think we need to create an aggregate schedule across the space and then find the max.

I think for each hour of the year we need to multiply EPD*Schedule Multiplier and sum them for each miscellaneous equipment id and then we need to take the maximum of this  aggregate schedule as the EPD for the space

docs/section4/4-14.md
supriyagoel: @claperle based on the discussion last week, i take back my comment. We don't need the conversion statement here.

supriyagoel (approved):
the one space area conversion statement needs to be removed, else all looks good

mkarpman (approved):

PR #402, opened 2022-03-31
Rs/wx/rule5 13

weilixu:
Rule 5-13 implementation

### DIFF
--- a/rct229/rules/section5/section5rule13.py
+++ b/rct229/rules/section5/section5rule13.py
@@ -69,7 +69,8 @@ class Section5Rule13(RuleDefinitionListIndexedBase):
             def __init__(self):
                 super(Section5Rule13.BuildingRule.FloorRule, self).__init__(
                     rmrs_used=UserBaselineProposedVals(False, True, False),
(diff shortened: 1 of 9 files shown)


yunjoonjung-PNNL (approved):

rct229/rules/section5/section5rule13.py
```diff
@@ -0,0 +1,126 @@
...
+                ),
+            }
+
+        class FloorRule(RuleDefinitionBase):
+
+            def __init__(self):
+                super(Section5Rule13.BuildingRule.FloorRule, self).__init__(
+                    rmrs_used=UserBaselineProposedVals(False, True, False),
+                    required_fields={"$": ["construction"],
+                                     "construction": ["u_factor"]},
```
charliepnnl, line 74: We are skipping Black in the git PR for now, but we should still use it before committing. This looks like non-Black formatting.

charliepnnl (approved):
Looks good to me.  I only noticed that it looks like Black was not run on this code before committing.

PR #403, opened 2022-04-01
Exception catch implementation

weilixu:
Exception handlings, should work with getattr_ to remove the requirement of None checks

### DIFF
--- a/rct229/rule_engine/rule_base.py
+++ b/rct229/rule_engine/rule_base.py
@@ -84,8 +84,8 @@ class RuleDefinitionBase:
                 id: string - A unique identifier for the rule; ommitted if None
                 description: string - The rule description; ommitted if None
                 rmr_context: string - a JSON pointer into the RMR; omitted if empty
-                result: string or list - One of the strings "PASS", "FAIL", "NA",
-                    or "REQUIRES_MANUAL_CHECK" or a list of outcomes for
+                result: string or list - One of the strings "PASS", "FAIL", "UNDERTERMINED",
+                    or a list of outcomes for
                     a list-type rule
             }
         """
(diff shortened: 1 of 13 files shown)


yunjoonjung-PNNL (approved):

rct229/rule_engine/rule_base.py
```diff
@@ -113,26 +114,33 @@ def evaluate(self, rmrs, data=None):
...
-                    else:
-                        # Evaluate the actual rule check
-                        result = self.rule_check(context, calc_vals, data)
-                        if isinstance(result, list):
-                            # The result is a list of outcomes
-                            outcome["result"] = result
-                        # Assume result type is bool
-                        elif result:
-                            outcome["result"] = "PASSED"
+                    try:
```
charliepnnl, line 117: I think this try should be moved up to include the applicability check.
The except clauses should then be moved down to match.

rct229/rule_engine/rule_base.py
```diff
@@ -113,26 +114,33 @@ def evaluate(self, rmrs, data=None):
...
+                                outcome["result"] = "FAILED"
+                    except MissingKeyException as ke:
+                        outcome["result"] = "UNDETERMINED"
+                        outcome["message"] = str(ke)
+                    except RCTFailureException as fe:
+                        outcome["result"] = "FAILED"
+                        outcome["message"] = str(fe)
                 else:
-                    outcome["result"] = "NA"
+                    outcome["result"] = "UNDETERMINED"
```
charliepnnl, line 143: Do we really want to make this UNDETERMINED?  It seems that NA is distinctly different. I suggest leaving is a NA until a further discussion with Supriya about it.

charliepnnl (approved):
I'm still not convinced of the need for an exception status, but it isn't hearting anything.

rct229/rule_engine/rule_base.py
weilixu: > I'm still not convinced of the need for an exception status, but it isn't hearting anything.

I can comment it out for now.

weilixu: Make sense, I will move the try and catch to include applicability check

PR #404, opened 2022-04-01
Rs/yj/rule5 11

yunjoonjung-PNNL:
I would appreciate it if you reviewed this pull request.


rct229/rules/section5/section5rule11.py
```diff
@@ -0,0 +1,125 @@
...
+                }
+
+            def manaul_check_required(self, context, calc_vals, data=None):
+                target_u_factor_res = calc_vals["target_u_factor_res"]
+                target_u_factor_nonres = calc_vals["target_u_factor_nonres"]
+
+                return (
+                    target_u_factor_res is not None
+                    and target_u_factor_nonres is not None
+                    and target_u_factor_res != target_u_factor_nonres
```
weilixu, line 118: Here, it seems we do not need the None check?
We are checking if `target_u_factor_res != target_u_factor_nonres` regardless of whether they are none or values.

weilixu (changes requested):

yunjoonjung-PNNL:
I updated the code based on the comments and would appreciate it if this branch could be reviewed again.

weilixu (approved):

charliepnnl (approved):
Approved with two comments that should be addressed.

PR #405, opened 2022-04-01
Rule 5-28

weilixu:
5-28 Rule definition implementation.


yunjoonjung-PNNL (approved):

rct229/rules/section5/section5rule28.py
```diff
@@ -0,0 +1,169 @@
...
+        class UnregulatedSurfaceRule(RuleDefinitionListIndexedBase):
+            def __init__(self):
+                super(
+                    Section5Rule28.BuildingRule.UnregulatedSurfaceRule, self
+                ).__init__(
+                    rmrs_used=UserBaselineProposedVals(False, True, True),
+                    required_fields={},
+                )
+
+            def create_context_list(self, context, data=None):
```
charliepnnl, line 93: In this case, we do not need to override `create_context_list`. Instead, we can simply include the field
```
list_path="subsurfaces[*]"
```
in the initializer method. We needed to override `create_context_list` in BuildingRule because we also wanted to restrict the surfaces to being UNREGULATED.

charliepnnl, line 93: In this case, we do not need to override `create_context_list`. Instead, we can simply include the field
```
list_path="subsurfaces[*]"
```
in the initializer method. We needed to override `create_context_list` in BuildingRule because we also wanted to restrict the surfaces to being UNREGULATED.

rct229/rules/section5/section5rule28.py
```diff
@@ -0,0 +1,169 @@
...
+                        )
+                        and std_equal(
+                            calc_vals["glazed_area"], calc_vals["glazed_area"]
+                        )
+                        and std_equal(
+                            calc_vals["opaque_area"], calc_vals["opaque_area"]
+                        )
+                    ):
+                        return True
+                    else:
```
charliepnnl, line 165: We'll have to talk about this `else` part.  I don't understand why the RDS has this part at all. It doesn't seem to add anything more that what is in the rule description. This is what it means for the rule to fail.  So in this case, I am pushing back against the need for raising this warning at all.  When we truly do need such a warning message, I think we should handle it differently.

charliepnnl, line 165: We'll have to talk about this `else` part.  I don't understand why the RDS has this part at all. It doesn't seem to add anything more that what is in the rule description. This is what it means for the rule to fail.  So in this case, I am pushing back against the need for raising this warning at all.  When we truly do need such a warning message, I think we should handle it differently.

charliepnnl (commented):
We need to discuss raising the warning message part at the end of the `rule_check` method.

weilixu:
Remove the warning message in the rule check function

rct229/rules/section5/section5rule28.py
weilixu: This part is removed and updated with a simpler format to only check true and false

yunjoonjung-PNNL (approved):

rct229/rules/section5/section5rule28.py
```diff
@@ -0,0 +1,139 @@
...
+                            ]
+                        },
+                    )
+
+                def get_calc_vals(self, context, data=None):
+                    subsurface_b = context.baseline
+                    subsurface_p = context.proposed
+
+                    return {
+                        "baseline_subsurface_u_factor": subsurface_b["u_factor"],
```
charliepnnl, line 114: I'm a little disoriented. Do all UNREGULATED subsurfaces have to have `u_factor`, `shgc`, `glazed_area`, and `opaque_area`? I can imagine that **if they have them**, they must be the same.  

Brainstorming:  If we use get() for all these, then we could eliminate these as required_fields, but `std_equal()` in the rule_check doesn't work with None arguments. Hmm.

charliepnnl (approved):
Approved with comments that should be addressed.

rct229/rules/section5/section5rule28.py
```diff
@@ -0,0 +1,139 @@
...
+                each_rule=Section5Rule28.BuildingRule.UnregulatedSurfaceRule(),
+                index_rmr="baseline",
+            )
+
+        def create_data(self, context, data=None):
+            building = context.baseline
+            # Merge into the existing data dict
+            return {
+                **data,
+                "surface_conditioning_category_dict": get_surface_conditioning_category_dict(
```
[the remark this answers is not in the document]
weilixu: the convention is adding a `_b` to indicate this is baseline and `_p` for proposed.
scc should be fine as long as the key stays in data. For output (calc_val), I prefer to have the full name spelled out.
So we will use `scc_dict_b` instead of `surface_conditioning_category_dict`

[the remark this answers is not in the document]
charliepnnl: Yes that sounds very good.

rct229/rules/section5/section5rule28.py
weilixu: hmm, this is a good question.
1. You are correct, not all the subsurfaces necessary to have all of these values. I will expect at least `glazed_area` and `opaque_area`, but not `shgc`, and probably `u_factor`. - maybe .get with a default value could work well.
2. In the rule_check I think we should have a strict equality as RDS suggested. She mentioned that if comparing values such as u_factor between RMDs, then we should not allow margins.  I need to confirm this with Maria again.

charliepnnl: I had forgotten about your #2.  So get for all of these should work well; with no extra argument the default is None.  None == None is True so we're all set.

rct229/rules/section5/section5rule28.py
```diff
@@ -0,0 +1,139 @@
...
+                        "proposed_subsurface_glazed_area": subsurface_p["glazed_area"],
+                        "baseline_subsurface_opaque_area": subsurface_b["opaque_area"],
+                        "proposed_subsurface_opaque_area": subsurface_p["opaque_area"],
+                    }
+
+                def rule_check(self, context, calc_vals, data=None):
+                    return std_equal(
+                            calc_vals["baseline_subsurface_u_factor"],
+                            calc_vals["proposed_subsurface_u_factor"],
+                        ) and std_equal(
```
charliepnnl, line 127: The rest of these comparisons are not correct. They reference calc_vals that do not exist.

charliepnnl (approved):

rct229/rules/section5/section5rule28.py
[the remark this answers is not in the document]
weilixu: Yes, I think we should use `_b` and `_p` in the calc_vals.

At one point, I think we will need to clean up all the existing rules in one PR.

PR #407, opened 2022-04-01
Rds/cml/4 16

claperle:


docs/section4/4-16.md
```diff
@@ -0,0 +1,50 @@
...
+- Eliminate duplicates in list of hvac systems: `applicable_hvac_with_virtual_heating_cooling = list(set(applicable_hvac_with_virtual_heating_cooling))`
+- For each HeatingVentilationAirconditioningSystem in the B_RMR check if this check is applicable: `for each hvac_b in B_RMR..HeatingVentilationAirConditioningSystem:`
+    - Check if hvac_b is in the list of applicable systems, if it is found then set applicability flag to true: `if hvac_b in applicable_hvac_with_virtual_heating_cooling: rule_applicability_flag = TRUE`   
+    - If rule_applicability_flag = TRUE for hvac_b: `if rule_applicability_flag == TRUE:`  
+        - Get operation during occupied periods: `operation_during_occupied_b = hvac_b.fan_systems[0].operation_during_occupied`  
+
+        **Rule Assertion:**  
+        - Case 1: For each applicable hvac system: `outcome == "UNDETERMINED" and raise_message "Check that heating or cooling system that is only being simulated in the P_RMR to meet the requirements described in Section G3.1-10 HVAC Systems proposed column c and d are modeled in the B_RMR to be cycled ON and OFF to meet heating and cooling loads during occupied hours."`
```
supriyagoel, line 47: @mkarpman since we're giving an undetermined outcome, I don't see the point of doing all of this evaluation. Maybe this should be a non-mandatory rule and the applicability could be determined based on the presented of a virtual H or C system in the P RMR.

supriyagoel: @mkarpman and @claperle I am also unclear on why we're not doing the additional step of checking if operation_during_occupied_b  = cycling. Looks like we could do that and give a pass outcome?

Let me know what I'm missing.

supriyagoel: as discussed, will leave this RDS as is. For incorporating the operation_during_occupied_b in the outcome message we could simply revise the outcome message to say 
Check that heating or cooling system that is only being simulated in the P_RMR to meet the requirements described in Section G3.1-10 HVAC Systems proposed column c and d are modeled in the B_RMR to be cycled ON and OFF to meet heating and cooling loads during occupied hours. The fan operation during occupied for this HVAC system is 'operation_during_occupied_b'

supriyagoel (approved):

PR #417, opened 2022-04-06
Rs/wx/rule5 31

weilixu:
Rule 5-31

Adding a message to the test results - in Rule 5-31, added a message to data in the manual check required function.
QNS: should we revise the engine to include messages?


rct229/rules/section5/section5rule31.py
```diff
@@ -0,0 +1,111 @@
...
+                # Both are none, not applicable.
+                # The logic indicates either or both surface_b and surface_p have subsurfaces to be applicable
+                # But XOR could fails the test (b != p) - the equality is not checked in this rule
+                # b != p will be captured by data error exception in this test.
+                return (surface_b.get("subsurfaces") is None) and (surface_p.get("subsurfaces") is None)
+
+            def manual_check_required(self, context, calc_vals=None, data=None):
+                surface_p = context.proposed
+                # raise data error exception if proposed surface has no subsurfaces
+                subsurfaces_p = assert_(surface_p.get("subsurfaces"), f"No subsurfaces found in the proposed surface {surface_p['id']}")
```
charliepnnl, line 55: `assert_` has no return value, so this will not work as expected.

FYI:  jsonpath has a slick way of getting items that have a specified value. For example, to get all the subsurfaces with `has_manual_interior_shades` set to True, you can use:
```
subsurfaces_with_manual_interior_shades_p = find_all("subsurfaces[?(@.has_manual_interior_shades=True)]", surface_p)
```
If you also use jsonpath the get all the subsurfaces with 
```
subsurfaces_p = findall("subsurfaces[*]", surface_p)
```
Then you can simply check that
```
len(subsurfaces_with_manual_interior_shades_p) != 0 and len(subsurfaces_with_manual_interior_shades_p) != len(subsurfaces_p)
```
If surface_p has no subsurfaces, then the findall will return [] for both. In that case, the manual check test will fail. That makes sense to me – no manual check is required if a surface has no subsurfaces.

charliepnnl (changes requested):
We may want to discuss this one further.

rct229/rules/section5/section5rule31.py
```diff
@@ -0,0 +1,111 @@
...
+                num_shades = 0
+                for subsurface in subsurfaces_p:
+                    # raise missing key exception if the subsurface has no has_manual_interior_shades
+                    if getattr_(subsurface, "subsurface", "has_manual_interior_shades"):
+                        num_shades += 1
+
+                data = {"manual_check": num_shades != 0 and num_shades != len(subsurfaces_p)}
+                data["message"] = MANUAL_CHECK_MSG if data["manual_check"] else ""
```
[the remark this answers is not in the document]
weilixu: For now, we settled with the static message method.

rct229/rule_engine/rule_base.py
```diff
@@ -120,6 +124,7 @@ def evaluate(self, rmrs, data=None):
                         # Determine if manual check is required
                         if self.manual_check_required(context, calc_vals, data):
                             outcome["result"] = "UNDETERMINED"
+                            outcome["message"] = self.manual_check_required_msg
```
charliepnnl, line 127: To reduce the size of the output, how about including the message conditionally?
```
if self.manual_check_required_msg:
    outcome["message"] = self.manual_check_required_msg
```

And the same for the NOT_APPLICABLE case.

rct229/rules/section5/section5rule31.py
```diff
@@ -0,0 +1,100 @@
+from rct229.rule_engine.rule_base import (
+    RuleDefinitionBase,
+    RuleDefinitionListIndexedBase,
+)
+from rct229.rule_engine.user_baseline_proposed_vals import UserBaselineProposedVals
+from rct229.utils.assertions import assert_, getattr_
+from rct229.utils.jsonpath_utils import find_all
+from rct229.utils.match_lists import match_lists_by_id
+
+MANUAL_CHECK_MSG = "SURFACE IN P-RMR HAS SUBSURFACES MODELED WITH DIFFERENT MANUAL SHADE STATUS. VERIFY IF SUBSURFACES MANUAL SHADE STATUS IN B-RMR ARE MODELED THE SAME AS IN P-RMR"
```
charliepnnl, line 10: In my opinion, this message should not be in all caps.  It only makes the message harder to read.
I understand that this is how it appears in the RDS.

charliepnnl (approved):
Approved with a few minor comments.

rct229/rule_engine/rule_base.py
```diff
@@ -18,6 +18,8 @@ def __init__(
         rmr_context="",
         required_fields=None,
         must_match_by_ids=[],
+        manual_check_required_msg="Manual Check Required",
```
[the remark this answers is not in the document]
weilixu: agree updated.

rct229/rule_engine/rule_base.py
weilixu: Updated

PR #422, opened 2022-04-09
Ruletest JSON updates for 5-13,15,17,18, and 21

jugonzal07:
Added updates for the rules mentioned in the PR. Changes should reflect other changes made by folks in master spreadsheet on the OneDrive as well.

Issues I found while QAing these rules:

NOTES:
- 5-17 required "unconditioned zones". Used parking area interior but that might not be best.
- 5-17 Remove HVAC system for unconditioned zone tests
- 5-17 F factors in 5-17 seem identical to u-factors. I would not think this is the case but maybe its ok since it's an equality test.

Added some other minor notes in the master spreadsheet.


weilixu:
All tests (available) are passed now.
1. Master TCD has an issue for f_factor units - it labeled linear_thermal_transmittances but used Btu/ft2-hr-R.
2. Zone conditioning category still has an issue when calculating surface_ua. We are still waiting for the subcommittee to confirm the ground surfaces. - however, at this point, to make it work, I add some work-around code.

weilixu:
Tests for 15, 18 and 21 will be address in another PR (if any of them have failures)

weilixu (approved):

jugonzal07:
regarding the f-factor units-- yes thank you. I made an edit to that on my local copy of master but forgot to put it on the OneDrive. I also did not push the one drive version of the spreadsheet to the repo. In the long term, I hoped to keep the spreadsheets off the repo because they cause so many problems with git.

PR #425, opened 2022-04-13
Rs/wx/rule5 34

weilixu:
Rule 5-34 implementation

@supriyagoel @mkarpman @xingcx 
I edited the RDS by adding an applicable check for the rule. The applicable check checks whether the skylight area ratio in the proposed model is less or equal to 0.03.
Let me know if it makes sense.


yunjoonjung-PNNL (approved):

yunjoonjung-PNNL (approved):

rct229/rules/section5/section5rule34.py
```diff
@@ -0,0 +1,107 @@
...
+    def __init__(self):
+        super(Section5Rule34, self).__init__(
+            rmrs_used=UserBaselineProposedVals(False, True, True),
+            required_fields={
+                "$": ["weather"],
+                "weather": ["climate_zone"],
+            },
+            each_rule=Section5Rule34.BuildingRule(),
+            index_rmr="baseline",
+            id="5-15",
```
charliepnnl, line 26: Should be `id="5-34"`

rct229/rules/section5/section5rule34.py
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+        class BuildingSegmentRule(RuleDefinitionBase):
+            def __init__(self):
+                super(Section5Rule34.BuildingRule.BuildingSegmentRule, self).__init__(
+                    rmrs_used=UserBaselineProposedVals(False, True, False),
+                )
+
+            def is_applicable(self, context, data=None):
+                proposed = context.proposed
```
charliepnnl, line 65: It might be a little more self-documenting to call this variable `building_segment_p` instead of `proposed`.  The same idea could also be applied in  `calc_vals()`.

rct229/rules/section5/section5rule34.py
```diff
@@ -0,0 +1,107 @@
...
+                ),
+                "skylight_roof_areas_dictionary_p": get_building_segment_skylight_roof_areas_dict(
+                    data["climate_zone"], proposed
+                ),
+            }
+
+        class BuildingSegmentRule(RuleDefinitionBase):
+            def __init__(self):
+                super(Section5Rule34.BuildingRule.BuildingSegmentRule, self).__init__(
+                    rmrs_used=UserBaselineProposedVals(False, True, False),
```
charliepnnl, line 61: I think this should be:
```
rmrs_used=UserBaselineProposedVals(False, True, True)
```
Proposed is missing.

charliepnnl (approved):
Approved with some small comments that need to be addressed.

rct229/rules/section5/section5rule34.py
weilixu: Agree, this will make the logic much clearer

rct229/rules/section5/section5rule34.py
```diff
@@ -0,0 +1,107 @@
+from rct229.rule_engine.rule_base import (
+    RuleDefinitionBase,
+    RuleDefinitionListIndexedBase,
+)
+from rct229.rule_engine.user_baseline_proposed_vals import UserBaselineProposedVals
+from rct229.ruleset_functions.get_building_segment_skylight_roof_areas_dict import (
+    get_building_segment_skylight_roof_areas_dict,
+)
```
mkarpman, line 8: I believe that instead of checking individual segments, the RDS needs to sum up all roof areas and skylight areas returned by the function for individual building_segments to get the total skylight to roof ratio for the project (which may include multiple buildings), and then apply the RDS logic based on this total ratio.

weilixu: That seems like the right approach but the current RDS is written using by building_segment.
Do you want to bring this up to the RDS meeting? Or else we can modify the code and RDS to reflect the change.

mkarpman: @weilixu I believe we discussed it at the last call and agreed that it should be aggregated to include all buildings and building segments. @supriyagoel can you please confirm?

PR #426, opened 2022-04-13
Rs/wx/rule5 36

weilixu:
Rule 5-36 completed


yunjoonjung-PNNL (approved):

charliepnnl (commented):
The description of this rule doesn't seem to match the rule check assertion. For example, the description says that the "skylight orientation and tilt shall be the same ansi n the proposed design" but these are not checked by the rule assertion.

weilixu:
> The description of this rule doesn't seem to match the rule check assertion. For example, the description says that the "skylight orientation and tilt shall be the same ansi n the proposed design" but these are not checked by the rule assertion.

It is updated to reflect the rule logic - the other deleted part shall be addressed in the umbrella check.

rct229/rules/section5/section5rule36.py
```diff
@@ -0,0 +1,163 @@
...
+                    )
+                    total_skylight_area_surface_p = sum(
+                        [
+                            subsurface.get("glazed_area", ZERO.AREA)
+                            + subsurface.get("opaque_area", ZERO.AREA)
+                            for subsurface in find_all("subsurfaces[*]", roof_p)
+                        ],
+                        ZERO.AREA,
+                    )
+
```
charliepnnl, line 145: We have a divide-by-zero issue here.
I suggest:
1. Put `total_skylight_area_b` and `total_skylight_area_p` in a `get_calc_vals` method at the buliding_segment level.
2. At the roof level, put `total_skylight_area_surface_b` and `total_skylight_area_surface_p` in `calc_vals`.
3. In `rule_check`:
  - Handle the case where both `total_skylight_area_b` and `total_skylight_area_p` are zero (a passing case I think - they are trivially in the same ratios).
  - Handle the case where only one of these is zero (failing I would say - they can't be in the same ratios).
  - Finally check the ratios.

I would write this in a single return statement:
```
return (both segment areas are zero) or ((neither segment area is zero) and (the surface ratios are the same))
```

charliepnnl (approved):
Approved with some comments that should be addressed.

rct229/rules/section5/section5rule36.py
```diff
@@ -0,0 +1,163 @@
...
+                )
+
+                return [
+                    UserBaselineProposedVals(None, surface_b, surface_p)
+                    for surface_b, surface_p in proposed_baseline_surface_pairs
+                    if get_opaque_surface_type(surface_b) == OST.ROOF
+                    and scc[surface_b["id"]] != SCC.UNREGULATED
+                ]
+
+            class SurfaceRule(RuleDefinitionBase):
```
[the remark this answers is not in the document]
weilixu: The RDS checks for regulated Roof. I will update this to RoofRule

docs/section5/Rule5-36.md
```diff
@@ -36,22 +36,20 @@
   
     - Get total skylight area for building segment in P_RMR: `total_skylight_area_p = skylight_roof_areas_dictionary_p[building_segment_p.id][0]`
 
-  - For each thermal block in building segment: `for thermal_block_b in building_segment_b.thermal_blocks:`
-
-    - For each zone in thermal block: `for zone_b in thermal_block_b.zones:`  
-
-      - For each surface in zone: `for surface_b in zone_b.surfaces:`  
+  - For each thermal block in building segment: `for zone_b in building_segment_b.zones:`
```
charliepnnl, line 39: This should be "For each zone **zone** in ..."

docs/section5/Rule5-36.md
```diff
@@ -2,7 +2,7 @@
 # Envelope - Rule 5-36  
 
 **Rule ID:** 5-36  
-**Rule Description:**  Skylight area must be allocated to surfaces in the same proportion in the baseline as in the proposed design; Skylight orientation and tilt shall be the same as in the proposed design.
+**Rule Description:**  Skylight area must be allocated to surfaces in the same proportion in the baseline as in the proposed design.
```
charliepnnl, line 5: Line 5 needs an extra space character at the end to force a line break. See the rendered html.

rct229/rules/section5/section5rule36.py
charliepnnl: Yes, you are right. My item 1. should have said to put the skylight area data in a create_data method.  Good catch.

rct229/rules/section5/section5rule36.py
```diff
@@ -0,0 +1,161 @@
...
+                    total_skylight_area_surface_p = calc_vals[
+                        "total_skylight_area_surface_p"
+                    ]
+
+                    return (
+                        # both segments have no skylight area
+                        total_skylight_area_b == 0 and total_skylight_area_p == 0
+                    ) or (
+                        # both segments have skylight area and ratios are the same
+                        total_skylight_area_b * total_skylight_area_p > 0
```
charliepnnl, line 156: Nice way to articulate that neither can be zero!
It might be good to add a comment explain that that is the purpose of looking at the product.

charliepnnl (approved):
Approved with very minor comments.

rct229/rules/section5/section5rule36.py
weilixu: added

PR #427, opened 2022-04-13
Add exception handling to function

weilixu:
Add exception handling to get opaque surface type function
 - I found it is a bit tricky to handle exceptions in this function. Better to have a proper review before merging into the developed branch.


yunjoonjung-PNNL (approved):

charliepnnl (approved):
Looks good.

PR #429, opened 2022-04-14
Rs/wx/compare schedule fn

weilixu:
Compare schedule function
Modified to limit the comparison to hourly-based schedule only.

One condition added: If the number of hours in schedule_1 and schedule_2 and mask_schedule do not match with each other and either one does not match to 8760 or 8784 (Depending on is_leap_year indicator), then raise an RCT data failure exception, which will eventually set a UNDETERMINED outcome for the rule checking.


yunjoonjung-PNNL (approved):

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,68 @@
+from rct229.utils.assertions import RCTFailureException
+
+REGULAR_YEAR_HOUR = 8760
+LEAP_YEAR_HOUR = 8784
+
+
+def compare_schedules(schedule_1: list, schedule_2: list, mask_schedule: list, comparison_factor: float, is_leap_year: bool):
```
charliepnnl, line 7: It might be convenient to default 
- comparison_factor (default to 1)
- is_leap_year (default to False)
- mask_schedule (all ones; would have to set this in the function body if mask_schedule is None; for that reason is should be the last argument). Defaulting this one is somewhat questionable to me, but the other two seems like very good candidates.

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,68 @@
...
+            the hourly value in schedule_1 shall be equal to that in schedule_2 times the comparison_factor. 1.0
+        is_leap_year: bool, indicate whether the comparison is in a leap year or not. True / False
+
+        Returns
+        -------
+        A dictionary containing the total number of hours that the function compares,
+        the number of hours schedule_1 matches schedule_2 with the comparison_factor,
+        EFLH difference for schedule_1 and schedule_2,
+        i.e. {
+            "TOTAL_HOURS_COMPARED": total_hours_compared,
```
charliepnnl, line 30: These shouldn't be all caps.  They should match the actual return values.

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,68 @@
...
+            eflh_schedule_1 += schedule_1[index]
+            eflh_schedule_2 += schedule_2[index]
+            if schedule_1[index] == schedule_2[index]:
+                total_hours_match += 1
+        elif hourly_value == 2:
+            eflh_schedule_1 += schedule_1[index]
+            eflh_schedule_2 += schedule_2[index] * comparison_factor
+            if schedule_1[index] == schedule_2[index] * comparison_factor:
+                total_hours_match += 1
+    eflh_difference = eflh_schedule_1 / eflh_schedule_2
```
charliepnnl, line 61: Edge case: This will be a divide by zero if `mask_schedule` is all zeros. I think there are other ways for this to occur also.

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,68 @@
...
+
+        Returns
+        -------
+        A dictionary containing the total number of hours that the function compares,
+        the number of hours schedule_1 matches schedule_2 with the comparison_factor,
+        EFLH difference for schedule_1 and schedule_2,
+        i.e. {
+            "TOTAL_HOURS_COMPARED": total_hours_compared,
+            "TOTAL_HOURS_MATCH": total_hours_match,
+            "EFLH_DIFFERENCE: EFLH_difference
```
charliepnnl, line 32: Can you explain what the `elfh_difference` represents?

charliepnnl (commented):
In order to complete my review, I need an explanation of what `eflh_difference` is intended to represent.

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,68 @@
...
+        num_hours = LEAP_YEAR_HOUR
+
+    if len(schedule_1) != len(schedule_2) or len(schedule_1) != len(mask_schedule) or len(schedule_1) != num_hours:
+        raise RCTFailureException(f"Failed when comparing hourly schedules with target number of hours. target number of hour: {num_hours}, "
+                                  f"number of hours of schedule_1 : {len(schedule_1)}; "
+                                  f"number of hours of schedule_2: {len(schedule_2)}; "
+                                  f"number of hours of mask_schedule: {len(mask_schedule)}")
+
+    total_hours_compared = 0.0
+    eflh_schedule_1 = 0.0
```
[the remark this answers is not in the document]
weilixu: EFLH stands for equivalent full load hours, It represents the number of hours an equipment would run at full load to consume the same amount of energy on the course of an entire year.

rct229/ruleset_functions/compare_schedules.py
weilixu: I have this on note for RDS.
By code, it seems the difference should be schedule_1 - schedule_2,
However, the RDS is schedule_1 / schedule_2

rct229/ruleset_functions/compare_schedules.py
weilixu: > It might be convenient to default
> 
> * comparison_factor (default to 1)
> * is_leap_year (default to False)
> * mask_schedule (all ones; would have to set this in the function body if mask_schedule is None; for that reason is should be the last argument). Defaulting this one is somewhat questionable to me, but the other two seems like very good candidates.

comparison_factor - agree, 1
is_leap_year - this shall come from RDS, I was thinking that this shall be taken from RDS, so I incline not to add default value, but it is probably good to discuss.
mask_schedule - probably a good idea to give default

I reviewed a few RDSs that have dependency on this function, it seems all mask_schedule is 1s' and comparison factor set to 1...

rct229/ruleset_functions/compare_schedules.py
weilixu: Correct, need to add zero division situation - will add the functionality.

rct229/ruleset_functions/compare_schedules.py
weilixu: After review of the RDSs, it seems the elfh_differences is the percentage difference between the two ELFH schedule. It is mostly used in the print statement such as:
`LIGHTING SCHEDULE EFLH IN P-RMR IS ${compare_schedules_result_dictionary[EFLH_DIFFERENCE]} OF THAT IN B-RMR.`

I originally thought the difference is schedule_1 - schedule_2 because standard mentioned 40 hours ELFH differences per week...

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,71 @@
...
+            eflh_schedule_1 += schedule_1[index]
+            eflh_schedule_2 += schedule_2[index] * comparison_factor
+            if schedule_1[index] == schedule_2[index] * comparison_factor:
+                total_hours_match += 1
+
+    if eflh_schedule_2 > 0:
+        eflh_difference = eflh_schedule_1 / eflh_schedule_2
+    else:
+        # only if the comparison factor is 0.0 or mask_schedule = [0] * 8760
+        eflh_difference = eflh_schedule_1
```
charliepnnl, line 64: Is this value meaningful?  Should we assert that `eflh_schedule_2 > 0` instead?

charliepnnl (approved):
Approved with comments.

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,83 @@
+from rct229.utils.assertions import RCTFailureException, assert_
+
+REGULAR_YEAR_HOURS = 8760
+LEAP_YEAR_HOURS = 8784
+
+
+def compare_schedules(
+    schedule_1, schedule_2, mask_schedule, comparison_factor: float, is_leap_year: bool
+):
+    """compare two schedules and determine if they match with or without a comparison factor when applicable
```
charliepnnl, line 10: "Compare" should be capitalized.

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,83 @@
...
+        the hourly value in schedule_1 shall be equal to that in schedule_2 times the comparison_factor. 1.0
+    is_leap_year: bool, indicate whether the comparison is in a leap year or not. True / False
+
+    Returns
+    -------
+    A dictionary containing the total number of hours that the function compares,
+    the number of hours schedule_1 matches schedule_2 with the comparison_factor,
+    EFLH difference for schedule_1 and schedule_2,
+    i.e. {
+        "TOTAL_HOURS_COMPARED": total_hours_compared,
```
charliepnnl, line 32: The capitalization should match the actual return value.

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,83 @@
...
+        "TOTAL_HOURS_MATCH": total_hours_match,
+        "EFLH_DIFFERENCE: EFLH_difference
+        }
+    """
+    num_hours = LEAP_YEAR_HOURS if is_leap_year else REGULAR_YEAR_HOURS
+
+    if (
```
charliepnnl, line 39: We have `assert_` for situations like this.

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,83 @@
...
+                total_hours_match += 1
+
+    eflh_difference = 0.0
+    if eflh_schedule_2 > 0:
+        eflh_difference = eflh_schedule_1 / eflh_schedule_2
+    else:
+        # only if the comparison factor is 0.0 or mask_schedule = [0] * 8760
+        assert_(
+            eflh_schedule_2 > 0,
+            f"EFLH_schedule_2 is 0, check comparison factor and mask_schedule. Comparison factor: #{comparison_factor}; total hours compared: #{total_hours_compared}",
```
charliepnnl, line 76: I don't think we need the "#" symbols.

rct229/ruleset_functions/compare_schedules.py
```diff
@@ -0,0 +1,83 @@
...
+                total_hours_match += 1
+        elif hourly_value == 2:
+            total_hours_compared += 1
+            eflh_schedule_1 += schedule_1[index]
+            eflh_schedule_2 += schedule_2[index] * comparison_factor
+            if schedule_1[index] == schedule_2[index] * comparison_factor:
+                total_hours_match += 1
+
+    eflh_difference = 0.0
+    if eflh_schedule_2 > 0:
```
charliepnnl, line 70: Since the if and the else are performing the same check, this could be simplified to:
```
assert_(
            eflh_schedule_2 > 0,
            f"EFLH_schedule_2 is 0, check comparison factor and mask_schedule. Comparison factor: #{comparison_factor}; total hours compared: #{total_hours_compared}",
        )

eflh_difference = eflh_schedule_1 / eflh_schedule_2
```

charliepnnl (approved):
Approved with minor comments.

rct229/ruleset_functions/compare_schedules.py
weilixu: RDS decided to use subtraction instead of division - there will be no zero division error

PR #437, opened 2022-04-25
Rs/wx/rule5 37

weilixu:


yunjoonjung-PNNL (approved):

rct229/rules/section5/section5rule37.py
```diff
@@ -0,0 +1,290 @@
...
+            return {
+                **data,
+                "building_scc_skylight_roof_ratios_dict_b": get_building_scc_skylight_roof_ratios_dict(
+                    climate_zone, building_b
+                ),
+                "surface_conditioning_category_dict_b": get_surface_conditioning_category_dict(
+                    climate_zone, building_b
+                ),
+            }
+
```
charliepnnl, line 73: By my reading of the RDS, all the output should be at the Building level.

I am very confused about the idea of looping over building segments. If anything, it seems that we would go straight to surfaces.

charliepnnl (changes requested):

rct229/rules/section5/section5rule37.py
weilixu: Its good that you have checked the RDS... I forgot the function returns building level SRR and the rule RDS has the same mistake. I will update both.

rct229/rules/section5/section5rule37.py
```diff
@@ -0,0 +1,258 @@
...
+        def __init__(self):
+            super(Section5Rule37.BuildingRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                each_rule=Section5Rule37.BuildingRule.RoofRule(),
+                index_rmr="baseline",
+                list_path="$..surfaces[*]",
+                manual_check_required_msg=MANUAL_CHECK_MSG,
+            )
+
+        def manual_check_required(self, context, calc_vals=None, data=None):
```
charliepnnl, line 62: I would prefer that we order the methods in the order in which they are called. In this case it would be

- create_data
- manual_check_required
- list_filter

rct229/rules/section5/section5rule37.py
```diff
@@ -0,0 +1,258 @@
...
+                    index_rmr="baseline",
+                    list_path="subsurfaces[*]",
+                )
+
+            def create_data(self, context, data=None):
+                surface_b = context.baseline
+                scc_type = data["surface_conditioning_category_dict_b"][surface_b["id"]]
+                return {**data, "scc_type": scc_type}
+
+            class SubsurfaceRule(RuleDefinitionBase):
```
charliepnnl, line 206: As I read the Rule Assertion - RMR section at the bottom of the RDS, the RDS seems to want an output for each roof, not each subsurface of the roof.  So I think all the logic needs to take place inside RoofRule. In other words, I don't think there should be a SubsurfaceRule here.

charliepnnl (changes requested):
I don't think there should be a SubsurfaceRule in this case.

rct229/rules/section5/section5rule37.py
weilixu: Isn't it manual_check_required go first?
In evaluate function of a RuleDefinitionListIndexedBase class, `manual_check_required` is called first, and then calls the `rule_check`, inside `rule_check`, it calls `create_data` first and then `create_context_list`.
Right after the `create_context_list`, we called `list_filter` to go into the next level `evaluate` function.

So I think the order should be
- manual_check_required
- create_data
- list_filter
?

rct229/rules/section5/section5rule37.py
weilixu: The [component level](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section5/Rule5-37.md?plain=1#L71-L81) rule assertion seems mixed surface and subsurfaces.
I think we still need to check each subsurface to verify the U-factor matches the table.

At the RMR level, it also seems we need to count the number of [subsurfaces ](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section5/Rule5-37.md?plain=1#L85-L89) for outcome reporting. My understanding is each subsurface has a rule check so we can count the number of fail, undetermined or success

charliepnnl (approved):

PR #439, opened 2022-04-25
Rs/ch/get manual check required msg

charliepnnl:
While I was here, I also implemented a `get_not_applicable_msg` method.


rct229/rule_engine/rule_base.py
```diff
@@ -452,6 +452,32 @@ def is_applicable(self, context, data=None):
 
         return True
 
+    def get_not_applicable_msg(self, context, calc_vals=None, data=None):
```
weilixu, line 455: Should we also change the message in the evaluate function?

weilixu (approved):

PR #442, opened 2022-04-27
RS/CH/get_building_scc_window_wall_ratios_dict

charliepnnl:
In the test file `get_building_scc_window_wall_ratios_dict_test.py` I mocked the `get_surface_conditioning_category_dict` and `get_opaque_surface_type` functions to return fixed values. This allowed me to avoid worrying about heating systems, etc. This technique may be useful in testing future ruleset_functions.

Notice that I made several changes to the RDS file that should be reviewed.


yunjoonjung-PNNL (approved):

xingcx:
@charliepnnl I need to update this function RDS to calculate the window-wall-ratio for RMR instead of building. The original PR is probably already merged. I'll create a new PR for this. Similarly for get_building_scc_window_wall_ratios_dict() (now renamed to get_rmr_scc_window_wall_ratios_dict()).

rct229/ruleset_functions/get_building_scc_window_wall_ratios_dict_test.py
```diff
@@ -0,0 +1,152 @@
...
+def test__get_building_scc_skylight_roof_ratios_dict(
+    mock_get_surface_conditioning_category_dict, mock_get_opaque_surface_type
+):
+    assert get_building_scc_window_wall_ratios_dict(CLIMATE_ZONE, TEST_BUILDING) == {
+        SCC.EXTERIOR_RESIDENTIAL: 0.5,
+        SCC.EXTERIOR_NON_RESIDENTIAL: 0.25,
+        SCC.EXTERIOR_MIXED: 0.25,
+        SCC.SEMI_EXTERIOR: 0.25,
+    }
+    assert mock_get_surface_conditioning_category_dict.called
```
weilixu, line 151: Do we need to leave the check for mock function in the test?

weilixu (approved):
Approved with some comments to address

rct229/ruleset_functions/get_building_scc_window_wall_ratios_dict_test.py
charliepnnl: It was useful while debugging, but I will remove it now.

rct229/ruleset_functions/get_building_scc_window_wall_ratios_dict.py
```diff
@@ -0,0 +1,108 @@
...
+                total_nonres_wall_area += surface_area
+                total_nonres_window_area += surface_window_area
+            elif scc == SCC.EXTERIOR_MIXED:
+                total_mixed_wall_area += surface_area
+                total_mixed_window_area += surface_window_area
+            elif scc == SCC.SEMI_EXTERIOR:
+                total_semi_exterior_wall_area += surface_area
+                total_semi_exterior_window_area += surface_window_area
+            else:
+                assert scc == SCC.UNREGULATED
```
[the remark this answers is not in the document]
charliepnnl: My logic was that if we get to this point, scc must be SCC.UNREGULATED because it is the only other SCC option. The assertion is intended to make that clear (I could add a comment too). If the assertion fails, it implies a bug in our code - probably because get_opaque_surface_type added another type that we are not handling here.

charliepnnl:
> @charliepnnl I need to update this function RDS to calculate the window-wall-ratio for RMR instead of building. The original PR is probably already merged. I'll create a new PR for this. Similarly for get_building_scc_window_wall_ratios_dict() (now renamed to get_rmr_scc_window_wall_ratios_dict()).

@weilixu How do you want to handle this. Should we merge this in and update it later, or wait for Xing to update the RDS?

PR #445, opened 2022-04-28
Rs/yj/rule5 48

yunjoonjung-PNNL:
I'd appreciate it if you could review this PR. 

One thing I'd like to highlight is that in the RDS, it says to use the **"match_data_element"** function (***zone_p = match_data_element(P_RMR, Zones, zone_b.id)***). However, I wasn't able to find this function so, I used one for-loop and f-else statement to make this code work. Please let me know if this is acceptable.

Further, I also modified the testing json file (***rule_5_48.json***) by filling up the missing parts.

Thank you!

### DIFF
--- a/rct229/rules/section5/section5rule48.py
+++ b/rct229/rules/section5/section5rule48.py
@@ -4,6 +4,9 @@ from rct229.rule_engine.rule_base import (
 )
 from rct229.rule_engine.user_baseline_proposed_vals import UserBaselineProposedVals
 from rct229.ruleset_functions.get_zone_conditioning_category_dict import (
+    ZoneConditioningCategory as ZCC,
+)
+from rct229.ruleset_functions.get_zone_conditioning_category_dict import (
     get_zone_conditioning_category_dict,
 )
 from rct229.utils.jsonpath_utils import find_all
@@ -36,6 +39,7 @@ class Section5Rule48(RuleDefinitionListIndexedBase):
                 rmrs_used=UserBaselineProposedVals(False, True, True),
                 each_rule=Section5Rule48.BuildingRule.ZoneRule(),
                 index_rmr="baseline",
+                list_path="$..zones[*]",
             )
 
         def create_data(self, context, data=None):
@@ -48,13 +52,18 @@ class Section5Rule48(RuleDefinitionListIndexedBase):
                 ),
             }
 
+        def list_filter(self, context_item, data=None):
+            scc_dict_b = data["scc_dict_b"]
+            zone_b = context_item.baseline
+            return scc_dict_b[zone_b["id"]] in [ZCC.UNCONDITIONED, ZCC.UNENCLOSED]
+
         class ZoneRule(RuleDefinitionBase):
             def __init__(self):
                 super(Section5Rule48.BuildingRule.ZoneRule, self).__init__(
                     rmrs_used=UserBaselineProposedVals(False, True, True),
                     required_fields={
-                        "$..zones[*]": ["infiltration"],
-                        "$..infiltration[*]": ["infiltration_flow_rate"],
+                        "$": ["infiltration"],
+                        "infiltration[*]": ["infiltration_flow_rate"],
                     },
                 )
 
@@ -62,19 +71,8 @@ class Section5Rule48(RuleDefinitionListIndexedBase):
                 zone_b = context.baseline
                 zone_p = context.proposed
 
-                baseline_zones = find_all("$..zones[*]", zone_b)
-                proposed_zones = find_all("$..zones[*]", zone_p)
-
-                if data["scc_dict_b"][baseline_zones[0]["id"]] in [
-                    "UNENCLOSED",
-                    "UNCONDITIONED",
-                ]:
-                    zone_b_infiltration = baseline_zones[0]["infiltration"][
-                        "infiltration_flow_rate"
-                    ]
-                    zone_p_infiltration = proposed_zones[0]["infiltration"][
-                        "infiltration_flow_rate"
-                    ]
+                zone_b_infiltration = zone_b["infiltration"]["infiltration_flow_rate"]
+                zone_p_infiltration = zone_p["infiltration"]["infiltration_flow_rate"]
 
                 return {
                     "baseline_infiltration": zone_b_infiltration,
--- a/rct229/rules/section5/section5rule48.py
+++ b/rct229/rules/section5/section5rule48.py
@@ -47,15 +47,15 @@ class Section5Rule48(RuleDefinitionListIndexedBase):
 
             return {
                 **data,
-                "scc_dict_b": get_zone_conditioning_category_dict(
+                "zcc_dict_b": get_zone_conditioning_category_dict(
                     data["climate_zone"], building_b
                 ),
             }
 
         def list_filter(self, context_item, data=None):
-            scc_dict_b = data["scc_dict_b"]
+            zcc_dict_b = data["zcc_dict_b"]
             zone_b = context_item.baseline
-            return scc_dict_b[zone_b["id"]] in [ZCC.UNCONDITIONED, ZCC.UNENCLOSED]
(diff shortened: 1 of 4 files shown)


rct229/rules/section5/section5rule48.py
```diff
@@ -0,0 +1,79 @@
...
+    class BuildingRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section5Rule48.BuildingRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, True),
+                required_fields={
+                    "$..zones[*]": ["infiltration"],
+                    "$..infiltration[*]": ["infiltration_flow_rate"],
+                },
+            )
+
```
charliepnnl, line 42: Since the RDS assertion applies to each zone, I think we need to use a nested class `ZoneRule` inside of `BuildingRule`.
One advantage of using a nested `ZoneRule` is that the zones will automatically be matched up.

rct229/rules/section5/section5rule48.py
```diff
@@ -0,0 +1,79 @@
...
+                    # used below for-loop because "match_data_element" function doesn't exit
+                    for zone_p in proposed_zones:
+                        if zone_p["id"] == zone_b["id"]:
+                            zone_p_infiltration = zone_p["infiltration"][
+                                "infiltration_flow_rate"
+                            ]
+                            break
+
+            return {
+                "baseline_infiltration": zone_b_infiltration,
```
charliepnnl, line 72: These return values are only for the last zone in the loop. Another indication that a nested `ZoneRule` is needed here.

charliepnnl (changes requested):
I believe a nested `ZoneRule` is needed for this one.

rct229/rules/section5/section5rule48.py
```diff
@@ -0,0 +1,97 @@
...
+                    data["climate_zone"], building_b
+                ),
+            }
+
+        class ZoneRule(RuleDefinitionBase):
+            def __init__(self):
+                super(Section5Rule48.BuildingRule.ZoneRule, self).__init__(
+                    rmrs_used=UserBaselineProposedVals(False, True, True),
+                    required_fields={
+                        "$..zones[*]": ["infiltration"],
```
charliepnnl, line 57: The context here is already a zone. So lines 57 and 58 should be
```
"$": ["infiltration"],
"infiltration[*]: ["measured_air_leakage_rate"]
```

rct229/rules/section5/section5rule48.py
```diff
@@ -0,0 +1,97 @@
...
+                        "$..infiltration[*]": ["measured_air_leakage_rate"],
+                    },
+                )
+
+            def get_calc_vals(self, context, data=None):
+                building_b = context.baseline
```
charliepnnl, line 63: The context is already a zone here, so the method can be very much simplified.

charliepnnl (changes requested):

rct229/rules/section5/section5rule48.py
```diff
@@ -0,0 +1,88 @@
...
+                )
+
+            def get_calc_vals(self, context, data=None):
+                zone_b = context.baseline
+                zone_p = context.proposed
+
+                baseline_zones = find_all("$..zones[*]", zone_b)
+                proposed_zones = find_all("$..zones[*]", zone_p)
+
+                if data["scc_dict_b"][baseline_zones[0]["id"]] in [
```
weilixu, line 68: instead of putting the conditions here, you can add `list_path` in the BuildingRule class and override the `list_filter` function to filter the qualified zones. Then you do not need to add the condition here. and also you do not need to use the `find_all` function, because `list_path` will take the code to the `Zone` data group.

```python
    class BuildingRule(RuleDefinitionListIndexedBase):
        def __init__(self):
            super(Section5Rule48.BuildingRule, self).__init__(
                rmrs_used=UserBaselineProposedVals(False, True, True),
                each_rule=Section5Rule48.BuildingRule.ZoneRule(),
                index_rmr="baseline",
                list_path = "$..zones[*]"
            )

        def create_data(self, context, data=None):
            building_b = context.baseline

            return {
                **data,
                "scc_dict_b": get_zone_conditioning_category_dict(
                    data["climate_zone"], building_b
                ),
            }
        
        def list_filter(self, context_item, data=None):
            scc_dict_b = data["scc_dict_b"]
            zone_b = context_item.baseline
            return (
                    scc_dict_b[zone_b["id"]] in [ZCC.UNCONDITIONED, ZCC.UNENCLOSED]
            )    
```

Then your `get_calc_val` method can be revised to:
```python
            def get_calc_vals(self, context, data=None):
                zone_b = context.baseline
                zone_p = context.proposed

                zone_b_infiltration = zone_b["infiltration"][
                    "infiltration_flow_rate"
                ]
                zone_p_infiltration = zone_p["infiltration"][
                    "infiltration_flow_rate"
                ]

                return {
                    "baseline_infiltration": zone_b_infiltration,
                    "proposed_infiltration": zone_p_infiltration,
                }
```

rct229/rules/section5/section5rule48.py
```diff
@@ -0,0 +1,88 @@
...
+                    "UNENCLOSED",
```
weilixu, line 69: For enums, we prefer to import the enum class for zone category, so import
```python
    get_zone_conditioning_category_dict, ZoneConditioningCategory as ZCC
)
```
and then in here, you can do:
```python
ZCC.UNENCLOSED
```

rct229/rules/section5/section5rule48.py
```diff
@@ -0,0 +1,88 @@
...
+                    zone_b_infiltration = baseline_zones[0]["infiltration"][
+                        "infiltration_flow_rate"
+                    ]
+                    zone_p_infiltration = proposed_zones[0]["infiltration"][
+                        "infiltration_flow_rate"
+                    ]
+
+                return {
+                    "baseline_infiltration": zone_b_infiltration,
+                    "proposed_infiltration": zone_p_infiltration,
```
weilixu, line 81: This part is not an error but just for the sake of consistency.

Charlie and I recently had a discussion about the naming convention. We both agree to put `_b` or `_p` at the end of the parameter. So in this case, instead of `zone_b_infiltration`, we will put it as `zone_infiltration_b`.

rct229/ruletest_engine/ruletest_engine.py
```diff
@@ -499,6 +499,6 @@ def run_envelope_tests():
     Results of envelope test are spit out to console
     """
 
-    envelope_test_json = "envelope_tests.json"
-
+    # envelope_test_json = "envelope_tests.json"
+    envelope_test_json = "section5/rule_5_48.json"
```
weilixu, line 503: Could you change it back to `envelope_test_json = "envelope_tests.json"`.

rct229/rules/section5/section5rule48.py
yunjoonjung-PNNL: I appreciate this comments, Charlie :).

rct229/rules/section5/section5rule48.py
yunjoonjung-PNNL: I appreciate your kind comments, Weili. After I saw this, I was able to better understand the usage of ```list_filter``` method.

rct229/rules/section5/section5rule48.py
```diff
@@ -0,0 +1,86 @@
...
+
+            def get_calc_vals(self, context, data=None):
+                zone_b = context.baseline
+                zone_p = context.proposed
+
+                zone_b_infiltration = zone_b["infiltration"]["infiltration_flow_rate"]
+                zone_p_infiltration = zone_p["infiltration"]["infiltration_flow_rate"]
+
+                return {
+                    "baseline_infiltration": zone_b_infiltration,
```
charliepnnl, line 78: This should follow our convention of putting "_b" or "_p" at the end of the variable everywhere. Also my comment above applies here also, so this should be:
```
"zone_infiltration_flow_rate_b": zone_b_infiltration,
```

charliepnnl (approved):
Approved with style/convention -type comments.

weilixu (approved):

PR #448, opened 2022-04-30
Update Rule 5-24

xingcx:


mkarpman (approved):

docs/section5/Rule5-24.md
```diff
@@ -18,79 +18,77 @@
...
+  - Else, request manual review: `else: manual_review_flag = TRUE`
 
-    - Else, request manual review: `else: manual_review_flag = TRUE`
+- Else, RMR does not have exterior mixed type wall surface: `else:`
```
mkarpman, line 37: Seems that line 37 would result in skipping the check for residential, non-residential and semi-heated surfaces. Suggest removing it.

docs/section5/Rule5-24.md
```diff
@@ -18,79 +18,77 @@
...
 
-            - Case 4: Else if wall is exterior non-residential type and vertical fenestration U-factor matches Table G3.4 requirement: `else if ( scc_dictionary_b[surface_b] == "EXTERIOR NON-RESIDENTIAL" ) AND ( subsurface_b.u_factor == target_u_factor_nonres ): PASS`
+          - Case 5: Else if wall is semi-exterior type and skylight U-factor matches Table G3.4 requirement: `else if ( scc_dictionary_b[surface_b] == "SEMI-EXTERIOR" ) AND ( subsurface_b.u_factor == target_u_factor_semiheated ): PASS`
 
-            - Case 5: Else if wall is semi-exterior type and skylight U-factor matches Table G3.4 requirement: `else if ( scc_dictionary_b[surface_b] == "SEMI-EXTERIOR" ) AND ( subsurface_b.u_factor == target_u_factor_semiheated ): PASS`
-
-            - Case 6: Else: `else: FAIL`
+          - Case 6: Else: `else: FAIL`
 
 **Rule Assertion - RMR:**
```
mkarpman, line 93: I don't think the entire rule should be tagged as "Undetermined" if at least one surface triggers such flag

docs/section5/Rule5-24.md
mkarpman: Discuss at the RDS call and agreed that no change is needed

docs/section5/Rule5-24.md
xingcx: Discussed at the RDS call, if one surface triggers a flag as mixed (res/non res) type, we will not be able to determine the actual res or non-res window-wall-ratio percentage, so the entire rule will be tagged as "undetermined".

supriyagoel (approved):

PR #449, opened 2022-04-30
Update Rule 5-26

xingcx:


docs/section5/Rule5-26.md
```diff
@@ -18,79 +18,77 @@
...
 
-            - Case 4: Else if wall is exterior non-residential type and vertical fenestration SHGC matches Table G3.4 requirement: `else if ( scc_dictionary_b[surface_b] == "EXTERIOR NON-RESIDENTIAL" ) AND ( subsurface_b.solar_heat_gain_coefficient == target_shgc_nonres ): PASS`
+          - Case 5: Else if wall is semi-exterior type and skylight SHGC matches Table G3.4 requirement: `else if ( scc_dictionary_b[surface_b] == "SEMI-EXTERIOR" ) AND ( subsurface_b.solar_heat_gain_coefficient == target_shgc_semiheated ): PASS`
 
-            - Case 5: Else if wall is semi-exterior type and skylight SHGC matches Table G3.4 requirement: `else if ( scc_dictionary_b[surface_b] == "SEMI-EXTERIOR" ) AND ( subsurface_b.solar_heat_gain_coefficient == target_shgc_semiheated ): PASS`
-
-            - Case 6: Else: `else: FAIL and NUMBER_OF_FAIL_COMPONENTS++`
+          - Case 6: Else: `else: FAIL and NUMBER_OF_FAIL_COMPONENTS++`
 
 **Rule Assertion - RMR:**
```
mkarpman, line 93: I don't think the entire rule should be tagged as "Undetermined" if at least one surface triggers such flag.

xingcx: Discussed at the RDS call, if one surface triggers a flag as mixed (res/non res) type, we will not be able to determine the actual res or non-res window-wall-ratio percentage, so the entire rule will be tagged as "undetermined".

supriyagoel (approved):

mkarpman (approved):

docs/section5/Rule5-26.md
```diff
@@ -18,79 +18,77 @@
...
+      - For each subsurface in wall: `for subsurface_b in surface_b.subsurfaces:`
 
-        - For each subsurface in wall: `for subsurface_b in surface_b.subsurfaces:`
+        - Check if subsurface is door and glazed area is more than 50% of the total door area, or subsurface is not door: `if (( subsurface_b.classification == "DOOR" ) AND ( subsurface_b.glazed_area > subsurface_b.opaque_area )) OR ( subsurface_b.classification != "DOOR" ):`
 
-          - Check if subsurface is door and glazed area is more than 50% of the total door area, or subsurface is not door: `if (( subsurface_b.classification == "DOOR" ) AND ( subsurface_b.glazed_area > subsurface_b.opaque_area )) OR ( subsurface_b.classification != "DOOR" ):`
+          **Rule Assertion - Component:**
 
-            **Rule Assertion - Component:**
+          - Case 1: For each subsurface that is vertical fenestration, if wall is exterior mixed type and the baseline requirements for residential and non-residential type SHGC for different window-wall-ratio are different: `if manual_review_flag AND ( scc_dictionary_b[surface_b] == "EXTERIOR MIXED" ): UNDETERMINED`
```
supriyagoel, line 82: Comment from @mkarpman 
One improvement that may be made to both 5-26 and 5-37 to reduce occurance of "undetermined" outcome is to check whether "mixed" fenestration would affect the WWR bracket for either residential or non-residential.  Currently, the indeterminate outcome is produced if the U-factor in Tables G3.4 differs for residential vs non-residential and WWR bracket. However, if area or "mixed" fenestration is small enough to not result in a change of WWR bracket for either residential or non-residential, the rule can produce PASS or FAIL outcome instead of UNDETERMINED.

supriyagoel: as discussed, since the function has already been implemented, we will leave it as is and add a comment to the RDS to clarify that this workaround can be implemented in the future

xingcx: Note added in R-26 and get_rmi_scc_window_wall_ratios().

Issue #454, opened 2022-05-05
5-45 logic


weilixu:
In [5-45](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section5/Rule5-45.md), the rule checks whether baseline matches the proposed.

It is recommended to to switch the logic from between line [23-29](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section5/Rule5-45.md?plain=1#L23-L29) to:
`- For each zone in the Baseline model: 'for zone_p in P_RMR...zones:'
   - Get zone infiltration: `infiltration_p = zone_p.infiltration`
   - Get matching zone in B_RMR: `zone_b = match_data_element(B_RMR, Zones, zone_p.id)`
   - ...
`
What do you think?

charliepnnl:
I think you meant `- For each zone in the Proposed model: ...`

xingcx:
I'm thinking if the rule is applicable to B_RMR, it should check B_RMR elements as P_RMR is the reference?
`**Applicability:** All required data elements exist for B_RMR`.

charliepnnl:
I think Xing is correct. The RDS says that the proposed matches the baseline. So we should index on the baseline as the RDS logic shows.

weilixu:
> I'm thinking if the rule is applicable to B_RMR, it should check B_RMR elements as P_RMR is the reference?
> `**Applicability:** All required data elements exist for B_RMR`.

The standard language is often phrased as the baseline should match the proposed model, which got me wondering why we are doing it the other way.

xingcx:
That is correct. The baseline should match the proposed model. If we are checking the baseline (whether the baseline matches the proposed), I assume we should loop through the baseline model instead of the proposed?

weilixu:
> That is correct. The baseline should match the proposed model. If we are checking the baseline (whether the baseline matches the proposed), I assume we should loop through the baseline model instead of the proposed?


I think the final rule check outcome is the same. The difference could be in the report.

Thinking a case where:
b_rmr contains {surface_1, surface_2}
p_rmr contains {surface_2, surface_4}

Loop b_rmr and try to find a matched p_rmr will raise error on no matching of surface_1
Loop p_rmr and try to find a matched p_rmr will raise error on no matching of surface_4

Both lead to a fail result, but reporting msgs are different.

mkarpman:
[...]

PR #455, opened 2022-05-05
Add rule 5-45, add fail message in the rule engine, update 5-45

weilixu:
#454 issue for RDS team review

### DIFF
--- a/rct229/rule_engine/rule_base.py
+++ b/rct229/rule_engine/rule_base.py
@@ -19,6 +19,7 @@ class RuleDefinitionBase:
         required_fields=None,
         must_match_by_ids=[],
         manual_check_required_msg="",
+        fail_msg="",
         not_applicable_msg="",
     ):
         """Base class for all Rule definitions
@@ -57,6 +58,7 @@ class RuleDefinitionBase:
         self.required_fields = required_fields
         self.manual_check_required_msg = manual_check_required_msg
         self.not_applicable_msg = not_applicable_msg
+        self.fail_msg = fail_msg
 
     def evaluate(self, rmrs, data=None):
         """Generates the outcome dictionary for the rule
@@ -142,6 +144,9 @@ class RuleDefinitionBase:
                                 outcome["result"] = "PASSED"
                             else:
                                 outcome["result"] = "FAILED"
+                                fail_msg = self.get_fail_msg(context, calc_vals, data)
+                                if fail_msg:
+                                    outcome["message"] = fail_msg
                     else:
                         outcome["result"] = "NOT_APPLICABLE"
                         not_applicable_msg = self.get_not_applicable_msg(context, data)
@@ -575,6 +580,32 @@ class RuleDefinitionBase:
         """
 
         raise NotImplementedError
+
+    def get_fail_msg(self, context, calc_vals=None, data=None):
+        """Gets the message to include in the outcome for the FAIL case.
+
+        This base implementation simply returns the value of
+        self.fail_msg, which defaults to the empty string.
+
+                This method should only be overridden if there is more than one string
+                used for the PASS or FAIL case. A fixed string can be given in the
+                `fail_msg` field passed to the initializer.
+
+                Parameters
+                ----------
+                context : UserBaselineProposedVals
+                    Object containing the contexts for the user, baseline, and proposed RMRs
+                calc_vals : dict or None
+
+                data : An optional data object. It is ignored by this base implementation.
+
+                Returns
+                -------
+                str
+                    The message associated with the Pass or Fail case
+                """
+
+        return self.fail_msg
 
 
 class RuleDefinitionListBase(RuleDefinitionBase):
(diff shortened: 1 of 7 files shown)


rct229/rules/section5/section5rule45.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def create_data(self, context, data=None):
+        rmr_proposed = context.proposed
+        return {"is_leap_year": rmr_proposed["calendar"]["is_leap_year"]}
+
+    def list_filter(self, context_item, data=None):
+        # baseline should match to proposed
+        zone_p = context_item.proposed
+        return (
+                zone_p.get("infiltration") is not None
```
charliepnnl, line 31: There are a couple of more direct, pythonic ways of making this check:
1. `"infiltration" in zone_p`
2. `zone_p.has_key("infiltration")`

In this case I would favor the second option because it is clearer to a non-python programer.

rct229/rules/section5/section5rule45.py
```diff
@@ -0,0 +1,75 @@
...
+            }
+            mask_schedule = [1.0] * (LEAP_YEAR_HOURS if data["is_leap_year"] else REGULAR_YEAR_HOURS)
+
+            infiltration_p = zone_p["infiltration"]
+            if zone_b:
+                # make sure matched baseline zone is not None (required_field does not check None)
+                infiltration_b = zone_b.get["infiltration"]
+                compare_schedules_result_dictionary = compare_schedules(infiltration_b["multiplier_schedule"], infiltration_p["multiplier_schedule"], mask_schedule, 1.0, data["is_leap_year"])
+
+            return {
```
charliepnnl, line 60: This can be simplified to:
```
return compare_schedules_result_dictionary
```

rct229/rules/section5/section5rule45.py
```diff
@@ -0,0 +1,75 @@
...
+
+            compare_schedules_result_dictionary = {
+                "total_hours_compared": 0.0,
+                "total_hours_match": 0.0,
+                "eflh_difference": 0.0
+            }
+            mask_schedule = [1.0] * (LEAP_YEAR_HOURS if data["is_leap_year"] else REGULAR_YEAR_HOURS)
+
+            infiltration_p = zone_p["infiltration"]
+            if zone_b:
```
charliepnnl, line 55: Two issues here.
1. We are indexing on baseline, so `zone_p` is the "matched" one of the two.
2. Issue one doesn't matter because `zone_b` and `zone_p` will not be None. Any non-matching proposed zones are flagged in the output and `get_calc_vals` will not be called for such zones.

rct229/rules/section5/section5rule45.py
```diff
@@ -0,0 +1,75 @@
...
+
+            return {
+                "total_hours_compared": compare_schedules_result_dictionary["total_hours_compared"],
+                "total_hours_match": compare_schedules_result_dictionary["total_hours_match"],
+                "eflh_difference": compare_schedules_result_dictionary["elfh_difference"]
+            }
+
+        def rule_check(self, context, calc_vals=None, data=None):
+            total_hours_compared = calc_vals["total_hours_compared"]
+            total_hours_match = calc_vals["total_hours_match"]
```
charliepnnl, line 68: This should be `total_hours_matched`.
I see that compare_schedules.py using "matched" in the documentation and "match" in the actual return value.  I think "matched" is more grammatically correct. "total_hours_match" would grammatically indicate a boolean value. 

While we are discussing compare_schedules.py, I see two other issues there:
- The equality check on line 63 should probably use std_equal() because of possible roundoff errors.
- This is very minor:  `total_hours_compared` and `total_hours_matched` are functionally integers so lines 48 and 51 should initialize the variable to 0 rather than 0.0.

rct229/rules/section5/section5rule45.py
```diff
@@ -0,0 +1,75 @@
...
+            # total_hours_compared needs to be greater than 0.0
+            return total_hours_compared > 0.0 and total_hours_compared == total_hours_match
```
charliepnnl, line 70: `total_hours_compared` is either LEAP_YEAR_HOURS or REGULAR_YEAR_HOURS, so it is never zero.

rct229/rule_engine/rule_base.py
```diff
@@ -576,6 +581,32 @@ def rule_check(self, context, calc_vals=None, data=None):
 
         raise NotImplementedError
 
+    def get_fail_msg(self, context, calc_vals=None, data=None):
+        """Gets the message to include in the outcome for the FAIL case.
+
+        This base implementation simply returns the value of
+        self.fail_msg, which defaults to the empty string.
+
+                This method should only be overridden if there is more than one string
```
charliepnnl, line 590: Please fix the indention. Shift it to the left.

charliepnnl (approved):
Approved with comments that need to be addressed.

rct229/rules/section5/section5rule45.py
weilixu: I will update both files.

PR #457, opened 2022-05-06
Rs/yj/rule5 42

yunjoonjung-PNNL:
Rule 5-42 was done referring to Rule 5-40. I'd appreciate it if you reviewed this PR.


weilixu (approved):

charliepnnl (approved):
Looks good to me!

@weilixu  Can you verify that we need to use std_equal in the rule_check?

PR #458, opened 2022-05-06
Update 5-18 + undetermined message logic (first attempt)

weilixu:
Update 5-18 based on RDS discussion

### DIFF
--- a/rct229/rules/section5/section5rule18.py
+++ b/rct229/rules/section5/section5rule18.py
@@ -1,3 +1,4 @@
+from rct229.data.schema_enums import schema_enums
 from rct229.data_fns.table_G3_111_fns import table_G3_1_1_1_lookup
 from rct229.rule_engine.rule_base import (
     RuleDefinitionBase,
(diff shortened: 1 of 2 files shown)


yunjoonjung-PNNL (approved):

rct229/rules/section5/section5rule18.py
```diff
@@ -101,24 +105,35 @@ def get_calc_vals(self, context, data=None):
...
-                if area_type is not "NONE":
-                    area_type_wwr = (
-                        area_type_window_wall_ratio_b[area_type]["total_window_area"]
-                        / area_type_window_wall_ratio_b[area_type]["total_wall_area"]
-                    )
+                area_type_wwr = (
+                    area_type_window_wall_ratio_b[area_type]["total_window_area"]
+                    / area_type_window_wall_ratio_b[area_type]["total_wall_area"]
+                )
+                if area_type != "NONE":
```
charliepnnl, line 113: Is there an area_type called "NONE"?  There is an "OTHER", but we have already filtered those out.

charliepnnl (approved):
Added comment about "NONE" area_type.

rct229/rules/section5/section5rule18.py
weilixu: The NONE is not an `area_type` from the schema, but is a [key ](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/ruleset_functions/get_area_type_window_wall_areas.md?plain=1#L27)returned from the `area_type_window_wall_areas` when dealing with data missing.

In this case, we just have to flag it out for manual checking.

PR #460, opened 2022-05-09
Create Rule23-9.md

xingcx:


mkarpman (approved):

docs/section23/Rule23-9.md
```diff
@@ -0,0 +1,45 @@
+
+# Airside System - Rule 23-9  
+
+**Schema Version:** 0.0.12  
+**Mandatory Rule:** True  
+**Rule ID:** 23-9  
+**Rule Description:** Systems 6&8: Supply air temperature setpoint shall be constant at the design condition.  
+**Rule Assertion:** B-RMR = expected value
```
supriyagoel, line 8: @mkarpman and @JasonGlazer this requirement is slightly confusing. G3.1.3.14 says 'The supply air temperature set point shall be constant at the design condition' and G3.1.3.12 says The air temperature for cooling shall be reset higher by 5°F under the minimum cooling load conditions.'
We use the same RMD data element - temperature_control under fan system to verify SAT under design conditions and under other conditions. This will lead to incorrect evaluations. 
Am i interpreting this requirement correctly?

mkarpman: @supriyagoel I believe G3.1.3.12 deals with cooling SAT (control of cooling coil), and G3.1.3.14 is for heating temperature for fan-powered boxes. 

I think the setpoint would be checked for Terminal.supply_temperature_setpoint. Do we need to have one for cooling and one for heating? Also, I think we need to add temperature_control to Terminal with the same options as for the corresponding FanSystem element.

supriyagoel: as discussed during the RDS call, MK will get clarity on this from the wider ECB grpup

supriyagoel: Added issue  https://github.com/open229/ruleset-model-description-schema/issues/114

supriyagoel (approved):

PR #464, opened 2022-05-13
Rs/wx/rule5 24

weilixu:
The implementation is temporary because we are still waiting for the confirmation of the ID issue to converting function input scope.

Regardless, rule5-24 is completed.
Once the ID issue confirmed, we can do another PR to address all rule definitions with the same issue.

### DIFF
--- a/rct229/rules/section5/section5rule24.py
+++ b/rct229/rules/section5/section5rule24.py
@@ -1,0 +1,149 @@
+from rct229.data.schema_enums import schema_enums
+from rct229.data_fns.table_G3_4_fns import table_G34_lookup
+from rct229.rule_engine.rule_base import RuleDefinitionListIndexedBase, RuleDefinitionBase
+from rct229.rule_engine.user_baseline_proposed_vals import UserBaselineProposedVals
+from rct229.ruleset_functions.get_building_scc_window_wall_ratios_dict import get_building_scc_window_wall_ratios_dict
+from rct229.ruleset_functions.get_opaque_surface_type import (get_opaque_surface_type, OpaqueSurfaceType as OST)
+from rct229.ruleset_functions.get_surface_conditioning_category_dict import (get_surface_conditioning_category_dict, SurfaceConditioningCategory as SCC)
+from rct229.utils.jsonpath_utils import find_all
+from rct229.utils.std_comparisons import std_equal
+
+DOOR = schema_enums["SubsurfaceClassificationType"].DOOR.name
+
+
+class Section5Rule24(RuleDefinitionListIndexedBase):
+    """Rule 24 of ASHRAE 90.1-2019 Appendix G Section 5 (Envelope)"""
+
+    def __init__(self):
+        super(Section5Rule24, self).__init__(
+            rmrs_used=UserBaselineProposedVals(False, True, False),
+            required_fields={
+                "$": ["weather"],
+                "weather": ["climate_zone"],
+            },
+            each_rule=Section5Rule24.BuildingRule(),
+            index_rmr="baseline",
+            id="5-24",
+            description="Vertical fenestration U-factors for residential, non-residential and semi-heated spaces in the baseline model must match the appropriate requirements in Table G3.4-1 through G3.4-8 for the appropriate WWR in the baseline RMD",
+            list_path="ruleset_model_instances[0].buildings[*]",
+        )
+
+    def create_data(self, context, data=None):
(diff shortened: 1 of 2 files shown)


yunjoonjung-PNNL (approved):

rct229/rules/section5/section5rule24.py
```diff
@@ -0,0 +1,259 @@
...
+                )
+
+            def create_data(self, context, data=None):
+                surface_b = context.baseline
+                scc_dict_b = data["scc_dict_b"]
+                return {**data, "scc": scc_dict_b[surface_b["id"]]}
+
+            def list_filter(self, context_item, data=None):
+                subsurface_b = context_item.baseline
+                return (
```
charliepnnl, line 215: `subsurface_b` guaranteed to not be None, so no need to check here.

rct229/rules/section5/section5rule24.py
```diff
@@ -0,0 +1,259 @@
...
+    class BuildingRule(RuleDefinitionListIndexedBase):
+        def __init__(self):
+            super(Section5Rule24.BuildingRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                each_rule=Section5Rule24.BuildingRule.AboveGradeWallRule(),
+                index_rmr="baseline",
+                list_path="$..surfaces[*]",
+            )
+
+        def manual_check_required(self, context, calc_vals=None, data=None):
```
charliepnnl, line 68: By my reading of the RDS, we need to set a `manual_review_flag` in `create_data()` rather than have a `manual_check_required()` method. The manual check is required at the subsurface level rather than the building level.

rct229/rules/section5/section5rule24.py
```diff
@@ -0,0 +1,259 @@
...
+                    == table_G34_lookup(
+                        climate,
+                        SCC.EXTERIOR_NON_RESIDENTIAL,
+                        "Vertical Glazing",
+                        wwr=0.1,
+                    )
+                )
+            )
+
+        def create_data(self, context, data=None):
```
charliepnnl, line 127: See my comment above. I think we need to set a `manual_review_flag` here.

rct229/rules/section5/section5rule24.py
```diff
@@ -0,0 +1,259 @@
...
+                    ).__init__(
+                        rmrs_used=UserBaselineProposedVals(False, True, False),
+                    )
+
+                def get_calc_vals(self, context, data=None):
+                    subsurface_b = context.baseline
+                    return {
+                        "subsurface_u_factor": subsurface_b["u_factor"],
+                    }
+
```
charliepnnl, line 235: I believe we need a `manual_check_required()` method here with:  
```
return data["manual_review_flag"] and scc == SCC.EXTERERIOR_MIXED
```

rct229/rules/section5/section5rule24.py
```diff
@@ -0,0 +1,259 @@
...
+                    if scc == SCC.EXTERIOR_MIXED:
+                        target_u_factor = data["target_u_factor_mix"]
+                    elif scc == SCC.EXTERIOR_RESIDENTIAL:
+                        target_u_factor = data["target_u_factor_res"]
+                    elif scc == SCC.EXTERIOR_NON_RESIDENTIAL:
+                        target_u_factor = data["target_u_factor_nonres"]
+                    elif scc == SCC.SEMI_EXTERIOR:
+                        target_u_factor = data["target_u_factor_semiheated"]
+                    else:
+                        assert (
```
charliepnnl, line 248: I think it might be slightly better to simplify this to:
```
else:
    # Getting here implies a bug in the Rule logic
    assert False, f"No matching surface category for: {scc}"
```

rct229/rules/section5/section5rule24.py
```diff
@@ -0,0 +1,259 @@
...
+
+                def get_calc_vals(self, context, data=None):
+                    subsurface_b = context.baseline
+                    return {
+                        "subsurface_u_factor": subsurface_b["u_factor"],
+                    }
+
+                def rule_check(self, context, calc_vals=None, data=None):
+                    scc = data["scc"]
+                    target_u_factor = 0.0
```
charliepnnl, line 238: Ii might slightly prefer to initialize this to None so that we'll get an exception if we somehow get to line 259 without truly setting this value.

charliepnnl (approved):
Approved with some comments that really need to be addressed.

PR #465, opened 2022-05-13
First PR for Rule5-26

weilixu:
Complete Rule 5-26
Add shgc to Table 34

Similar issue as Rule 5-24


yunjoonjung-PNNL (approved):

rct229/rules/section5/section5rule26.py
```diff
@@ -0,0 +1,274 @@
...
+            # get standard code data
+            target_shgc_mix = (
+                table_G34_lookup(
+                    climate_zone,
+                    SCC.EXTERIOR_RESIDENTIAL,
+                    "VERTICAL GLAZING",
+                    wwr=bldg_scc_wwr_ratio[SCC.EXTERIOR_MIXED],
+                )["solar_heat_gain_coefficient"]
+                if bldg_scc_wwr_ratio[SCC.EXTERIOR_MIXED] > 0
+                else 0.0
```
charliepnnl, line 152: I guess the `0.0` value is saying that a wall with no fenestration has a solar heat gain coefficient of `0.0`.  That makes sense.

Should this logic be added to the RDS?

rct229/rules/section5/section5rule26.py
```diff
@@ -0,0 +1,274 @@
...
+                table_G34_lookup(
+                    climate_zone,
+                    SCC.EXTERIOR_RESIDENTIAL,
+                    "VERTICAL GLAZING",
+                    wwr=bldg_scc_wwr_ratio[SCC.EXTERIOR_RESIDENTIAL],
+                )["solar_heat_gain_coefficient"]
+                if bldg_scc_wwr_ratio[SCC.EXTERIOR_RESIDENTIAL] > 0
+                else 0.0
+            )
+            target_shgc_nores = (
```
charliepnnl, line 164: It doesn't matter much since this variable name does not appear in the report, but I believe we have used `nonres` elsewhere instead of `nores`.

rct229/rules/section5/section5rule26.py
```diff
@@ -0,0 +1,274 @@
...
+            def __init__(self):
+                super(Section5Rule26.BuildingRule.AboveGradeWallRule, self).__init__(
+                    rmrs_used=UserBaselineProposedVals(False, True, False),
+                    each_rule=Section5Rule26.BuildingRule.AboveGradeWallRule.SubsurfaceRule(),
+                    index_rmr="baseline",
+                    list_path="subsurfaces[*]",
+                    required_fields={
+                        "$.subsurfaces[*]": [
+                            "classification",
+                            "glazed_area",
```
charliepnnl, line 214: As I recall, in one or more of the ruleset functions, we assumed a missing `glazed_area` or `opaque_area` was zero.  We should think about whether that is what we really want and be consistent.

charliepnnl (approved):
Approved with one comment about a misspelled variable, one comment about the RDS, and a few minor, stylistic comments.

rct229/rules/section5/section5rule26.py
weilixu: I actually think None may work better in this case to avoid confusions in the report.

PR #466, opened 2022-05-13
Rs/yj/rule5 49

yunjoonjung-PNNL:
I'd appreciate it if you could review the Rule 5-49. Thank you!


rct229/rules/section5/section5rule49.py
```diff
@@ -0,0 +1,146 @@
...
+)
+from rct229.ruleset_functions.get_zone_conditioning_category_dict import (
+    ZoneConditioningCategory as ZCC,
+    get_zone_conditioning_category_dict,
+)
+
+from rct229.utils.jsonpath_utils import find_all
+from rct229.utils.pint_utils import ONE, ZERO
+from rct229.utils.std_comparisons import std_equal
+
```
charliepnnl, line 18: We should define the constant
```
TARGET_AIR_LEAKAGE_COEFF = 0.6 * ureg("CFM / foot**2")
```
and use this in place of the 0.6.

rct229/rules/section5/section5rule49.py
```diff
@@ -0,0 +1,146 @@
...
+                **data,
+                "scc_dict_p": get_surface_conditioning_category_dict(
+                    data["climate_zone"], building_p
+                ),
+                "zcc_dict_p": get_zone_conditioning_category_dict(
+                    data["climate_zone"], building_p
+                ),
+            }
+
+        class ZoneRule(RuleDefinitionBase):  # change to the zoneRule
```
charliepnnl, line 63: Because the `building_total_measured_air_leakage_rate` needs to accumulate at the building level, and because the `empty_measrued_air_leakage_rate_flow_flag` needs to be set at the building level, I think we do not want a `ZoneRule` for this one. Instead, we should loop of the zones inside `get_calc_vals` in BuildingRule.

charliepnnl (changes requested):
I think we should loop over the zones in the BuildingRule instead of using a ZoneRule.

In the RDS Rule Description, `(I75Pa)` should be: `(<sub>75Pa</sub>)` so the 75Pa shows up as a subscript.

rct229/rules/section5/section5rule49.py
```diff
@@ -0,0 +1,132 @@
...
+            )
+
+            building_total_envelope_area = 0.0
+            building_total_air_leakage_rate = 0.0
+            building_total_measured_air_leakage_rate = 0.0
+            empty_measured_air_leakage_rate_flow_flag = False
+
+            building_total_envelope_area = sum(
+                [
+                    surface.get("area")
```
charliepnnl, line 70: I think you need to give a default for the area:
```
surface.get("area", ZERO.AREA)
```

rct229/rules/section5/section5rule49.py
```diff
@@ -0,0 +1,132 @@
...
+            )
+
+            for zones in find_all("$..zones[*]", building_p):
+                if zones["infiltration"].get("measured_air_leakage_rate"):
+                    building_total_measured_air_leakage_rate += zones["infiltration"][
+                        "measured_air_leakage_rate"
+                    ]
+                else:
+                    empty_measured_air_leakage_rate_flow_flag = True
+                    if zcc_dict_p[zones["id"]] in [
```
charliepnnl, line 84: I think this if statement belongs above the if statement on line 78.

rct229/rules/section5/section5rule49.py
```diff
@@ -0,0 +1,132 @@
...
+                building_total_air_leakage_rate
+                != 0.112 * target_air_leakage_rate_75pa_p
+                and empty_measured_air_leakage_rate_flow_flag == False
+                and std_equal(
+                    building_total_air_leakage_rate,
+                    0.112 * building_total_measured_air_leakage_rate,
+                )
+            )
+
+        def get_fail_msg(self, context, calc_vals=None, data=None):
```
charliepnnl, line 131: Since this is a fixed message, you can put it in the class initializer as 
```
fail_msg = " ...",
```

rct229/rules/section5/section5rule49.py
```diff
@@ -0,0 +1,132 @@
...
+                TARGET_AIR_LEAKAGE_COEFF
+            ) * building_total_envelope_area
+
+            return {
+                "building_total_air_leakage_rate": building_total_air_leakage_rate,
+                "building_total_measured_air_leakage_rate": building_total_measured_air_leakage_rate,
+                "target_air_leakage_rate_75pa_p": target_air_leakage_rate_75pa_p,
+                "empty_measured_air_leakage_rate_flow_flag": empty_measured_air_leakage_rate_flow_flag,
+            }
+
```
charliepnnl, line 103: I think we need a `manual_check_required` method that checks for Case 2 in the Rule Assertion.

charliepnnl (changes requested):

rct229/rules/section5/section5rule49.py
```diff
@@ -0,0 +1,132 @@
...
+                [
+                    surface.get("area")
+                    for surface in find_all("$..surfaces[*]", building_p)
+                    if scc_dict_p[surface["id"]] != SCC.UNREGULATED
+                ],
+                ZERO.AREA,
+            )
+
+            for zones in find_all("$..zones[*]", building_p):
+                if zones["infiltration"].get("measured_air_leakage_rate"):
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Instead, I used ```measured_air_leakage_rate = zone["infiltration"].get(
                    "measured_air_leakage_rate"
                )```
because there could be a case when ```measured_air_leakage_rate``` doesn't exist and this case caused an error.  I'd appreciate it if you would review this when I re-request this PR review.

rct229/rules/section5/section5rule49.py
```diff
@@ -0,0 +1,145 @@
...
+                building_total_air_leakage_rate
+                != 0.112 * target_air_leakage_rate_75pa_p
+                and empty_measured_air_leakage_rate_flow_flag == False
+                and std_equal(
+                    building_total_air_leakage_rate,
+                    0.112 * building_total_measured_air_leakage_rate,
+                )
+            )
+
+        def manual_check_required(self, context, calc_vals=None, data=None):
```
charliepnnl, line 130: Since `manual_check_required()` will be executed before `rule_check()`, it would be nicer to place it above `rule_check()` in this file. That makes it easier to keep the correct logic flow in mind.

rct229/rules/section5/section5rule49.py
```diff
@@ -0,0 +1,145 @@
...
+            rmrs_used=UserBaselineProposedVals(False, False, True),
+            required_fields={
+                "$": ["weather"],
+                "weather": ["climate_zone"],
+            },
+            each_rule=Section5Rule49.BuildingRule(),
+            index_rmr="proposed",
+            id="5-49",
+            description="The proposed air leakage rate of the building envelope (I75Pa) at a fixed building pressure differential of 0.3 in. of water shall be 0.6 cfm/ft2 for buildings providing verification in accordance with Section 5.9.1.2. The air leakage rate of the building envelope shall be converted to appropriate units for the simulation program using one of the methods in Section G3.1.1.4. Exceptions: When whole-building air leakage testing, in accordance with Section 5.4.3.1.1, is specified during design and completed after construction, the proposed design air leakage rate of the building envelope shall be as measured..",
+            list_path="ruleset_model_instances[0].buildings[*]",
```
charliepnnl, line 35: I think we need a `manual_check_required_msg` field.

charliepnnl (approved):
Approved with comments that should be addressed.
@weilixu This is ready for your review.

weilixu (approved):

weilixu:
@yunjoonjung-PNNL  need to resolve the conflict for the merge

PR #467, opened 2022-05-17
Updating lighting JSON to 0.0.12

jugonzal07:
Could not resolve issue with 6-8. The `status_type` parameter under `spaces` would not pass schema validation for "NEW" and it's unclear why.


weilixu:
I will merge the test cases into the develop except those section 6 are not tested - I will create a issue for section 6 while RT team is developing the rule tests.

weilixu:
I will merge the test cases into the develop except those section 6 are not tested - I will create a issue for section 6 while RT team is developing the rule tests.

PR #469, opened 2022-05-26
Rs/wx/rule6 3

weilixu:


rct229/rules/section6/section6rule3.py
```diff
@@ -0,0 +1,61 @@
...
+GUEST_ROOM = schema_enums["LightingSpaceType2019ASHRAE901TG37"].GUEST_ROOM.name
+DORMITORY_LIVING_QUARTERS = schema_enums["LightingSpaceType2019ASHRAE901TG37"].DORMITORY_LIVING_QUARTERS.name
+DWELLING_UNITS = schema_enums["LightingSpaceType2019ASHRAE901TG37"].DWELLING_UNIT.name
+
+class Section6Rule3(RuleDefinitionListIndexedBase):
+    """Rule 3 of ASHRAE 90.1-2019 Appendix G Section 6 (Lighting)"""
+
+    def __init__(self):
+        super(Section6Rule3, self).__init__(
+            rmrs_used=UserBaselineProposedVals(True, True, False),
```
charliepnnl, line 15: This rule uses user and proposed, so this should be:
`rmrs_used=UserBaselineProposedVals(True, False, True),`

Likewise for the nested rules.

rct229/rules/section6/section6rule3.py
```diff
@@ -0,0 +1,61 @@
...
+
+        def get_calc_vals(self, context, data=None):
+            space_p = context.proposed
+            space_u = context.user
+
+            # get allowance
+            lighting_power_allowance_p = table_9_6_1_lookup(space_p["lighting_space_type"])
+
+            # sum of space_p lighting power density
+            interior_lighting_space_p = sum(interior_lighting["power_per_area"] for interior_lighting in space_p["interior_lighting"])
```
charliepnnl, line 46: Is there a reason not to use the same variable names for these two as in the RDS?  The RDS uses `space_lighting_power_per_area_p`, for example.

charliepnnl (approved):
Approved with minor comments.

rct229/rules/section6/section6rule3.py
weilixu: Updated.

PR #472, opened 2022-06-01
Add 5-47

weilixu:
Completed 5-47


yunjoonjung-PNNL (approved):

rct229/rules/section5/section5rule47.py
```diff
@@ -0,0 +1,87 @@
...
+
+    class BuildingRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section5Rule47.BuildingRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, False, True),
+                required_fields={"$..zones[*]": ["surfaces"]},
+            )
+
+        def get_calc_vals(self, context, data=None):
+            building_p = context.proposed
```
charliepnnl, line 44: Only the baseline is used in this rule, so all "p" subscripts should be "b"s.

rct229/rules/section5/section5rule47.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+
+    def create_data(self, context, data=None):
+        rmr_proposed = context.proposed
+        return {"climate_zone": rmr_proposed["weather"]["climate_zone"]}
+
+    class BuildingRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section5Rule47.BuildingRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, False, True),
```
charliepnnl, line 39: This rule only uses the baseline, so the arguments should be `(False, True, False)`.

charliepnnl (approved):
Approved with comments.

rct229/rules/section5/section5rule47.py
weilixu: Not sure why I did proposed...you are correct, it should baseline.

PR #474, opened 2022-06-03
Rs/yj/rule6-7 (Previously 6-12)

yunjoonjung-PNNL:
I'd appreciate it if you reviewed this PR. I also got the testing JSON file fixed in this PR. Thank you.


charliepnnl:
I find the RDS for Rule6-12 in this branch but not in develop.  We seem to have lost some RDS files from develop.

rct229/rules/section6/section6rule12.py
```diff
@@ -0,0 +1,95 @@
...
+    """Rule 12 of ASHRAE 90.1-2019 Appendix G Section 6 (Lighting)"""
+
+    def __init__(self):
+        super(Section6Rule12, self).__init__(
+            rmrs_used=UserBaselineProposedVals(False, False, True),
+            each_rule=Section6Rule12.BuildingRule(),
+            index_rmr="proposed",
+            id="6-12",
+            description="User building is modeled with daylighting controls directly or through schedule adjustments.",
+            rmr_context="ruleset_model_instances/0/buildings",
```
charliepnnl, line 23: The RDS rule assertion seems to apply to each zone in the RMD. So I think we need to use 
`list_path="ruleset_model_instances[0]..zones[*]`
and create a `ZoneRule` instead of a `BuildingRule`.

charliepnnl (changes requested):
I think this needs a ZoneRule instead of a BuildingRule.

rct229/rules/section6/section6rule12.py
```diff
@@ -0,0 +1,102 @@
...
+                has_daylight_control_flag = False
+                for subsurface in find_all(
+                    "$..surfaces[*].subsurfaces[?classification != 'DOOR']", building_p
+                ):
+                    daylight_flag_u = True
+
+                interior_lighting_u = False
+                for lighting in find_all("$..spaces[*].interior_lighting", building_p):
+                    interior_lighting_u = lighting[0][
+                        "are_schedules_used_for_modeling_daylighting_control"
```
charliepnnl, line 66: @weilixu This is a problem. `are_schedules_used_for_modeling_daylighting_control` is not a property of `InteriorLighting` in schema version 0.0.12, so we won't be able to test this rule with schema-valid test json.  I don't know when it was introduced, but I see that it is in the current version 0.0.17.

rct229/rules/section6/section6rule12.py
```diff
@@ -0,0 +1,102 @@
...
+                daylight_flag_u = False
+                has_daylight_control_flag = False
+                for subsurface in find_all(
+                    "$..surfaces[*].subsurfaces[?classification != 'DOOR']", building_p
+                ):
+                    daylight_flag_u = True
+
+                interior_lighting_u = False
+                for lighting in find_all("$..spaces[*].interior_lighting", building_p):
+                    interior_lighting_u = lighting[0][
```
charliepnnl, line 65: I don't understand the purpose of defining `interior_lighting_u` here.

rct229/rules/section6/section6rule12.py
```diff
@@ -0,0 +1,102 @@
...
+                        "are_schedules_used_for_modeling_daylighting_control"
+                    ]
+                    if lighting[0]["daylighting_control_type"] != "NONE":
```
charliepnnl, line 68: I don't think the `[0]` is needed here. `lighting` is an instance of `InteriorLighting`, which has a `daylighting_control_type` property.

rct229/rules/section6/section6rule12.py
```diff
@@ -0,0 +1,102 @@
...
+                        has_daylight_control_flag = True
+
+                return {
+                    "daylight_flag_u": daylight_flag_u,
+                    "has_daylight_control_flag": has_daylight_control_flag,
+                    "interior_lighting_u": interior_lighting_u,
```
charliepnnl, line 74: @weilixu  There is a problem with the RDS. My guess is that they mean to have another flag such as `schedules_used_for_modeling_daylighting_flag` that is set if any space has an interior_lighting with `are_schedules_used_for_modeling_daylighting_control` property set to `True`. This flag would then be used in the Rule Assertion Cases 1 and 2.

rct229/rules/section6/section6rule12.py
```diff
@@ -0,0 +1,102 @@
...
+                }
+
+            def manual_check_required(self, context, calc_vals=None, data=None):
+                daylight_flag_u = calc_vals["daylight_flag_u"]
+
+                return daylight_flag_u
```
charliepnnl, line 80: I think this is more complicated in the RDS. It involves both flags, and probably the third one I mentioned in my comment above.

rct229/rules/section6/section6rule12.py
```diff
@@ -0,0 +1,102 @@
...
+                    else:
+                        manual_check_msg = MSG_WARN_NO_DAYLIGHT
+
+                return manual_check_msg
+
+            def rule_check(self, context, calc_vals, data=None):
+                daylight_flag_u = calc_vals["daylight_flag_u"]
+                has_daylight_control_flag = calc_vals["has_daylight_control_flag"]
+
+                return not daylight_flag_u and not has_daylight_control_flag
```
charliepnnl, line 102: I think we need a `get_fail_msg()` method defined to provide the message from Case3.

charliepnnl (changes requested):

rct229/rules/section6/section6rule12.py
```diff
@@ -0,0 +1,102 @@
...
+class Section6Rule12(RuleDefinitionListIndexedBase):
+    """Rule 12 of ASHRAE 90.1-2019 Appendix G Section 6 (Lighting)"""
+
+    def __init__(self):
+        super(Section6Rule12, self).__init__(
+            rmrs_used=UserBaselineProposedVals(False, False, True),
+            each_rule=Section6Rule12.BuildingRule(),
+            index_rmr="proposed",
+            id="6-12",
+            description="User building is modeled with daylighting controls directly or through schedule adjustments.",
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: I just found that the RDS was updated. Now the RDS has "Proposed" instead of "User".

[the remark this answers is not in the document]
yunjoonjung-PNNL: Fixed. I appreciate this finding.

PR #482, opened 2022-06-15
Rs/wx/rule6 9 previous 6 14)

weilixu:
Complete 6-9 (previously 6-14)


rct229/rules/section6/section6rule9.py
```diff
@@ -0,0 +1,135 @@
...
+                        return f"Space models at least one daylight control using schedule. Verify if other programmable lighting control is modeled correctly using schedule. Lighting schedule EFLH in P-RMD is {elfh_differeces} of that in B-RMD"
+
+                    def rule_check(self, context, calc_vals=None, data=None):
+                        total_hours_compared = calc_vals["total_hours_compared"]
+                        total_hours_matched = calc_vals["total_hours_matched"]
+                        return total_hours_matched == total_hours_compared
+
+                    def get_fail_msg(self, context, calc_vals=None, data=None):
+                        eflh_difference = calc_vals["eflh_difference"]
+                        return f"Space lighting schedule EFLH in P-RMD is {eflh_difference} of that in B-RMD"
```
yunjoonjung-PNNL, line 135: period(.) is missing here. Could you also check if we need to update the ```B-RMR``` to ```B-RMD``` in the RDS?

yunjoonjung-PNNL (approved):

rct229/rules/section6/section6rule9.py
```diff
@@ -0,0 +1,135 @@
+from rct229.rule_engine.rule_base import RuleDefinitionListIndexedBase, RuleDefinitionBase
+from rct229.rule_engine.user_baseline_proposed_vals import UserBaselineProposedVals
+from rct229.ruleset_functions.compare_schedules import compare_schedules
+from rct229.ruleset_functions.normalize_interior_lighting_schedules import normalize_interior_lighting_schedules
+from rct229.utils.assertions import getattr_
+from rct229.utils.jsonpath_utils import find_all, find_one_with_field_value
+from rct229.utils.pint_utils import ZERO, pint_sum
```
charliepnnl, line 7: We can use `sum` in place of `pint_sum`.

rct229/rules/section6/section6rule9.py
```diff
@@ -0,0 +1,135 @@
...
+import functools
+
+FLOOR_AREA_LIMIT = 5000 # square foot
```
charliepnnl, line 10: Should be `5000 * ureg('ft2')`.

rct229/rules/section6/section6rule9.py
```diff
@@ -0,0 +1,135 @@
...
+            each_rule=Section6Rule9.RuleSetModelInstanceRule(),
+            index_rmr="proposed",
+            id="6-9",
+            description="Proposed building is modeled with other programmable lighting controls through a 10% schedule reduction in buildings less than 5,000sq.ft.",
+            list_path="ruleset_model_instances[0]",
+            required_fields={"$": ["calendar"], "calendar": ["is_leap_year"]},
+        )
+
+    def create_data(self, context, data=None):
+        rmd_p = context.proposed
```
charliepnnl, line 28: I have the same comment here as I did for Rule 6-8. What level of the json file is an RMD?

rct229/rules/section6/section6rule9.py
```diff
@@ -0,0 +1,135 @@
...
+                        each_rule=Section6Rule9.RuleSetModelInstanceRule.BuildingRule.ZoneRule.SpaceRule(),
+                        index_rmr="proposed",
+                        list_path="spaces[*]",
+                    )
+
+                def create_data(self, context, data=None):
+                    zone_p = context.proposed
+                    return {
+                        **data,
+                        "avg_space_height": zone_p.get("volume", 0.0) / pint_sum(
```
charliepnnl, line 84: As with Rule6-8, perhaps we should create a new ruleset function, `get_avg_zone_height()`.

rct229/rules/section6/section6rule9.py
```diff
@@ -0,0 +1,135 @@
...
+                building_p = context.proposed
+                return pint_sum(find_all("spaces[*].floor_area", building_p), ZERO.AREA) >= FLOOR_AREA_LIMIT
+
+            def create_data(self, context, data=None):
+                building_p = context.proposed
+                schedules_p = data["schedules_p"]
+                return {
+                    **data,
+                    "building_open_schedule_p":  getattr_(
+                        find_one_with_field_value("$", "id", building_p["building_open_schedule"], schedules_p),
```
charliepnnl, line 67: As with Rule6-8, I think we need to use `find_exactly_one_with_field_value()` here.

rct229/rules/section6/section6rule9.py
```diff
@@ -0,0 +1,135 @@
...
+
+                def create_data(self, context, data=None):
+                    zone_p = context.proposed
+                    return {
+                        **data,
+                        "avg_space_height": zone_p.get("volume", 0.0) / pint_sum(
+                            find_all("spaces[*].floor_area", zone_p), ZERO.AREA)
+                    }
+
+                def list_filter(self, context_item, data=None):
```
charliepnnl, line 88: This doesn't seem to be in the RDS. How would I know to make this restriction?

Since we are indexing on proposed, should the check be for `space_p`?
Or do you want `space_b.get("interior_lighting") and space_p.get("interior_lighting")?

The second argument of `get` defaults to `None`, so the `None` is optional here.

rct229/rules/section6/section6rule9.py
```diff
@@ -0,0 +1,135 @@
...
+                        avg_space_height = data["avg_space_height"]
+                        schedules_b = data["schedules_b"]
+                        schedules_p = data["schedules_p"]
+                        building_open_schedule_p = data["building_open_schedule_p"]
+
+                        normalized_schedule_b = normalize_interior_lighting_schedules(space_b, avg_space_height, schedules_b)
+                        normalized_schedule_p = normalize_interior_lighting_schedules(space_p, avg_space_height, schedules_p)
+                        schedule_comparison_result_dictionary = compare_schedules(normalized_schedule_p, normalized_schedule_b,
+                                                                       building_open_schedule_p, 1.0)
+                        daylight_control = functools.reduce(lambda daylight_control_a, daylight_control_b: daylight_control_a | daylight_control_b, find_all("interior_lighting[*].are_schedules_used_for_modeling_daylighting_control", space_p))
```
charliepnnl, line 111: I appreciate the use of functools.  I am always looking for good uses for it.
In this case, however, I think the logic could be simplified considerably:
```
daylight_control = any(
    find_all(
        "interior_lighting[*].are_schedules_used_for_modeling_daylighting_control", 
        space_p
    )
)
```

charliepnnl (approved):
Approved with comments.

rct229/rules/section6/section6rule9.py
weilixu: My 2 cents was spaces with no interior lighting should be skipped for processing, but I just realize the normalized schedule would take care of this scenario. I will remove this filter.

rct229/rules/section6/section6rule9.py
weilixu: Great to offer this elegant and simple solution. I never knew this built-in function before. Thanks!

rct229/rules/section6/section6rule9.py
weilixu: update the RMI to rmi_b and rmi_p

PR #483, opened 2022-06-18
Clean up old lighting ruletests

jugonzal07:
We've updated all the numbering on ruletests and this branch would clear out all the old test JSONs and begin the process of introducing verified lighting tests


yunjoonjung-PNNL (approved):

yunjoonjung-PNNL:
Thanks for doing this Juan!

PR #484, opened 2022-06-18
Ruletest JSON 6-3

jugonzal07:
Updated rule 6-3 to line up with latest Rule numbering convention


yunjoonjung-PNNL (approved):
I checked that the 6-3 testing file includes all the required keys (e.g., ```lighting_space_type```, ```interior_lighting```, and ```power_per_area```).

PR #487, opened 2022-06-22
Add compare standard value ruleset function

weilixu:


rct229/ruleset_functions/compare_standard_val.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Returns
+    -------
+    bool
+        True if the comparison matches to the defined relationship or equal.
+    """
+    if ahj_ra_compare:
+        assert_(operator, "Need to provide operator (operator.lt, operator.gt, operator.le, operator.ge) for "
+                          "stringent code comparison")
+        return std_equal(std_val, val) or operator(val, std_val)
```
claperle, line 29: @weilixu Good morning! I thought we were also going to include the global tolerance in the function logic as well. Maybe I am missing it. 

Also, I have created RDSs that assume a different order of input parameters but the meaning of each is clearly explained in the RDS. Is this OK or do I need to go back and update it.

charliepnnl:
@weilixu  I have given this some more thought.  Here is an alternative approach wrt the function arguments:
```
def compare_standard_val(operator, val, std_val) -> bool:
```
I've eliminated the `ahj_ra_compare` argument because specifying the operator seems to have the same effect. In particular,  operator.eq can be used when equality is desired. Specifying the operator as the first argument has some advantages from a programming standpoint. For example, if desired, we could then easily create short-cut functions:
```
std_eq = partial(compare_standard_val(operator.eq))
std_lt = partial(compare_standard_val(operator.lt))
std_le = partial(compare_standard_val(operator.le))
std_gt = partial(compare_standard_val(operator.gt))
std_ge = partial(compare_standard_val(operator.ge))
```

So we could do a less than or equal to comparison as:
```
std_le(val, std_val)
```

weilixu:
> @weilixu I have given this some more thought. Here is an alternative approach wrt the function arguments:
> 
> ```
> def compare_standard_val(operator, val, std_val) -> bool:
> ```
> 
> I've eliminated the `ahj_ra_compare` argument because specifying the operator seems to have the same effect. In particular, operator.eq can be used when equality is desired. Specifying the operator as the first argument has some advantages from a programming standpoint. For example, if desired, we could then easily create short-cut functions:
> 
> ```
> std_eq = partial(compare_standard_val(operator.eq))
> std_lt = partial(compare_standard_val(operator.lt))
> std_le = partial(compare_standard_val(operator.le))
> std_gt = partial(compare_standard_val(operator.gt))
> std_ge = partial(compare_standard_val(operator.ge))
> ```
> 
> So we could do a less-than-or-equal-to comparison simply as:
> 
> ```
> std_le(val, std_val)
> ```
> 
> If we want the option of specifying the tolerance, I suggest putting it as the final argument and setting a default (the most common value) that can be overridden. So the function could be expanded to:
> 
> ```
> def compare_standard_val(operator, val, std_val, tol = DEFAULT_TOLERANCE) -> bool:
> ```

I think this solution is great that we actually avoid writing operators in the rule definition. I have two comments:

1. We will still need `ahj_ra_compare`. The idea is AHJ should determine whether to have stringent code check (ahj_ra_compare) or regular equality check. This means based on AHJ's preference, it could be equality check or less than, or greater than check.
2. Global tolerance shall be applied in either case. operator.eq does not seem to have tolerance option.

I modified to address the two issues
```
AHJ_RA_COMPARE = False
DEFAULT_PERCENT_TOLERANCE = 0.005

def compare_standard_val(val, std_val, operator) -> bool:
    if AHJ_RA_COMPARE:
        return std_eq(val, std_val) or operator(val, std_val)
    return std_eq(val, std_val)

std_lt = partial(compare_standard_val, operator=operator.lt)
std_le = partial(compare_standard_val, operator=operator.le)
std_gt = partial(compare_standard_val, operator=operator.gt)
std_ge = partial(compare_standard_val, operator=operator.ge)
```

rct229/ruleset_functions/compare_standard_val.py
```diff
@@ -0,0 +1,31 @@
+from rct229.utils.assertions import assert_
+from rct229.utils.std_comparisons import std_equal
+
```
yunjoonjung-PNNL, line 3: Isn't ```import operator``` required because ```operator(val, std_val)``` part in line 29? I also see ```import operator``` in ```rct229/ruleset_functions/compare_standard_val_test.py``` file. I'd appreciate it if you could double-check this.

yunjoonjung-PNNL (approved):

rct229/ruleset_functions/compare_standard_val.py
weilixu: I think the `operator` will need to be imported once I updated code following my previous comments. I kind of like the way it was implemented to avoid importing operators in the rules. @charliepnnl @yunjoonjung-PNNL Let me know if you think the same way. I can update the function.

rct229/utils/std_comparisons.py
```diff
@@ -1,4 +1,5 @@
 DEFAULT_PERCENT_TOLERANCE = 0.05
+AHJ_RA_COMPARE = False
```
weilixu, line 2: Add comments to this line to explain what is AHJ_RA_COMPARE, and the meaning for True and False.

PR #497, opened 2022-06-28
Ruletest JSON 6-6

jugonzal07:
Added ruletest for Rule 6-6


jugonzal07:
Noticed the schedules in the test case descriptions dont line up with any schedules in the rest of the RMD-- not sure why they were added. Might need to investigate.

yunjoonjung-PNNL (approved):
Thank you for doing this Juan.

I have two comments and would appreciate if if you could double-check on these.

1) Would you be able to check the ``` test_description``` key? I beliebe ```Project has one building segment with one zone and one space. The building gross floor area is 1000 ft2. Baseline building  > 5000 ft2 is modeled with automatic shutoff controls``` and ```Baseline building  > 5000 ft2 is not modeled with automatic shutoff controls, contrary to rule requirement``` don't line up with the 6-6 rule.

2) 6-6-c doesn't have "interior_lighting" key. Could you double-check this?

jugonzal07:
These TCDs appear to be for 6-5. I believe the TCDs in the Lighting TCD draft spreadsheet have incorrect numbers and that is why this is confusing. Closing this PR as these test cases don't line up.

PR #499, opened 2022-06-29
Update normalize_interior_lighting function and compare schedule func…

weilixu:


weilixu:
This has been addressed in the other PR.

PR #500, opened 2022-06-29
Rs/ch/enumerations as classes

charliepnnl:
1. Model the enumerations with a class instead of Enum. This allows us to avoid the need to append `.name` to the enumeration when it is used.
2. Updated all the files that use the enumerations.
3. Removed unused imports of `schema_enum`.
4. Fixed `assert` syntax in `jasonpath_utils.py`.

### DIFF
--- a/rct229/data/schema_enums.py
+++ b/rct229/data/schema_enums.py
@@ -4,6 +4,30 @@ from os.path import dirname, join
 
 from jsonpath_ng import parse
 
+"""This module exports the dictionary schema_enums that provides access to the
+enumerations in the schema files.
+
+The keys of schema_enums are the names of the enumeration objects; each value
+is a class with an attribute for each item in the enumeration. The value
+of the attribute is the same as the attribute name.
+"""
+
+
+class _ListEnum:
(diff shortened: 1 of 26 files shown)


rct229/data/schema_enums.py
```diff
@@ -45,10 +63,10 @@ def print_schema_enums():
...
-        for e in schema_enums[key]:
-            print(f"    {e.name}: {e.value}")
+        for e in schema_enums[key].get_list():
+            print(f"    {e}")
         print()
 
 
 # Uncomment this for checking the enumerations after a schema change
-# print_schema_enums()
+print_schema_enums()
```
weilixu, line 72: Great solution. Thanks for working on it. 
One comment, should this line be commented out?

weilixu (approved):
Approve with a minor comment

yunjoonjung-PNNL (approved):
I aprrreciate this PR, Charlie! I went through your code and it looks perfect to me!

rct229/data/schema_enums.py
charliepnnl: Good catch!  Done.

PR #503, opened 2022-07-01
Rs/yj/rule6 7 (Previously 6-12)

yunjoonjung-PNNL:
I re-created 6-7 PR (previously 6-12) because I renamed the branch. I'd appreciate it if you reviewed this PR. Thank you!


rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,113 @@
...
+class Section6Rule7(RuleDefinitionListIndexedBase):
+    """Rule 7 of ASHRAE 90.1-2019 Appendix G Section 6 (Lighting)"""
+
+    def __init__(self):
+        super(Section6Rule7, self).__init__(
+            rmrs_used=UserBaselineProposedVals(False, False, True),
+            each_rule=Section6Rule7.BuildingRule(),
+            index_rmr="proposed",
+            id="6-7",
+            description="User building is modeled with daylighting controls directly or through schedule adjustments.",
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Fixed. Thanks for this finding.

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,113 @@
...
+    class BuildingRule(RuleDefinitionListIndexedBase):
+        def __init__(self):
+            super(Section6Rule7.BuildingRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, False, True),
+                each_rule=Section6Rule7.BuildingRule.ZoneRule(),
+                index_rmr="proposed",
+                list_path="$..zones[*]",
+            )
+
+        class ZoneRule(RuleDefinitionBase):
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: It totally makes sense! Thanks for this kind comment.

yunjoonjung-PNNL:
I reflected on all Weili's comments and would appreciate it if you could review this PR one more time. Thank you!

rct229/rules/section6/__init__.py
```diff
@@ -7,10 +7,13 @@
     "section6rule3",
     "section6rule4",
     "section6rule6",
+    "section6rule7",
     "section6rule8",
     "section6rule9",
 ]
 
+from rct229.rules.section6 import section6rule7
```
charliepnnl, line 15: Was this being used during development and should not be removed?

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,113 @@
...
+                    },
+                )
+
+            def get_calc_vals(self, context, data=None):
+                zone_p = context.proposed
+
+                daylight_flag_p = False
+                has_daylight_control_flag = False
+                for subsurface in find_all(
+                    "$..surfaces[?adjacent_to = 'EXTERIOR'].subsurfaces[?classification != 'DOOR']",
```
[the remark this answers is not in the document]
charliepnnl: I couldn't agree more!

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,131 @@
...
+                    },
+                    index_rmr="proposed",
+                    list_path="$..spaces[*]",
+                )
+
+            def create_data(self, context, data=None):
+                zone_p = context.proposed
+
+                return {
+                    "has_daylight_control_flag": len(
```
charliepnnl, line 54: According to the RDS, shouldn't this be `zone_pdaylight_flag_p`?
Otherwise, I love this part of the code.

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,131 @@
...
+                        find_all(
+                            "$..surfaces[?adjacent_to = 'EXTERIOR'].subsurfaces[?classification != 'DOOR']",
+                            zone_p,
+                        )
+                    )
+                    > 0
+                }
+
+            class SpaceRule(RuleDefinitionBase):
```
charliepnnl, line 63: It looks like the rule assertion cases in the RDS all refer to the zone. So I don't think we want a SpaceRule. Instead, I believe we want to:
- change the `create_data` above to `get_calc_vals`,
- move the calc_vals needed from `get_calc_vals` below to the `get_calc_vals` above at the zone level.

charliepnnl (changes requested):
I believe BulidingRule and SpaceRule are not needed.

rct229/rules/section6/__init__.py
yunjoonjung-PNNL: I don't know why this line was added here, but I really appreciate for pointing this out :).

rct229/rules/section6/section6rule7.py
yunjoonjung-PNNL: Correct! I just realized that ```zone_pdaylight_flag_p``` should be ```daylight_flag_p``` in the RDS. Hence, I changed ```has_daylight_control_flag``` to ```daylight_flag_p```. I appreciate for finding this.

yunjoonjung-PNNL:
I reflected on Charlie's comments and would appreciate it if you could review this PR one more time. Thank you!

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,94 @@
...
+                        "$..surfaces[?adjacent_to = 'EXTERIOR'].subsurfaces[?classification != 'DOOR']",
+                        zone_p,
+                    )
+                )
+                > 0
+            )
+
+            has_daylight_control_flag = False
+            for lighting in find_all("$..interior_lighting[*]", zone_p):
+                interior_lighting_p = getattr_(
```
charliepnnl, line 50: I'm finding the intent of the RDS a little hard to follow with respect to `interior_lighting_p.are_schedules_used_for_modeling_daylighting_control` that appears in the Rule Assertion Cases 1 and 2.

The problem is that `interior_lighting_p.are_schedules_used_for_modeling_daylighting_control` refers to a specific space within the zone, but the Rule Assertions refer to the zone as a whole.
Is `interior_lighting_p.are_schedules_used_for_modeling_daylighting_control` intended as another flag which should be set if this is True for ANY space in the zone?

The implementation here resets `interior_lighting_p` in each iteration of the loop. So on the last one is returned.

charliepnnl (changes requested):
I'm stuck on `interior_lighting_p.are_schedules_used_for_modeling_daylighting_control` - see my comment.

yunjoonjung-PNNL:
I updated the code and RDS based on the feedback in Issue #567. I'd appreciate it if you could review it one more time. Thank you!

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,94 @@
...
+                len(
+                    find_all(
+                        "$..surfaces[?adjacent_to = 'EXTERIOR'].subsurfaces[?classification != 'DOOR']",
+                        zone_p,
+                    )
+                )
+                > 0
+            )
+
+            has_daylight_control_flag = False
```
charliepnnl, line 48: Following the pattern for *daylight_flag_p* above, couldn't this be:
```
has_daylight_control_flag = (
        len(
            find_all(
                "$..spaces[*].interior_lighting[*][?daylighting_control_type != 'NONE']",
                zone_p,
            )
        )
        > 0
    )
```
and do something similar for *daylight_schedule_adjustment_flag*?

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,94 @@
...
+from rct229.rule_engine.rule_list_indexed_base import RuleDefinitionListIndexedBase
+from rct229.rule_engine.user_baseline_proposed_vals import UserBaselineProposedVals
+from rct229.utils.assertions import getattr_
+from rct229.utils.jsonpath_utils import find_all
+
+MSG_WARN_DAYLIGHT_NO_SCHEDULE = "Some of the spaces in zone are modeled with window(s) and/or skylight(s) and have daylighting controls modeled explicitly in the simulation tool. Verify that the mandatory lighting control requirements are met."
+MSG_WARN_DAYLIGHT = "Some of the spaces in zone are modeled with window(s) and/or skylight(s) and have daylighting controls modeled via schedule adjustment. Verify that the mandatory lighting control requirements are met, and that the supporting documentation is provided for the schedule adjustment."
+MSG_WARN_NO_DAYLIGHT = "Some of the spaces in zone are modeled with fenestration but no daylighting controls. The design must include mandatory daylighting controls unless any of the exceptions to 90.1 section 9.4.1.1€ apply."
+
+DAYLIGHT_CONTROL_TYPE = schema_enums["LightingDaylightingControlType"].NONE
```
charliepnnl, line 12: It might more self documenting to set 
```
DAYLIGHT_CONTROL_TYPE = schema_enums["LightingDaylightingControlType"]
```
here and then use DAYLIGHT_CONTROL_TYPE.NONE where this is used in the code.

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,94 @@
...
+                    "are_schedules_used_for_modeling_daylighting_control",
+                )
+
+                if lighting["daylighting_control_type"] != DAYLIGHT_CONTROL_TYPE:
+                    has_daylight_control_flag = True
+
+            return {
+                "daylight_flag_p": daylight_flag_p,
+                "has_daylight_control_flag": has_daylight_control_flag,
+                "interior_lighting_p": interior_lighting_p,
```
charliepnnl, line 62: I don't think *interior_lighting_p* is used in the Rule Assertions any more, so I don't think this should be added to calc_vals.
But we do need *daylight_schedule_adjustment_flag*.

charliepnnl (commented):
I think there is a mistake regarding *interior_lighting_p* vs *daylight_schedule_adjustment_flag* in *get_calc_vals*.

rct229/rules/section6/section6rule7.py
yunjoonjung-PNNL: This is brilliant. I appreciate for letting me know about this :).

rct229/rules/section6/section6rule7.py
yunjoonjung-PNNL: Fixed. Thanks for finding this and I should've been more meticulous.

rct229/rules/section6/section6rule7.py
yunjoonjung-PNNL: Great point! This part was removed because we now use ```find_all``` function instead of for-loop.

charliepnnl (approved):
Looks good to me

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,107 @@
...
+                rmrs_used=UserBaselineProposedVals(False, False, True),
+            )
+
+        def get_calc_vals(self, context, data=None):
+            zone_p = context.proposed
+
+            daylight_flag_p = (
+                len(
+                    find_all(
+                        "$..surfaces[?adjacent_to = 'EXTERIOR'].subsurfaces[?classification != 'DOOR']",
```
weilixu, line 39: I will prefer to use enum instead of string here. See [5-24](https://github.com/pnnl/ruleset-checking-tool/blob/develop/rct229/rules/section5/section5rule24.py) as an example.
You can replace both `adjacent_to` and `classification' to the enum.

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,107 @@
...
+                        zone_p,
+                    )
+                )
+                > 0
+            )
+
+            has_daylight_control_flag = (
+                len(
+                    find_all(
+                        "$..spaces[*].interior_lighting[?daylighting_control_type!='NONE']",
```
weilixu, line 49: Use enum if possible.

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,107 @@
...
+            )
+
+            daylight_schedule_adjustment_flag = (
+                len(
+                    find_all(
+                        "$..spaces[*].interior_lighting[?are_schedules_used_for_modeling_daylighting_control == true]",
+                        zone_p,
+                    )
+                )
+                > 0
```
weilixu, line 63: Minor comment
You may also use any() function here (unless the `are_schedules_used_for_modeling_daylighting_control` is equal to a string, I don't quite remember how it is in the schema now) - this is a personal preference choice
```python
any(find_all("$..spaces[*].interior_lighting[*].are_schedules_used_for_modeling_daylighting_control", zone_p))
```

rct229/rules/section6/section6rule7.py
```diff
@@ -0,0 +1,107 @@
...
+                "daylight_schedule_adjustment_flag"
+            ]
+
+            if daylight_flag_p and has_daylight_control_flag:
+                if daylight_schedule_adjustment_flag:
+                    manual_check_msg = MSG_WARN_DAYLIGHT
+                else:
+                    manual_check_msg = MSG_WARN_DAYLIGHT_NO_SCHEDULE
+
+            return manual_check_msg
```
weilixu, line 91: We tend to add the final logic to the return statement.

You do not need to worry about `if daylight_flag_p and has_daylight_control_flag:` because the `get_manual_check_required_msg` will only be called when `manual_check_required` is True.
This means, the `daylight_flag_p = True` and `has_daylight_control_flag = True` is the precondition for the `get_manual_check_required_msg` function in this case.

So you can rewrite this function to:
```python
def get_manual_check_required_msg(self, context, calc_vals=None, data=None):
    daylight_schedule_adjustment_flag = calc_vals[
                "daylight_schedule_adjustment_flag"
            ]
    return MSG_WARN_DAYLIGHT if  daylight_schedule_adjustment_flag else MSG_WARN_DAYLIGHT_NO_SCHEDULE
```

weilixu (approved):
Approved with minor comments for styling and coding preferences

rct229/rules/section6/section6rule7.py
yunjoonjung-PNNL: Great point! I learned new thing :).

rct229/rules/section6/section6rule7.py
yunjoonjung-PNNL: Great point. I reflected this in the code.

rct229/rules/section6/section6rule7.py
yunjoonjung-PNNL: Great point! Fixed.

rct229/rules/section6/section6rule7.py
yunjoonjung-PNNL: Fixed.

yunjoonjung-PNNL: Fixed.

rct229/rules/section6/section6rule7.py
yunjoonjung-PNNL: Great point! Fixed.

PR #505, opened 2022-07-03
RS/CH/split-rule_base

charliepnnl:
- Split the classes in rct229/rule_engine/rule_base.py into separate files.
- Refactored the import of RuleDefinitionListIndexedBase in all affected rule files.

The rule test run with these changes, although I notice that the tests for Section 6 have issues with missing "volume" elements.


yunjoonjung-PNNL (approved):
Thank you for doing this, Charlie! We have much more readable rulebase-related classes.

PR #506, opened 2022-07-04
RS/CH/update-black

charliepnnl:
- Updates the version of Black to 22.6.0.
- Updated README.md to reflect the new install steps (it is simpler now that Black is not longer a beta version).
- Developers should
  - Delete Pipfile.lock
  - Run `pipenv install --dev`
  - I suggest having your editor automatically  run black whenever you save a file
  - Before committing for a pull request, run `pipenv run isort .` and `pipenv run black .`

After all developers have updated to the new version of Black and all active PRs have been created using this new version, I will again set up git hub to automatically check isort and Black when creating a PR.

### DIFF
--- a/rct229/data_fns/table_G3_7_fns_test.py
+++ b/rct229/data_fns/table_G3_7_fns_test.py
@@ -18,27 +18,21 @@ watts_per_sqft = ureg("watt / foot**2")
 
 # Testing table_G3_7_lookup() ----------------------------------------
 def test__table_G3_7_lookup__with_w_per_ft_null():
-    assert (
-        table_G3_7_lookup(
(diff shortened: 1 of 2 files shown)


weilixu:
Are the file format changes because the latest black?

@charliepnnl Thanks for updating black! I think it is better send it out in email to notify everyone about the workflow update. Github comments seems not so effective.

charliepnnl:
> Are the file format changes because the latest black?
> 
> @charliepnnl Thanks for updating black! I think it is better send it out in email to notify everyone about the workflow update. Github comment seems not so effective.

Yes the file format changes were made by the new version of Black.

PR #507, opened 2022-07-05
Update 0.0 to pint ZERO and remove unncessary imports.

weilixu:
Update numeric 0.0 to Zero quantity class.

### DIFF
--- a/rct229/ruleset_functions/get_building_segment_skylight_roof_areas_dict.py
+++ b/rct229/ruleset_functions/get_building_segment_skylight_roof_areas_dict.py
@@ -12,10 +12,9 @@ from rct229.ruleset_functions.get_zone_conditioning_category_dict import (
 from rct229.ruleset_functions.get_zone_conditioning_category_dict import (
     get_zone_conditioning_category_dict,
 )
-from rct229.schema.config import ureg
 from rct229.utils.assertions import getattr_
 from rct229.utils.jsonpath_utils import find_all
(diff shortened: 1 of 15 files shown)


rct229/ruleset_functions/get_building_segment_skylight_roof_areas_dict.py
```diff
@@ -12,10 +12,9 @@
 from rct229.ruleset_functions.get_zone_conditioning_category_dict import (
     get_zone_conditioning_category_dict,
 )
-from rct229.schema.config import ureg
 from rct229.utils.assertions import getattr_
 from rct229.utils.jsonpath_utils import find_all
-from rct229.utils.pint_utils import ZERO, pint_sum
```
charliepnnl, line 18: I think you do want to use pint_sum because there should be a third argument for the default in the case that the find_all returns the empty list.  
```
sum([]) = 0
```
which is not a pint Quantity.

charliepnnl (approved):
Approved, but be careful to use `pint_sum` to provide a pint Quantity if the list is empty.

rct229/ruleset_functions/get_building_segment_skylight_roof_areas_dict.py
weilixu: that is correct. I deleted because the function did not use pint_sum. I will add using pint_sum in this case cause it is important to get the default as quantity object. Thanks for catching this.

PR #508, opened 2022-07-05
RS/CH/data_items

charliepnnl:
The easiest way to explain what I had in mind was to create a pull request.

The primary use case is adding climate_zone to the data dictionary. In this, and similar cases, we can include a line in the Rule class initializer such as 
```
data_items={"climate_zone": ("baseline", "weather/climate_zone")}
```
instead of overriding the `create_data()` method.
The way I have implemented this, you can either override `create_data` or include `data_items`, but not both.

Another aspect of this proposed change is that `data` is explicitly a dictionary. (I originally thought we might want it to be a list or other type of object, but we seem to have exclusively used it as a dictionary.)  Moreover, this change treats the return from `create_data` as new data only – that is automatically merged into the existing data dictionary.

If there is agreement that this change would be beneficial, I will adjust the rules accordingly.  The only rule I modified for this PR is Rule5-5, which is a good place to see how the change will be implemented.


weilixu:
I need some helps on understanding the pros and cons of this approach.
Here is what I understand:
The abstract function `create_data()` now have default implementation. We now just need to give a pointer (assign to `data_item`) instead of overriding the `create_data`

The pros I see is it is cleaner and the overall structure is simpler.
The cons? I think it might be harder for people to understand / read and harder for new developers to understand?

yunjoonjung-PNNL (approved):
I appreciate this suggestion and PR, Charlie. I believe this approach will make our source code much cleaner and reduce redundancy. I also agree with Weili's cons point mentioned in the above comment, but (I think) this addition doesn't complicate the source code much. I will leave this discussion to both of you and follow the decision. Thank you!

charliepnnl:
I have updated all the rules to take advantage of the new data_items initializer argument where possible.
I also removed any **data parts in the create_data methods since merging takes place automatically now.

weilixu (approved):
Looks good, thanks for updating all the rules with create_data function.

PR #512, opened 2022-07-07
Create are_all_terminal_fans_null.md

claperle:


docs/ruleset_functions/is_terminal_fan_null(terminal.id).md
```diff
@@ -0,0 +1,21 @@
+# is_terminal_fan_null   
+
+**Description:** Returns TRUE if the list of fans associated with the terminal unit is Null. It returns FALSE if the list of fans associated with the terminal unit is anything other than Null.
```
mkarpman, line 3: I wonder if this shouldn't be part of the baseline HVAC system type check, but rather checked in the RDSs that verify the baseline fan power value and that it is allocated between supply, exhaust and return fans in the same proportion in the baseline as in the proposed design. For example, what if a zone (e.g., bathroom) has an exhaust fan in the proposed design but is served by a multi-zone system in the baseline. How would this scenario be modeled?

Also, I think all sub-functions that apply to terminals should loop through all terminals associated with a given HVAC system.

supriyagoel: @mkarpman and @claperle I agree that a exhaust fan shouldnt be modeled as a terminal fan and the exhaust airflow should  be specified at the zone level.  However, this might be worth a discussion.
Other than that, this function looks good to me.

mkarpman: Function looks good to me too, but there seems to be a potential issue with schema as there seems to be no way to model zonal exhaust fan.

claperle: I think we would not send this function terminal units that we determined were exhaust fans if we end up going with option 1 in the end (if the working groups rejects our suggested change).

mkarpman (approved):

supriyagoel (approved):

PR #513, opened 2022-07-07
Create are_all_terminal_types_CAV.md

claperle:


docs/ruleset_functions/is_terminal_type_CAV(terminal.id).md
```diff
@@ -0,0 +1,21 @@
+# is_terminal_type_CAV
+
+**Description:** Returns TRUE if the terminal unit type is constant air volume (CAV). It returns FALSE if the terminal unit is a type other than constant air volume (CAV).   
+
```
mkarpman, line 4: So for a constant volume system such as PTAC, we are expecting Termina.type=CONSTANT_AIR_VOLUME and FanSystem.fan_control=CONSTANT? And for VAV, both should be variable?

claperle: I agree with your summary there. Let's ask group.

supriyagoel: I guess terminal type is slightly confusing. CAV, VAV seems to be terminal airflow control and radiant, FPFCU, TPFCU, baseboard seems to terminal equipment type.
I guess we should options for 'terminal airflow control', which could be constant, variable, none. we would use constant for PTAC, FPFCU, TPFCU.
we would use none for baseboard and radiant.

mkarpman (approved):

supriyagoel (approved):

PR #515, opened 2022-07-07
Create are_all_terminal_types_four_pipe.md

claperle:


docs/ruleset_functions/is_terminal_type_four_pipe(terminal.id).md
```diff
@@ -0,0 +1,21 @@
...
+- **terminal_b.id**: The id of the terminal unit to evaluate.  
+
+**Returns:**  
+- **is_terminal_type_four_pipe**: The function returns TRUE if the terminal unit type is FOUR_PIPE_FAN_COIL_UNIT. It returns FALSE if the terminal unit is a type other than FOUR_PIPE_FAN_COIL_UNIT.    
+ 
+**Function Call:** None  
+
+## Logic: 
+- Create an object for the terminal unit: `terminal_b = terminal_b.id`  
+- Check if the terminal unit type is FOUR_PIPE_FAN_COIL_UNIT: `if terminal_b.type == "FOUR_PIPE_FAN_COIL_UNIT": is_terminal_type_four_pipe = TRUE`
```
mkarpman, line 16: Should we additionally check that terminal has cooling source = chilled water loop and heating source = hot water loop? That would also affect the check for Systems 1a and 1c. 

Also, consider looping through all terminal units served by a given HVAC system.

claperle: I add this to system_1 function logic

supriyagoel: @mkarpman and @claperle I think it is correct to have the HW and CHW defined at the HVAC system level and not the terminal level. This is the way we do it for TSPR as well
If you think the correct answer is terminal, then what should the heating and cooling source at HVAC system level be?

supriyagoel (commented):
function looks good to me, I'll wait for @mkarpman 's response to my comment before approving.

docs/ruleset_functions/is_terminal_type_four_pipe(terminal.id).md
claperle: @supriyagoel  The answer to your question is yes. The master runs a function that get all of the terminals associated with the hvac system and then passes the list to the terminal related functions.

docs/ruleset_functions/are_all_terminal_types_four_pipe(hvac.id).md
```diff
@@ -0,0 +1,24 @@
+# are_all_terminal_types_four_pipe
```
claperle, line 1: Function related to how terminal type is defined.

mkarpman: Based on the latest decisions this function is obsolete.

PR #516, opened 2022-07-07
Create get_terminal_unit_for_SZ_baseline_HVAC(hvac_b.id).md

claperle:


docs/ruleset_functions/get_terminal_unit_for_SZ_baseline_HVAC(hvac_b.id).md
```diff
@@ -0,0 +1,25 @@
+# get_terminal_unit_for_SZ_baseline_HVAC
+
+**Description:** Get the terminal unit object associated with a single zone hvac system in the baseline model.  The function returns the terminal unit object associated with the hvac system ID input to the function. If there is more than one terminal unit associated with the HVAC system or no terminal units associated with the HVAC system then return Null.
+
```
mkarpman, line 4: I would consider refashioning this function so that it determins whether the system is SZ (has one zone with one terminal) or multizone (has more than one zone with each zone having one terminal). It would return dictinary of zones with the list of terminals for each zone.

claperle: I decided to take a different approach and just have it return a dictionary with HVAC IDs and noting whether the unit is single zone or multizone.

supriyagoel (approved):
function looks good to me though I am not sure on how this will be used in thr master function. Guess that will become clear with time

claperle:
This function is obsolete.

*Christina LaPerle* PE, BEMP, CEM, CBCP
________________
[image: image.png]
Phone 860.266.6220

PR #519, opened 2022-07-08
Create is_baseline_system_4(hvac.id).md

claperle:


supriyagoel (approved):

supriyagoel:
here we're not checking the supplementary heating type, I assume that would happen under a separate RDS for rule G3.1.3.1

mkarpman (approved):

PR #520, opened 2022-07-11
Update normalize schedule and compare schedule functions

weilixu:
Update the normalized interior lighting schedule and compare schedule.
Update the RDS for rule 6-5.
reformat rules and functions with black


docs/ruleset_functions/normalize_interior_lighting_schedules.md
```diff
@@ -5,6 +5,7 @@ Description: This function would determine a normalized interior lighting schedu
 
 Inputs:
   - **Space.data_element**: The space data element that needs to determine a normalized schedule, e.g. space.interior_lighting.
+  - **adjust_for_credit**: Flag to indicate whether the schedule normalization needs to add control credit adjustment.
```
mkarpman, line 8: Shouldn't the input be Space.ID and not Space.data_element, since we are looking for a normalized schedule representing all the interior lighting objects associated with a given space?

docs/ruleset_functions/normalize_interior_lighting_schedules.md
```diff
@@ -30,7 +31,7 @@ Logic:
 
     - Add lighting power per area to space total: `space_total_power_per_area += power_per_area`
 
-  - Check if lighting uses schedule to model occupancy control, get occupancy control credit: `if interior_lighting.are_schedules_used_for_modeling_occupancy_control: control_credit = data_lookup(table_G3_7, space.lighting_space_type, interior_lighting.occupancy_control_type)`
+  - Check if adjust_for_credit is true and lighting uses schedule to model occupancy control, get occupancy control credit: `if adjust_for_credit and interior_lighting.are_schedules_used_for_modeling_occupancy_control: control_credit = data_lookup(table_G3_7, space.lighting_space_type, interior_lighting.occupancy_control_type)`
 
   - Else, set occupancy control credit to 0: `else: control_credit = 0`
```
mkarpman, line 37: I think we also need to define bonus_adjustment variable =1.25 if InteriorLighting.occupancy_control_type=PARTIAL_AUTO_ON OR MANUAL ON, and =1 otherwise, and then in rows 40 and 44 update the forumal to hourly_use_per_area_array.append(hourly_value / (1 - control_credit***bonus_adjustment**) *....

I see now that your note 1 suggests to add column to Table 3.7 to account for bonus adjustment, but I am not sure how this will work because we need to know for each interior lighting object whether it has OS controls that qualify it for this additional credit.

docs/section6/Rule6-5.md
```diff
@@ -45,19 +45,19 @@
...
 
         - Get matching space in P_RMR: `space_p = match_data_element(P_RMR, Spaces, space_b.id)`  
 
-          - Get normalized space lighting schedule in P_RMR: `normalized_schedule_p = normalize_space_schedules(space_p.interior_lighting)`
+          - Get normalized space lighting schedule in P_RMR: `normalized_schedule_p = normalize_interior_lighting_schedules(space_p.interior_lighting, true)`
 
-        - Check if automatic shutoff control is modeled in space during building closed hours (i.e. if lighting schedule hourly value in B_RMR is equal to P_RMR during building closed hours): `schedule_comparison_result = compare_schedules(normalized_schedule_b, normalized_schedule_p, building_open_schedule_b, -111)`  
+        - Check if automatic shutoff control is modeled in space during building closed hours (i.e. if lighting schedule hourly value in B_RMR is equal to P_RMR during building closed hours): `schedule_comparison_result = compare_schedules(normalized_schedule_b, normalized_schedule_p, inverse(building_open_schedule_b))`
```
mkarpman, line 55: Perfect!

docs/section6/Rule6-5.md
```diff
@@ -45,19 +45,19 @@
...
-        - Check if automatic shutoff control is modeled in space during building closed hours (i.e. if lighting schedule hourly value in B_RMR is equal to P_RMR during building closed hours): `schedule_comparison_result = compare_schedules(normalized_schedule_b, normalized_schedule_p, building_open_schedule_b, -111)`  
+        - Check if automatic shutoff control is modeled in space during building closed hours (i.e. if lighting schedule hourly value in B_RMR is equal to P_RMR during building closed hours): `schedule_comparison_result = compare_schedules(normalized_schedule_b, normalized_schedule_p, inverse(building_open_schedule_b))`  
 
           **Rule Assertion:**
 
-          - Case 1: For building closed hours, if lighting schedule hourly value in B_RMR is equal to P_RMR: `if schedule_comparison_result == "MATCH": PASS`  
+          - Case 1: For building closed hours, if lighting schedule hourly value in B_RMR is equal to P_RMR: `if schedule_comparison_result["total_hours_compared"] == schedule_comparison_result["total_hours_matched"]: PASS`  
 
-          - Case 2: Else: `else: CAUTION and return schedule_comparison_result`  
+          - Case 2: Else: `else: Failed`
```
mkarpman, line 60: We discussed providing a flag to jurisdictions for whether to permit deviation that results in a more conservative outcome. If the flag is set to YES, then the RDS should also pass when EFLH_difference<1. (EFLH_difference = EFLH_schedule_1 / EFLH_schedule_2. Schedule_1 is baseline, Schedule_2 is proposed, so EFLH-difference<1 if baseline runtime hours are less than they should be.) 

Also, seems like Future Change Note should be removed (building_open_schedule was not moved to building_segment), and the temporary function note below also should be removed as it reflects an outdated logic.
  
Otherwise this RDS looks good!

docs/ruleset_functions/compare_schedules.md
```diff
@@ -7,7 +7,6 @@ Inputs:
 - **schedule_1**: First schedule.
 - **schedule_2**: Second schedule.
 - **mask_schedule**: The schedule that defines comparison mode for all 8760 hours in a year, i.e. if hourly value is 1, schedule_1 is evaluated to be equal to schedule_2; if hourly value is 2, schedule_1 is evaluated to be equal to schedule_2 times the comparison factor; if hourly value is 0, comparison was skipped for that particular hour (example when evaluating shut off controls, only he building closed hrs are evaluated).
```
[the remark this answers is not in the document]
weilixu: Updated.

docs/ruleset_functions/normalize_interior_lighting_schedules.md
weilixu: I can update this to space.interior_lightings since the function only focus on lightings, will that work?

docs/ruleset_functions/normalize_interior_lighting_schedules.md
weilixu: The note might be confusing but the function [table_G3_7_fns](https://github.com/pnnl/ruleset-checking-tool/blob/develop/rct229/data_fns/table_G3_7_fns.py#L167-L175) takes care of the control credit bonus logic and the [database ](https://github.com/pnnl/ruleset-checking-tool/blob/develop/rct229/data/ashrae_90_1_prm_2019.prm_interior_lighting.json#L1246-L1248)we are using have the bonus credit.

Note 1 was updated by Xing 11 months ago. I can remove it since we do not have this issue anymore.?

docs/ruleset_functions/normalize_interior_lighting_schedules.md
weilixu: Update it to space

docs/section6/Rule6-5.md
weilixu: Put it on hold until everything is #487 is merged.

docs/ruleset_functions/normalize_interior_lighting_schedules.md
```diff
@@ -30,7 +31,7 @@ Logic:
 
     - Add lighting power per area to space total: `space_total_power_per_area += power_per_area`
 
-  - Check if lighting uses schedule to model occupancy control, get occupancy control credit: `if interior_lighting.are_schedules_used_for_modeling_occupancy_control: control_credit = data_lookup(table_G3_7, space.lighting_space_type, interior_lighting.occupancy_control_type)`
+  - Check if adjust_for_credit is true and lighting uses schedule to model occupancy control, get occupancy control credit: `if adjust_for_credit and interior_lighting.are_schedules_used_for_modeling_occupancy_control: control_credit = data_lookup(table_G3_7, space.lighting_space_type, interior_lighting.occupancy_control_type)`
```
mkarpman, line 34: I don't see this function checking interiorlighting.occupancy_control_type, but I beleive you showed me in the code that this is verified. Also, I don't beleive we resolved the question whether bonus credit (1.25 adjustment) is allowed for any space type or just for certain space types. I think it should be any space type.

docs/section6/Rule6-5.md
```diff
@@ -45,38 +45,23 @@
 
       - For each space in zone: `space_b in zone_b.spaces:`  
 
-        - Get normalized space lighting schedule: `normalized_schedule_b = normalize_space_schedules(space_b.interior_lighting)`  
+        - Get normalized space lighting schedule: `normalized_schedule_b = normalize_interior_lighting_schedules(space_b.interior_lighting, false)`  
 
         - Get matching space in P_RMR: `space_p = match_data_element(P_RMR, Spaces, space_b.id)`  
 
-          - Get normalized space lighting schedule in P_RMR: `normalized_schedule_p = normalize_space_schedules(space_p.interior_lighting)`
+          - Get normalized space lighting schedule in P_RMR: `normalized_schedule_p = normalize_interior_lighting_schedules(space_p.interior_lighting, true)`
```
mkarpman, line 52: During unoccupied hours schedules should be the same between baseline and proposed, so I think normalized_schedule_p = normalize_interior_lighting_schedules(space_p.interior_lighting, **false**)

docs/ruleset_functions/normalize_interior_lighting_schedules.md
weilixu: We did not have a data_lookup function RDS. Instead, each table search has its own implementation.
In `table_g3_7_lookup` function, the [code](https://github.com/pnnl/ruleset-checking-tool/blob/develop/rct229/data_fns/table_G3_7_fns.py#L167-L177) checks if the `occupancy_control_type` is one of [enum](LightingOccupancyControlOptions). Also, there is an exception handling function that will set the output to UNDETERMINED if the `occupancy_control_type` is not available when `are_schedules_used_for_modeling_occupancy_control` is `True`

As bonus credit goes, I thought the discussion was that the code and database takes care of the bonus credit.
Line 167 - 171 in `table_g3_7_lookup`
```python
 manon_or_partauto = osstd_entry["manon_or_partauto"]
    if occupancy_control_type in [PARTIAL_AUTO_ON, MANUAL_ON]:
        if manon_or_partauto:
            control_credit = osstd_entry["occup_sensor_savings"]
        else:
            control_credit = osstd_entry["occup_sensor_auto_on_svgs"]
    elif occupancy_control_type == FULL_AUTO_ON:
        control_credit = osstd_entry["occup_sensor_auto_on_svgs"]
    else:
        # no credit
        control_credit = 0.0
```

The logic is if the `interior_lighting.occupancy_control_type` is one of the `[PARTIAL_AUTO_ON, MANUAL_ON]`, then we will get the value from `occup_sensor_savings` as control credit.
The `occup_sensor_savings` is the credit that included the bonus adjustment in the [database](https://github.com/pnnl/ruleset-checking-tool/blob/develop/rct229/data/ashrae_90_1_prm_2019.prm_interior_lighting.json#L1238-L1250):
```json
 {
      "template": "90.1-PRM-2019",
      "lpd_space_type": "office - enclosed <= 250 sf",
      "primary_space_type": "Office",
      "secondary_space_type": "Enclosed",
      "w/ft^2": 1.1,
      "w/ft": null,
      "isresidential": 0,
      "manon_or_partauto": 1,
      "occup_sensor_auto_on_svgs": 0.3,
      "occup_sensor_savings": 0.375,
      "notes": null
    },
```
In short, the `data_lookup` described in the RDS has the capabilities of automatically apply bonus credit based on the `interior_lighting.occupancy_control_type`.

Let me know if you think the implementation shall be different than the current logic.

docs/section6/Rule6-5.md
weilixu: That is correct. Thanks for pointing it out!

docs/ruleset_functions/normalize_interior_lighting_schedules.md
mkarpman: I understand that the logic is implemented via expanded table G3.7, and this is totally fine just makes it a little harder to review the RDS. Few comments:
1. I think most BEM tools will not populate "occupancy_control_type", so producing UNDETERMINED outcome if the occupancy_control_type is missing when are_schedules_used_for_modeling_occupancy_control is True will result in flagging majority of projects. Since these controls are mandatory, I suggest that in this case we assume that the required controls are specified (OS reduction in 90.1 Table G3.7 should be applied) but not give the bonus credit (by multiplying adjustment by 1.25). @supriyagoel please chime in if you agree. 

2. I think we agreed that the bonus credit would be allowed for all space types, but it doesn't seem to be the case - e.g. for Attic (copied below)  "occup_sensor_auto_on_svgs"="occup_sensor_savings"=0.1. I think credit should be allowed for any space type that has the controls. Also, "Attic" space type is not in G3.7, so where 0.1 adjustment is coming from?

     "template": "90.1-PRM-2019",
      "lpd_space_type": "attic - unoccupied",
      "primary_space_type": "Attic",
      "secondary_space_type": "General",
      "w/ft^2": 0.0,
      "w/ft": null,
      "isresidential": 0,
      "manon_or_partauto": 0,
      "occup_sensor_auto_on_svgs": 0.1,
      "occup_sensor_savings": 0.1,
      "notes": null

weilixu: 1. I can adjust the code to take assume NONE `occupancy_control_type` in the case where the type is missing. Will that work?
2. I was consulting the OSSTD team and seems my understanding was incorrect in some portions. I am not 100% understanding how the data is structured but below is the feedback I got from the team: 

> `manon_or_partauto = 0` means that neither Manual ON nor Partial Automatic ON are required in Chapter 9 of 90.1-2019. I think the idea was that if those are required (=1), then they need to be in the Appendix G Proposed model. If those are in the proposed model (propose is adjusted by occupancy sensor value), then the baseline is calculated from the proposed schedule based the Appendix G lighting table occupancy sensor values with adjustment of 1.25. If they are not in the proposed model, then Appendix G table occupancy sensor values applied. The column `occup sensor savings` accounts for the `manon_or_partauto`. You can see where that value is one, the `occup sensor savings` col is greater than the `occup sensor auto on svgs` col by factor of 1.25.

Based on the description, I think the logic can be simplified to:

```python
    if occupancy_control_type in [PARTIAL_AUTO_ON, MANUAL_ON, FULL_AUTO_ON]:
            control_credit = osstd_entry["occup_sensor_savings"]
    else:
        # no credit
        control_credit = 0.0
```

Is that OK?

The Attic seems using the credit from Atrium. It seems common for BEM software to do that because they have to work with much broader range of space types than table G3.7. However, inside the RCT, it will never search this data record because the `space.lighting_space_type` does not have `Attic` space type.

mkarpman: Weili, I suggest assuming occupancy_control_type=Other when control is not specified. This will result in using base adjustment in Table G3.7 but not the bonus adjustment. 

I think the logic should be as follows:
if occupancy_control_type in [PARTIAL_AUTO_ON, MANUAL_ON, FULL_AUTO_ON]:
            control_credit = osstd_entry["occup_sensor_savings"]
    else:
        ****default credit from Table G3.7****
        control_credit = osstd_entry[**"occup_sensor_auto_on_svgs"**]

PR #524, opened 2022-07-15
Create are_all_terminal_heat_sources_hot_water(hvac.id).md

claperle:


supriyagoel (approved):

docs/ruleset_functions/are_all_terminal_heat_sources_hot_water(hvac.id).md
```diff
@@ -0,0 +1,24 @@
...
+ 
+**Function Call:**  None   
+
+## Logic:  
+- Set are_all_terminal_heat_sources_hot_water = TRUE: `are_all_terminal_heat_sources_hot_water = TRUE`  
+- For each terminal_b in the list of terminal units: `For terminal_b in terminal_unit_id_list:`  
+    - Create an object for the terminal unit: `terminal_b = terminal_b.id`  
+    - Check if the heat source associated with the terminal unit does not equal HOT_WATER: `if terminal_b.heating_source != "HOT_WATER":`  
+        - Set are_all_terminal_heat_sources_hot_water = FALSE: `are_all_terminal_heat_sources_hot_water = FALSE`  
+        - Leave the loop: `break`
```
supriyagoel, line 20: @claperle I guess we will have separate functions to check if the terminal HW is being provided through PHW or HW Boiler?
For system 6, 8, the boiler should be the same as the central heating boiler, so we might need a function for that as well- correct?

claperle: @supriyagoel 
First question, yes.

Second, we check that the HW loop at the terminal level is connected to a boiler but we do not check if it is the same boiler as is serving the HW coil at the system level. Is this a baseline HVAC check or a check for another section?

mkarpman: I think this is checked by RDS 21-16 which confirms that there is only one HW plant.

mkarpman (approved):

PR #525, opened 2022-07-15
Rds/cml/function are all terminal cool sources chilled water

claperle:


docs/ruleset_functions/are_all_terminal_cool_sources_chilled_water(hvac.id).md
```diff
@@ -0,0 +1,24 @@
...
+
+**Description:** Returns TRUE if the cool source associated with all terminal units is CHILLED_WATER. It returns FALSE if any terminal unit has a cool source other than CHILLED_WATER.   
+
+**Inputs:**  
+- **B-RMR**: To evaluate if the cool source associated with all terminal units sent to this function is CHILLED_WATER.     
+- **terminal_unit_id_list**: List of terminal units to assess.
+
+**Returns:**  
+- **are_all_terminal_cool_sources_chilled_water**: The function returns TRUE if the cool source associated with all terminal units sent to this function is CHILLED_WATER. It returns FALSE if any terminal unit has a cool source other than CHILLED_WATER.  
+
```
supriyagoel, line 11: not ssure what we need this function for. which baseline system requires the terminal cooling source?

claperle: @supriyagoel I created then when we said the terminal type would be four pipe for the case where there is purchased HW and purchased CHW for system 1. Sounds like we are leaning towards the HW/CHW being defined at the hvac system and not the terminal so this function would become obsolete.

mkarpman (approved):

docs/ruleset_functions/are_all_terminal_cool_sources_chilled_water(hvac.id).md
supriyagoel: update: for FPfcu, chw will be defined at terminal now

supriyagoel (approved):

PR #526, opened 2022-07-15
Create does_each_zone_have_only_one_terminal(hvac.id).md

claperle:


docs/ruleset_functions/does_each_zone_have_only_one_terminal(hvac.id).md
```diff
@@ -0,0 +1,24 @@
...
+**Returns:**  
+- **does_each_zone_have_only_one_terminal**: The function returns TRUE if each zone input to this function only has one terminal unit. It returns FALSE if any zone has more than one terminal unit.   
+ 
+**Function Call:**  None    
+
+## Logic: 
+- Set does_each_zone_have_only_one_terminal = TRUE: `does_each_zone_have_only_one_terminal = TRUE`  
+- For each zone id in zone_id_list: `For zone_id in zone_id_list:`   
+    - Create zone object: `zone = zone_id`
+    - Check if there is not one terminal unit associated with the zone: `if len(zone.terminals) != 1:`
```
supriyagoel, line 18: small comment to reword line 18 as 'check if there is more than one terminal associated with the zone'

supriyagoel (approved):

docs/ruleset_functions/does_each_zone_have_only_one_terminal(hvac.id).md
claperle: Done but I adjusted slightly.

docs/ruleset_functions/does_each_zone_have_only_one_terminal(hvac.id).md
```diff
@@ -0,0 +1,24 @@
...
+- **zone_id_list**: List of zones to evaluate.  
+
+**Returns:**  
+- **does_each_zone_have_only_one_terminal**: The function returns TRUE if each zone input to this function only has one terminal unit. It returns FALSE if any zone has more than one terminal unit.   
+ 
+**Function Call:**  None    
+
+## Logic: 
+- Set does_each_zone_have_only_one_terminal = TRUE: `does_each_zone_have_only_one_terminal = TRUE`  
+- For each zone id in zone_id_list: `For zone_id in zone_id_list:`
```
mkarpman, line 16: Based on a recent exchange (below), exhaust fan may be modeled via a dedicated terminal:

Maria: We do not seem to have a field in the RMD to define zonal exhaust fans. These are important for verifying types of fans in the baseline and that the fan power is allocated to these fans in the same proportion as in the proposed design.
SG: Zone has a ‘exhaust airflow rate’, and ‘terminals’. Would it be correct to specify the exhaust airflow rate to the zone.exhaust_airflow_rate data element and then use terminals to connect it with a terminal that only includes a fan?
The fan would then be connected to a fan system that categorizes it as an exhaust fan

mkarpman (approved):

PR #535, opened 2022-07-19
Rds/cml/function is hvac sys fan sys vsd

claperle:


docs/ruleset_functions/is_hvac_sys_fan_sys_VSD.md
```diff
@@ -0,0 +1,23 @@
...
+
+**Returns:**  
+- **is_hvac_sys_fan_sys_VSD**: Returns TRUE if the HVAC system fan system his variable speed drive control. Returns FALSE if the HVAC system has a fan system that is anything other than variable speed drive controlled.   
+ 
+**Function Call:** None  
+
+## Logic:   
+- Set is_hvac_sys_fan_sys_VSD = FALSE: `is_hvac_sys_fan_sys_VSD = FALSE`  
+- Create an object associate with the fan_system associated with hvac_b: `fan_system_b = hvac_b.fan_system[0]`
+- Check if the system control is variable speed drive control, if yes then is_hvac_sys_fan_sys_VSD equals TRUE  : `if fan_system_b.fan_control == "VARIABLE_SPEED_DRIVE":is_hvac_sys_fan_sys_VSD = TRUE`
```
mkarpman, line 17: Suggest adding a check that there is one fan_system, analogous to checks for single heating and cooling system. As an alternative to having a separate sub-function it could be added here to avoid modifying parent functions is_bseline_system_type_X

claperle: I don't think this is needed because there is only allowed to be one fan system per HVAC system.

supriyagoel: Yes, looks like only one fan system per HVAC system is allowed. Though I do see some disconnect between fan system an fan.
A FanSystem could have multiple fans- supply fans, return fans etc. and this is a list, so a HVAC system would have one fan system which can have multiple supply fans. Maybe we need a test to make sure there is only one supply fan for all baseline systems?
Another related question (adding @JasonGlazer ) , fan system includes data elements for 'fan volume reset type' 'fan volume reset fraction'. However, these should be characteristics of the fan and not the fan system, correct? By having this at the fan system level we assume that the supply fan, return fan, exhaust fan etc all have these characteristics.

mkarpman: @supriyagoel, to your first question, would it make sense to only allow one fan of each type?

claperle: Discussed on call. Outcome, as is is okay.

mkarpman (approved):

supriyagoel (approved):

PR #540, opened 2022-07-19
Rds/cml/function is baseline system 5

claperle:


mkarpman (approved):

docs/ruleset_functions/is_baseline_system_5.md
```diff
@@ -0,0 +1,47 @@
...
+12. are_all_terminal_fans_null()  
+13. are_all_terminal_types_VAV()  
+
+ 
+## Logic:    
+- Create an object associated with the hvac system: `hvac_b = hvac_b.id`  
+- Set is_baseline_system_5 = Not_Sys_5: `is_baseline_system_5 = "Not_Sys_5"`    
+- Check that there is one preheat system per G3.1.3.19, if there is then carry on: `if Len(hvac_b.preheat_system) == 1:`   
+    - Check if the preheat system is a fluid loop, if yes then carry on: `if is_hvac_sys_preheating_type_fluid_loop(B_RMR, hvac_b.id) == TRUE:`
+        - Check if the cooling system type is DX, if yes then carry on: `if is_hvac_sys_cooling_type_DX(B_RMR, hvac_b.id) == TRUE:`
```
mkarpman, line 35: Should we check that there is only one cooling system?

docs/ruleset_functions/is_baseline_system_5.md
```diff
@@ -0,0 +1,47 @@
...
+
+ 
+## Logic:    
+- Create an object associated with the hvac system: `hvac_b = hvac_b.id`  
+- Set is_baseline_system_5 = Not_Sys_5: `is_baseline_system_5 = "Not_Sys_5"`    
+- Check that there is one preheat system per G3.1.3.19, if there is then carry on: `if Len(hvac_b.preheat_system) == 1:`   
+    - Check if the preheat system is a fluid loop, if yes then carry on: `if is_hvac_sys_preheating_type_fluid_loop(B_RMR, hvac_b.id) == TRUE:`
+        - Check if the cooling system type is DX, if yes then carry on: `if is_hvac_sys_cooling_type_DX(B_RMR, hvac_b.id) == TRUE:`  
+            - Check if fansystem is variable speed drive controlled, if yes then carry on: `if is_hvac_sys_fan_sys_VSD(B_RMR, hvac_b.id) == TRUE:`  
+                - Check if the hvac system is multizone and that each zone only has one terminal unit: `if is_hvac_system_multizone(B_RMR, zone_id_list) == TRUE AND does_each_zone_have_only_one_terminal(B_RMR,zone_id_list) == TRUE:`
```
mkarpman, line 37: I don't think we need this check - see analogous comment for System 6 function. There may be scenarios when a multizone system serves just one zone.

docs/ruleset_functions/is_baseline_system_5.md
mkarpman: Also, should we check that there is no heating system?

mkarpman (changes requested):

docs/ruleset_functions/is_baseline_system_5.md
claperle: Updated

docs/ruleset_functions/is_baseline_system_5.md
```diff
@@ -0,0 +1,50 @@
...
+12. are_all_terminal_fans_null()  
+13. are_all_terminal_types_VAV()
+14. is_there_only_one_hvac_sys_cooling_system ()  
+
+ 
+## Logic:    
+- Create an object associated with the hvac system: `hvac_b = hvac_b.id`  
+- Set is_baseline_system_5 = Not_Sys_5: `is_baseline_system_5 = "Not_Sys_5"`    
+- Check that there is no heating system, if there is none then carry on: `if len(hvac_b.heating_system) == Null or hvac_b.heating_system[0].heating_system_type == "NONE":`    
+    - Check that there is one preheat system per G3.1.3.19, if there is then carry on: `if Len(hvac_b.preheat_system) == 1:`
```
supriyagoel, line 35: @claperle I believe I might have asked this before as well. For system type 5 (or 6-8) do we assume that central  heating = NULL? I always thought that central heating is also the same as the reheat source.

claperle: @supriyagoel  My understanding is it is because they have a preheat coil they would not have heating coil. They would not have a heating coil since they have the terminal reheat. Note that per G3.1.3.19 systems 5 through 8 have preheat coils.

docs/ruleset_functions/is_baseline_system_5.md
supriyagoel: @mkarpman I have a question related to the check for there is no heating system. I always assumed that all MZ  systems include preheat, central heat and reheat. Though I agree it doesn't make sense to include all 3. I assume this is a well understood fact that MZ systems only include preheat and reheat and central heat is not modeled?

docs/ruleset_functions/is_baseline_system_5.md
claperle: Discuss on call

docs/ruleset_functions/is_baseline_system_5.md
claperle: Comment closed, discussed as group on call. There should be preheat, reheat and that's it for heating.

docs/ruleset_functions/is_baseline_system_5.md
claperle: Comment closed, discussed as group on call. There should be preheat, reheat and that's it for heating.

docs/ruleset_functions/is_baseline_system_5.md
claperle: Removed check for 5-8

mkarpman (approved):

supriyagoel (approved):

PR #541, opened 2022-07-20
Rs/wx/apply messages

weilixu:
Add messages to some rule definitions based on RDS.


rct229/rules/section5/section5rule15.py
```diff
@@ -71,6 +71,11 @@ def __init__(self):
                         "$": ["construction"],
                         "construction": ["f_factor"],
                     },
+                    manual_check_required_msg="Zone has both residential and non-residential spaces and the "
+                                              "construction requirements for slab-on-grade floor are different. "
+                                              "Verify construction is modeled correctly.",
+                    fail_msg='Baseline slab F-factor is not as expected for slabs that are less than 24" below grade. '
```
charliepnnl, line 77: It looks like you need to run Black on this. I expect it will replace the single quotes with double quotes.

rct229/rules/section6/section6rule3.py
```diff
@@ -10,6 +10,8 @@
 from rct229.utils.std_comparisons import std_equal
 
 OFFICE_OPEN_PLAN = schema_enums["LightingSpaceType2019ASHRAE901TG37"].OFFICE_OPEN_PLAN
+FAIL_MSG = "Lighting exists or is submitted with design documents. Lighting power density in P_RMR does not match U_RMR."
```
charliepnnl, line 13: We should be consistent whether to use a constant for the messages at the top of the file and then use the constant in the initializer, or to write out the messages in the initializer itself.

I don't have a strong preference other than to be consistent.

charliepnnl (approved):
Approved with style comments.

rct229/rules/section5/section5rule15.py
weilixu: Ran Black did not change the single quotes.
Should be safe I guess...

PR #542, opened 2022-07-20
Replace create_context with list_filter

weilixu:
Update RD using create_context with list_filter


yunjoonjung-PNNL (approved):
Thank you for doing this, Weili. I think this PR is perfect!

charliepnnl (approved):
Looks good!

PR #547, opened 2022-07-26
RS/CH/unique-ids

charliepnnl:
This adds a check for unique id values immediately after schema validation.

The strategy is to first determine "generic" json paths to all the lists inside a RuleSetModelInstance.  An example of such a path would be `"buildings[*].building_segments[*].zones"`.

Then we use _find_all_ to collect all the ids in these lists, as with 
```
find_all("buildings[*].building_segments[*].zones[*].id", <a RuleSetModelInstance>)
```

Finally, we check that this list of ids has no duplicates.


yunjoonjung-PNNL (approved):
I appreciate for doing this Charlie. Your logic in the ```check_unique_ids_in_ruleset_model_instances``` function totally makes sense and everything in this PR looks perfect to me.

weilixu (approved):
Approve - looks good!

PR #558, opened 2022-07-28
Rds/cml/function is baseline system 10

claperle:


supriyagoel (approved):
same comment as #557 related to exhaust fans. this RDS might need to be updated once we identify the approach for modeling exhaust fans

mkarpman (approved):

docs/ruleset_functions/is_baseline_system_10.md
```diff
@@ -0,0 +1,40 @@
...
+- Check that there is one heating system, if there is then carry on: `if is_there_only_one_hvac_sys_heating_system(B_RMR, hvac_b.id) == TRUE:`   
+    - Check that there is no preheat system, if there is none then carry on: `if len(hvac_b.preheat_system) == Null or hvac_b.preheat_system[0].heating_system_type = "NONE" :`    
+        - Check if fansystem is constant volume, if yes then carry on: `if is_hvac_sys_fan_sys_CV(B_RMR, hvac_b.id) == TRUE:`  
+            - Check if the hvac system serves a single zone and that the zone only has one terminal unit: `if does_hvac_system_serve_single_zone(B_RMR, zone_id_list) == TRUE AND does_each_zone_have_only_one_terminal(B_RMR,zone_id_list) == TRUE:`     
+                - Check that the data elements associated with the terminal unit align with system 10: `if are_all_terminal_heat_sources_none_or_null(B_RMR,terminal_unit_id_list) == TRUE AND are_all_terminal_cool_sources_none_or_null(B_RMR,terminal_unit_id_list) == TRUE And are_all_terminal_fans_null(B_RMR,terminal_unit_id_list) == TRUE AND are_all_terminal_types_CAV(B_RMR,terminal_unit_id_list) == TRUE:`
```
mkarpman, line 33: I wonder if for single zone systems we need to check for CV vs VAV at both fan system level and terminal level, or only once, at the level where fans are defined. For example, here we would only check it at the fan system level (is_hvac_sys_fan_sys_CV(B_RMR, hvac_b.id) == TRUE) but not at the terminal level (are_all_terminal_types_CAV(B_RMR,terminal_unit_id_list) == TRUE)

supriyagoel: that would be consistent with how we verify FPFCU, i would be ok with that change. I'll hold off on marking this as ready to merge till we resolve this

PR #563, opened 2022-07-30
Rt/yj/ruletest 6 2

yunjoonjung-PNNL:
Changes in this PR are
1. 6-2 testing JSON files are updated based on the RDS.
2. some parts in 6-2 Python code that didn't reflect on some parts in the RDS.  So, I made some updates.

I appreciate it if you could check whether these updates make sense. Thank you!


jugonzal07 (approved):

rct229/rules/section6/section6rule2.py
```diff
@@ -20,11 +22,8 @@ def __init__(self):
...
             index_rmr="proposed",
             id="6-2",
-            description="Spaces in proposed building with hardwired lighting, including Hotel/Motel Guest Rooms, "
-            "Dormitory Living Quarters, Interior Lighting Power >= Table 9.6.1 or the lighting design in "
-            "the user model, whichever is greater; For Dwelling Units, Interior Lighting Power >= "
-            "0.6W/sq.ft. or the lighting design in the user model, whichever is greater",
-            list_path="ruleset_model_instances[0].buildings[*].zones[*].spaces[*]",
+            description="SSpaces in proposed building with hardwired lighting, including Hotel/Motel Guest Rooms, Dormitory Living Quarters, Interior Lighting Power >= Table 9.6.1; For Dwelling Units, Interior Lighting Power >= 0.6W/sq.ft.",
+            list_path="ruleset_model_instances[0].buildings[*].building_segments[*].zones[*].spaces[*]",
         )
```
charliepnnl, line 28: @weilixu   Do we want to use
```
def list_filter(self, , context_item, data):
    lighting_space_type_p = context_item.proposed["lighting_space_type"]
    return lighting_space_type_p in [
        GUEST_ROOM,
        DORMITORY_LIVING_QUARTERS
        DWELLING_UNIT
    ]
```
instead of using an _is_applicable_ method in SpaceRule?
I honestly don't know which is better here.
If we use _list_filter_, only the spaces with these lighting space types will be in the outcome.
If we use an _is_applicable_ method in SpaceRule, then all the other spaces will be included in the outcome and be set to not applicable.

My own personal preference would be to keep the outcome shorter by using the list_filter.

rct229/rules/section6/section6rule2.py
```diff
@@ -51,18 +51,23 @@ def get_calc_vals(self, context, data=None):
...
-                space_p["lighting_space_type"]
-            )
+            if space_p["lighting_space_type"] in [
+                GUEST_ROOM,
+                DORMITORY_LIVING_QUARTERS,
+            ]:
+                lighting_power_allowance_p = table_9_6_1_lookup(
+                    space_p["lighting_space_type"]
+                )["lpd"]
+            elif space_p["lighting_space_type"] == DWELLING_UNIT:
```
charliepnnl, line 61: Since the filtering should force the  lighting space type to be DWELLING_UNIT at this point, I would favor logic such as
```
else:
    assert space_p["lighting_space_type"] == DWELLING_UNIT
    lighting_power_allowance_p = 0.6
```
Also, the 0.6 should be a constant defined at the top of the file such as 
```
DWELLING_UNIT_MIN_LIGHTING_POWER_PER_AREA = 06 * ureg("W/ft2")
```

rct229/rules/section6/section6rule2.py
```diff
@@ -51,18 +51,23 @@ def get_calc_vals(self, context, data=None):
...
+                lighting_power_allowance_p = 0.6
 
-            # sum of space_p lighting power density
-            space_lighting_power_per_area_p = sum(
-                interior_lighting["power_per_area"]
-                for interior_lighting in space_p["interior_lighting"]
+            space_lighting_power_per_area_p = pint_sum(
+                find_all("$..interior_lighting[*].power_per_area", space_p),
```
charliepnnl, line 65: Since _interior_lighitng_ is directly under the Space, you don't need the "$.." at the beginining of the JSON path.  The same is true for line 69 below.

charliepnnl (approved):
Approved with comments that are largely stylistic

yunjoonjung-PNNL:
@charliepnnl Thank you for the PR review and the comments were great. I addressed all of your comments. I'd appreciate it if you could verify the updates one more time.

rct229/rules/section6/section6rule2.py
weilixu: Following the RDS, I think we should use `is_applicable` method in SpaceRule since this is part of the applicability check in the RDS.

However, I agree that there could be too many `not_applicable` cases that could make the export report look bad.

I think we shall be consistent, I like the `list_filter` better. In that case, we need to update the RDS to remove the applicability check.

BTW, the `list_filter` shall be under the `Section6Rule2` class. It is misplaced under the `SpaceRule` class.
@yunjoonjung-PNNL

rct229/rules/section6/section6rule2.py
```diff
@@ -51,18 +54,24 @@ def get_calc_vals(self, context, data=None):
...
-            )
+            if space_p["lighting_space_type"] in [
+                GUEST_ROOM,
+                DORMITORY_LIVING_QUARTERS,
+            ]:
+                lighting_power_allowance_p = table_9_6_1_lookup(
+                    space_p["lighting_space_type"]
+                )["lpd"]
+            else:
+                assert space_p["lighting_space_type"] == DWELLING_UNIT
```
weilixu, line 65: I don't think this `assert` is needed if the `Section21Rule6` class uses `list_filter`.
When using the `list_filter` it sets the precondition for the logic that the `lighting_space_type` shall only be `GUEST_ROOM`, `DORMITORY_LIVING_QUARTERS` and `DWELLING_UNIT`
In that case, we can just say:
```python
if space_p["lighting_space_type"] in [
                 GUEST_ROOM,
                 DORMITORY_LIVING_QUARTERS,
             ]:
                 lighting_power_allowance_p = table_9_6_1_lookup(
                     space_p["lighting_space_type"]
                 )["lpd"]
             else:
                 lighting_power_allowance_p = DWELLING_UNIT_MIN_LIGHTING_POWER_PER_AREA
```

weilixu (changes requested):

rct229/rules/section6/section6rule2.py
charliepnnl: @weilixu This is a fairly big question.  We haven't been very consistent about this so far.  Logically,  _list_filter_ and _is_applicable_ are interchangeable.  Going forward, on what basis do we decide which to use for a particular rule?

So far, we have tended to use list_filter when the applicability test _feels_ like filtering.  I guess "feels like" here means that the filter code is simple. Another determiner seems to have been how explicit the RDS is about the applicability check.  But I suspect they write the RDS in a way that seems natural and are not necessarily trying to indicate how they want the outcome to look in this respect.

Going forward, I'd like for us to have a clear guideline for which approach to take so we don't second guess each other in the review process and add extra work.

weilixu: > @weilixu This is a fairly big question. We haven't been very consistent about this so far. Logically, _list_filter_ and _is_applicable_ are interchangeable. Going forward, on what basis do we decide which to use for a particular rule?
> 
> So far, we have tended to use list_filter when the applicability test _feels_ like filtering. I guess "feels like" here means that the filter code is simple. Another determiner seems to have been how explicit the RDS is about the applicability check. But I suspect they write the RDS in a way that seems natural and are not necessarily trying to indicate how they want the outcome to look in this respect.
> 
> Going forward, I'd like for us to have a clear guideline for which approach to take so we don't second guess each other in the review process and add extra work.

I agree.
Generally, I believe the way to use `is_applicable` is when RDS explicitly added the logic of applicability check, and use `list_filer` when RDS use ` if xxx == xxx`, example -> `If surface.type == Floor`.
This has been consistent until this RDS where the same condition appears in both applicability check and `list_filter`. 
One thing we can do is to clarify it with the RDS team to give a clear instruction on which one to choose in this particular RDS?

rct229/rules/section6/section6rule2.py
yunjoonjung-PNNL: Fixed Great point!

yunjoonjung-PNNL:
The updates are
i) ```list_filter``` method was moved.
ii) Applicability check was removed in the RDS.

I'd appreciate it if you could review it one more time. Thank you!

rct229/rules/section6/section6rule2.py
weilixu: Confirm with the [issue ](https://github.com/pnnl/ruleset-checking-tool/issues/605) here,
For any rule that checks a subset of data group:
1. loop through the full set of data group
2. Skip the data group that do not match the selection criteria
3. At rule assertion level (post-process), add logic that no results returned, then set the rule outcome to `NOT_APPLICABLE`

weilixu (approved):

PR #585, opened 2022-08-19
RS/CH/outcome_units

charliepnnl:
- I used Rule6-4 and Rule5-5 as examples.
- The conversion to a string takes place in _rule_engine/engine/evaluate_rules()_  were the unit system could be specified.
- The _UNIT_CONVENTIONS are in _utils/pint_utils.js_; since this is somewhat ruleset-specific, we might prefer to put this elsewhere. We could pass this structure in to _evaluate_rules()_ also.
- I replaced each "-" with "*", and used parentheses in the denominators, in the unit strings to make Pint happier; we could convert them back after the string is created if desired. To be honest, I never liked the "-"s anyway.


rct229/ruletest_engine/ruletest_engine.py
```diff
@@ -508,6 +508,6 @@ def run_envelope_tests():
     Results of envelope stest are spit out to console
     """
 
-    envelope_test_json = "envelope_tests.json"
```
weilixu, line 511: Let's keep the `envelope_tests.json` to be the default in the developer branch. Could you change the `lighting_test_json` to `lighting_tests.json` as well? Thanks

weilixu (approved):
Thanks for the implementation.
approved with a minor comment.

yunjoonjung-PNNL (approved):
I reviewed the updates and they look perfect to me Charlie :). I appreciate this effort!

PR #588, opened 2022-08-24
RS/YJ/RULE 21-11

yunjoonjung-PNNL:
I'd appreciate it if you could review this PR. Thank you!


rct229/rules/section21/section21rule11.py
```diff
@@ -0,0 +1,82 @@
...
+
+    class HeatingFluidLoopRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section21Rule11.HeatingFluidLoopRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+            )
+
+        def get_calc_vals(self, context, data=None):
+            heating_fluid_loop_b = context.baseline
+            return {"child_loop": heating_fluid_loop_b.get("child_loops")}
```
charliepnnl, line 79: I think the intent is to check that there are no child loops.  So I think we need logic such as
```
return {"has_child_loops":  bool(heating_fluid_loop_b.get("child_loops"))
```
This works because both None and [] are falsey.

Then the rule_check logic becomes:
```
has_child_loops = calc_vals["has_child_loops"]
return not has_child_loops
```

charliepnnl (approved):
Approved with one comment that needs to be addressed.

rct229/rules/section21/section21rule11.py
yunjoonjung-PNNL: This is a very elegant way to solve this problem. Thank you so much for letting me know about this :).

weilixu (approved):

weilixu:
This looks good to me now.

PR #589, opened 2022-08-24
RS/YJ/RULE 21-12

yunjoonjung-PNNL:
I'd appreciate it if you could review this PR. Thank you!


rct229/rules/section21/section21rule12.py
```diff
@@ -0,0 +1,96 @@
...
+            flow_control = heating_fluid_loop_b["heating_design_and_control"][
+                "flow_control"
+            ]
+            operation = heating_fluid_loop_b["heating_design_and_control"]["operation"]
+            return {"flow_control": flow_control, "operation": operation}
+
+        def rule_check(self, context, calc_vals=None, data=None):
+            return (
+                True
+                if calc_vals["flow_control"] == "VARIABLE_FLOW"
```
weilixu, line 93: I prefer here to use enum from the schema. 
`flow_control` -> `FluidLoopFlowControlOptions`
`operation` -> `FluidLoopOperationOptions`

Same as the `operation`.

weilixu (approved):
Approved with a style modification

rct229/rules/section21/section21rule12.py
yunjoonjung-PNNL: Fixed. Great point!

rct229/rules/section21/section21rule12.py
```diff
@@ -0,0 +1,97 @@
+from rct229.data.schema_enums import schema_enums
+from rct229.rule_engine.rule_base import RuleDefinitionBase
+from rct229.rule_engine.rule_list_indexed_base import RuleDefinitionListIndexedBase
+from rct229.rule_engine.user_baseline_proposed_vals import UserBaselineProposedVals
+from rct229.ruleset_functions.get_baseline_system_types import get_baseline_system_types
+from rct229.utils.assertions import getattr_
+
+VARIABLE_FLOW = schema_enums["FluidLoopFlowControlOptions"].VARIABLE_FLOW
```
charliepnnl, line 8: I'd prefer for us to generally do something like:
```
FLUID_LOOP_FLOW_CONTROL = schema_enums["FluidLoopFlowControlOptions"]
```
and then use FLUID_LOOP_FLOW_CONTROL.VARIABLE_FLOW in the code.
If we do that consistently, I think it will make the intent a little bit clearer.

rct229/rules/section21/section21rule12.py
```diff
@@ -0,0 +1,97 @@
...
+    "SYS-12B",
+    "SYS-13B",
+    "SYS-1C",
+    "SYS-3C",
+    "SYS-7C",
+    "SYS-11C",
+    "SYS-12C",
+    "SYS-13C",
+]
+FLUID_LOOP = schema_enums["FluidLoopOptions"]
```
charliepnnl, line 39: Yea, like you've done here.

rct229/rules/section21/section21rule12.py
```diff
@@ -0,0 +1,97 @@
...
+        fluid_loop_b = context_item.baseline
+        return getattr_(fluid_loop_b, "FluidLoop", "type") == FLUID_LOOP.HEATING
+
+    class HeatingFluidLoopRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section21Rule12.HeatingFluidLoopRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                required_fields={
+                    "$": ["heating_design_and_control"],
+                    "$.heating_design_and_control": ["flow_control", "operation"],
```
charliepnnl, line 79: The "$." part of line 79 is not needed.

rct229/rules/section21/section21rule12.py
```diff
@@ -0,0 +1,97 @@
...
+        def get_calc_vals(self, context, data=None):
+            heating_fluid_loop_b = context.baseline
+            flow_control = heating_fluid_loop_b["heating_design_and_control"][
+                "flow_control"
+            ]
+            operation = heating_fluid_loop_b["heating_design_and_control"]["operation"]
+            return {"flow_control": flow_control, "operation": operation}
+
+        def rule_check(self, context, calc_vals=None, data=None):
+            return (
```
charliepnnl, line 92: This logic can be simplified to:
```
return calc_vals["flow_control"] == VARIABLE_FLOW and calc_vals["operation"] == CONTINUOUS
```
Although, we have typically collected the needed pieces from calc_vals before the return statement to make the return statement as easy to read as possible. So I think we have typically done something like:
```
flow_control = calc_vals["flow_control"]
operation = calc_vals["operation"]

return flow_control  == VARIABLE_FLOW and operation == CONTINUOUS
```

charliepnnl (approved):
Approved with only a few stylistic comments

yunjoonjung-PNNL:
I addressed all of Charlie's comments and would appreciate it if you could verify the changes make sense. Thank you!

weilixu:
Looks good

PR #598, opened 2022-08-28
RS/CH/calc_q_for_all

charliepnnl:
This adds CalcQ to all rules as needed.
I was able to test them all with the ruletests except for Rule6-1 and Rule6-2, whose ruletests to not appear to be working independently of the addition of CalcQ.


yunjoonjung-PNNL (approved):
I went over the entire files and these updates look great to me. Thanks for this effort, Charlie!

weilixu (approved):

PR #600, opened 2022-08-30
Rds/cml/update 22 33

claperle:


docs/section22/Rule22-33.md
```diff
@@ -4,50 +4,60 @@
 **Schema Version:** 0.0.10  
 **Mandatory Rule:** True  
 **Rule ID:** 22-33  
-**Rule Description:** Baseline chilled water system that does not use purchased chilled water must only have no more than one CHW plant.  
+**Rule Description:** Baseline chilled water system that does not use purchased chilled water must only have no more than one CHW plant and be modeled as primary seconday and baseline purchased CHW systems must only have one external fluid loop.
```
supriyagoel, line 7: @mkarpman and @claperle should this be 3 rules? Not sure why we're combining 1. one CHW plant 2.  P&S loops + 3. one external fluid loop

mkarpman: Yes, I agree. And it seems that we already have primary/secondary check as [RDS 22-7](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section22/Rule22-7.md).

claperle: @supriyagoel  @mkarpman  Okay, I think I just need to remove the purchased CHW part and move it somewhere else or create a new rule. I also need to restore the rule description (I don't think I changed anything in terms of the previous logic with the exception of adding the check on the purchased CHW system have one loop). Due to rule name changes looks like wires got crossed.

claperle: Per phone call, add a separate rule that checks for one purchased CHW loop. Only add applicability check. Non primary rule. Reset this back to what it was.

PR #602, opened 2022-08-30
RS/YJ/Rule21-17

yunjoonjung-PNNL:
I'd appreciate it if you reviewed this PR. Thank you!


rct229/rules/section21/section21rule17.py
```diff
@@ -0,0 +1,99 @@
...
+    "SYS-7",
+    "SYS-11.2",
+    "SYS-12",
+    "SYS-1A",
+    "SYS-7A",
+    "SYS-11.2A",
+    "SYS-12A",
+]
+
+BOILER_RATED_CAPACITY_300000 = 300000 * ureg("Btu/hr")
```
charliepnnl, line 21: I would prefer that the names of the constants not be dependent on the choice of units.  Here the 300000 assumes Btu/hr is used.  How about `BOILER_RATED_CAPACITY_LOW_LIMIT` and `BOLER_RATED_CAPACITY_HIGH_LIMIT` or other terms that make more sense to you.

rct229/rules/section21/section21rule17.py
```diff
@@ -0,0 +1,99 @@
...
+            boiler_rated_capacity_b = calc_vals["boiler_rated_capacity_b"]
+            boiler_efficiency_metric_b = calc_vals["boiler_efficiency_metric_b"]
+            boiler_efficiency_b = calc_vals["boiler_efficiency_b"]
+
+            return (
+                (
+                    boiler_rated_capacity_b < BOILER_RATED_CAPACITY_300000
+                    and boiler_efficiency_metric_b
+                    == BOILER_EFFICIENCY_METRIC.ANNUAL_FUEL_UTILIZATION
+                    and std_equal(boiler_efficiency_b, BOILER_EFFICIENCY_80)
```
charliepnnl, line 86: @weilixu  Do we need to use `std_equal` here? I'm thinking that `boiler_efficiency_b` must be exactly 0.8 in this case, but I could easily be wrong. Similarly for the other cases.

charliepnnl (approved):
Approved with one style comment and one question for Weili.

rct229/rules/section21/section21rule17.py
yunjoonjung-PNNL: Great point! Fixed.

rct229/rules/section21/section21rule17.py
yunjoonjung-PNNL: I will be waiting for Weili's opinion and will also reflect on this opinion as well.

weilixu: Agree, we don't need `std_equal` here.

weilixu: However, an AHJ comparison is coming up soon, so in this case, we will need to use the AHJ comparison to allow greater and equal.
The current implementation is OK, and Once the AHJ comparison is merged, we will need to create a new PR to address all the comparisons in the RCT.

yunjoonjung-PNNL: I removed ```std_equal``` to prevent any confusion. Once the AHJ comparison function is developed, I will get this rule updated.

weilixu (approved):

PR #603, opened 2022-08-30
RS/YJ/Rule21-18

yunjoonjung-PNNL:
I'd appreciate it if you reviewed this PR. Thank you!


rct229/rules/section21/section21rule18.py
```diff
@@ -0,0 +1,66 @@
...
+    "SYS-7",
+    "SYS-11.2",
+    "SYS-12",
+    "SYS-1A",
+    "SYS-7A",
+    "SYS-11.2A",
+    "SYS-12A",
+]
+
+FUEL_SOURCE_OPTION = schema_enums["EnergySourceOptions"]
```
charliepnnl, line 18: I think we've been not including the work "OPTION" in the names of our enumeration constants.  Then it is used as `FUEL_SOURCE. NATURAL_GAS`, which seems natural (no pun intended).

charliepnnl (approved):
Only one tiny tiny style comment.

rct229/rules/section21/section21rule18.py
yunjoonjung-PNNL: It totally makes sense to me and this one was fixed.

weilixu (approved):

PR #607, opened 2022-09-01
complete is there only one hvac sys heating system function

weilixu:


yunjoonjung-PNNL (approved):
This looks great to me. Thank you for this effort!

weilixu:
We do not need this function - I will move it to archive

PR #611, opened 2022-09-02
Update are_all_terminal_types_VAV.md

claperle:


supriyagoel (approved):

claperle:
Change name to indicate _or_Null?

docs/ruleset_functions/are_all_terminal_types_VAV.md
```diff
@@ -1,21 +1,21 @@
...
-**Description:** Returns TRUE if all of the terminal unit types input to this function are variable air volume (VAV). It returns FALSE if any of the terminal units are of a type other than variable air volume (VAV).   
+**Description:** Returns TRUE if all of the terminal unit types input to this function are variable air volume (VAV) or are equal to null. It returns FALSE if any of the terminal units are of a type other than variable air volume (VAV) or null.   
 
 **Inputs:**  
 - **B-RMR**: To evaluate if the terminal unit types are variable air volume (VAV).    
 - **terminal_unit_id_list**: List of terminal units to assess.  
 
 **Returns:**  
-- **are_all_terminal_types_VAV**: The function returns TRUE if all of the terminal unit types input to this function are variable air volume (VAV). It returns FALSE if any of the terminal units are of a type other than variable air volume (VAV).      
+- **are_all_terminal_types_VAV**: The function returns TRUE if all of the terminal unit types input to this function are variable air volume (VAV) or are equal to null. It returns FALSE if any of the terminal units are of a type other than variable air volume (VAV) or null.
```
mkarpman, line 10: @claperle why does it return TRUE if terminal unit type is null?

claperle: @mkarpman Because we agreed that for the terminals if it is equal to what it should be or not populated (i.e., null) then it would pass (i.e., equal true).

mkarpman: OK, I remember now. This is because in the RDS this function is always called after "if is_hvac_sys_fan_sys_VSD(B_RMR, hvac_b.id) == TRUE", so we are not requiring that VAV is defined both on system and terminal level. Perhaps add a note to this effect to this function to avoid future confusion, if you think it makes sense.

mkarpman (approved):

docs/ruleset_functions/are_all_terminal_types_VAV.md
claperle: @mkarpman Note added at the bottom of the RDS

PR #613, opened 2022-09-02
Rs/wx/is hvac sys cooling type fluid loop

weilixu:


rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/is_hvac_sys_cooling_type_fluid_loop.py
```diff
@@ -0,0 +1,42 @@
...
+    # Get the hvac system
+    hvac_b = find_exactly_one_with_field_value(
+        "$.buildings[*].building_segments[*].heating_ventilation_air_conditioning_systems",
+        "id",
+        hvac_b_id,
+        rmi_b,
+    )
+    # Check if hvac_b has preheat system
+    cooling_system = hvac_b.get("cooling_system")
+    if cooling_system:
```
charliepnnl, line 35: This is correct, but could be more succint:
```
is_hvac_sys_cooling_type_fluid_loop_flag =
    cooling_system and 
    cooling_system.get("chilled_water_loop") and
    cooling_system["cooling_system_type"] == COOLING_SYSTEM_TYPE.FLUID_LOOP
```
And then line 24 could also be deleted.

charliepnnl (approved):
Approved with one stylistic comment.
Also, the RDS should be updated to remove the use of _is_there_only_one_cooling_system()_ and the "_[0]_" from _hvac_b.cooling_system[0]_.

weilixu:
> Approved with one stylistic comment. Also, the RDS should be updated to remove the use of _is_there_only_one_cooling_system()_ and the "_[0]_" from _hvac_b.cooling_system[0]_.

Good catch - 

One thing is we decided to remove all `is_there_only_one xxx` functions since the data group is not array.

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/is_hvac_sys_cooling_type_fluid_loop.py
weilixu: It seems:
```
cooling_system and 
    cooling_system.get("chilled_water_loop") and
    cooling_system["cooling_system_type"] == COOLING_SYSTEM_TYPE.FLUID_LOOP
```
returns `None` if `cooling_system` is `None` or `cooling_system.get("chilled_water_loop")` is `None`, which cannot be assign to the flag.

I was thinking to flip the default flag to `True` so the logic can be written in a more clean way, but give up because I don't want to alter the RDS.

charliepnnl: Good point about returning None.
How about:
``` 
cooling_system is not None and
cooling_system.get("chilled_water_loop") is not None and
cooling_system["cooling_system_type"] == COOLING_SYSTEM_TYPE.FLUID_LOOP
```
?

weilixu: True, this will work. I can condense the if statement to the one you proposed. Seems cleaner to me.

yunjoonjung-PNNL (approved):
It looks great to me. Thank you!

Issue #627, opened 2022-09-05
Logic for does_each_zone_have_only_one_terminal


weilixu:
RDS [link](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/does_each_zone_have_only_one_terminal.md)

In the description, the function says:

Returns TRUE if each zone input to this function only has one terminal unit. It returns FALSE if any zone **has more than one terminal unit**.

In the rule assertion logic, it states:

check if there is more than one terminal **or no terminal units associated with the zone**:

The question is, when a zone has no terminal associate: 
```python 
zone.terminals is None or len(zone.terminals == 0)
```
, should the function return `True` or `False`?

claperle:
@weilixu @mkarpman I agree with Weili's suggestion. Maria, do you?

mkarpman:
Yes, I agree.

PR #628, opened 2022-09-05
Complete does_each_zone_have_only_one_terminal

weilixu:
One logic issue relate to this function is [posted](https://github.com/pnnl/ruleset-checking-tool/issues/627). Waiting for RDS team response


rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/does_each_zone_have_only_one_terminal.py
```diff
@@ -0,0 +1,25 @@
...
+    bool
+        True: each zone input to this function only has one terminal unit or no terminal
+        False: any zone has more than one terminal
+    """
+    does_each_zone_have_only_one_terminal_flag = True
+    for zone_id in zone_id_list:
+        zone = find_exactly_one_with_field_value(
+            "$.buildings[*].building_segments[*].zones", "id", zone_id, rmi_b
+        )
+        if zone.get("terminals") is not None and len(zone["terminals"]) > 1:
```
yunjoonjung-PNNL, line 22: Could this logic also determine ```no terminal units with the zone``` case?

yunjoonjung-PNNL (approved):
Approved. I'd appreciate it if you could double-check the comment.

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/does_each_zone_have_only_one_terminal.py
weilixu: It is possible. I posted this issue [here ](https://github.com/pnnl/ruleset-checking-tool/issues/627)for the `no terminal` situation

charliepnnl (approved):
Approved contingent on the response from the RDS team.
A more appropriate name for this function, as currently defined by the RDS, would be _does_each_zone_have_at_most_one_terminal_.

weilixu:
> Approved contingent on the response from the RDS team. A more appropriate name for this function, as currently defined by the RDS, would be _does_each_zone_have_at_most_one_terminal_.

I have discussed the issue with RDS, they agree that this shall be only one terminal (null, 0 or more than 1 shall return false.) I will update the RDS language.

PR #631, opened 2022-09-06
Complete are_all_terminal_heating_loops_purchased_heating

weilixu:


yunjoonjung-PNNL (approved):
Looks great to me!

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_heating_loops_purchased_heating.py
```diff
@@ -0,0 +1,63 @@
...
+    -------
+    bool
+        True: fluid loop associated with the heating_from_loop associated with each terminal unit is purchased heating
+        False: otherwise
+    """
+    are_all_terminal_heating_loops_purchased_heating_flag = True
+    purchased_heating_loop_list_b = []
+
+    external_fluid_sources = rmi_b.get("external_fluid_source")
+    if external_fluid_sources:
```
charliepnnl, line 26: An alternate way of getting this list would be:
```
purchased_heating_loop_list_b = [
    *find_all(f'external_fluid_source[*][?(@.type={EXTERNAL_FLUID_SOURCE.HOT_WATER})].loop, rmi_b),
    *find_all(f'external_fluid_source[*][?(@.type={EXTERNAL_FLUID_SOURCE.STEAM})].loop, rmi_b),
]
```

I am exploring an alternate Python jsonpath implementation that would further simplify something like this by allowing us to combine the two find_all expressions into one.
This alternate implementation is actively supported by Mark Borkum of PNNL.

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_heating_loops_purchased_heating.py
```diff
@@ -0,0 +1,63 @@
...
+                external_fluid_source.get("type")
+                in [
+                    EXTERNAL_FLUID_SOURCE.HOT_WATER,
+                    EXTERNAL_FLUID_SOURCE.STEAM,
+                ]
+                and external_fluid_source.get("loop") is not None
+            ):
+                purchased_heating_loop_list_b.append(external_fluid_source["loop"])
+
+    for terminal_b_id in terminal_unit_id_list:
```
charliepnnl, line 38: The RDS is a bit sketchy at this point. It does not indicate to loop over the terminal_unit_id_list, but can see that they did mean to.

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_heating_loops_purchased_heating.py
```diff
@@ -0,0 +1,63 @@
...
+        terminal_b = find_exactly_one_with_field_value(
+            "$.buildings[*].building_segments[*].zones[*].terminals",
+            "id",
+            terminal_b_id,
+            rmi_b,
+        )
+        heating_from_loop = terminal_b.get("heating_from_loop")
```
charliepnnl, line 45: Should we call this variable _heating_from_loop_id_?

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_heating_loops_purchased_heating.py
```diff
@@ -0,0 +1,63 @@
...
+                heating_from_loop,
+                rmi_b,
+            )
+            if (
+                getattr_(fluid_loop, "fluid loop", "type") != FLUID_LOOP_TYPE.HEATING
+                or heating_from_loop not in purchased_heating_loop_list_b
+            ):
+                are_all_terminal_heating_loops_purchased_heating_flag = False
+                break
+        else:
```
charliepnnl, line 59: The RDS doesn't seem to explicitly include this else part.  I suspect that the RDS should be updated to include this case.

charliepnnl (approved):
Approved with comments

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_heating_loops_purchased_heating.py
weilixu: Unfortunately it seems jsonpath_ng does not support OR operator.
I did a slight change to make it work in one line, let me know if you found other package that works withe alternate jsonpath.

```python
    purchased_heating_loop_id_list_b = [*find_all(f'(external_fluid_source[?(@.type=={EXTERNAL_FLUID_SOURCE.HOT_WATER})].loop) | (external_fluid_source[?(@.type=={EXTERNAL_FLUID_SOURCE.STEAM})].loop)', rmi_b)]
```

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_heating_loops_purchased_heating.py
weilixu: This is the case when a terminal has no heating from loop and RDS seems to be missing this logic. I will add it to the document.

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_heating_loops_purchased_heating.py
charliepnnl: Since you've combined the two find_all's into one, you no longer need to wrap it with [* ... ].

Correct, jsonpath-ng does not support an OR operator.
https://github.com/pacifica/python-jsonpath2 does support OR (most conveniently as a comma).  With this package, the find_all would be:
```
find_all(f"external_fluid_source[?(@.type=={EXTERNAL_FLUID_SOURCE.HOT_WATER}, @.type=={EXTERNAL_FLUID_SOURCE. STEAM})].loop")
```

charliepnnl (approved):

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_heating_loops_purchased_heating.py
weilixu: Does the jsonpath2 supports AND operator? I find some use cases that may need to intersect two jsonpaths.

charliepnnl: Yes. Although I think we can do that using jsonpath-ng too by appending restrictions `[?(one_restriction)][?(another_restriction)]`

PR #632, opened 2022-09-06
Complete are_all_terminal_types_VAV

weilixu:


rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_types_VAV.py
```diff
@@ -0,0 +1,39 @@
...
+    are_all_terminal_types_VAV_flag = True
+
+    for terminal_b_id in terminal_unit_id_list:
+        terminal_b = find_exactly_one_with_field_value(
+            "$.buildings[*].building_segments[*].zones[*].terminals",
+            "id",
+            terminal_b_id,
+            rmi_b,
+        )
+        if (
```
charliepnnl, line 32: This could be simplified to
```
if terminal_b.get("type") != TERMINAL_TYPE.VARIABLE_AIR_VOLUME:
```
since if the type is None, then it is not TERMINAL_TYPE.VARIABLE_AIR_VOLUME:

charliepnnl (approved):
Approved with one comments for shortening the code.

PR #635, opened 2022-09-07
RS/YJ/RULE21-9

yunjoonjung-PNNL:
I'd appreciate it if you could review this PR. Thanks.


rct229/rules/section21/section21rule9.py
```diff
@@ -0,0 +1,80 @@
...
+
+    class HeatingFluidLoopRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section21Rule9.HeatingFluidLoopRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+            )
+
+        def get_calc_vals(self, context, data=None):
+            fluid_loop_b = context.baseline
+            pump_power_per_flow_rate = fluid_loop_b["pump_power_per_flow_rate"]
```
weilixu, line 73: It seems we need to add "pump_power_per_flow_rate" to the required field or using `getattr_`.

weilixu (approved):
Approved with one comment must be addressed and a stylistic comment.

rct229/rules/section21/section21rule9.py
yunjoonjung-PNNL: Fixed. Great catch!

rct229/rules/section21/section21rule9.py
```diff
@@ -0,0 +1,80 @@
...
+        # FIXME: replace with baseline_system_types = get_baseline_system_types(rmi_b) when get_baseline_system_types
+        #  is ready.
+        baseline_system_types = {
+            "SYS-7": ["hvac_sys_7"],
+            "SYS-12": ["hvac_sys_12"],
+        }
+        # if any system type found in the APPLICABLE_SYS_TYPES then return applicable.
+        return any(
+            [key in APPLICABLE_SYS_TYPES for key in baseline_system_types.keys()]
+        )
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Fixed. Thank you for letting me know!

rct229/rules/section21/section21rule9.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        boilers = find_all("$.boilers[*]", rmi_b)
+        loop_boiler_dict = {}
```
charliepnnl, line 52: The RDS is overly complicated here. It creates this dictionary and then only uses its keys.  It would be more straight forward to simply create a list of boiler loop ids.
```
boiler_loop_ids = find_all("boilers[*].loop", rmi_b)
```
@weilixu  What do you think?

rct229/rules/section21/section21rule9.py
```diff
@@ -0,0 +1,83 @@
...
+    "SYS-5",
+    "SYS-7",
+    "SYS-11.2",
+    "SYS-12",
+    "SYS-1A",
+    "SYS-7A",
+    "SYS-11.2A",
+    "SYS-12A",
+]
+PUMP_POWER_PER_FLOW_RATE = 19.0 * ureg("W/gpm")
```
charliepnnl, line 19: Would it make sense to call this `REQUIRED_PUMP_POWER_PER_FLOW_RATE`?  I think that would help the readability of the rule_check.

charliepnnl (approved):
Approved with some comments that should be considered

rct229/rules/section21/section21rule9.py
charliepnnl: In that case, the RDS should probably be updated to match.

yunjoonjung-PNNL: Fixed. Thank you for this comment. This is brilliant.

PR #636, opened 2022-09-08
Create Rule 19-1.md

claperle:


docs/section19/Rule 19-1.md
```diff
@@ -0,0 +1,52 @@
...
+            - Check if the baseline hvac system type has a furnace or heat pump coil at the HVAC system level: `if sys_type in ["SYS-2", "SYS-3", "SYS-3a", "SYS-4", "SYS-9"]:`  
+                - Get the over sizing factor associated with the heating coil: `heating_oversizing_factor = hvac.heating_system.oversizing_factor`     
+            - Else: `Else: heating_oversizing_factor = "n/a"` 
+            - Check if the baseline hvac system type has a DX cooling coil at the HVAC system level: `if sys_type in ["SYS-1", "SYS-1b", "SYS-2", "SYS-3", "SYS-3b", "SYS-4", "SYS-5", "SYS-5b", "SYS-6", "SYS-6b"]:`  
+                - Get the over sizing factor associated with the cooling coil: `cooling_oversizing_factor = hvac.cooling_system.oversizing_factor`        
+            - Else: `Else: cooling_oversizing_factor = "n/a"`   
+
+            - **Rule Assertion:** 
+            - Case 1: If heating_oversizing_factor = 25% and cooling_oversizing_factor = 15% then pass: `if heating_oversizing_factor == 25% AND cooling_oversizing_factor == 15%: outcome = "PASS"`
```
mkarpman, line 40: Since another rule related to sizing (19-3) is not primarly, I think we should also check is_oversized=TRUE for both HeatingSystem and CoolingSystem

docs/section19/Rule 19-1.md
```diff
@@ -0,0 +1,52 @@
...
+- Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`  
+- Check if this rule is applicable to any HVAC systems in the B_RMR, if not then the outcome is IN_APPLICABLE for all HVAC systems: `If any(hvac.id in baseline_hvac_system_dict[sys_type] for sys_type in ["SYS-1", "SYS-1b","SYS-2", "SYS-3", "SYS-3a", "SYS-4", "SYS-5", "SYS-5b", "SYS-6", "SYS-6b", "SYS-9"]):`   
+    - For each hvac system in B_RMR: `for hvac in B_RMR...HeatingVentilationAirConditioningSystem:`    
+        - Reset heating_oversizing_factor: `heating_oversizing_factor = ""`  
+        - Reset cooling_oversizing_factor: `cooling_oversizing_factor = ""`  
+        - Get baseline system type: `sys_type = list(baseline_hvac_system_dict.keys())[list(baseline_hvac_system_dict.values()).index(hvac.id)]`  
+        **Applicability Check 2:** At the component level
+        - Check if the baseline system type has a furnace, heat pump, or DX cooling coil, ELSE then rule outcome for this HVAC system is NOT_APPLICABLE: `if sys_type in ["SYS-1", "SYS-1b","SYS-2", "SYS-3", "SYS-3a", "SYS-4", "SYS-5", "SYS-5b", "SYS-6", "SYS-6b", "SYS-9"]:`
```
supriyagoel, line 31: @claperle for checks like these where we're identifying whether heating system or cooling system is a certain type, is there value in simply calling the sub-function for is cooling DX or is heating furnace rather than the system type function? i do not know if one works better than the other

claperle: @supriyagoel This is a good point. I'll think about this going forward as I am not sure which is better in terms of ease of creating the RDS and code efficiency at the moment. Are you on board with leaving this one as is?

docs/section19/Rule 19-1.md
claperle: @mkarpman I'm thinking you meant is_autosized and I updated accordingly.

docs/section19/Rule 19-1.md
```diff
@@ -0,0 +1,52 @@
+# Section 19 - Rule 19-1  
+**Schema Version:** 0.0.21  
+**Mandatory Rule:** True  
+**Rule ID:** 19-1   
+**Rule Description:** HVAC system coil capacities for the baseline building design shall be oversized by 15% for cooling and 25% for heating..  
+**Rule Assertion:** Options are Pass/Fail/IN_APPLICABLE
```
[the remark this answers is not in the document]
claperle: Revised.

mkarpman (approved):

KarenWGard:
line 29: "Get baseline system type: sys_type = list(baseline_hvac_system_dict.keys())[list(baseline_hvac_system_dict.values()).index(hvac.id)]"

appears to be no longer necessary ("sys_type" is not referenced anywhere else in the rule)

docs/section19/Rule 19-1.md
```diff
@@ -0,0 +1,50 @@
...
+**Applicability Check 1:**  Checks each HVAC system for applicability  (does it have a furnace, heatpump, and/or DX cooling) otherwise NOT_APPLICABLE
+- For each hvac system in B_RMR: `for hvac in B_RMR...HeatingVentilationAirConditioningSystem:`    
+    - Reset heating_oversizing_factor: `heating_oversizing_factor = ""`  
+    - Reset cooling_oversizing_factor: `cooling_oversizing_factor = ""`  
+    - Check if the baseline system type has a furnace, heat pump, and/or DX cooling coil, ELSE then rule outcome for this HVAC system is NOT_APPLICABLE: `if is_hvac_sys_heating_type_furnace(B_RMR, hvac.id) == TRUE OR is_hvac_sys_heating_type_heat_pump(B_RMR, hvac.id) ==  TRUE OR is_hvac_sys_cooling_type_DX(B_RMR, hvac.id) == TRUE:`   
+        - Check if the baseline hvac system type has a furnace or heat pump coil at the HVAC system level: `if is_hvac_sys_heating_type_furnace(B_RMR, hvac.id) == TRUE OR is_hvac_sys_heating_type_heat_pump(B_RMR, hvac.id) ==  TRUE:`  
+            - Get the over sizing factor associated with the heating coil: `heating_oversizing_factor = hvac.heating_system.oversizing_factor`     
+        - Else: `Else: heating_oversizing_factor = "n/a"`
```
weilixu, line 32: If possible, we want to avoid change data type of a parameter. In this case, it seems the oversizing factor is assigned to a number (if success) or a string (if failed). Is there a particular reason for that?

claperle: @weilixu great point. I updated the structure. I was not sure how to reset the number variable, I guess I could use del but I figure this is easy for the RCT to take care of when implementing.

weilixu (approved):

PR #638, opened 2022-09-09
RS/YJ/RULE_21_7

yunjoonjung-PNNL:


charliepnnl (approved):
Approved with comments that should be addressed

charliepnnl:
I forgot about the need to use `CalcQ` in `get_calc_vals` from now on.
We should add a pytest for CalcQ to make sure it handles measured temperatures correctly.

rct229/schema/config_functions.py
```diff
@@ -11,7 +11,7 @@ def get_pint_unit_registry():
     path_to_units = os.path.join(
         os.path.dirname(__file__), "resources", "unit_registry.txt"
     )
-    ureg = pint.UnitRegistry(path_to_units)
+    ureg = pint.UnitRegistry(path_to_units, autoconvert_offset_to_baseunit=True)
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: When this was not set, there was an error with ```pint.errors.OffsetUnitCalculusError: Ambiguous operation with offset unit (degree_Fahrenheit, ). See https://pint.readthedocs.io/en/latest/nonmult.html for guidance.``` message. The ```pint``` document says that offset units can't be created like other quantities by multiplication of magnitude. And the ```autoconvert_offset_to_baseunit=True``` setting doesn't convert the unit to base unit but remains unchanged, which doesn't cause the error. Please also see this [link ](https://pint.readthedocs.io/en/latest/nonmult.html#f1:~:text=Multiplication%20of%20a%20quantity%20with%20a%20single%20offset%20unit%20with%20order%20%2B1%20by%20a%20number%20or%20ndarray%20yields%20the%20quantity%20in%20the%20given%20unit.)for better explanation.

weilixu:
> I forgot about the need to use `CalcQ` in `get_calc_vals` from now on. We should add a pytest for CalcQ to make sure it handles measured temperatures correctly.

One comment to the CalcQ is that we should always to cautious specifying values as quantity in any calculations.
The current workflow calls CalcQ to convert `quantity` to string. If a output value calculated in the calc_val is not quantity, then an exception will raise and stop the execution.

For example:
```
total_space_area = pint_sum([space_area in space in spaces], 0.0) 
```
In the above line, the total_space_area will raise exception in the output.

```
total_space_area = pint_sum([space_area in space in spaces], ZERO.AREA) 
```
A correct way is to define the default to ZERO class.

We have made a revision to all such values in the previous round of enhancement. Going forward, we should stick to this programming guidance.

docs/section21/Rule21-7.md
```diff
@@ -29,7 +30,7 @@
 
 ## Rule Logic:  
 
-- For each boiler in B_RMR, save boiler to loop boiler dictionary: `for boiler_b in B_RMR.RulesetModelInstance.boilers: loop_boiler_dict[boiler_b.loop].append(boiler_b)`
+- For each boiler in B_RMI, save boiler to loop boiler dictionary: `loop_boiler_dict = find_all(B-RMI)`
```
weilixu, line 33: The modification does not seems to save boilers to loop boiler dictionary.

If this is a list, we should explicitly called this out as a list.

weilixu: One more comment is: `find_all` is a utility function. Generally, in the RDS we only put `ruleset specific utility` function. A utility function is usually expanded in the RDS or give a more explicit explanations.
In this case, the line can be updated to:

```
loop_boiler_ids = find_all(boiler.id, B_RMI)
```
or
```
for boiler_b in B_RMI.RulesetModelInstance.boilers:
    loop_boiler_ids.append(boiler_b.id)
```

weilixu (approved):
Approved with comments on the RDS needs to be addressed

docs/section21/Rule21-7.md
yunjoonjung-PNNL: Fixed. Thanks for letting me know about this Weili.

PR #568, opened 2022-08-02
Rt/yj/ruletest 6 1

yunjoonjung-PNNL:
This PR includes 
1. Updated testing JSON file
2. The code updates that reflect the 6-1 RDS

The changes from the 6-2 JSON testing file should be fine because these changes are identical to the ones in the 6-2 PR.

I'd appreciate it if you reviewed this PR. Thank you!

PR #655, opened 2022-09-20
RS+RT/YJ/RULE 22-3

yunjoonjung-PNNL:
I made a mistake while I was working on 6-1 (PR https://github.com/pnnl/ruleset-checking-tool/pull/568) - I worked on this task right after working on 6-1. However, I believe this shouldn't be a problem if this branch or the 6-1 branch is merged.

Further, for the JSON file, based on the system type 7 template, I added cooling_source key under the terminals and id, heating_system_type keys under the heating_system to make this JSON file also fall into system 11.1. This is because system type 7 doesn't include some keys that are required for system 11.1.


yunjoonjung-PNNL:
I made a mistake while I was working on 6-1 (PR #568). However, I believe this shouldn't be a problem if this branch or 6-1 branch is merged.

rct229/rules/section22/section22rule3.py
```diff
@@ -0,0 +1,88 @@
...
+]
+NOT_APPLICABLE_SYS_TYPES = [
+    "SYS-11.1",
+    "SYS-11.2",
+    "SYS-11.1A",
+    "SYS-11.2A",
+    "SYS-11B",
+    "SYS-11C",
+]
+TEMP_RETURN_TYPE = schema_enums["TemperatureResetOptions"]
```
charliepnnl, line 25: Did you mean this to be `Temp_REST_TYPE`?

rct229/rules/section22/section22rule3.py
```diff
@@ -0,0 +1,88 @@
...
+        return any(
+            [key in APPLICABLE_SYS_TYPES for key in baseline_system_types.keys()]
+        ) and not any(
+            [key in NOT_APPLICABLE_SYS_TYPES for key in baseline_system_types.keys()]
+        )
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        chiller_loop_ids = find_all("chillers[*].cooling_loop", rmi_b)
+        return {"loop_chiller_dict": chiller_loop_ids}
```
charliepnnl, line 60: I think you mean:
```
return {"chiller_loop_ids": chiller_loop_ids}
```
?

rct229/rules/section22/section22rule3.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def list_filter(self, context_item, data):
+        fluid_loop_b = context_item.baseline
+        loop_chiller_dict = data["loop_chiller_dict"]
```
charliepnnl, line 64: `chiller_loop_ids`?

charliepnnl (approved):
Approved `section22rule3.py` with minor comments.

rct229/rules/section22/section22rule3.py
weilixu, line 60: Since this is a list, I suggest to change the `_dict` to `_ids_list`.

rct229/rules/section22/section22rule3.py
```diff
@@ -0,0 +1,88 @@
...
+        return fluid_loop_b["id"] in loop_chiller_dict
```
weilixu, line 65: See comments above

rct229/rules/section22/section22rule3.py
```diff
@@ -0,0 +1,88 @@
...
+
+    class ChillerFluidLoopRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section22Rule3.ChillerFluidLoopRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                required_fields={
+                    "$": ["cooling_or_condensing_design_and_control"],
+                    "cooling_or_condensing_design_and_control": [
```
weilixu, line 73: I agree the `cooling_or_condensing_design_and_control` is a required field in a chiller fluid loop but the `temperature_reset_type` may not the same case.
In a scenario where the `cooling_or_condensing_design_and_control` is missing means, the chiller has no control, which is nearly impossible.

However, there could be various types of controls under this data group. If the RMD uses the wrong control, which means it used a scheduled temperature instead of a temperature reset, we want to capture this error and report it in the outcome. Putting in the required field will just flags the context as a data missing error, which seems not the intention of this rule check.

rct229/rules/section22/section22rule3.py
```diff
@@ -0,0 +1,88 @@
...
+                        "temperature_reset_type"
+                    ],
+                },
+            )
+
+        def get_calc_vals(self, context, data=None):
+            fluid_loop_b = context.baseline
+            temperature_reset_type = fluid_loop_b[
+                "cooling_or_condensing_design_and_control"
+            ]["temperature_reset_type"]
```
weilixu, line 83: In the case, we can use `find_one` function.

`find_one("$cooling_or_condensing_design_and_control.temperature_reset_type", fluid_loop_b)`

If the data is missing, the `find_one` will return None, which is OK.because `None == TEMP_RETURN_TYPE.OUTSIDE_AIR_RESET` is false - and this will be captured in the output report as a failed case.

weilixu (approved):
Approved with some comments need to be addressed

rct229/rules/section22/section22rule3.py
yunjoonjung-PNNL: Fixed.

rct229/rules/section22/section22rule3.py
yunjoonjung-PNNL: Fixed.

rct229/rules/section22/section22rule3.py
yunjoonjung-PNNL: Fixed.

rct229/rules/section22/section22rule3.py
yunjoonjung-PNNL: Fixed.

rct229/rules/section22/section22rule3.py
yunjoonjung-PNNL: Great point! Thank you for letting me know about this.

rct229/rules/section22/section22rule3.py
yunjoonjung-PNNL: Fixed.

weilixu:
Thanks for addressing the comments - this can be merged now

PR #661, opened 2022-09-21
Complete function are_all_terminal_types_cav

weilixu:


yunjoonjung-PNNL (approved):
Approved. Thank you for this effort.

docs/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_types_CAV.md
```diff
@@ -7,7 +7,7 @@
 - **terminal_unit_id_list**: List of terminal units to assess.  
 
 **Returns:**  
-- **are_all_terminal_types_CAV**: The function returns TRUE if all of the terminal unit types input to this function are constant air volume (CAV) if this data element is undefined. It returns FALSE if any of the terminal units are of a type other than constant air volume (CAV).
```
charliepnnl, line 10: On develop, this says "... are constant air volume (CAV) **or** if this data element is undefined. ..."
Not sure where the "or" went, but with it in, isn't it correct and helpful?

charliepnnl (approved):
Approved with one minor comment.

docs/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_types_CAV.md
weilixu: I agree the write-up is very confusing here but looking at the logic at [line 18](https://github.com/pnnl/ruleset-checking-tool/blob/59e5a13de35988791b2f38dca82af9f1e1986db7/docs/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/are_all_terminal_types_CAV.md?plain=1#L18), it seems the intent is the terminal unit type is not undefined and it is constant air volume (CAV).
This is should be true for all the other sub functions: `are_all_terminal_types_VAV` or `are_all_terminal_types_CAV_With_None_Equal_to_Null` but none of these functions mentioned the type NULL check. I wanted to be consistent here so I deleted this line.

PR #663, opened 2022-09-21
Complete function is_hvac_sys_heating_type_furnace

weilixu:


yunjoonjung-PNNL (approved):
Approved.

docs/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/is_hvac_sys_heating_type_furnace.md
```diff
@@ -7,16 +7,13 @@
 - **hvac_b.id**: The id of the hvac system to evaluate.
```
charliepnnl, line 7: I think the Description should be updated also to remove "or if it has more than 1 heating system".

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/is_hvac_sys_heating_type_furnace.py
```diff
@@ -0,0 +1,33 @@
...
+            RMD at RuleSetModelInstance level
+        hvac_b_id : str
+            The HVAC system ID.
+        Returns
+        -------
+        bool
+            True: the HVAC system heating system has furnace as the heating type
+            False: the HVAC system has a heating system type other than furnace
+    """
+    hvac_b = find_exactly_one_with_field_value(
```
charliepnnl, line 23: It occurs to me that we are seeing this nearly identical logic in several functions.
What would you think of creating a utility function like:
```
def find_hvac_system(rmi, hvac_id):
    return find_exactly_one_with_field_value(
         "$.buildings[*].building_segments[*].heating_ventilation_air_conditioning_systems[*]",
            "id",
            hvac_b_id,
            rmi_b,
    )

It is ruleset specific, so it would go in the ruleset_functions folder somewhere.

charliepnnl (approved):
Approved with minor comments.

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/is_hvac_sys_heating_type_furnace.py
weilixu: I think it is the case but there are cases we repeatedly using the same function to extract terminals or zones too.
It might be good to thinking from the top view and see which direction we would like to go  - can discuss on tomorrow's standup.

PR #665, opened 2022-09-23
RS/YJ/RULE 22-4

yunjoonjung-PNNL:
I'd appreciate it if you reviwed this PR.


rct229/rules/section22/section22rule4.py
```diff
@@ -0,0 +1,148 @@
...
+]
+NOT_APPLICABLE_SYS_TYPES = [
+    "SYS-11.1",
+    "SYS-11.2",
+    "SYS-11.1A",
+    "SYS-11.2A",
+    "SYS-11B",
+    "SYS-11C",
+]
+REQUIRED_OUTDOOR_HIGH_LOOP_SUPPLY_TEMP_RESET = 80 * ureg("degF")
```
charliepnnl, line 27: Now I think I understand why there was an issue with the measured temperatures like this.  I think the preferred way to express a measured temperature would be `ureg("80 degF")`. This avoids the multiplication that was causing the problem.

rct229/rules/section22/section22rule4.py
```diff
@@ -0,0 +1,148 @@
...
+        return any(
+            [key in APPLICABLE_SYS_TYPES for key in baseline_system_types.keys()]
+        ) and not any(
+            [key in NOT_APPLICABLE_SYS_TYPES for key in baseline_system_types.keys()]
+        )
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        chiller_loop_ids = find_all("chillers[*].cooling_loop", rmi_b)
+        return {"loop_chiller_dict": chiller_loop_ids}
```
charliepnnl, line 65: Do you mean `return {"chiller_loop_ids": chiller_loop_ids}`?

rct229/rules/section22/section22rule4.py
```diff
@@ -0,0 +1,148 @@
...
+        # if any system type found in the APPLICABLE_SYS_TYPES then return applicable.
+        return any(
+            [key in APPLICABLE_SYS_TYPES for key in baseline_system_types.keys()]
+        ) and not any(
+            [key in NOT_APPLICABLE_SYS_TYPES for key in baseline_system_types.keys()]
+        )
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        chiller_loop_ids = find_all("chillers[*].cooling_loop", rmi_b)
```
charliepnnl, line 64: The RDS should be updated to match the use of a list instead of a dictionary.

charliepnnl (approved):
Approved with minor comments.

rct229/rules/section22/section22rule4.py
yunjoonjung-PNNL: Was this intention of the change to get rid of ```.to(ureg.kelvin)``` part in the ```rule_check``` method? I tested whether I can remove the Kelvin conversion part, However, I still got an error after the adjustment.

If the intention was to remove the Kelvin conversion part, I will dig more into this. If not, I will leave the adjustment as is. I'd appreciate it if @charliepnnl let me know about this!

weilixu (approved):
Looks good, thanks!

rct229/rules/section22/section22rule4.py
```diff
@@ -0,0 +1,148 @@
...
+        loop_chiller_dict = data["loop_chiller_dict"]
+        return fluid_loop_b["id"] in loop_chiller_dict
+
+    class ChillerFluidLoopRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section22Rule4.ChillerFluidLoopRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                required_fields={
+                    "$": ["cooling_or_condensing_design_and_control"],
+                    "cooling_or_condensing_design_and_control": [
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: This totally makes sense and I appreciate for letting me know about this!

PR #666, opened 2022-09-26
RS+RT/YJ/Rule 22-5

yunjoonjung-PNNL:
I'd appreciate it if you could review this PR.


rct229/rules/section22/section22rule5.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        chiller_loop_ids_list = find_all("chillers[*].cooling_loop", rmi_b)
+        return {"chiller_loop_ids_list": chiller_loop_ids_list}
+
+    def list_filter(self, context_item, data):
+        fluid_loop_b = context_item.baseline
+        loop_chiller_list = data["chiller_loop_ids_list"]
```
charliepnnl, line 51: Wouldn't it be clearer to call this `chiller_loop_ids_list` instead of `loop_chiller_list`?

rct229/rules/section22/section22rule5.py
```diff
@@ -0,0 +1,77 @@
...
+
+APPLICABLE_SYS_TYPES = [
+    "SYS-11.1",
+    "SYS-11.2",
+    "SYS-11.1A",
+    "SYS-11.2A",
+    "SYS-11B",
+    "SYS-11C",
+]
+TEMP_RESET_OPTION = schema_enums["TemperatureResetOptions"]
```
charliepnnl, line 15: I wonder if it is better to call this `TEMPERATURE_RESET` without the `_OPTION`. Since this is an enumeration, the `_OPTION` or similar word is implied.

charliepnnl (approved):
Looks good. I just had a couple of comments you could consider.

rct229/rules/section22/section22rule5.py
yunjoonjung-PNNL: Fixed. Thanks for this Charlie.

weilixu (approved):
Looks good. Thanks

PR #672, opened 2022-09-27
Rds/cml/rule 19 2

claperle:


docs/section19/19-2.md
```diff
@@ -0,0 +1,46 @@
...
+1. Applies to buildings with the following baseline system types: 1, 1a, 5, 5a, 7, 7a, 7b, 8, 8b, 11.1, 11b, 11.2, 11.2a, 12, 12a, 12b, 13 (buildings with chiller or boiler plants)  
+
+**Function Calls:**  
+1. get_baseline_system_types()  
+
+
+## Rule Logic:  
+**Applicability Check 1:**  
+- Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`  
+- Check if this rule is applicable to any HVAC systems in the B_RMR, if not then the outcome is IN_APPLICABLE: `If any(hvac.id in baseline_hvac_system_dict[sys_type] for sys_type in ["SYS-1", "SYS-1a","SYS-5", "SYS-5a", "SYS-7", "SYS-7a", "SYS-7b", "SYS-8", "SYS-8b", "SYS-11.1", "SYS-11b", "SYS-11.2", "SYS-11.2a", "SYS-12", "SYS-12a", "SYS-12b", "SYS-13"]):`
```
mkarpman, line 25: We don't have 5a?

docs/section19/19-2.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+## Rule Logic:  
+**Applicability Check 1:**  
+- Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`  
+- Check if this rule is applicable to any HVAC systems in the B_RMR, if not then the outcome is IN_APPLICABLE: `If any(hvac.id in baseline_hvac_system_dict[sys_type] for sys_type in ["SYS-1", "SYS-1a","SYS-5", "SYS-5a", "SYS-7", "SYS-7a", "SYS-7b", "SYS-8", "SYS-8b", "SYS-11.1", "SYS-11b", "SYS-11.2", "SYS-11.2a", "SYS-12", "SYS-12a", "SYS-12b", "SYS-13"]):`   
+    Create list of plant loops:
+    - For each boiler in B_RMR, save boiler to loop boiler list: `for boiler_b in B_RMR.RulesetModelInstance.boilers: loop_boiler_list.append(boiler_b.loop)`  
+    - For each chiller in B_RMR, save chiller to loop chiller list: `for chiller_b in B_RMR.RulesetModelInstance.chillers: loop_chiller_list.append(chiller_b.cooling_loop)` 
+    - Create a list of all plant loops by combining boiler and chiller lists and eliminating duplicates: `plant_loops_list_b = list(set(list(loop_chiller_list + loop_boiler_list)))`
```
mkarpman, line 29: I think we need to exclude loops that are unrelated to HVAC, e.g.  serve process heating or cooling. So I think we need to isolate loops referenced by the baseline HVAC systems and apply the logic to these loops only.

docs/section19/19-2.md
```diff
@@ -0,0 +1,46 @@
...
+    - For each chiller in B_RMR, save chiller to loop chiller list: `for chiller_b in B_RMR.RulesetModelInstance.chillers: loop_chiller_list.append(chiller_b.cooling_loop)` 
+    - Create a list of all plant loops by combining boiler and chiller lists and eliminating duplicates: `plant_loops_list_b = list(set(list(loop_chiller_list + loop_boiler_list)))`  
+    - For each plant loop in the plant loop list: `for plant_loop_b in plant_loops_list_b:`    
+        - Reset FluidLoopDesignAndControl_cool_b: `FluidLoopDesignAndControl_cool_b = ""`
+        - Reset FluidLoopDesignAndControl_heat_b: `FluidLoopDesignAndControl_heat_b = ""`  
+        - if plant_loop_b.type is cooling then: `if plant_loop_b.type == "COOLING":`  
+            - Get cooling_or_condensing_design_and_control FluidLoopDesignAndControl object: `FluidLoopDesignAndControl_cool_b = plant_loop_b.cooling_or_condensing_design_and_control`  
+        - Elif plant_loop_b.type is heating then: `elif plant_loop_b.type == "HEATING":`  
+            - Get heating_design_and_control FluidLoopDesignAndControl object: `FluidLoopDesignAndControl_heat_b = plant_loop_b.heating_design_and_control`
```
mkarpman, line 36: I wonder at which point do we check that each FluidLoop has a cooling_or_condensing_design_and_control object not equal to Null?  This RDS assumes that it's present for each loop. This is kind of checked in the assertion because it produces FAIL outcome if we have one of the applicable HVAC systems in the B_RMR but there are no FluidLoopDesignAndControl object. But if there are two loops and one is missing the  FluidLoopDesignAndControl then it will not be flagged.

docs/section19/19-2.md
supriyagoel: same comment as #636 iss there value in checking if H/C provided through fluid loop instead of the baseline system type?
essentially, a HVAC might not get categorized as baseline system X due to some inconsistency but having a fluid loop would allow us to verify 19-2

claperle: @supriyagoel Now that I see this maybe I will update 19-1

docs/section19/19-2.md
claperle: Updated

docs/section19/19-2.md
claperle: @mkarpman Updated this RDS completely. I think this will flag if this object is not defined. Let me know if you disagree.

KarenWGard:
are district plants not included?  There is still some autosizing (loop flow & volume, pump size, etc), even if boilers / chillers are not sized

docs/section19/19-2.md
```diff
@@ -0,0 +1,74 @@
...
+5. is_hvac_sys_preheating_type_fluid_loop()  
+6. is_hvac_sys_preheat_fluid_loop_attached_to_boiler()  
+7. are_all_terminal_heating_loops_attached_to_boiler()
+
+
+
+## Rule Logic:   
+- Cycle through each HVAC system in the B-RMR to create a list of HW loops for space heating served by boiler(s) and a list of CHW loop for space cooling served by chiller(s) and a list of condenser water loops: `for each hvac in B_RMR...HeatingVentilatingAirConditioningSystem:`  
+    - Check if the hvac system heating coil is a fluid loop: `if is_hvac_sys_heating_type_fluid_loop(B_RMR, hvac.id)  == TRUE:`  
+        - Check if the fluid loop is attached to a boiler, if yes then add to list of HW fluid loops: `if is_hvac_sys_fluid_loop_attached_to_boiler(B_RMR, hvac.id) ==  TRUE: HW_fluid_loop_list =  HW_fluid_loop_list.append(hvac.heating_system.hot_water_loop)`
```
weilixu, line 32: if using append, we do not need to add `HW_fluid_loop_list = `. This line can be modified:
`if is_hvac_sys_fluid_loop_attached_to_boiler(B_RMR, hvac.id) == TRUE: HW_fluid_loop_list.append(hvac.heating_system.hot_water_loop)`

docs/section19/19-2.md
```diff
@@ -0,0 +1,74 @@
...
+        - Reset FluidLoopDesignAndControl_cool_b: `FluidLoopDesignAndControl_cool_b = ""`
+        - Reset FluidLoopDesignAndControl_heat_b: `FluidLoopDesignAndControl_heat_b = ""`  
+        - if the loop is in the CHW fluid loop: `if fluid_loop in CHW_fluid_loop_list:`  
+            - Get cooling_or_condensing_design_and_control FluidLoopDesignAndControl object: `FluidLoopDesignAndControl_cool_b = fluid_loop.cooling_or_condensing_design_and_control`  
+        - Elif the loop is in the CW fluid loop: `if fluid_loop in CW_fluid_loop_list:`  
+            - Get cooling_or_condensing_design_and_control FluidLoopDesignAndControl object: `FluidLoopDesignAndControl_cool_b = fluid_loop.cooling_or_condensing_design_and_control`  
+        - Elif the loop is in the HW fluid loop: `if fluid_loop in HW_fluid_loop_list:`  
+            - Get heating_design_and_control FluidLoopDesignAndControl object: `FluidLoopDesignAndControl_heat_b = fluid_loop.heating_design_and_control`  
+
```
weilixu, line 63: I dont quite understand the applicability check here.
Does the actual check ends at this line?
`if fluid_loop in HW_fluid_loop_list OR fluid_loop in CHW_fluid_loop_list OR fluid_loop in CW_fluid_loop_list:`

docs/section19/19-2.md
```diff
@@ -0,0 +1,74 @@
...
+        - **Rule Assertion:** 
+        - Case 1: If FluidLoopDesignAndControl_cool_b.is_sized_using_coincident_load = TRUE and FluidLoopDesignAndControl_heat_b = "" then pass: `if FluidLoopDesignAndControl_cool_b.is_sized_using_coincident_load == TRUE AND FluidLoopDesignAndControl_heat_b == "": outcome = "PASS"`  
+        - Case 2: Else if FluidLoopDesignAndControl_heat_b.is_sized_using_coincident_load = TRUE and FluidLoopDesignAndControl_cool_b = "" then pass: `if FluidLoopDesignAndControl_heat_b.is_sized_using_coincident_load == TRUE AND FluidLoopDesignAndControl_cool_b == "": outcome = "PASS"`
```
weilixu, line 66: Could we have just one parameter: `fluid_loop_design_and_control` and use `type` under the `fluid_loop` to distinguish heat and cool?

This could be possibly rewrite as:

```python
is_sized_using_coincident_load = False
if fluid_loop.type == COOLING and fluid_loop_id in CHW_fluid_loop_list:
    is_sized_using_coincident_load = fluid_loop.cooling_or_condensing_design_and_control.is_sized_using_concident_load
elif fluid_loop.type == HEATING and fluid_loop_id in HW_fluid_loop_list:
      is_sized_using_coincident_load = fluid_loop.heating_design_and_control.is_sized_using_concident_load
...

In rule assertion at component level:
is_sized_using_coincident_load == True -> PASS
else:
FAIL.
```

docs/section19/19-2.md
claperle: Thanks!

docs/section19/19-2.md
claperle: Yes exactly.

docs/section19/19-2.md
claperle: Updated.

docs/section19/19-2.md
weilixu: Got it. Thanks

weilixu (approved):

docs/section19/19-2.md
```diff
@@ -0,0 +1,74 @@
...
+            - Add CHW fluid loop to chilled water loop list: `CHW_fluid_loop_list =  CHW_fluid_loop_list.append(hvac.cooling_system.chilled_water_loop)`    
+            - Add condenser water fluid loop to condenser water loop list: `CW_fluid_loop_list =  CW_fluid_loop_list.append(hvac.cooling_system.condenser_water_loop)`  
+            - Add the child loops to the CHW loop list: `For CHW_child_loop in hvac.cooling_system.chilled_water_Loop.child_loops:`  
+                - Add to the list of chilled water loops: `CHW_fluid_loop_list =  CHW_fluid_loop_list.append(CHW_child_loop)`             
+- Cycle through each terminal unit in the B-RMR and create list of HW loops: `for terminal in B_RMR...Terminal:`  
+    - Reset list with the one terminal unit: `terminal_list = []` 
+    - Check if heating source is hot water: `if terminal.heating_source == "HOT_WATER":`  
+        - Create a list with the one terminal unit: `terminal_list = [terminal.id]`  
+        - Check if the loop is attached to a boiler: `if are_all_terminal_heating_loops_attached_to_boiler(B_RMR,terminal_list) ==  TRUE:`
```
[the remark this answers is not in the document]
claperle: @mkarpman We are only sending that function one terminal unit. The only reason I create a list with one value is because that function expects a list but there is only one terminal being sent to the function. Does this make your comment obsolete?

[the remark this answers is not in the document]
KarenWGard: instead of creating terminal_list, you could amend line 45 to read:

`if are_all_terminal_heating_loops_attached_to_boiler(B_RMR,[terminal.id]) ==  TRUE:`

this makes it a little more obvious.

Also, the check on line 46 is redundant, lines 48-50 already eliminate duplicates

mkarpman (approved):

PR #675, opened 2022-09-28
Complete is_hvac_system_multizone

weilixu:


rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/is_hvac_system_multizone.py
```diff
@@ -0,0 +1,16 @@
+def is_hvac_system_multizone(rmi_b, zone_id_list):
```
charliepnnl, line 1: As I've mentioned elsewhere: do we really want to `_b` since this function would equally apply to any RMI?

charliepnnl (approved):
Approved with one question in a comment.

rct229/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/is_hvac_system_multizone.py
weilixu: The intention of these functions are only applies to baseline model, not proposed or user model.

PR #678, opened 2022-09-28
RS+RT/YJ/RULE 22-19

yunjoonjung-PNNL:


rct229/rules/section22/section22rule19.py
```diff
@@ -0,0 +1,74 @@
...
+    "SYS-11.2",
+    "SYS-12",
+    "SYS-13",
+    "SYS-7B",
+    "SYS-8B",
+    "SYS-11B",
+    "SYS-12B",
+    "SYS-13B",
+]
+REQUIRED_TEMP_RESET = schema_enums["TemperatureResetOptions"]
```
charliepnnl, line 20: I suggest using the name of the enum without the "Options" at the end.  So in this case it would be `TEMPERATURE_RESET`

rct229/rules/section22/section22rule19.py
```diff
@@ -0,0 +1,74 @@
...
+        }
+        # if any system type found in the APPLICABLE_SYS_TYPES then return applicable.
+        return any(
+            [key in APPLICABLE_SYS_TYPES for key in baseline_system_types.keys()]
+        )
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        temp_reset_type_dict_b = {
+            heat_rejection_loop: find_exactly_one_with_field_value(
```
charliepnnl, line 53: We should probably create a `find_exactly_one_heat_rejection(rmi, heat_rejection_id)` function like Weily has done for hvac systems and others.

rct229/rules/section22/section22rule19.py
```diff
@@ -0,0 +1,74 @@
...
+                "$..fluid_loops[*]", "id", heat_rejection_loop, rmi_b
+            )["cooling_or_condensing_design_and_control"]["temperature_reset_type"]
```
charliepnnl, line 55: We should discuss how best to handle cases where the `cooling_or_condensing_desing_and_control` or `temperature_rest_type` fields are missing. I believe these should be required fields. I believe that `loop` should also be a required field for all `heat_rections[*]`.

charliepnnl (approved):
Approved with some comments that should be addressed.

rct229/rules/section22/section22rule19.py
```diff
@@ -0,0 +1,74 @@
...
+
+        def get_calc_vals(self, context, data=None):
+            heat_rejection_b = context.baseline
+            loop_b = heat_rejection_b["loop"]
+            temperature_reset_type = data["temp_reset_type_dict_b"][loop_b]
+            return {"temperature_reset_type": temperature_reset_type}
+
+        def rule_check(self, context, calc_vals=None, data=None):
+            temperature_reset_type = calc_vals["temperature_reset_type"]
+            return temperature_reset_type == REQUIRED_TEMP_RESET.CONSTANT
```
weilixu, line 74: It is OK to have `None` here because `None` will return Failed.

rct229/rules/section22/section22rule19.py
weilixu: Update `temp_reset_type_dict_b` to `heat_rejection_loop_dict`
```python
heat_rejection_loop_dict = {
    heat_rejection_loop_id: find_exactly_one_with_field_value("$..fluid_loops[*]", "id", heat_rejection_loop, rmi_b)
     for heat_rejection_loop_id in find_all("heat_rejections[*].loop, rmi_b)
```

rct229/rules/section22/section22rule19.py
```diff
@@ -0,0 +1,74 @@
...
+    class HeatRejectionRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section22Rule19.HeatRejectionRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+            )
+
+        def get_calc_vals(self, context, data=None):
+            heat_rejection_b = context.baseline
+            loop_b = heat_rejection_b["loop"]
+            temperature_reset_type = data["temp_reset_type_dict_b"][loop_b]
```
weilixu, line 69:
```python
data["heat_rejection_loop_dict][loop_b] = heat_rejection_loop
getattr_(,....).
```

rct229/rules/section22/section22rule19.py
yunjoonjung-PNNL: I will constantly use (schema_enums_name without options). Thanks for this Charlie.

rct229/rules/section22/section22rule19.py
yunjoonjung-PNNL: Great point! I reflected on this in the code. Thank you for letting me know about this :).

weilixu (approved):
Looks good thanks

PR #683, opened 2022-09-30
Create zone_meets_G3_1_1c

KarenWGard:
G3_1_1c exception test for a zone


docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
+# zone_meets_G3_1_1c
+
+**Schema Version:** 0.0.21
+
+**Description:** determines whether a given zone meets the G3_1_1c exception "If the baseline HVAC system type is 5, 6, 7, 8, 9, 10, 11, 12, or 13 use separate single-zone systems conforming with the requirements of system 3 or system 4 (depending on building heating source) for any spaces that have occupancy or process loads or schedules that differ significantly from the rest of the building. Peak thermal loads that differ by 10 Btu/h·ft2 (2.930710 W/sf) or more from the average of other spaces served by the system, or schedules that differ by more than 40 equivalent full-load hours per week from other spaces served by the system, are considered to differ significantly. Examples where this exception may be applicable include but are not limited to natatoriums and continually occupied security areas. This exception does not apply to computer rooms."
+"average" is not particularly well defined in this exception. 
+for the peak W/sf calculation I've assumed:
+	1. the zone being tested is not included
```
mkarpman, line 8: agreed

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
...
+ 	2. the average: (each space coincident peak W) / total_area
+
+**Inputs:** 
+- **RMR**
```
mkarpman, line 12: Should we start calling it RMD? (I think Christina also keeps using RMR abbreviation.)
Also, we need to indicate that this function applies to the B_RMD, not P_ or U_

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
...
+- **zone**
+
+**Returns:**  
+- **result**: a string - either "C" or "No"
```
mkarpman, line 16: I would consider using "Yes" instead of "C" for clarity, so that the result is a direct answer to the function name "zone_meets_G3_1_1c".

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
...
+- **get_zone_peak_internal_load**
+
+## Logic:
+- set the result variable to "No" - only a positive test can give it a different value: `result = "No"`
+- get the building_segment the zone is in:
+`for building_segmentin RMR.building.buildings:`
+	- check if the zone is in the building_segment:
+	`if zone in building_segment.zones:`
+		- break - this means the variable "building_segment" is the building_segment that contains the zone:
+		`break`
```
mkarpman, line 29: There seems to be an existing function that you can use https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/ruleset_functions/get_list_hvac_systems_associated_with_zone.md. Seems like we need the following logic:
1. Call get_list_hvac_systems_associated_with_zone. This will return all HVAC systems associated with the zone.  Since in the baseline we expect to have one system serving each zone, there should only be one system in the list but perhaps we can structure this function to also work for scenarios where there is more than one system.
2. For each system in the list,
      - check if it is multizone (5, 5b, 7, 7a, etc.) via get_baseline_system_types(RMR). If yes,
      - get_dict_of_zones_and_terminal_units_served_by_hvac_sys (or perhaps get_hvac_zone_list_w_area would work better) to get all other zones served by this HVAC system
      - see if the test zone meets either exception criteria

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
...
+- create set of system types that are eligible for this rule AND included:
+`eligible_systems_in_RMR = (set(segment_system_types_b) & set(eligible_primary_system_types))`
+- check if any of the eligible_primary_system_types is in the system_types_b:
+`if((len(eligible_systems_in_RMR)) > 0):`
+	- create a variable to hold the sum total of all peak loads in the system:
+	`system_total_peak_load = 0`
+	- create a variable to hold the sum total of the system area:
+	`system_total_area = 0`
+	- create a list of the EFLH's for each zone `eflh = []`
+	- look through all zones in the building segment:
```
mkarpman, line 51: The function should only consider zones served by the given HVAC system, not all zones in the building segment.

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
...
+	`for zone_b in building_segment.zones:`
+		- check that zone_b is not the zone being investigated:
+		`if zone_b != zone:`
+			- add the zone internal loads to the system_total_peaks:
+			`a = get_zone_peak_internal_load(zone_b)`
+			`system_total_peak += a[0]`
+			`system_total_area += a[1]`
```
mkarpman, line 58: For floor area I would use get_hvac_zone_list_w_area function.

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
...
+			- add the eflh to the eflh list.  We are multiplying the schedule by the zone area so that later we can divide the total by the total area.  Otherwise small zones with very different schedules have too great an impact
+			`eflh.append(a[2] * a[1])`
```
mkarpman, line 60: I am not following how eflh are calculated. It seems that we need a function for that is suggested in line 20. 90.1 language seems to suggest simple average instead of weighted average. Let's discuss on the call.

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
...
+	- now calculate the average internal load: `avg_internal_load = system_total_peak / system_total_area`
+	- and calculate the averge eflh: `avg_eflh = sum(eflh)/len(eflh)`
+	- get the internal load of the reference zone: `internal_loads = get_zone_peak_internal_load(zone)`
+	- now do the rule checks:
+	- if the zone peak differs by more than 2.930710 W/sf from the average, then it meets the exception:
```
mkarpman, line 65: After RCT parces the RMD it will convert all numeric values from metric to IP, so within the RDSs we directly use the values prescribed in the 90.1 IP version.

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
...
+	`if zone in building_segment.zones:`
+		- break - this means the variable "building_segment" is the building_segment that contains the zone:
+		`break`
+- make a list of the heating_ventilation_air_conditioning_systems in the building_segment:
+`relevant_systems = [a.id for a in building_segment.heating_ventilation_air_conditioning_systems]`
+
+
+
+- first check if the system type is one of the following:
+`eligible_primary_system_types = ["SYS-5","SYS-5b","Sys-6","Sys-6b","Sys-7","Sys-7a","Sys-7b","Sys-7c","Sys-8","Sys-8a","Sys-8b","Sys-8c","Sys-8d","Sys-9","Sys-9b","Sys-10","Sys-11.1","Sys-11.1a","Sys-11b","Sys-11c","Sys-12","Sys-12a","Sys-12b","Sys-12c","Sys-13","Sys-13a"]`
```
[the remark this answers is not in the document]
KarenWGard: the text of G3.1.1.b lists 5-13:

If the baseline HVAC system type is 5, 6, 7, 8, 9, 10, 11, 12, or 13 use separate
single-zone systems conforming with the requirements of system 3 or system 4
for any HVAC zones that have occupancy, internal gains, or schedules that differ
significantly from the rest of the HVAC zones served by the system. The total
peak internal gains that differ by 10 Btu/h·ft2 or more from the average of other
HVAC zones served by the system, or schedules that differ by more than 40
equivalent full-load hours per week from other spaces served by the system, are
considered to differ significantly. Examples where this exception may be applicable
include but are not limited to natatoriums and continually occupied security
areas. This exception does not apply to computer rooms.

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
KarenWGard: ok, will change

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
KarenWGard: #1 - awesome.  Much faster
#2 - this line: `segment_system_types_b = {key: system_types_b[0] for key in system_types_b if(len(set(system_types_b[key])&set(relevant_systems)) > 0)` does some fancy code to do the first part

the second part - yes, good points.  I've updated the code

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
KarenWGard: yes, updated.  Thanks

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
KarenWGard: good idea.  Changed

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
KarenWGard: changed to 10 btu/hr/ft2

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
+# zone_meets_G3_1_1c
+
+**Schema Version:** 0.0.21
+
+**Description:** determines whether a given zone meets the G3_1_1c exception "If the baseline HVAC system type is 5, 6, 7, 8, 9, 10, 11, 12, or 13 use separate single-zone systems conforming with the requirements of system 3 or system 4 (depending on building heating source) for any spaces that have occupancy or process loads or schedules that differ significantly from the rest of the building. Peak thermal loads that differ by 10 Btu/h·ft2 (2.930710 W/sf) or more from the average of other spaces served by the system, or schedules that differ by more than 40 equivalent full-load hours per week from other spaces served by the system, are considered to differ significantly. Examples where this exception may be applicable include but are not limited to natatoriums and continually occupied security areas. This exception does not apply to computer rooms."
```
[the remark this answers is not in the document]
KarenWGard: split the eflh part to `get_zone_eflh`

docs/ruleset_functions/G311_exceptions/zone_meets_G3_1_1c
```diff
@@ -0,0 +1,74 @@
...
+**Inputs:** 
+- **RMR**
+- **zone**
+
+**Returns:**  
+- **result**: a string - either "C" or "No"
+ 
+**Function Call:**
+- **get_baseline_system_types**
+- **get_zone_peak_internal_load**
```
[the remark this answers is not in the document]
KarenWGard: created `get_zone_eflh` following these rules

PR #686, opened 2022-10-04
Complete table G3_1_3_7 and G3_5_3

weilixu:


weilixu:
@yunjoonjung-PNNL @charliepnnl 
This is the tables needed for 20 and 22.

yunjoonjung-PNNL (approved):
It looks great to me! Thanks for this effort Weili.

rct229/data/ashrae_90_1_table_G3_1_3_7.json
```diff
@@ -0,0 +1,80 @@
+{
+    "heat_rejection_leaving_water_temp":[
+        {
+            "climate_zone": "CZ0A",
```
charliepnnl, line 4: I suggest dropping the "CZ" prefix for the climate zones.  Table G3.1.3.11 does not have the "CZ" nor is it used in other JSON tables.

rct229/data/ashrae_90_1_table_G3_1_3_7.json
```diff
@@ -0,0 +1,80 @@
...
+            "degree_Fahrenheit": 80
```
charliepnnl, line 5: Should this field be called "leaving_water_temperature" to match the standard table? Perhaps add the suffix "_dF"?

rct229/data_fns/table_3_1_3_7_fns.py
```diff
@@ -0,0 +1,25 @@
...
+    -------
+    dict
+        { leaving_water_temperature: Quantity – the leaving water temperature defined by Table 3.1.3.7 [degree-F]}
+    """
+    osstd_entry = find_osstd_table_entry(
+        [("climate_zone", climate_zone_enum_val)],
+        osstd_table=data["ashrae_90_1_table_G3_1_3_7"],
+    )
+    degree_fahrenheit = osstd_entry["degree_Fahrenheit"]
+    leaving_water_temperature = degree_fahrenheit * ureg("degree_Fahrenheit")
```
charliepnnl, line 23: I don't think we want to multiply the units in this case.
How about:
```
ureg(f"{degree_fahrenheit} F")
```
?

rct229/data_fns/table_3_1_3_7_fns_test.py
```diff
@@ -0,0 +1,24 @@
...
+
+from rct229.data_fns.table_3_1_3_7_fns import table_3_1_3_7_lookup
+from rct229.schema.config import ureg
+
+degree_fahrenheit = ureg("degree_Fahrenheit")
+
+# Testing table_3_1_3_7------------------------------------------
+def test__table_3_1_3_7_CZ0A():
+    assert table_3_1_3_7_lookup("CZ0A") == {
+        "leaving_water_temperature": 80 * degree_fahrenheit
```
charliepnnl, line 11: I think it is better to write these as `ureg("80 F")`.

rct229/data/ashrae_90_1_table_G3_5_3.json
```diff
@@ -0,0 +1,46 @@
+{
+  "prm_water_chiller_min_efficiency": [
+    {
+      "cooling_type": "displacement",
+      "minimum_capacity": 0.0,
```
charliepnnl, line 5: Should we specify the units by adding a suffix "_tons"?

rct229/data/ashrae_90_1_table_G3_5_3.json
```diff
@@ -0,0 +1,46 @@
+{
+  "prm_water_chiller_min_efficiency": [
+    {
+      "cooling_type": "displacement",
```
charliepnnl, line 4: Should this be "equipment_type" to match the standard table?

rct229/data/ashrae_90_1_table_G3_5_3.json
```diff
@@ -0,0 +1,46 @@
...
+      "minimum_capacity": 0.0,
+      "maximum_capacity": 149.99,
```
charliepnnl, line 6: Another way to express the strict inequality from the table would be to call the capacity fields "inclusive_minimum_capacity" and "exclusive_maximimum_capacity".

rct229/data_fns/table_3_5_3_fns.py
```diff
@@ -0,0 +1,75 @@
...
+        }
+    """
+    compressor_category = water_chiller_compressor_type_map[compressor_type]
+
+    minimum_capacity = 0
+    for min_threshold in minimum_capacity_threshold_list:
+        if capacity > min_threshold:
+            minimum_capacity = min_threshold
+
+    maximum_capacity = maximum_capacity_threshold_list[-1]
```
charliepnnl, line 49: I think you need `maximum_capacity = maximum_capacity_threshold_list[0]`. Otherwise, the `maximum_capacity` corresponding to a capacity of 1000000 would be 149.99.

charliepnnl (approved):
Approved with one comment that needs to be addressed and several other comments that are more a matter of style.

rct229/data/ashrae_90_1_table_G3_1_3_7.json
weilixu: "CZ" prefix is from the [schema](https://github.com/open229/ruleset-model-description-schema/blob/master/docs229/Enumerations2019ASHRAE901.schema.json#L872-L893).

The reason I added CZ in the Json was trying to avoid to create a map in the table 3.1.3.11 function.
It seems the table 3.2 already have a similar map there. I don't have strong feeling about one or another, but just want to know your preference?

PR #687, opened 2022-10-06
Create 19-3.md

claperle:


docs/section19/19-3.md
```diff
@@ -0,0 +1,27 @@
...
+**Function Calls:**  None
+
+## Rule Logic:   
+- For each Building in the B_RMR: `for building in B_RMR...Building:` 
+    - **Rule Assertion:** 
+    - All cases: Conduct manual check that rule is met:  `outcome = "UNDETERMINED" and raise_message "Check that the weather conditions used in sizing runs to determine baseline equipment capacities is based either on hourly historical weather files containing typical peak conditions or on design days developed using 99.6% heating design temperatures and 1% dry-bulb and 1% wet-bulb cooling design temperatures."      `  
+
+
+**Notes/Questions:**  
+1. Does this apply at the building level as I created it?
```
weilixu, line 25: The sizing runs is often used to determine the performance parameters in HVAC systems that supply heating, cooling and ventilations to multiple buildings. I think this can be at ruleset model instance level?

docs/section19/19-3.md
```diff
@@ -0,0 +1,26 @@
+# Section 19 - Rule 19-3     
+**Schema Version:** 0.0.23  
+**Mandatory Rule:** False    
+**Rule ID:** 19-2     
+**Rule Description:** Weather conditions used in sizing runs to determine baseline equipment capacities shall be based either on hourly historical weather files containing typical peak conditions or on design days developed using 99.6% heating design temperatures and 1% dry-bulb and 1% wet-bulb cooling design temperatures.    
+**Rule Assertion:** Options are UNDETERMINED     
+**Appendix G Section:** G3.1.2.2.1      
+**90.1 Section Reference:** None
```
mkarpman, line 8: Now that committee agreed that some of the rules will be excluded from the scope of 229, I think this rule shuld be one of these because it always applies and cannot be automated.

supriyagoel: so is the consensus to move this to rules not evaluated?

docs/section19/19-3.md
```diff
@@ -0,0 +1,26 @@
+# Section 19 - Rule 19-3     
+**Schema Version:** 0.0.23  
+**Mandatory Rule:** False    
+**Rule ID:** 19-2
```
jugonzal07, line 4: Think this should read 19-3, not 19-2

docs/section19/19-3.md
claperle: Discussed on call on and decided to make no evaluated.

PR #690, opened 2022-10-06
Rds/kjw/function get hvac building area types and zones

KarenWGard:
This function gives the total area, of each type of HVAC building area type, plus a list of the zones associated with the building area type.  This function will serve many functions and rules for section 18.


docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
```diff
@@ -0,0 +1,177 @@
...
+	- `lighting_space_lookup["SPORTS_ARENA_PLAYING_AREA_CLASS_I_FACILITY"] = "PUBLIC_ASSEMBLY"`
+	- `lighting_space_lookup["SPORTS_ARENA_PLAYING_AREA_CLASS_II_FACILITY"] = "PUBLIC_ASSEMBLY"`
+	- `lighting_space_lookup["SPORTS_ARENA_PLAYING_AREA_CLASS_III_FACILITY"] = "PUBLIC_ASSEMBLY"`
+	- `lighting_space_lookup["SPORTS_ARENA_PLAYING_AREA_CLASS_IV_FACILITY"] = "PUBLIC_ASSEMBLY"`
+	- `lighting_space_lookup["TRANSPORTATION_FACILITY_BAGGAGE_CAROUSEL_AREA"] = "OTHER_NON_RESIDENTIAL"`
+	- `lighting_space_lookup["TRANSPORTATION_FACILITY_AIRPORT_CONCOURSE"] = "OTHER_NON_RESIDENTIAL"`
+	- `lighting_space_lookup["TRANSPORTATION_FACILITY_TICKET_COUNTER"] = "OTHER_NON_RESIDENTIAL"`
+	- `lighting_space_lookup["WAREHOUSE_STORAGE_AREA_MEDIUM_TO_BULKY_PALLETIZED_ITEMS"] = "HEATED-ONLY_STORAGE"
+	- `lighting_space_lookup["WAREHOUSE_STORAGE_AREA_SMALLER_HAND_CARRIED_ITEMS"] = "HEATED-ONLY_STORAGE"
+
```
mkarpman, line 151: We usually don't include look-up tables in the functions or RDSs (e.g., do not have lookup for U-factors depending on building type and climate zone). But perhaps it's justified here since this is not coming from 90.1. @weilixu what do you think?

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
```diff
@@ -0,0 +1,177 @@
...
+		- add zone_id to list_of_zones: `list_of_zones.append(zone.id)`
+		- create space_area_type variable and set equal to "None": `space_area_type = "None"`
+		- check if there is an HVAC Building Area Type: `if building_segment.area_type_heating_ventilation_air_conditioning_system != "Null":`
+			- set space_area_type equal to HVAC BAT: `space_area_type = building_segment.area_type_heating_ventilation_air_conditioning_system`
+		- else:
+			- look for a lighting_building_area_type: `if building_segment.lighting_building_area_type != "Null":`
+				- set space_area_type based on the lookup table: `space_area_type = lighting_space_lookup[building_segment.lighting_building_area_type`
+		- For each space in zone: `for space in zone.spaces:`
+			- check if there is a value in space_area_type: `if space_area_type != "None":`
+				- add space area: `list_buildiung_area_types_with_total_area_and_zones[space_area_type]["AREA"] += space.floor_area`
```
mkarpman, line 164: I wonder if it's an issue when zone has spaces with different lighting area type. Not sure we zound be splitting a zone into different HVAC area types.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
```diff
@@ -0,0 +1,177 @@
...
+			- otherwise, we'll need to map the building area type off of the ligthing category: `else:`
```
mkarpman, line 165: Based on line 163, Else here seems to refer to space_area_type = "None"?

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
```diff
@@ -0,0 +1,177 @@
...
+				- set the space area type based on lighting space area: `space_area_type = lighting_space_lookup[space.lighting_space_type]`
+				- add space to the area based on space.lighting_space_type: `list_buildiung_area_types_with_total_area_and_zones[space_area_type]["AREA"] += space.floor_area`
+
+	- Append zone_ids to the dict: `list_buildiung_area_types_with_total_area_and_zones[space_area_type]["ZONE_IDS"] = list_of_zones`
+	
+	 **Returns** `return list_buildiung_area_types_with_total_area_and_zones`  
+
+**Notes/Questions:**  
+1. "Heated-Only Storage" was not listed in the schema under HeatingVentilatingAirConditioningBuildingAreaOptions2019ASHRAE901 when this function was written - double check spelling
+2. a map was provided for mapping LightingBuildingAreaOptions2019ASHRAE901T951TG38, but not LightingSpaceOptions2019ASHRAE901TG37.  Double check my mapping
```
mkarpman, line 175: @claperle - can you plese provide mapping for #2 if you haven't already?

claperle: @mkarpman I provided what we have. We do not have mapping for space types.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
KarenWGard: yes, might be more straightforward to do: 

`if space_area_type == "None":`
...
`else:`
...

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
KarenWGard: the concern was what if there is no HVAC BAT entered at the building_segment level.  Christina provided mapping of lighting building area types to HVAC BAT (I think this was provided by ASHRAE).  I took a shot at extending the mapping to lighting space types as most of them are sub-categories of lighting building space types.

We absolutely need an HVAC BAT, so the other option would be to make it a required field in the schema - thoughts @JasonGlazer ?

@weilixu - I see your point about storage can be cooled.  G3.1.1.f takes care of this "If the baseline HVAC system type is 9 or 10, use additional system types for all HVAC zones that are mechanically cooled in the proposed design."

mkarpman: @KarenWGard I don't think HVAC BAT should be a required part of schema because it will mean that files that don't have this specified are rejected by the RCT. Even if the HVAC-related rules can't be checked, RCT can still be used to check envelope and lighting rules.

mkarpman: > Generally the approach I think is ok, but I am not sure if RCT should provide such mapping since it is not part of Standard, and it is not entirely clear how some of the lighting spaces can be mapped to HVAC building type

@weilixu I agree, but perhaps if we based HVAC building area type on lighting space type mapping we can have a note to this effect as part of outcome, to suggest that reviewer manually double-check the HVAC building area type.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
mkarpman: I agree. Since it's generally unlikely that there is more than one space per zone in the model, I would set zone area type based on the lighting area type of spaces that account for the greatest area withing a zone.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
```diff
@@ -0,0 +1,177 @@
...
+		- For each space in zone: `for space in zone.spaces:`
+			- check if there is a value in space_area_type: `if space_area_type != "None":`
+				- add space area: `list_buildiung_area_types_with_total_area_and_zones[space_area_type]["AREA"] += space.floor_area`
+			- otherwise, we'll need to map the building area type off of the ligthing category: `else:`
+				- set the space area type based on lighting space area: `space_area_type = lighting_space_lookup[space.lighting_space_type]`
+				- add space to the area based on space.lighting_space_type: `list_buildiung_area_types_with_total_area_and_zones[space_area_type]["AREA"] += space.floor_area`
+
+	- Append zone_ids to the dict: `list_buildiung_area_types_with_total_area_and_zones[space_area_type]["ZONE_IDS"] = list_of_zones`
+	
+	 **Returns** `return list_buildiung_area_types_with_total_area_and_zones`
```
mkarpman, line 171: I think this function should also return a flag to indicate whether the determination was made based on buildingsegment.area_type_heating_ventilating_air_conditioning_system or our homecooked mapping. Or perhaps this could be handled by the calling RDS which would check if buildingsegment.area_type_heating_ventilating_air_conditioning_system is blank.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
mkarpman: Yes, I thought about that also. In general, the current schema architecture implies that each BuildingSegment represent one occupancy type . (The main point of this data object is to group zones within a mixed use building by occupancy type to allow specifying  area_type_vertical_fenestration, lighting_building_area_type and area_type_heating_ventilating_air_conditioning_system that apply to all zones within the segment.) Few questions along these lines:

1. For a multifamily building, are we expecting that corridors, stairwells, etc. are included into a separate building segment from apartments?
2. . If area_type_heating_ventilating_air_conditioning_system is provided, should we be checking that it is consistent with the lighting_space_area_type? E.g., if area_type_heating_ventilating_air_conditioning_system is set to residential, but lighting_space_area_type suggests that it cannot be that? Similarly, should we check for general conistence between HVAC, lighting and SWH building area types specified for a given segment? I think we should, otherwise we can overlook impactful mistakes.
3. Should we use area_type_vertical_fenestration and lighting_building_area_type as proxies for a missing area_type_heating_ventilating_air_conditioning_system before we check space type?
4. Since the entire BuildingSegment is meant to represent single building occupancy type, should the function determine a single area_type_heating_ventilating_air_conditioning_system for each BuildingSegment based on the predominant space type?
5. The function returns list_buildiung_area_types_with_total_area_and_zones that are segregated by BuildingSegment. Should the last step of this function be to aggregate the same bluilding area types across segments? E.g., if there are two building segments that are both "Other-Nonresidential", they would be combined for the purpose of determining building area type.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
KarenWGard: @mkarpman - I agree.  I think that if we don't do the mapping, but we require HVAC BAT, the simulation software developers will do their own mapping, and it will be inconsistent.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
KarenWGard: > I agree. Since it's generally unlikely that there is more than one space per zone in the model, I would set zone area type based on the lighting area type of spaces that account for the greatest area withing a zone.

done

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
KarenWGard: > Yes, I thought about that also. In general, the current schema architecture implies that each BuildingSegment represent one occupancy type . (The main point of this data object is to group zones within a mixed use building by occupancy type to allow specifying area_type_vertical_fenestration, lighting_building_area_type and area_type_heating_ventilating_air_conditioning_system that apply to all zones within the segment.) Few questions along these lines:
> 
> 1. For a multifamily building, are we expecting that corridors, stairwells, etc. are included into a separate building segment from apartments?

**I believe that corridors serving multi-family are considered residential HVAC_BAT, right?**

> 2. . If area_type_heating_ventilating_air_conditioning_system is provided, should we be checking that it is consistent with the lighting_space_area_type? E.g., if area_type_heating_ventilating_air_conditioning_system is set to residential, but lighting_space_area_type suggests that it cannot be that? Similarly, should we check for general conistence between HVAC, lighting and SWH building area types specified for a given segment? I think we should, otherwise we can overlook impactful mistakes.

**should this be a separate check or rule?**

> 3. Should we use area_type_vertical_fenestration and lighting_building_area_type as proxies for a missing area_type_heating_ventilating_air_conditioning_system before we check space type?
> 4. Since the entire BuildingSegment is meant to represent single building occupancy type, should the function determine a single area_type_heating_ventilating_air_conditioning_system for each BuildingSegment based on the predominant space type?

**good question.  It is easy to program, just let me know what you would like me to do.**

> 5. The function returns list_buildiung_area_types_with_total_area_and_zones that are segregated by BuildingSegment. Should the last step of this function be to aggregate the same bluilding area types across segments? E.g., if there are two building segments that are both "Other-Nonresidential", they would be combined for the purpose of determining building area type.

**I updated the function for clarity.  It is building-wide**

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
KarenWGard: I think this should be a separate rule or test.  The function can still return this flag if necessary, or the new rule test could simply look for missing HVAC_BAT & create an alert that they're missing, which means that the HAC BAT is determined based on lighting

KarenWGard:
in response to: [Maria's comment above](https://github.com/pnnl/ruleset-checking-tool/pull/690#discussion_r1013014658):

1.  I don't know, but if they are and there is no HVAC_BAT in a very large building, corridors and other common spaces could reach the 20,000ft2 threshold for "OTHER_NON_RESIDENTIAL"
2. if we do a check, I think it should be it's own rule
3. These space types look good & might be better than lighting space type.  They still won't pick up the case where a corridor is a support corridor for a residential building (corridors and stairways get non-residential construction types even in residential buildings, if I'm remembering correctly)
4. I'm happy to do this.  Would you like me to change it?
5. I'm 99% sure that the function list_buildiung_area_types_with_total_area_and_zones returns a list for the entire building.  I've edited it for clarity.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
```diff
@@ -0,0 +1,177 @@
+# get_HVAC_building_area_types_and_zones
+
+**Description:** Get the hvac_ids associated with each building area type associated with U_RMR, P_RMR, or B_RMR. Also returns the total floor area of each building area type
+- used to verify the correct type of HVAC baseline system (or systems)
```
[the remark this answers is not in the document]
mkarpman: There is still a mismatch between description in Line 3& 4, function name, and what is being returned based on line 178. Comment about the function app;ling only to B-RMI also still stands.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
mkarpman: So I think the plan is to do the mapping. For transparency, let's also include the mapping into documentation outside of the RDSs. Perhaps it should be included with the 229 package. Perhaps also make it an exterinal library table similar to how ASHRAE tables are handled, so that it is not incapsulated into the RDS.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
mkarpman: > > 3\. . If area_type_heating_ventilating_air_conditioning_system is provided, should we be checking that it is consistent with the lighting_space_area_type? E.g., if area_type_heating_ventilating_air_conditioning_system is set to residential, but lighting_space_area_type suggests that it cannot be that? Similarly, should we check for general conistence between HVAC, lighting and SWH building area types specified for a given segment? I think we should, otherwise we can overlook impactful mistakes.
> 
> **should this be a separate check or rule?**

Let's discuss with the team. Perhaps this should be a separate rule (and an RDS) as you suggest.

mkarpman: > > 1. For a multifamily building, are we expecting that corridors, stairwells, etc. are included into a separate building segment from apartments?
> 
> **I believe that corridors serving multi-family are considered residential HVAC_BAT, right?**

They are actually not considered residential, and are only lumped with residential if they are less than 20K SF.

mkarpman: > 3\. Should we use area_type_vertical_fenestration and lighting_building_area_type as proxies for a missing area_type_heating_ventilating_air_conditioning_system before we check space type?

Let's discuss on the call.

mkarpman: > 4\. Since the entire BuildingSegment is meant to represent single building occupancy type, should the function determine a single area_type_heating_ventilating_air_conditioning_system for each BuildingSegment based on the predominant space type?

Let's discuss on the call.

mkarpman: > I updated the function for clarity. It is building-wide

Can you please point me to where this is done?

mkarpman: > There are a few, that are less obvious. For example "CORRIDOR" has sub-categories for hospital and manufacturing, but not for residential or retail, so corridors in these buildings would be categorized as "OTHER-NONRESIDENTIAL" unless we can find a way to classify them better. One way to do this would be to figure out the predominant space type on a given floor (or in a given building) and classify these types of spaces (corridors, copy rooms, storage) as the same BAT as the predominant.

Perhaps this is the way to go. We can tag spaces of type corridor, stairwells, etc. as "supporting nonresidential" and then add them to the predominant non-residentail space type. I am generally leaning toward not using HVAC building area type at all and use lighting space type instead.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
mkarpman: I agree

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
KarenWGard: function has been updated with mapping.  Description was updated with a clearer description of how mapping happens.  Mapping re-written to be more easily readable.  I've noted where I am unclear on what mapping is correct.  Please review where noted.

Support spaces are initially categorized as `OTHER_UNDETERMINED` and then re-mapped to the predominant space area type at the end of the function.  Please review what is mapped as `OTHER_UNDETERMINED`

KarenWGard:
Function updated based on comments from 12/29 call - 

function first looks at space lighting area classification, and uses the lookup table.  Support type spaces inherit the main building occupancy type, unless it is residential.  If the main building occupancy type is residential, the support spaces inherit the 2nd most common building occupancy type (if there is one) OR OTHER_NONRESIDENTIAL of there is none

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
```diff
@@ -0,0 +1,203 @@
...
+		- create a dict of space area types: `zone_space_area_types = {}`
+		- create a variable for the total floor area of the zone: `zone_total_floor_area = 0`
+		- loop through the spaces: `for space in zone.spaces:`
+			- add the space floor area to the zone total floor area: `zone_total_floor_area += space.floor_area
+			- look up the area type using `lighting_space_lookup`: `area_type = lighting_space_lookup[space.lighting_space_type]`
+			- create the integer for this space area type if it doesn't exist yet: `zone_space_area_types[area_type] = zone_space_area_types[area_type] or 0`
+			- add the space area to the dict under `area_type`: `zone_space_area_types[area_type] = zone_space_area_types[area_type] + space.floor_area`
+		- assign the key with the maximum floor area to `space_area_type`: `space_area_type = max(zone_space_area_types, key = zone_space_area_types.get)`
+		- if the space_area_type is NONE, we look at the building_segment.area_type_heating_ventilating_air_conditioning_system: `space_area_type = building_segment.area_type_heating_ventilating_air_conditioning_system`
+
```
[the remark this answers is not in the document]
mkarpman: I think the method that Supriya proposed is more accurate.
Line 171 allows setting the mapping based on building_segment.area_type_heating_ventilating_air_conditioning_system. Do we want to allow that? Them can overlook some exceptions and be actually incorrectly determined by modeler.

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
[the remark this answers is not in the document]
mkarpman: This doesn't seem to have been addressed.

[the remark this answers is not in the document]
supriyagoel: @KarenWGard the function description still needs to be changed

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
```diff
@@ -0,0 +1,196 @@
+# get_HVAC_building_area_types_and_zones
+**Scheme Version:** 0.0.23  
+
+**Description:** Get the hvac_ids associated with each building area type associated with U_RMR, P_RMR, or B_RMR. Also returns the total floor area of each building area type
+- used to verify the correct type of HVAC baseline system (or systems)
+- HVAC building area type is determined first by looking at the building segment HVAC_BAT.  If this variable does not exist, the function next looks at building segment lighting_building_area_type.  If this variable doesn't exist, the function will look at the space lighting type.  Support spaces such as corridors, stairwells and other support spaces will inherit the predominant building area type EXCEPT in buildings where the predominant HVAC_BAT is residential.  In residential buildings, support spaces such as corridors and stairwells will be classified as Other Non-Residential.
+- the entry "CLASSIFICATION_SOURCE" in the **building_area_types_with_total_area_and_zones_dict** passes information about how the HVAC_BAT was determined for each segment along
```
supriyagoel, line 7: In line 6 we say 'In residential buildings, support spaces such as corridors and stairwells will be classified as Other Non-Residential.'.
90.1-2022 has clarified that such support spaces are called 'residential associated HVAC zone' and classified as residential. However, i am not sure if 90.1 2019  is clear about how such spaces shall be classified. am i missing some requirmeent?

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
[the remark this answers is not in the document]
KarenWGard: @supriyagoel - updated

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
KarenWGard: the way the function is written, if a building segment has the HVAC_BAT "RESIDENTIAL" - the spaces will be classified as residential, regardless of the space type.  However, if there is no user-given HVAC_BAT, these spaces will be classified as OTHER_NON_RESIDENTIAL - I've edited the description for clarity.  Is this consistent with the requirements of ASHRAE 90.1?

supriyagoel: I am not sure. Might need to be clarified by @mkarpman or @JasonGlazer

mkarpman: Corridors and other non-residential spaces in residential buildings are considered non-residential. This was discussed in comments to line 194. There was an official interpretation for this, and it is also (somewhat) consistent with 90.1 2022 where such spaces are also not modeled with System 1/2. I beleive this is how it's currently done in the RDS (line 182)

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
```diff
@@ -0,0 +1,196 @@
...
+		- add the area from the OTHER_UNDETERMIEND list to the OTHER_NON_RESIDENTIAL area: `building_area_types_with_total_area_and_zones_dict[OTHER_NON_RESIDENTIAL]["AREA"] += building_area_types_with_total_area_and_zones_dict[OTHER_UNDETERMINED]["AREA"]
+	- otherwise, we can add the are and zones from OTHER_UNDETERMINED to the predominate_HVAC_BAT: `else:`
+		- add the zones from the OTHER_UNDETERMINED list to the predominate_HVAC_BAT list: `building_area_types_with_total_area_and_zones_dict[predominate_HVAC_BAT]["ZONE_IDS"] += building_area_types_with_total_area_and_zones_dict[OTHER_UNDETERMINED]["ZONE_IDS"]`
+		- add the area from the OTHER_UNDETERMIEND list to the predominate_HVAC_BAT area: `building_area_types_with_total_area_and_zones_dict[predominate_HVAC_BAT]["AREA"] += building_area_types_with_total_area_and_zones_dict[OTHER_UNDETERMINED]["AREA"]
+	- remove the key OTHER_UNDETERMINED from building_area_types_with_total_area_and_zones_dict: `del building_area_types_with_total_area_and_zones_dict[OTHER_UNDETERMINED]`
+
+	 **Returns** `return buildiung_area_types_with_total_area_and_zones_dict`
```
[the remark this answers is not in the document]
KarenWGard: yes, this is how it works

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
```diff
@@ -0,0 +1,196 @@
...
+		- add the area from the OTHER_UNDETERMIEND list to the OTHER_NON_RESIDENTIAL area: `building_area_types_with_total_area_and_zones_dict[OTHER_NON_RESIDENTIAL]["AREA"] += building_area_types_with_total_area_and_zones_dict[OTHER_UNDETERMINED]["AREA"]
+	- otherwise, we can add the are and zones from OTHER_UNDETERMINED to the predominate_HVAC_BAT: `else:`
+		- add the zones from the OTHER_UNDETERMINED list to the predominate_HVAC_BAT list: `building_area_types_with_total_area_and_zones_dict[predominate_HVAC_BAT]["ZONE_IDS"] += building_area_types_with_total_area_and_zones_dict[OTHER_UNDETERMINED]["ZONE_IDS"]`
+		- add the area from the OTHER_UNDETERMIEND list to the predominate_HVAC_BAT area: `building_area_types_with_total_area_and_zones_dict[predominate_HVAC_BAT]["AREA"] += building_area_types_with_total_area_and_zones_dict[OTHER_UNDETERMINED]["AREA"]
+	- remove the key OTHER_UNDETERMINED from building_area_types_with_total_area_and_zones_dict: `del building_area_types_with_total_area_and_zones_dict[OTHER_UNDETERMINED]`
+
+	 **Returns** `return building_area_types_with_total_area_and_zones_dict`
```
mkarpman, line 190: The dictionary that is returned by this function lists BAT, list of zones, and floor area for each segment. Shouldn't it then aggergate the matching values across the segments? For example, if building has two segments that both have the same BAT, should the returned dictionary include that BAT, list of all zones in the two segments, and the combined area of both segments?

KarenWGard: @mkarpman - the function follows your logic.  Lets' review on the call.

supriyagoel: discussed on RDS call and Karen clarified that the function follows the logic MK has laid out

docs/ruleset_functions/get_HVAC_building_area_types_and_zones.md
supriyagoel: @weilixu will move this out when the RDS is implemented and update the RDS to not have this mapping

mkarpman (approved):

supriyagoel (approved):

PR #693, opened 2022-10-07
RS+RT/YJ/RULE 22-20

yunjoonjung-PNNL:


rct229/rules/section22/section22rule20.py
```diff
@@ -0,0 +1,92 @@
...
+            each_rule=Section22Rule20.HeatRejectionRule(),
+            index_rmr="baseline",
+            id="22-20",
+            description="The baseline minimum condenser water reset temperature is per Table G3.1.3.11.",
+            list_path="ruleset_model_instances[0].heat_rejections[*]",
+            data_items={"climate_zone": ("baseline", "weather/climate_zone")},
+        )
+
+    def is_applicable(self, context, data=None):
+        rmi_b = context.baseline
```
charliepnnl, line 44: Since you, correctly, didn't set the `rmr_context` above, the context here is at the RMD level. So
```
rmd_b = context.baseline
rmi_b = rmd_b["ruleset_model_instances"][0]
```
You should probably also include
```
required_fields={
    "$": ["ruleset_model_instances"]
}
```
in the initializer above.  Unfortunately, we can't use a required field to ensure the `[0]` ruleset_model_instance exists. You will have to use `assert_` between the rmd_b and rmi_b assignments above:
```
assert_(
    # Should be a non-empty list
    ruleset_model_instance,
    "ruleset_model_instance list is empty"
)
```

charliepnnl (approved):
Approved with one comment that needs to be addressed.

weilixu (approved):
Looks good, thanks

rct229/rules/section22/section22rule20.py
yunjoonjung-PNNL: Thank you so much for catching this up, @charliepnnl.
One remaining question is for the ```required_fields``` in the comment, the problem is ```required_fields``` is also used in the ```HeatRejectionRule``` level,. In other words, we use ```required_fields``` two times, which causes an error. In this case, how we could solve this?

rct229/rules/section22/section22rule20.py
```diff
@@ -0,0 +1,101 @@
...
+            # required_fields={
+            #     "$": ["ruleset_model_instances"],
+            # },
+            data_items={"climate_zone": ("baseline", "weather/climate_zone")},
+        )
+
+    def is_applicable(self, context, data=None):
+        rmd_b = context.baseline
+        rmi_b = rmd_b["ruleset_model_instances"][0]
+        assert_(
```
charliepnnl, line 49: I think you need to move lines 49-52 up between lines 47 and 48 because line 48 will raise an unhandled exception if the `ruleset_model_instances` list is empty. Oh, then you'll need to modify the code on line 50 too.

rct229/rules/section22/section22rule20.py
```diff
@@ -0,0 +1,101 @@
...
+                "$": ["weather"],
+                "weather": ["climate_zone"],
+            },
+            each_rule=Section22Rule20.HeatRejectionRule(),
+            index_rmr="baseline",
+            id="22-20",
+            description="The baseline minimum condenser water reset temperature is per Table G3.1.3.11.",
+            list_path="ruleset_model_instances[0].heat_rejections[*]",
+            # required_fields={
+            #     "$": ["ruleset_model_instances"],
```
charliepnnl, line 41: You can change line 32 to `"$": ["ruleset_model_instances", "weather"],`

rct229/rules/section22/section22rule20.py
charliepnnl: I added a couple of new comments. Do those answer your questions here?
BTW: It is perfectly fine for each level to include its own `required_fields`. Generally, the required_fields should go in the lowest level possible.

rct229/rules/section22/section22rule20.py
yunjoonjung-PNNL: Fixed. Thanks for letting me know about this Charlie!

rct229/rules/section22/section22rule20.py
yunjoonjung-PNNL: Fixed. Thanks!

rct229/rules/section22/section22rule20.py
yunjoonjung-PNNL: Yes! Fixed.

PR #705, opened 2022-10-10
creation of function

KarenWGard:


docs/ruleset_functions/zone_meets_G3_1_1e.md
```diff
@@ -0,0 +1,60 @@
+# zone_meets_G3_1_1e
```
weilixu, line 1: Could we change this function by adding does as prefix?
`does_zone_meets_G3_1_1e`? The name could indicate this function is returns boolean

docs/ruleset_functions/zone_meets_G3_1_1e.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+**Returns** `result`
+
+
+**Notes/Questions:**  
+1. the "is_zone_a_vestibule" function gives only "NO" and indeterminite answers.  Is there a way to alert to confirm that the zone is a vestibule?
+2. Transfer air does not indicate whether it is transferring from an air conditioned zone or not, therefore cannot do a complete check, only an alert
+	221006 - Jason suggests that the schema gets modified to indicate where the transfer air is coming from ("transfer_air_from_zone")
+3. does a HeatingVentilationAirConditioningSystem always have a CoolingSystem, or does it return "Null" if there is no cooling system?
```
weilixu, line 58: `cooling_system` is not a required data in the schema, so it can be empty.
In python, getting a non-existent `cooling system` will return `None`. 

When dealing with non-required data, we can decide whether to stop the process and return `UNDETERMINED` decision or accept `None` as one possible outcome.

docs/ruleset_functions/zone_meets_G3_1_1e.md
```diff
@@ -0,0 +1,60 @@
...
+	- however, if there is transfer air into the space from an air conditioned zone, it is not eligible, so look for transfer air into the space: `if zone.transfer_airflow_rate > 0:`
+		- there is no way of knowing where the transfer air is coming from, so for now will give the result the tag "E_T" to signal that there is transfer air.  When the rule is written, "E_T" from this function will trigger an "UNDETERMINED" result and indicate manual checking for transfer air from air-conditioned space: `result = "E_T"
+
+
+**Returns** `result`
+
+
+**Notes/Questions:**  
+1. the "is_zone_a_vestibule" function gives only "NO" and indeterminite answers.  Is there a way to alert to confirm that the zone is a vestibule?
+2. Transfer air does not indicate whether it is transferring from an air conditioned zone or not, therefore cannot do a complete check, only an alert
```
weilixu, line 56: The current one is OK,  but this function needs to be updated once the schema adds the transfer air portion.

docs/ruleset_functions/zone_meets_G3_1_1e.md
```diff
@@ -0,0 +1,60 @@
...
+	- set result to "E": `result = "E"`
+	- however, if there is transfer air into the space from an air conditioned zone, it is not eligible, so look for transfer air into the space: `if zone.transfer_airflow_rate > 0:`
+		- there is no way of knowing where the transfer air is coming from, so for now will give the result the tag "E_T" to signal that there is transfer air.  When the rule is written, "E_T" from this function will trigger an "UNDETERMINED" result and indicate manual checking for transfer air from air-conditioned space: `result = "E_T"
+
+
+**Returns** `result`
+
+
+**Notes/Questions:**  
+1. the "is_zone_a_vestibule" function gives only "NO" and indeterminite answers.  Is there a way to alert to confirm that the zone is a vestibule?
```
weilixu, line 55: I suggest setting it to `False` first instead of a string. A boolean is always a preferred way to string :)

docs/ruleset_functions/zone_meets_G3_1_1e.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Logic:
+- create list of eligible space types: `eligible_space_types = ["STORAGE_ROOM_HOSPITAL", "STORAGE_ROOM_SMALL", "STORAGE_ROOM_LARGE", "WAREHOUSE_STORAGE_AREA_MEDIUM_TO_BULKY_PALLETIZED_ITEMS", "WAREHOUSE_STORAGE_AREA_SMALLER_HAND_CARRIED_ITEMS", "STAIRWELL", "ELECTRICAL_MECHANICAL_ROOM", "RESTROOM_FACILITY_FOR_THE_VISUALLY_IMPAIRED", "RESTROOM_ALL_OTHERS"]
+- set the result variable to "No" - only a positive test can give it a different value: `result = "No"`
+- set eligibility boolean to False: `eligible = TRUE`
+- For each space in zone: `for space_p in zone.Spaces:`
+	- check if the space has an eligible lighting type: `if space_p.lighting_space_type not in eligible_space_types:`
+		- set eligibility to False - any non-compliant space type will result in a non-eligible zone: `eligible = FALSE`
+
+- if the zone is not eligible, check if it is a vestibule: `if eligible == FALSE:`
```
weilixu, line 25: The code can be `if not eligible`

docs/ruleset_functions/zone_meets_G3_1_1e.md
```diff
@@ -0,0 +1,60 @@
...
+	- set variable is_a_vestibule = is_zone_a_vestibule: `is_a_vestibule = is_zone_a_vestibule(zone, B-RMR)`
```
weilixu, line 26: Does this function has to take in `B-RMD`?
If geometry is the same, I think `P-RMD` could work? By setting it to `P-RMD`, we can save one argument.

docs/ruleset_functions/zone_meets_G3_1_1e.md
```diff
@@ -0,0 +1,60 @@
...
+		`if zone in building_segment.zones:`
+			- break - this means the variable "building_segment" is the building_segment that contains the zone:
+			`break`
+	- loop through each heating system: `for system_p in building_segment.heating_ventilation_air_conditioning_systems:`
+		- look for a cooling system: `if system_p.cooling_system != "Null":`
+			- check to see if the cooling_system_options equals "None": `if system_p.cooling_system.cooling_system_options == "None":`
+				- the zone is eligible
+			- otherwise: `else:`
+				- there is a cooling system and the zone is not eligible: `eligible = FALSE`
+
```
weilixu, line 44: The above logic could result in a zone connected to an HVAC system with heating and cooling systems but still identified as an eligible zone.

Suggest to add:
1. search for terminals in the zone and get the `served_by_heating_ventilating_air_conditioning_system` list. 
2. Searched the HVAC systems inside the `building_segment.heating_ventilation_air_conditioning_systems` and filter the ones matched to the `served_by_heating_ventilating_air_conditioning_system` list.
3. Check if those HVAC systems are heating-only HVAC systems.

docs/ruleset_functions/zone_meets_G3_1_1e.md
```diff
@@ -0,0 +1,60 @@
+# zone_meets_G3_1_1e
+**Schema Version:** 0.0.22  
+
+**Description:** determines whether a given zone meets the G3_1_1e exception "Thermal zones designed with heating-only systems in the proposed design serving storage rooms, stairwells, vestibules, electrical/mechanical rooms, and restrooms not exhausting or transferring air from mechanically cooled thermal zones in the proposed design shall use system type 9 or 10 in the baseline building design."
+
+**Inputs:**
+- **P-RMR**
+- **B-RMR**
+- **zone**
```
weilixu, line 9: Depends on how the function is used 
In this case, If `P-RMD` is an argument, a zone seems unnecessary. A `zone_id` should do well here.

KarenWGard: in line 21, the function iterates through the spaces in the zone, so makes sense to me to pass a zone, instead of an id:

`for space_p in zone.spaces:`

docs/ruleset_functions/zone_meets_G3_1_1e.md
KarenWGard: sounds good.

docs/ruleset_functions/zone_meets_G3_1_1e.md
KarenWGard: good point.  changed.

docs/ruleset_functions/zone_meets_G3_1_1e.md
KarenWGard: ok, I took a crack at changing this.

- if the zone is still eligible: `if eligible:`
	- check if the proposed has a heating-only system
	- get the building_segment_p the zone is in:
	`for building_segment_p in P-RMR.building.building_segments:`
		- check if the zone is in the building_segment:
		`if zone in building_segment.zones:`
			- break - this means the variable "building_segment" is the building_segment that contains the zone:
			`break`
	- create a list of hvac system serving the zone: `hvac_systems_serving_zone = []`
	- loop through the terminals in the zone: `for terminal in zone.terminals:`
		- get the hvac system served by the terminal: `hvac_systems_serving_zone.append(terminal.served_by_heating_ventilating_air_conditioning_system)`
	- loop through each heating system: `for system_p in building_segment.heating_ventilation_air_conditioning_systems:`
		- check if system_p is in the list of hvac systems serving the zone: `if system_p in hvac_systems_serving_zone:`
			- look for a cooling system: `if system_p.cooling_system != "Null":`
				- check to see if the cooling_system_options equals "None": `if system_p.cooling_system.cooling_system_options == "None":`
					- the zone is eligible
				- otherwise: `else:`
					- there is a cooling system and the zone is not eligible: `eligible = FALSE`

docs/ruleset_functions/zone_meets_G3_1_1e.md
weilixu: I realize this issue when reviewing `is_zone_a_vestibule` function. We can discuss it in the coming call. Thanks!

docs/ruleset_functions/zone_meets_G3_1_1e.md
weilixu: The `P-RMD` contains the zone data. So passing a `zone_id` can locate the zone in the `P-RMD` by calling `zone=find_zone(zone_id)`. 
However, this is a detail implementation. Let's leave it for now. No worries.

KarenWGard: ah, that's a useful function for me to know!  Wouldn't it be more efficient from a programming perspective to pass a zone (which is just a pointer) rather than an id and use a function to find the zone?

Although the logic works either way & no need to spend time on it, just curious.

weilixu: That is very good question. It makes sense in low-level languages such as C or C++, which uses pointers ( in that case, we should label them as *zone) pointing to a chunk of computer memory that stores the object. However, in python, this makes not much difference in terms code efficiency. Passing a zone ID is probably cleaner since the zone object data is also contained in B_RMD.

docs/ruleset_functions/G311_exceptions/does_zone_meet_G3_1_1e.md
```diff
@@ -0,0 +1,56 @@
...
+- **P-RMD**
+- **B-RMD**
+- **zone_id**
+
+**Returns:**  
+- **result**: an ENUM YES, NO, LIKELY_VESTIBULE or MAYBE_VESTIBULE
+ 
+**Function Call:**
+- **is_a_vestibule**
+- **is_zone_mechanically_cooled**
```
mkarpman, line 16: I would create a placeholder for a function "does_zone_gets_cooled_transfer_air", and call it from here so that only that new function is affected once the schema is finalized.

docs/ruleset_functions/G311_exceptions/does_zone_meet_G3_1_1e.md
```diff
@@ -0,0 +1,56 @@
...
+		- set eligibility to False - any non-compliant space type will result in a non-eligible zone: `eligible = FALSE`
+
+- if the zone is not eligible, check if it is a vestibule: `if not eligible:`
+	- set variable is_a_vestibule = is_zone_a_vestibule: `is_a_vestibule = is_zone_a_vestibule(zone_b, B-RMD)`
+	- if the result is not equal to no, then it is eligible to be a vestibule: `if is_a_vestibule != NO:`
+		- set eligible to true: `eligible = TRUE`
+
+- if the zone is still eligible: `if eligible:`
+	- check if the proposed has a heating-only system by using the is_zone_mechanically_cooled function: `if is_zone_mechanically_cooled(P-RMD,zone_p)`
+		- mechanically cooled zones are not eligible: `eligible = FALSE`
```
mkarpman, line 36: Here, we would call the proposed new function "does_zone_gets_cooled_transfer_air", if yes, then `eligible = FALSE`

docs/ruleset_functions/G311_exceptions/does_zone_meet_G3_1_1e.md
```diff
@@ -0,0 +1,56 @@
...
+
+- if the zone is still eligible: `if eligible:`
+	- set result to YES: `result = YES`
+	- we still need to signal if the zone is LIKELY a vestible.  We only do this if the zone is eligible in all the other checks.  So if the proposed zone is mechanically cooled, it doesn't matter if this is a vestibule or not: `if is_a_vestibule == LIKELY:`
+		- set result to LIKELY_VESTIBULE: result = `LIKELY_VESTIBULE`
+	- otherwise, if the zone is MAYBE a vestibule: `if is_a_vestibule == MAYBE:`
+		- - set result to MAYBE_VESTIBULE: result = `MAYBE_VESTIBULE`
+	- however, if there is transfer air into the space from an air conditioned zone, it is not eligible, so look for transfer air into the space: `if zone_p.transfer_airflow_rate > 0:`
```
mkarpman, line 44: Per above, suggest that we add a separate function for transfer air and remove lines 44-47.

docs/ruleset_functions/G311_exceptions/does_zone_meet_G3_1_1e.md
KarenWGard: created

docs/ruleset_functions/G311_exceptions/does_zone_meet_G3_1_1e.md
KarenWGard: done

PR #709, opened 2022-10-13
RCT/CH/remove-jsonpath-ng

charliepnnl:
This removes most of the remaining uses of jsonpath-ng.
It is still being used in schema_utils.py. Removing it from scheam_utils.py will take a bit more work.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -16,7 +16,6 @@ pytest-watch = "*"
 
 [packages]
 rct229 = {editable = true, path = "."}
-jsonpath-ng = "*"
 jsonschema="*"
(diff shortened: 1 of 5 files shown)


weilixu (approved):
This looks good. How much work is involved in getting rid of it in schema_util? It seems the pipfile already removed the `jsonpath-ng` package.

charliepnnl:
It looks like it will take another couple of hours.  My thinking would be to come back to this whenever there is a lull in priority items.

yunjoonjung-PNNL (approved):
Looks great to me! Thank you Charlie.

PR #710, opened 2022-10-16
Complete the report module infrastructure.

weilixu:
Report module draft.

This is a draft for the report module infrastructure.
Add RCT outcome label to replace the outcome strings,
Add rct_report.py as the template for various types of reports - function `generate` orders the report generation process.
Add three types of reports in the reports folder: ASHRAE901_2019_detail_report, engine raw output, and engine raw summary.

Add code to manufacture reports in the engine - `evaluate_all_rules.`

Overall, I think this is a good way to support report modules.
1. provide flexibility to register and de-register various types of reports.
2. `generate` orchestra the report generation process while template functions created inside `generate` customize how the report can be generated.
3. No dependency on the rule definition module - easy to modify and create a custom report without worrying about the logic in the rule definition.

I can see many code optimizations needed around the `generate` function and how to read the RCT output.
So this PR wants to confirm the report architecture with the team, and at the same time, serve as a starting point for code optimization.


rct229/rule_engine/rct_report.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Parameters
+        ----------
+        ruleset_report: type depends on the initialized ruleset report
+        rule_report: type depends on the intialized ruleset re
+        rule_outcome: dict, contains counts for four RCT outputs
+        -------
+
+        """
+        pass
```
charliepnnl, line 102: Should the implementation here be:
```
raise NotImplementedError()
```
And the same for `save_ruleset_report`?

rct229/rule_engine/rct_report.py
```diff
@@ -0,0 +1,139 @@
...
+    def generate_rule_report(self, rule_outcome, outcome_dict):
+        """
+        Function to generate a rule's report
+
+        Parameters
+        ----------
+        rule_outcome: json contains a rule's raw report
+        outcome_dict: outcome dictionary, default is
+
+        Returns: rule re
```
charliepnnl, line 73: What did you mean to write for Returns: ?

rct229/rule_engine/rct_report.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Returns: data structure, default is list
+        -------
+
+        """
+        return list()
+
+    def add_rule_to_ruleset_report(self, ruleset_report, rule_report, rule_outcome):
+        """
+        Add rule to the ruleset report
```
charliepnnl, line 92: I assume that subclasses MUST implement this method.  It would be good to add a comment to that effect.
Same for the `save_ruleset_report` method.

rct229/rule_engine/rct_report.py
```diff
@@ -0,0 +1,139 @@
...
+        Parameters
+        ----------
+        ruleset_report: user-defined data structure (should be consistent with the data defined in initialize_ruleset_report function
+        report_dir: report file path.
+
+        Returns NA
+        -------
+
+        """
+        pass
```
charliepnnl, line 139: It may become clear as I continue my review, but at this point, it isn't clear why an implementation couldn't be given here.

Indeed, I now see that the report is not always a json file, so saving may need to be handled differently.

charliepnnl (commented):
My main issue/question is regarding where the report generation code lives and how it is called.

rct229/rule_engine/rct_report.py
weilixu: That is correct. I will update this.

rct229/rule_engine/rct_report.py
weilixu: Yes, that was the intent to have this function - The AHJ summary will be saved as strings so we will need this function to handle json and strings differently.

rct229/rule_engine/rct_report.py
weilixu: Return report data that can be either string or json or other data types.
Right now I can only think of STRING and JSON. Should I limit the return data to these two types?

rct229/reports/utils.py
```diff
@@ -1,3 +1,22 @@
+def calc_vals_converter(calc_vals):
```
charliepnnl, line 1: Do we really need `calc_vals_converter`?

charliepnnl (approved):

PR #720, opened 2022-10-19
Rds/kjw/function is cz 0 to 3a

KarenWGard:
simple function to determine whether a climate zone is CZ 0 to 3a


docs/functions/is_CZ_0_to_3a.md
```diff
@@ -0,0 +1,23 @@
+# is_CZ_0_to_3a
+**Schema Version:** 0.0.22  
+
+**Description:** Determines whether the building is in climate zone 0 to 3a - used for Appendix G Table G3.1.1-3
+
+**Inputs:**
+- **climate_zone**
```
weilixu, line 7: We could add the climate zone as the `ClimateZoneOptions2019ASHRAE901`

so 
**climate_zone**: `ClimateZoneOptions2019ASHRAE901`

KarenWGard: I'm not exactly sure how this should be changed.  Seems like the climate_zone doesn't need to be passed as a variable and can be accessed in the function using: `ASHRAE229.weather.climate_zone`, which returns a `ClimateZoneOptions2019ASHRAE901` object

supriyagoel (approved):

mkarpman (approved):

PR #721, opened 2022-10-19
Rds/kjw/function get number of floors

KarenWGard:
simple function to determine whether a climate zone is CZ 0 to 3a


weilixu (approved):

docs/ruleset_functions/get_number_of_floors.md
```diff
@@ -0,0 +1,27 @@
...
+ 
+**Function Call:**
+
+## Logic:
+- create a set of the floor names.  We use a set because only unique elements can be added to the set: `floor_names = set()`
+- loop through the building segments: `for segment in RMD.building.building_segments:`
+	- loop through the zones: `for zone in segment.zones:`
+		- add the floor name to the set of floor names.  If the floor name is already in the set, nothing happens at this step: `floor_names.add(zone.floor_name)`
+- the number of items in the set equals the number of floors in the building.  Set result equal to the length of the set: `result = len(floor_names)`
+
```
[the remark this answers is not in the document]
KarenWGard: we need the number of floors in the entire building.  We are expecting multiple building segments on one floor, and simply returning the BuildingSegment.number_of_floors_above_grade doesn't tell us whether we are double-counting

> Why not use BuildingSegment.number_of_floors_above_grade+BuildingSegment.number_of_floors_below_grade?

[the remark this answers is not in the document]
mkarpman: Not sure using the floor name would work. In eQUEST, side-by-side portion of the building may be modeled as two separate shells and each floor within a shell will have a unique name even if it has the same elevation as another floor object. Perhaps add number of floors to Building? Or add Zone.elevation or Space.elevation? Let's discuss on the call.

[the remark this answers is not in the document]
KarenWGard: updated the function logic based on new schema variables: `zone.elevation` and `zone.ceiling_height`

docs/ruleset_functions/get_number_of_floors.md
```diff
@@ -0,0 +1,54 @@
...
+**Returns:**  
+- **result**: an integer indicating the number of floors
+ 
+**Function Call:**
+
+## Logic:
+- make a list of eligible zones in the building: `zone_list = []`
+	- loop through each zone in the input list: `for zone in zone_input_list: `
+		- eligible zones are zones with ceiling height greater than 7.5 (2.286m)': `if zone.ceiling_height > 2.286`
+			- loop throught the spaces in the zone: `for space in zone:`
```
supriyagoel, line 18: is the height check to filter out plenum zones?

docs/ruleset_functions/get_number_of_floors.md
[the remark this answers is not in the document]
supriyagoel: @mkarpman and @KarenWGard I am not sure what was discussed and the reasoning for going with zone.elevation instead of number of above grade and below grade floors at the building level. It would help to discuss that during the call today

docs/ruleset_functions/get_number_of_floors.md
KarenWGard: ASHRAE defines a zone as being at least 7.5' tall, so yes.

mkarpman: @supriyagoel, plenums are not supposed to be modeled - see note to the "adjacent_to" data element in the RMD.
@KarenWGard I am not sure why you are using ceilign height criteria. 
There is definition of "story" in 90.1 "story: portion of a building that is between one finished floor level and the next higher finished floor level or the roof, provided, however, that a basement or cellar shall not be considered a story."
I think we should use it as the criteria here. 7.5' height only comes up for the gross floor area definition, so I don't think it applies here. Perhaps we need to add building.number_of_floors

docs/ruleset_functions/get_number_of_floors.md
```diff
@@ -0,0 +1,54 @@
+# get_number_of_floors
+**Schema Version:** 0.0.23  
+
+**Description:** gets the number of floors in the building - logic is based on looking at the elevation of zones - zones with an elevation of <0 are considered underground and are not counted.  Parking Garages are not counted
```
mkarpman, line 4: I think we need to count floors below grade. based on G3.1.1.1#2

docs/ruleset_functions/get_number_of_floors.md
```diff
@@ -0,0 +1,54 @@
...
+
+**Inputs:**
+- **zone_input_list** - the list of the zones for which we want to get the number of floors (changed from RMI to this so that we can get the number of floors for a limited list of zones, for example for the function get_hvac_systems_5_6_serving_multiple_floors_b)
```
mkarpman, line 7: It is my understanding that we need the total number of floors in the building, not number of floors for the BAT.

docs/ruleset_functions/get_number_of_floors.md
```diff
@@ -0,0 +1,54 @@
...
+					- add the zone to the zone_list: `zone_list.append(zone)`
+					- break out of the loop: `break`
+
+- sort the list by zone elevation: `zone_list.sort(key = lambda x: x.elevation)`
+
+- create an integer num_of_floors: `num_of_floors = 0`
+
+- when we find floors, we'll remove the zones from the zone_list, until there are no more zones in the list, so we create a loop using `while` and the condition `len(zone_list) > 0`: `while len(zone_list) > 0:`
+	- the first zone in the list will be the one with the lowest elevation, create variable elevation equal to this first zone's elevation: `elevation = zone_list[0].elevation`
+	- until we find a zone with an elevation greater than 7.5' higher than this base elevation, count the zones as being on the same floor - this step insures that zones that have a slightly different elevation, such as a single step up or down do not get counted as being on different floors: `while zone_list[0].elevation <= elevation + 7.5:`
```
mkarpman, line 29: I am not sure why we are using 7.5' in the calculation. Based on the definition of "story" in 90.1 copied below, wouldn't the number of floors in the building be equal to the number of unique "elevations"?  I do not think we should worry about plenums because it is my understanding that they should be combined with the adjacent spaces in the RMD and not listed as a separate zone - see note to adjacent_to field in the schema/

story: portion of a building that is between one finished floor level and the next higher finished floor level or
the roof, provided, however, that a basement or cellar shall not be considered a story.

docs/ruleset_functions/get_number_of_floors.md
KarenWGard: ok.  updated

docs/ruleset_functions/get_number_of_floors.md
KarenWGard: removed the zone_input_list parameter

docs/ruleset_functions/get_number_of_floors.md
KarenWGard: if we are counting below-grade floors, and we the user / UI to name the floors correctly, we can use floor_name and not bother with elevation

(I was using 7.5' to capture spaces that have a slightly different elevation, such as a renovation where there are 1-2 steps up or down to the new part of the building - these zones would have a different elevation, but should still count as the same floor)

docs/ruleset_functions/get_number_of_floors.md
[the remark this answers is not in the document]
mkarpman: The logic above still refers to floor names, so I am confused. I thought we decided to use zone.elevation criteria?

docs/ruleset_functions/get_number_of_floors.md
supriyagoel: as discussed on RDS call, we will use zone elevation instead of the zone name

docs/ruleset_functions/get_number_of_floors.md
supriyagoel: as discussed on RDS call, we will use zone elevation instead of the zone name

docs/ruleset_functions/get_number_of_floors.md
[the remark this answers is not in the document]
KarenWGard: Per the call on 5/1/2023 - I have emailed Mike Rosenberg to ask if he has a consistent definition of the way to define number of floors.  If we don't have a clear and consistent definition, revert to using floor names.

docs/ruleset_functions/get_number_of_floors.md
```diff
@@ -0,0 +1,35 @@
...
+- **RMI**: The baseline ruleset model instance
+
+**Returns:**  
+- **result**: an integer indicating the number of floors
+ 
+**Function Call:**
+
+## Logic:
+- check to see if Building.number_of_floors_above_grade and Building.number_of_floors_below_grade exisits, if so, the sum of these two numbers will be returned. `if Building.number_of_floors_above_grade != null && Building.number_of_floors_below_grade != null:`
+	- set result equal to the number of floors above grade plus the number of floors below grade: `result = Building.number_of_floors_above_grade + null && Building.number_of_floors_below_grade`
```
mkarpman, line 16: This method does not account for the fact that there may be garage floors. Also, what us returned by the function if neither number of floors nor floor names are included in the RMD? Also, what if lighting_space_types are not entered, and we are not able to tell if any zone are garages?

KarenWGard: the note regarding `Building.number_of_floors_above_grade` and `Building.number_of_floors_below_grade` specifically states that parking-only floors should be excluded.

I would argue that if neither is entered, it is not a valid RMD.

Regarding the question about lighting_space_type - @weilixu can correct me here, but I believe that for each function, his team has a list of data elements that are required for the function to evaluate correctly, otherwise there will be some sort of error thrown.  Again, if this information is not provided, I don't think that we can evaluate section 18.

mkarpman: Discussed at 6/1 call that this will not be the required fields in order for an RDSs to be valid. Weili confirmed that if these values are not entered, the RDS that require them will return "undetermined".

mkarpman (approved):

docs/ruleset_functions/get_number_of_floors.md
```diff
@@ -0,0 +1,36 @@
...
+		- loop through the zones: `for zone in segment.zones:`
+			- loop through the spaces in the zone: `for space in zone:`
+				- eligible zones are zones that are not parking garage zones if any of the spaces are not PARKING_GARAGE, then this zone is not exclusively a parking garage space: `if space.lighting_space_type != PARKING_GARAGE:`
+					- add the floor name to the set of floor names.  It doesn't matter if the floor name is already in the set, because a set, by definition can only have one of each item: `floor_names.add(zone.floor_name)`
+	- the number of items in the set equals the number of floors in the building.  Set result equal to the length of the set: `result = len(floor_names)`
+
+**Returns** `result`
+
+**Notes/Questions:**  
+1. Relies on a variable that is not currently in the schema: zone.elevation
```
mkarpman, line 29: This note no longer applies I think.

docs/ruleset_functions/get_number_of_floors.md
```diff
@@ -0,0 +1,39 @@
...
+	- create a set of the floor names.  We use a set because only unique elements can be added to the set: `floor_names = set()`
+	- loop through the building segments: `for segment in RMI.building.building_segments:`
+		- loop through the zones: `for zone in segment.zones:`
+			- using zone_conditioning_category_dict, check if the zone is conditioned or semi-heated. `if zone_conditioning_category_dict[zone.id] == "CONDITIONED RESIDENTIAL" || zone_conditioning_category_dict[zone.id] == "CONDITIONED NON-RESIDENTIAL" || zone_conditioning_category_dict[zone.id] == "CONDITIONED MIXED" || zone_conditioning_category_dict[zone.id] == "SEMI-HEATED"`:
+				- loop through the spaces in the zone: `for space in zone:`
+					- eligible zones are zones that are not parking garage zones if any of the spaces are not PARKING_GARAGE, then this zone is not exclusively a parking garage space: `if space.lighting_space_type != PARKING_GARAGE:`
+						- add the floor name to the set of floor names.  It doesn't matter if the floor name is already in the set, because a set, by definition can only have one of each item: `floor_names.add(zone.floor_name)`
```
supriyagoel, line 26: @KarenWGard the lighting space type should be 'parking_area_interior' correct?
image

We could also consider checking he zone conditioning category returned through the function, if 'unenclosed' then it would imply a parking garage.

docs/ruleset_functions/get_number_of_floors.md
mkarpman: Now that we agreed that we need to count number of floors that are not solely taken by garage and unconditioned space, the logic in line 16 can only work if we update instructions for the corresponding fields in the schema. Is that the plan?

mkarpman (approved):

docs/ruleset_functions/get_number_of_floors.md
KarenWGard: updated to parking_area_interior

we area already filtering for any zones that are not one of the acceptable conditioning categories (CONDITIONED... or SEMI-HEATED), so any unenclosed zones will not make it to the line 25 check.

mkarpman (approved):

supriyagoel (approved):

PR #723, opened 2022-10-19
Create is_zone_mechanically_cooled.md

KarenWGard:
this function will be called twice by the does_zone_meet_G3_1_1e function


KarenWGard:
creating a function to compartmentalize the check for a cooling system, as this needs to happen multiple times in the does_zone_meet_G3_1_1_e function

docs/ruleset_functions/is_zone_mechanically_cooled.md
```diff
@@ -0,0 +1,30 @@
...
+
+**Returns:**  
+- **result**: a boolean - TRUE or FALSE
+ 
+**Function Call:**
+- **get_list_hvac_systems_associated_with_zone**
+
+## Logic:
+- set result to FALSE: `result = FALSE`
+- get the hvac systems serving the zone by using the get_list_hvac_systems_associated_with_zone function: `list_hvac_systems = get_list_hvac_systems_associated_with_zone(RMD,zone)`
```
weilixu, line 18: Is this a function that hasn't been written?

I am not sure how many times this function will be called, but we can also write here as 

for terminal in zone.terminals:
    list_hvac_system_ids.append(terminal.served_by_heating_ventilation_air_condioining_system).

A list of hvac ids should give a good starting point to continue the process below.

docs/ruleset_functions/is_zone_mechanically_cooled.md
```diff
@@ -0,0 +1,30 @@
...
+	- loop through each heating system: `for system in list_hvac_systems:`
+		- look for a cooling system: `if system_p.cooling_system != "Null":`
```
weilixu, line 20: Should we just use `system` instead of `system_p` here?

docs/ruleset_functions/is_zone_mechanically_cooled.md
```diff
@@ -0,0 +1,30 @@
...
+			- check to see if the cooling_system_options equals "None": `if system_p.cooling_system.cooling_system_options != "None":`
+				- there is a cooling system in the zone: `result = TRUE`
```
weilixu, line 22: I am not clear on the definition of mechanically cooled in this clause.
When defining whether a zone is conditioned or not, one threshold is cooling capacity shall be greater than 3.4 Btu/(hr * ft2). Does the threshold applies here too?

If so, it will resolve the question in the Note.

KarenWGard: that's a good question.  G3.1.1.e also mentions zones that receive transfer air from mechanically cooled zones, so I am led to believe that it's a zone with any mechanical cooling capability.  Probably a question for @mkarpman

docs/ruleset_functions/is_zone_mechanically_cooled.md
KarenWGard: changed

docs/ruleset_functions/is_zone_mechanically_cooled.md
KarenWGard: this is a function that exists already

docs/ruleset_functions/is_zone_mechanically_cooled.md
```diff
@@ -0,0 +1,30 @@
+# is_zone_mechanically_cooled
+**Schema Version:** 0.0.22  
+
+**Description:** determines whether a zone is mechanically cooled or not.  DOES NOT CHECK FOR TRANSFER AIR
```
mkarpman, line 4: I don't think transfer air needs to be considered here, so I am OK with it as is.

docs/ruleset_functions/is_zone_mechanically_cooled.md
mkarpman: The distinctions that you discuss above all apply to definition of conditioned/indirectly conditionsed/semiheated spaces so don't think they apply here. I would take the language of this exception literaly - a zone should have a heating system but no cooling system in P-RMD in order to meet this exception.

docs/ruleset_functions/is_zone_mechanically_cooled.md
```diff
@@ -0,0 +1,30 @@
...
+**Returns:**  
+- **result**: a boolean - TRUE or FALSE
+ 
+**Function Call:**
+- **get_list_hvac_systems_associated_with_zone**
+
+## Logic:
+- set result to FALSE: `result = FALSE`
+- get the hvac systems serving the zone by using the get_list_hvac_systems_associated_with_zone function: `list_hvac_systems = get_list_hvac_systems_associated_with_zone(RMD,zone)`
+	- loop through each heating system: `for system in list_hvac_systems:`
```
mkarpman, line 19: I beleive here you mean loop through each HeatingVentilatingAirconditionsSystem? 
I think for the exception should check that zone (1) has a heating system and (2) does not have cooling system. The function now only checks for (2).

docs/ruleset_functions/is_zone_mechanically_cooled.md
mkarpman, line 4: Since schema for transfer air is in flux, I would make the transfer air check a separate function.

docs/ruleset_functions/is_zone_mechanically_cooled.md
KarenWGard: I updated the function and renamed it for clarity to `is_zone_mechanically_heated_and_not_cooled`

docs/ruleset_functions/is_zone_mechanically_cooled.md
KarenWGard: yes.  updated

PR #725, opened 2022-10-20
add error handling

weilixu:
Handle issues
1. When HVAC system do not match any baseline system types (model error since it is baseline system only)
2. When HVAC system matches multiple baseline system types (model error or RCT bug.. will need to think how to handle the RCT bug)

### DIFF
--- a/rct229/ruleset_functions/get_baseline_system_types.py
+++ b/rct229/ruleset_functions/get_baseline_system_types.py
@@ -89,22 +89,18 @@ def get_baseline_system_types(rmi_b):
                 continue
 
             # Add error handling
-            # no matching system warning
-            iterator = iter(
+            hvac_sys_count = len(
                 [
-                    hvac_b_id in baseline_hvac_system_dict[hvac_sys_key]
+                    hvac_b_id
                     for hvac_sys_key in baseline_hvac_system_dict.keys()
+                    if hvac_b_id in baseline_hvac_system_dict[hvac_sys_key]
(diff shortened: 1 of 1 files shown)


rct229/ruleset_functions/get_baseline_system_types.py
```diff
@@ -87,4 +88,25 @@ def get_baseline_system_types(rmi_b):
                 baseline_hvac_system_dict[hvac_sys].append(hvac_b_id)
                 continue
 
+            # Add error handling
```
charliepnnl, line 91: This is a really clever use of iterators!
As much as I do like it, I think it might be more straight forward to get a count of the hvac systems:
```
hvac_sys_count = len([
    hvac_b_id 
    for hvac_sys_key in baseline_hvac_system_dict.keys()
    for hvac_b_id in baseline_hvac_system_dict[hvac_sys_key]
])
if hvac_sys_count == 0:
     raise ...
elif hvac_sys_count > 1:
    raise ...

charliepnnl (approved):
Approved with one comment to consider

yunjoonjung-PNNL (approved):
Looks great to me. Thanks for this effort!

rct229/ruleset_functions/get_baseline_system_types.py
charliepnnl: I think you need a `for` instead of `if` in the list comprehension.

PR #727, opened 2022-10-20
RS+RT/YJ/RULE22-30

yunjoonjung-PNNL:


charliepnnl (changes requested):
My comments for this rule are essentially the same as for Rule22-29.

weilixu:
> My comments for this rule are essentially the same as for Rule22-29.

I second - these two are quite similar

yunjoonjung-PNNL:
I added ```CondensingFluidLoopRule``` instead of ```ChillerRule``` because rule check is performed at the condensing ```fluid_loops``` level rather than chiller level.

weilixu (approved):

PR #730, opened 2022-10-21
Rds/kjw/function get zone eflh

KarenWGard:
creation of full load hours function


get_zone_eflh.md
```diff
@@ -0,0 +1,41 @@
...
+- **flh**: a number equal to the total equivalent full load hours for the year
+ 
+**Function Call:**
+- **get_list_hvac_systems_associated_with_zone**
+
+## Logic:
+- create a list to hold information about the occupancy of each space: `people_info_for_spaces = []`
+- create a variable for the total number of zone occupants: `total_zone_occupants = 0`
+- loop through the spaces in the zone collecting information about occupancy: `for space in zone.spaces:`
+  - append the information about the occupancy to the list: `people_info_for_spaces.append([space.number_of_occupants, space.occupant_multiplier_schedule.hourly_values])`
```
weilixu, line 20: In this case (`people_info_for_spaces`) I think RDS usually use `dictionary` instead of a 2D `list` @claperle could you confirm if that is the case?

get_zone_eflh.md
```diff
@@ -0,0 +1,41 @@
...
+  - add the occupants of this space to the total number of zone occupants: `total_zone_occupants += space.number_of_occupants`
+- get the list of the hvac systems associated with the zone: `hvac_systems_list = get_list_hvac_systems_associated_with_zone(RMD,zone.id)`
+- create an index to which each full load hour will be added: `flh = 0`
+- now calculate flh for each hour of the year: `for hour in range(8760):`
```
weilixu, line 24: This range will need to be updated with the calendar object (`is_leap_year`).

Maybe we just need to get an occupancy schedule and loop over the values instead of using a `range` since RMD only supports hourly schedule (in RCT we decided to not deal with event-based schedule for now).

get_zone_eflh.md
```diff
@@ -0,0 +1,41 @@
...
+    - loop through the hvac_systems_list: `for hvac_system in hvac_systems_list:`
+      - check if the HVAC system fan operating schedule is greater than 1 this hour: `if hvac_system.fan_system.operation_schedule[hour] == 1:`
+        - the system is operational this hour, set the boolean to TRUE: `hvac_systems_operational_this_hour = TRUE`
+    - check if the boolean is true: `if hvac_systems_operational_this_hour:`
+      - add the hour to the flh: `flh += 1`
+
+**Returns** `flh`
+
+**Questions:**
+# 1.  Do we need to do a check to see if FanSystem and it's operating schedule exist before calling hvac_system.fan_system.operation_schedule?
```
weilixu, line 39: RCT handles missing data as undetermined if the data is critical to the logic.

For this case, in Eplus, if an operation schedule is not provided, Eplus uses 1 as default for every hour. Does eQuest has the same behavior?

If software uses user schedule or default value for fan system, then I believe the RCT should expect this to be supplied from the model - thus undetermined if a `fan_system` is missing `operation_schedule`.

@mkarpman @JasonGlazer what do you think?

get_zone_eflh.md
KarenWGard: this brings up the question about if the RMD is for a model that was simulated for less than a year?  I've updated to reference the number of values in any one of the occupant schedules.

get_zone_eflh.md
KarenWGard: so, should I check whether fan_system is "undetermined" in this function?

get_zone_eflh.md
KarenWGard: changed to make it a dict

get_zone_eflh.md
weilixu: I am not sure such case could happen (theoretically) but almost all standards require for a full year simulation.

docs/ruleset_functions/get_zone_eflh.md
```diff
@@ -0,0 +1,43 @@
...
+## Logic:
+- create a list to hold information about the occupancy of each space: `people_info_for_spaces = {}`
+- create a variable for the total number of zone occupants: `total_zone_occupants = 0`
+- loop through the spaces in the zone collecting information about occupancy: `for space in zone.spaces:`
+  - append the information about the occupancy to the list: `people_info_for_spaces[space]["NUM_OCCUPANTS] = space.number_of_occupants`
+  - `people_info_for_spaces[space]["SCHEDULE_HOURLY_VALUES"] = space.occupant_multiplier_schedule.hourly_values`
+  - add the occupants of this space to the total number of zone occupants: `total_zone_occupants += space.number_of_occupants`
+- get the list of the hvac systems associated with the zone: `hvac_systems_list = get_list_hvac_systems_associated_with_zone(RMD,zone.id)`
+- create an index to which each full load hour will be added: `flh = 0`
+- now create a variable num_hours_in_year which will take the number of hours in the schedule for the first (really any) space in people_info_for_spaces: `num_hours_in_year = len(people_info_for_spaces[list(people_info_for_spaces.keys())[0]]["SCHEDULE_HOURLY_VALUES"])`
```
weilixu, line 25: Should we include a check here?

```python
if not all(len(people_info_for_spaces[key]["schedule_hourly_values"]) == num_hours_in_year for key in people_info_for_spaces)
raise exception "Hourly schedules have different number of hours in a year"
```
This might be either addressed in the umbrella check or address in here.

KarenWGard: > Should we include a check here?
> 
> ```python
> if not all(len(people_info_for_spaces[key]["schedule_hourly_values"]) == num_hours_in_year for key in people_info_for_spaces)
> raise exception "Hourly schedules have different number of hours in a year"
> ```
> 
> This might be either addressed in the umbrella check or address in here.

@mkarpman - thoughts?

@weilixu - can you explain how exceptions are handled?  Does this simply give the user an alert, or does it stop the RCT from completing the check?

docs/ruleset_functions/get_zone_eflh.md
```diff
@@ -0,0 +1,43 @@
+# get_zone_eflh
+
+**Description:** provides the equivalent full load hours of the zone.  Equivalent full load hours are defined as: any hour where the occupancy fraction is greater than 5% AND the HVAC system is in occupied mode.  For this function, we are recognizing the HVAC system as being in occupied mode if ANY of the HVAC systems serving the zone are in occupied mode.
+**Applicability Note:** hvac_system.fan_system.operation_schedule is being used as the HVAC operation schedule.  Therefore, this check will not work for radiant systems or other systems that do not include a fan
```
mkarpman, line 4: I am not clear what this functions is for. Let's discuss at the call. Also, given that zones may have multiple spaces and occupancy is defined at space level, should 5% occupancy criteria apply at the space level or zone level? Have not reviewed the function.

KarenWGard: the logic is based on what is outlined in this [PNNL document](https://www.pnnl.gov/main/publications/external/technical_reports/PNNL-26917.pdf IN THE COMMENTS)

docs/ruleset_functions/get_zone_eflh.md
```diff
@@ -0,0 +1,43 @@
...
+**Function Call:**
+- **get_list_hvac_systems_associated_with_zone**
+
+## Logic:
+- create a dictionary to hold information about the occupancy of each space: `people_info_for_spaces = {}`
+- create a variable for the total number of zone occupants: `total_zone_occupants = 0`
+- loop through the spaces in the zone collecting information about occupancy: `for space in zone.spaces:`
+  - append the information about the occupancy to the list: `people_info_for_spaces[space]["NUM_OCCUPANTS] = space.number_of_occupants`
+  - `people_info_for_spaces[space]["SCHEDULE_HOURLY_VALUES"] = space.occupant_multiplier_schedule.hourly_values`
+  - add the occupants of this space to the total number of zone occupants: `total_zone_occupants += space.number_of_occupants`
```
mkarpman, line 22: It seems that 5% occupancy criteria should be determined as follows:
1. Calculating design occupancy of each zone as a product of space.number_of_occupants and max(1, maximum occupant_multiplier_schedule fraction 
2. Determine aggregated occupancy schedule by multiplying space.number_of_occupants by occupant_multiplier_schedule fraction and then summing up the product across all zones to get 8760 occupancy schedule. 
3. For each hour, determine hourly fraction as a ratio of item 2 to item 1. 

Seems that it's aligned with what you are doing except for 1, where you do not account for the schedule value over 1.

docs/ruleset_functions/get_zone_eflh.md
mkarpman: Seems like this should be part of RMD validity check when the file is parsed by RCT. If ScheduleSequenceOption=hourly, but the schedule has fewer than 8760 values, the RMD should be rejected with a report listing all areas where RMD violates the spec. .

docs/ruleset_functions/get_zone_eflh.md
```diff
@@ -0,0 +1,43 @@
...
+    - loop through the hvac_systems_list: `for hvac_system in hvac_systems_list:`
+      - check if the HVAC system fan operating schedule is greater than 1 this hour: `if hvac_system.fan_system.operation_schedule[hour] == 1:`
+        - the system is operational this hour, set the boolean to TRUE: `hvac_systems_operational_this_hour = TRUE`
+    - check if the boolean is true: `if hvac_systems_operational_this_hour:`
+      - add the hour to the flh: `flh += 1`
+
+**Returns** `flh`
+
+**Questions:**
+# 1.  Do we need to do a check to see if FanSystem and it's operating schedule exist before calling hvac_system.fan_system.operation_schedule?
```
mkarpman, line 41: Yes, we need to check for it I think, because if it's missing than the system is continuously on based on the note for the FanSystem.operating_schedule

docs/ruleset_functions/get_zone_eflh.md
KarenWGard: ok, will update

docs/ruleset_functions/get_zone_eflh.md
```diff
@@ -0,0 +1,44 @@
...
+- get the list of the hvac systems associated with the zone: `hvac_systems_list = get_list_hvac_systems_associated_with_zone(RMI,zone.id)`
+- create an index to which each full load hour will be added: `flh = 0`
+- now create a variable num_hours_in_year which will take the number of hours in the schedule for the first (really any) space in people_info_for_spaces: `num_hours_in_year = len(people_info_for_spaces[list(people_info_for_spaces.keys())[0]]["SCHEDULE_HOURLY_VALUES"])`
+- now calculate flh for each hour of the year: `for hour in range(num_hours_in_year):`
+  - calculate the number of people in the zone this hour: `occupants_this_hour = 0`
+  - loop through the people_info_for_spaces: `for space in people_info_for_spaces:`
+    - multiply the max number of people in this space by the schedule value and add it to people_this_hour: `occupants_this_hour += space_people_info[space]["NUM_OCCUPANTS"] * space_people_info[space]["SCHEDULE_HOURLY_VALUES"][hour]`
+  - check if the people this hour is greater than 5%: `if((occupants_this_hour / total_zone_occupants)>0.05:`
```
supriyagoel, line 31: @KarenWGard If I understand what's happening here- we're created a merged scheduled based on actual number of occupants when multiplied by the hourly occupancy fraction. 
And then we check if the occupancy is greater than 5% based on occupants_this_hour / total_zone_occupants


I guess the other way this could've been done is to check if the hourly occupancy fraction >0.05 and if the fan schedule = 1. That would count as FLH.
For the zone occupancy schedule we could have created a merged occupancy schedule where all space level schedules would be merged to create an average zone level schedule. I am sure I am missing something in this logic, and your approach looks good to me.

supriyagoel (approved):

docs/ruleset_functions/get_zone_eflh.md
KarenWGard: @supriyagoel - the reason not to simply check the schedule value is so that we account for the situations in which a modeler has used a schedule with values greater than 1.

PR #731, opened 2022-10-22
Rds/kjw/function get zone target baseline system

KarenWGard:


docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,97 @@
...
+		- `for zone in building_segment.zones:`
+			- `zones_and_systems[zone] = []`
+- get the list of HVAC building area types and zones: `list_building_area_types_and_zones = get_HVAC_building_area_types_and_zones(B-RMR)`
+- get the predominant building area type: `predominant_building_area_type = get_predominant_HVAC_building_area_type(list_building_area_types_and_zones)`
+- get the total number of floors for the building: `num_floors = get_number_of_floors(B-RMR)`
+- first get the expected system type for the predominant building area type.  To do this, we need the total area of the predominant building area type, which includes all BAT's less than 20000 ft2
+- set area equal to the area of the predominant building area type: `area = list_building_area_types_and_zones[predominant_building_area_type]["AREA"]`
+- loop through the building area types: `for bat in list_building_area_types_and_zones:`
+	- if the bat is not equal to the predominant building area type: `if bat != predominant_building_area_type:`
+		- and if the area of this bat < 20,000 ft2: `if list_building_area_types_and_zones[bat]["AREA"] <= 20000:`
```
weilixu, line 41: logic is not consistent in the description vs. in the pseudo code. Should this be `< 20000` ?

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,97 @@
...
+- expected_system_type_list = expected_system_type_from_Table_G3_1_1(predominant_building_area_type,num_floors,area,climate_zone)
+- fill the zones_and_systems list with the expected system and description string for all zones (we'll overwrite the other zones in the next step): `for zone in zones_and_systems:`
+	- set the value to expected_system_type_list: `zones_and_systems[zone] = expected_system_type_list`
+	
+- now we need to go through each 'exception' to Table_G3_1_1 in order.
+
+- G3.1.1b "Use additional system types for nonpredominant conditions (i.e., residential/nonresidential) if those conditions apply to more than 20,000 ft2 of conditioned floor area." uses the same logic that we just used for the expected_system_type_list, except that we select systems based on non-predominant building area types and the area of these non-predominant conditions
+- loop throught the building area types: `for bat in list_building_area_types_and_zones:`
+	- look for any bat's that have more than 20000 ft2 area: `if list_building_area_types_and_zones[bat]["AREA"] >=20000:`
```
weilixu, line 51: do we need to put if-skip to the predominant BAT here?

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,97 @@
...
+- loop through the zones_and_systems to see if any of the zones meets Ge.1.1.c: `for zone in zones_and_systems:`
+	- use the function does_zone_meet_G3_1_1c to determine if this zone meets this requirement: `if does_zone_meet_G3_1_1c(B_RMI,zone,zones_and_systems[zone][0] == YES):`
+		- the zone meets the G3_1_1c requirements, choose system 3 or 4 based on climate zone: `if is_CZ_0_to_3a(climate_zone):`
+			- set the system to "SYS-3" and source to "G3_1_1c": `zones_and_systems[zone] = ["SYS-3",zones_and_systems[zone][0] + " G3_1_1c"]`
```
weilixu, line 61: Is this a string or a list assigning to the key?

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,97 @@
...
+		- else if the climate zone is_CZ_3b_3c_and_4_to_8: `elsif is_CZ_3b_3c_and_4_to_8(climate_zone):`
+			- set the system to "SYS-4" and source to "G3_1_1c": `zones_and_systems[zone] = ["SYS-4",zones_and_systems[zone][0] + " G3_1_1c"]`
+				
+- G3.1.1d "For laboratory spaces in a building having a total laboratory exhaust rate greater than 15,000 cfm, use a single system of type 5 or 7 serving only those spaces.  The lab exhaust fan shall be modeled as constant horsepower reflecting constantvolume stack discharge with outdoor air bypass."
+- not sure how to apply this one
+
+
+- G3.1.1e "Thermal zones designed with heating-only systems in the proposed design serving storage rooms, stairwells, vestibules, electrical/mechanical rooms, and restrooms not exhausting or transferring air from mechanically cooled thermal zones in the proposed design shall use system type 9 or 10 in the baseline building design."
```
weilixu, line 69: I think the first layer of logic is to identify whether the zone is heating-only. Is it in the `zone_meets_G3_1_1e` function?

KarenWGard: yes, it's in the `zone_meets_G3_1_1e` function

docs/ruleset_functions/get_zone_target_baseline_system.md
KarenWGard: yes.  thanks

docs/ruleset_functions/get_zone_target_baseline_system.md
KarenWGard: there are two strings, the first is the expected system type "SYS-3", the second is a string that will be matched to the `target_string` for the individual rules.

However, I've reformatted this to be a dict instead of a paired list

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,97 @@
...
+- **Table G3.1.1**
+
+## Logic:
+- get climate zone (not sure this is the correct path): `climate_zone = ASHRAE229.weather.climate_zone`
+- create zones_and_systems list: `zones_and_systems = {}`
+- pre-fill the list with an empty string:
+	- `for building_segment in RMR.building.building_segments:`
+		- `for zone in building_segment.zones:`
+			- `zones_and_systems[zone] = []`
+- get the list of HVAC building area types and zones: `list_building_area_types_and_zones = get_HVAC_building_area_types_and_zones(B-RMR)`
```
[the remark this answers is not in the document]
KarenWGard: updated to add this logic

docs/ruleset_functions/get_zone_target_baseline_system.md
KarenWGard: The logic is correct.  Description is wrong.

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,127 @@
...
+- get the predominant building area type: `predominant_building_area_type = get_predominant_HVAC_building_area_type(list_building_area_types_and_zones)`
+- get the total number of floors for the building: `num_floors = get_number_of_floors(B-RMI)`
+- first get the expected system type for the predominant building area type.  To do this, we need the total area of the predominant building area type, which includes all BAT's less than 20000 ft2
+- set area equal to the area of the predominant building area type: `area = list_building_area_types_and_zones[predominant_building_area_type]["AREA"]`
+- loop through the building area types: `for bat in list_building_area_types_and_zones:`
+	- if the bat is not equal to the predominant building area type: `if bat != predominant_building_area_type:`
+		- and if the area of this bat <= 20,000 ft2: `if list_building_area_types_and_zones[bat]["AREA"] <= 20000:`
+			- add this area to the area value for determining predominant HVAC system type: `area = area + list_building_area_types_and_zones[bat]["AREA"]`
```
mkarpman, line 40: System types are determined based on area and number of floors of the entire building, not just area and number of floors of that BAT.

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,127 @@
...
+- expected_system_type_list = expected_system_type_from_Table_G3_1_1(predominant_building_area_type,num_floors,area)
+- fill the zones_and_systems list with the expected system and description string for all zones (we'll overwrite the other zones in the next step): `for zone in zones_and_systems:`
+	- set the value to expected_system_type_list: `zones_and_systems[zone] = expected_system_type_list`
+	
+- now we need to go through each 'exception' to Table_G3_1_1 in order.
+
+- G3.1.1b "Use additional system types for nonpredominant conditions (i.e., residential/nonresidential) if those conditions apply to more than 20,000 ft2 of conditioned floor area." uses the same logic that we just used for the expected_system_type_list, except that we select systems based on non-predominant building area types and the area of these non-predominant conditions
```
mkarpman, line 47: Per above, system types are determined based on area and number of floors of the entire building, not just area and number of floors of that BAT.

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,127 @@
...
+- first check that the building is greater than or equal to 40,000 ft2, otherwise it's impossible to have non-predominant space types.  Create variable total_area: `total_area = 0`
+- add the area of all space types: `for bat in list_building_area_types_and_zones:`
+	- `total_area += list_building_area_types_and_zones[bat]["AREA"]`
+- now check whether the total building area is greater than or equal to 40,000 ft2: `if total_area >= 40000:`
+	- check if it is the predominant bat (we have already done this case): `if bat == predominant_building_area_type:`
+		- skip this loop using continue: `continue`
+	- loop throught the building area types: `for bat in list_building_area_types_and_zones:`
+		- look for any bat's that have more than 20000 ft2 area: `if list_building_area_types_and_zones[bat]["AREA"] >=20000:`
+			- we select a secondary system type for this bat: `secondary_system_type = expected_system_type_from_Table_G3_1_1(bat,num_floors,list_building_area_types_and_zones[bat]["AREA"])`
```
mkarpman, line 56: Would need to pass total building area to that function, not BAT area.

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,127 @@
...
+		- append the zone_id to the zones_that_meetG3_1_1c_list: `zones_that_meetG3_1_1c_list.append(zone)`
+- loop through zones_that_meetG3_1_1c_list and change the expected system type and origin for the zones that meet G3.1.1c: `for zone in zones_that_meetG3_1_1c_list:`
+	- change the system origin string: `zones_and_systems[zone]["SYSTEM_ORIGIN"] = "G3_1_1c"` 
+	- the zone meets the G3_1_1c requirements, choose system 3 or 4 based on climate zone: `if is_CZ_0_to_3a():`
+		- set the system to SYS_4: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = SYS_4`
+	- else (the climate zone is_CZ_3b_3c_and_4_to_8): `else:`
+		- set the system to SYS_3: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = SYS_3`
+				
+- G3.1.1d "For laboratory spaces in a building having a total laboratory exhaust rate greater than 15,000 cfm, use a single system of type 5 or 7 serving only those spaces.  The lab exhaust fan shall be modeled as constant horsepower reflecting constantvolume stack discharge with outdoor air bypass."
+	- not sure how to apply this one
```
mkarpman, line 75: Should this be evaluated for applicability only?

KarenWGard: LABORATORY is one of the SpaceFunctionOptions.  I've taken a crack at creating a function - [does_zone_meet_G3_1_1d](https://github.com/pnnl/ruleset-checking-tool/pull/1095)

If we are evaluating for applicability only, I think it makes sense to do this at the rule level - [Rule 18-26](https://onedrive.live.com/edit.aspx?resid=796769604E905084!462&cid=796769604e905084&authkey=!Ap-zMmcPgQlWt0Q&CT=1683226254245&OR=ItemsView)

docs/ruleset_functions/get_zone_target_baseline_system.md
KarenWGard: yes, thank you.  updated

docs/ruleset_functions/get_zone_target_baseline_system.md
KarenWGard: yes, thank you.  updated

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Logic:
+- create zones_and_systems list: `zones_and_systems = {}`
+- pre-fill the list with an empty string:
+	- `for building_segment in RMR.building.building_segments:`
+		- `for zone in building_segment.zones:`
+			- `zones_and_systems[zone] = {}`
+- get the list of HVAC building area types and zones: `list_building_area_types_and_zones = get_HVAC_building_area_types_and_zones(B-RMR)`
+- get the predominant building area type: `predominant_building_area_type = get_predominant_HVAC_building_area_type(list_building_area_types_and_zones)`
+- get the total number of floors for the building: `num_floors = get_number_of_floors(B-RMI)`
```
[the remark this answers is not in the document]
KarenWGard: we have the function... but not the logic in the function get!  But if we can use the function as a placeholder while we figure out the logic, we can at least move forwards with other tasks ;-0

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,135 @@
...
+- loop through zones_that_meetG3_1_1c_list and change the expected system type and origin for the zones that meet G3.1.1c: `for zone in zones_that_meetG3_1_1c_list:`
+	- change the system origin string: `zones_and_systems[zone]["SYSTEM_ORIGIN"] = "G3_1_1c"` 
+	- the zone meets the G3_1_1c requirements, choose system 3 or 4 based on climate zone: `if is_CZ_0_to_3a():`
+		- set the system to SYS_4: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = SYS_4`
+	- else (the climate zone is_CZ_3b_3c_and_4_to_8): `else:`
+		- set the system to SYS_3: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = SYS_3`
+				
+- G3.1.1d "For laboratory spaces in a building having a total laboratory exhaust rate greater than 15,000 cfm, use a single system of type 5 or 7 serving only those spaces.  The lab exhaust fan shall be modeled as constant horsepower reflecting constantvolume stack discharge with outdoor air bypass."
+- select assigned system type based on total building area and principal occupancy type.
+- if the building is OTHER_NON_RESIDENTIAL and 4-5 floors, and < 25,000 sf: `if( predominant_building_area_type == OTHER_NON_RESIDENTIAL && num_floors > 3 && num_floors < 6 && area < 25000 ):`
```
mkarpman, line 75: I don't think we should check for "other non-residential" here. What if building is residential or public assembly or retail, or other non-residential 3 floors or less and 25K SF or less but meets the total exhaust rate stipulated in this exception? I would change the logic to have System 5 if building is (less then 5 floors + less than 150K SF) and Ststem 7 in other cases.

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,135 @@
...
+	- then the system type is SYS-5: `G3_1_1d_expected_system_type = SYS_5`
+- otherwise if the building is 5 floors or fewer and between 25,000 and 150,000 sf: `elif( num_floors < 6 && area >= 25000 && area <= 150000 ):`
+	- then the system type is also SYS-5: `G3_1_1d_expected_system_type = SYS_5`
+- all other cases: `else:`
+	- the system type is SYS-7: `G3_1_1d_expected_system_type = SYS_7`
+- loop through the zones_and_systems to see if any of the zones meets G3.1.1.d: `for zone in zones_and_systems:`
+	- use the function does_zone_meet_G3_1_1d to determine if this zone meets this requirement: `if does_zone_meet_G3_1_1d(B_RMI,zone,zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] == YES):`
+		- change the system origin string: `zones_and_systems[zone]["SYSTEM_ORIGIN"] = "G3_1_1d"` 
+		- set the system to SYS_4: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = G3_1_1d_expected_system_type`
```
mkarpman, line 84: I think line 84 should say "set the system to 5 or 7".

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,135 @@
...
+
+- G3.1.1f "If the baseline HVAC system type is 9 or 10, use additional system types for all HVAC zones that are mechanically cooled in the proposed design."
+- loop through the zones and systems, check only the zones that are SYS-9 or 10: `for zone in zones_and_systems:`
+	- check if it is SYS-9 or SYS-10: `if zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] in [SYS_9, SYS_10]:`
+		- use the function does_zone_meet_G3_1_1f to see if the zone meets the requirements of G3_1_1f: `if does_zone_meet_G3_1_1f(P-RMI,B-RMI,zone.id) == YES:`
+			- change the system origin string: `zones_and_systems[zone]["SYSTEM_ORIGIN"] = "G3_1_1f"`
+			- the zone meets the G3_1_1f requirements, choose system 3 or 4 based on climate zone: `if is_CZ_0_to_3a():`
+				- set the system to SYS_4: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = SYS_4
+			- else (the climate zone is_CZ_3b_3c_and_4_to_8): `else:`
+				- set the system to SYS_3: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = SYS_3`
```
mkarpman, line 104: I think when exception f applies, we need to determine the baseline HVAC system for all such zones combined. So depending on occupancy type and size, it may be a multizone system. For example, if 1M SF warehouse has 100K SF taken by offices (but is not a mixed use building), the baseline system type for the office areas would be determined based on Table G3.1.1. This is clarified in 90.1 2022 G3.2.1.2 d: "d. If the baseline HVAC system type is 9 or 10, use additional system types for all HVAC zones that are mechanically cooled in the proposed design. The baseline HVAC system types for such zones shall be determined based on the building area type determined in accordance with Section G3.2.1.1(a) and the requirements of Section G3.2.1.1(c). "

KarenWGard: discussed on call - will select system type based on zone HVAC building area type (use `list_building_area_types_and_zones`), building total number of floors, and building total area

KarenWGard: updated

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,131 @@
...
+		- else (the climate zone is_CZ_3b_3c_and_4_to_8): `else:`
+			- set the system to SYS_9: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = SYS_9`
+
+
+- G3.1.1f "If the baseline HVAC system type is 9 or 10, use additional system types for all HVAC zones that are mechanically cooled in the proposed design."
+- loop through the zones and systems, check only the zones that are SYS-9 or 10: `for zone in zones_and_systems:`
+	- check if it is SYS-9 or SYS-10: `if zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] in [SYS_9, SYS_10]:`
+		- use the function does_zone_meet_G3_1_1f to see if the zone meets the requirements of G3_1_1f: `if does_zone_meet_G3_1_1f(P-RMI,B-RMI,zone.id) == YES:`
+			- change the system origin string: `zones_and_systems[zone]["SYSTEM_ORIGIN"] = "G3_1_1f"`
+			- now determine the baseline system type using the zone HVAC building area type, the total building area and the total building number of floors.  First by getting the bat by looping through the building area types: `for bat in list_building_area_types_and_zones:`
```
mkarpman, line 98: The example that I have previously provided (and that is shown in "Conversations" tab but not on this tab) is a warehouse project with some office zones that have cooling. Consider the two scenarios below. The logic works for #1 but not for #2:
1. If the project is treated as a mixed use building, have separate BAT specified for warehouse zones versus office zones, and office zones do not qualify as non-predominant condition (<20K SF), the baseline HVAC system type would be properly updated.
2. If the project is not treated as a mixed use building and has "warehouse" BAT assigned to all zones, your logic with will result in mapping office zones back to Sysem 9/10 instead of 3-8. 

I would updat the logic to check if the mapping again points to 9/10, and if so establish BAT based on zone types. You can do that if you isolate the portion of get_HVAC_building_area_types_and_zones that starts with "else look at each zone, and then each space and determine the building_segment_HVAC_BAT using the largest space.lighting_space_type" into a separate function and call it here (and modify get_HVAC_building_area_types_and_zones to call this function also).

KarenWGard: @mkarpman - I started implementing this change, but it is not quite that simple to break out the mapping zone-by zone.  if there is no building segment HVAC_BAT, get_HVAC_building_area_types_and_zones looks at each space lighting type and area, then assigns the building segment HVAC_BAT based on the largest HVAC_BAT as calculated based on spaces.  Let's discuss on the call.

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,131 @@
...
+- loop through the zones_and_systems to see if any of the zones meets this requirement: `for zone in zones_and_systems:`
+	- use the function does_zone_meet_G3_1_1e to determine whether to change the system type (we are using != NO instead of YES, because the function can return MAYBE.  Here we will assign any "maybe's to the new system type.  The rule itself will deal with the uncertainty of vestibules using the function is_zone_a_vestibule): `if does_zone_meet_G3_1_1e(P-RMI,B-RMI,zone) != NO:`
+		- - change the system origin string: `zones_and_systems[zone]["SYSTEM_ORIGIN"] = "G3_1_1e"`
+		- choose system 9 or 10 based on climate zone: `if is_CZ_0_to_3a():`
+			- set the system to SYS_10: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = SYS_10`
+		- else (the climate zone is_CZ_3b_3c_and_4_to_8): `else:`
+			- set the system to SYS_9: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = SYS_9`
+
+
+- G3.1.1f "If the baseline HVAC system type is 9 or 10, use additional system types for all HVAC zones that are mechanically cooled in the proposed design."
```
mkarpman, line 95: Prehaps we should check for G3.1.1f before we do G3.1.1e to avoid havings zones that met exception G3.1.1e re-considered for exception (f). But also OK to keep the current order because the outcome will still be correct.

docs/ruleset_functions/get_zone_target_baseline_system.md
```diff
@@ -0,0 +1,131 @@
...
+				- set the system to SYS_3: `zones_and_systems[zone]["EXPECTED_SYSTEM_TYPE"] = SYS_3`
+
+
+
+**Returns** `zones_and_systems`
+
+
+**Notes/Questions:**  
+1.  Unresolved how to handle G3.1.1d
+2.  For G3.1.1f - which system type should be assigned when the zone meets the requirements?
```
mkarpman, line 129: Both questions are now resolved, right? Then please delete.

mkarpman (approved):

docs/ruleset_functions/get_zone_target_baseline_system.md
KarenWGard: I would like to keep the order because this is the order specified in Section G3.1.1

supriyagoel (approved):

PR #575, opened 2022-08-10
Ruletest 21-5 test JSON

jugonzal07:
Completed first draft of boiler section 21-5 test JSON

PR #735, opened 2022-10-26
Rs/wx/rule21 5

weilixu:
I have to glue many pieces in the `create_data` function to make the rule checking work.
Many of them are intermedia steps and wondering if I should also update RDS for those steps.
Overall, it is tested against the correct json tests (the actual PR #575 contains wrong test json and needs to be updated.)

### DIFF
--- a/rct229/rules/section21/section21rule5.py
+++ b/rct229/rules/section21/section21rule5.py
@@ -48,13 +48,10 @@ class Section21Rule5(RuleDefinitionListIndexedBase):
             data_items={"climate_zone": ("baseline", "weather/climate_zone")},
         )
 
-    class RulesetModelInstanceRule(RuleDefinitionListIndexedBase):
+    class RulesetModelInstanceRule(RuleDefinitionBase):
         def __init__(self):
-            super(Section21Rule5.RulesetModelInstanceRule, self).__init__(
+            super(Section21Rule5.RulesetModelInstanceRule, self,).__init__(
                 rmrs_used=UserBaselineProposedVals(False, True, False),
-                each_rule=Section21Rule5.RulesetModelInstanceRule.BoilerLoop(),
-                index_rmr="baseline",
-                list_path="$.fluid_loops[*]",
             )
 
         def is_applicable(self, context, data=None):
@@ -73,17 +70,20 @@ class Section21Rule5(RuleDefinitionListIndexedBase):
                 ]
             )
 
-        def create_data(self, context, data):
+        def get_calc_vals(self, context, data=None):
             rmi_b = context.baseline
             climate_zone = data["climate_zone"]
-            # join multiple dicts as temp solution. zone_cond_cat function may upgrade to rmi level
+
+            # get zone conditions from buildings
             zone_conditioning_category_dict = {}
             for bldg in find_all("$.buildings[*]", rmi_b):
                 zone_conditioning_category_dict = {
                     **zone_conditioning_category_dict,
                     **get_zone_conditioning_category_dict(climate_zone, bldg),
                 }
+
             loop_zone_list_w_area_dict = get_hw_loop_zone_list_w_area(rmi_b)
+
             # boiler to loop dict
             loop_attach_boiler_dict = {}
             for boiler in find_all("$.boilers[*]", rmi_b):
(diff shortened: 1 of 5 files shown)


charliepnnl (changes requested):
By my reading of the RDS, the rule assertion should be at the RulesetModelInstance level - the top level. So I don't think there should be any nested rules.

rct229/rules/section21/section21rule5.py
```diff
@@ -0,0 +1,161 @@
...
+    """Rule 5 of ASHRAE 90.1-2019 Appendix G Section 21 (Hot water loop)"""
+
+    def __init__(self):
+        super(Section21Rule5, self).__init__(
+            rmrs_used=UserBaselineProposedVals(False, True, False),
+            each_rule=Section21Rule5.RulesetModelInstanceRule(),
+            index_rmr="baseline",
+            id="21-5",
+            description="The baseline building design boiler plant shall be modeled as having a single boiler if the baseline building design plant serves a conditioned floor area of 15,000sq.ft. or less, and as having two equally sized boilers for plants serving more than 15,000sq.ft.",
+            list_path="ruleset_model_instances[*]",
```
charliepnnl, line 47: Elsewhere we have assumed that there is only one RuleModelInstance.
If we assume that here, you can set `list_path="ruleset_model_instances[0]"`. It will then provide a list with one element to the nested rule.

Better yet, if we are assuming only one RuleModelInstance,  the RuleModelInstanceRule could be eliminated and all logic could be performed at the top level.

rct229/rules/section21/section21rule5.py
```diff
@@ -0,0 +1,161 @@
...
+    class RulesetModelInstanceRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section21Rule5.RulesetModelInstanceRule, self,).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+            )
+
+        def is_applicable(self, context, data=None):
+            rmi_b = context.baseline
+            baseline_system_types_dict = get_baseline_system_types(rmi_b)
+            # create a list contains all HVAC systems that are modeled in the rmi_b
```
charliepnnl, line 60: "contains" should be "containing" on line 60

rct229/rules/section21/section21rule5.py
```diff
@@ -0,0 +1,161 @@
...
+            available_type_lists = [
```
charliepnnl, line 61: As with Rule22-10, I suggest naming this variable `available_types` or `available_types_list`.

rct229/rules/section21/section21rule5.py
```diff
@@ -0,0 +1,161 @@
...
+            zone_conditioning_category_dict = {}
+            for bldg in find_all("$.buildings[*]", rmi_b):
+                zone_conditioning_category_dict = {
+                    **zone_conditioning_category_dict,
+                    **get_zone_conditioning_category_dict(climate_zone, bldg),
+                }
+
+            loop_zone_list_w_area_dict = get_hw_loop_zone_list_w_area(rmi_b)
+
+            # boiler to loop dict
```
charliepnnl, line 87: I would think of this a a mapping from the loop id to the boiler id. So in the comment, I might say "loop to boiler dict".

rct229/rules/section21/section21rule5.py
```diff
@@ -0,0 +1,161 @@
...
+            loop_attach_boiler_dict = {}
```
charliepnnl, line 88: You only use the keys from this dictionary. So a list would be more natural. Something like:
```
boiler_loop_ids = [
    getattr_(boiler, "boiler", "loop") 
    for boiler in find_all("$.boilers[*]", rmi_b)
]
```

rct229/rules/section21/section21rule5.py
```diff
@@ -0,0 +1,161 @@
...
+            # The heating_loop_conditioned_zone_area will include sum of all connected zones and indirectly zone areas.
+            heating_loop_conditioned_zone_area = ZERO.AREA
+            # The connected zones list, zones in this list can be residential, nonresidential, mixed or semi-heated
+            loop_zone_list = []
+
+            for fluid_loop in find_all("$.fluid_loops[*]", rmi_b):
+                # Make sure heating loop, and its heating is supplied by a boiler(s)
+                if (
+                    getattr_(fluid_loop, "fluid_loops", "type") == FLUID_LOOP.HEATING
+                    and fluid_loop["id"] in loop_attach_boiler_dict.keys()
```
charliepnnl, line 105: Should the RDS be updated to include the logic of ensuring that the loop is attached to a boiler?

rct229/rules/section21/section21rule5.py
```diff
@@ -0,0 +1,161 @@
...
+            heating_loop_conditioned_zone_area = calc_vals[
+                "heating_loop_conditioned_zone_area"
+            ]
+            num_boilers = calc_vals["num_boilers"]
+            boiler_capacity_list = calc_vals["boiler_capacity_list"]
+            return {
+                heating_loop_conditioned_zone_area
+                <= HEATING_LOOP_CONDITIONED_AREA_THRESHOLD
+                and num_boilers == 1
+                or num_boilers == 2
```
charliepnnl, line 158: Some parentheses would be helpful to the reader if Black will allow them.

charliepnnl (approved):
Approved with some comments

rct229/rules/section21/section21rule5.py
weilixu: that is weird, I thought I changed it. Will update now

weilixu: Oh I got it. It was copy & paste issue.

rct229/rules/section21/section21rule5.py
weilixu: That is correct. Updated

rct229/rules/section21/section21rule5.py
weilixu: Yep, that is allowed. Updated

rct229/rules/section21/section21rule5.py
weilixu: The plan for the ruleset_model_isntances is not entirely clear. I will keep the list_path = 'ruleset_model_instances[0]' for now

PR #593, opened 2022-08-25
Ruletest 21-10 JSON 

yunjoonjung-PNNL:
I believe [this part](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section21/Rule21-10.md#:~:text=Check%20if%20pump%20serves%20heating%20hot%20water%20loop%3A%20if%20pump_b.loop_or_piping.type%20%3D%3D%20%22HEATING%22%3A) in the RDS is wrong. I think like other section 21 rules, this part should also be ```fluid_loops```. 

I would appreciate it if you reviewed this.

PR #736, opened 2022-10-26
Rs/wx/rule21 10

weilixu:
Working as tests shows,
Test json PR needs to be updated. #593

### DIFF
--- a/docs/section21/Rule21-10.md
+++ b/docs/section21/Rule21-10.md
@@ -24,7 +24,7 @@
 
 - Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`
   
-  - Check if B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11.2, 12, 1a, 7a, 11.2a, 12a, 1b, 3b, 5b, 6b, 7b, 8b, 9b, 11.1b, 12b, 1c, 3c, 7c, 11.1c, 12c, continue to rule logic: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-1", "SYS-5", "SYS-7", "SYS-11.2", "SYS-12", "SYS-1A", "SYS-7A", "SYS-11.2A", "SYS-12A", "SYS-1B", "SYS-3B", "SYS-5B", "SYS-6B", "SYS-7B", "SYS-8B", "SYS-9B", "SYS-11B", "SYS-12B", "SYS-13B", "SYS-1C", "SYS-3C", "SYS-7C", "SYS-11C", "SYS-12C", "SYS-13C"]): CHECK_RULE_LOGIC`
+  - Check if B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11.2, 12, 1a, 7a, 11.2a, 12a, 1b, 3b, 5b, 6b, 7b, 8b, 9b, 11.1b, 12b, 1c, 3c, 7c, 11.1c, 12c, continue to rule logic: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-1", "SYS-5", "SYS-7", "SYS-11.2", "SYS-12", "SYS-1A", "SYS-7A", "SYS-11.2A", "SYS-12A", "SYS-1B", "SYS-3B", "SYS-5B", "SYS-6B", "SYS-7B", "SYS-8B", "SYS-9B", "SYS-11.1B", "SYS-12B", "SYS-1C", "SYS-3C", "SYS-7C", "SYS-11.1C", "SYS-12C"]): CHECK_RULE_LOGIC`
 
   - Else, rule is not applicable to B-RMR: `else: RULE_NOT_APPLICABLE`
 
--- a/docs/section21/Rule21-10.md
+++ b/docs/section21/Rule21-10.md
@@ -24,7 +24,7 @@
 
 - Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`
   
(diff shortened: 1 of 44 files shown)


rct229/rules/section21/section21rule10.py
```diff
@@ -0,0 +1,112 @@
...
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        # to avoid pumps in service water heating system
+        loop_zone_list_w_area_dict = get_hw_loop_zone_list_w_area(rmi_b)
+
+        return {"loop_zone_list_w_area_dict": loop_zone_list_w_area_dict}
+
+    def list_filter(self, context_item, data):
+        pump = context_item.baseline
+        loop_or_piping = getattr_(pump, "Pump", "loop_or_piping")
```
charliepnnl, line 84: Since all pumps must have a `loop_or_piping` field, I think it makes sense to make this a required_field in the initializer. Then this list_filter could be simplified a bit to:
```
pump = context_item.baseline
return pump["loop_or_piping"] in data["loop_zone_list_w_area_dict"].keys()
```

rct229/rules/section21/section21rule10.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        # to avoid pumps in service water heating system
+        loop_zone_list_w_area_dict = get_hw_loop_zone_list_w_area(rmi_b)
+
+        return {"loop_zone_list_w_area_dict": loop_zone_list_w_area_dict}
+
+    def list_filter(self, context_item, data):
+        pump = context_item.baseline
```
charliepnnl, line 83: It might be good to add a comment explaining that this method is filtering for pumps with HEATING loops, using the fact that the  keys in `loop_zone_list_w_area_dict` are ids to `HEATING` loops.

rct229/rules/section21/section21rule10.py
```diff
@@ -0,0 +1,112 @@
...
+
+    class PumpRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section21Rule10.PumpRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False)
+            )
+
+        def get_calc_vals(self, context, data=None):
+            pump_b = context.baseline
+            # list_filter has checked the data
```
charliepnnl, line 96: Can you expand on this comment? Checked the data for what?

rct229/rules/section21/section21rule10.py
```diff
@@ -0,0 +1,112 @@
...
+            pump_loop_or_piping = pump_b["loop_or_piping"]
```
charliepnnl, line 97: It might help to name this variable `pump_loop_or_piping_id`.

rct229/rules/section21/section21rule10.py
```diff
@@ -0,0 +1,112 @@
...
+            loop_zone_list_w_area_dict = data["loop_zone_list_w_area_dict"]
+            total_area = loop_zone_list_w_area_dict[pump_loop_or_piping]["total_area"]
+            target_pump_type = PUMP_SPEED_CONTROL.VARIABLE_SPEED
```
charliepnnl, line 100: A slightly more pythonic way do set `target_pump_type` would be
```
target_pump_type = 
    PUMP_SPEED_CONTROL.FIXED_SPEED
    if total_area < PUMP_CONFIGURATION_THRESHOLD
    else PUMP_SPEED_CONTROL.VARIABLE_SPEED
```

charliepnnl (approved):
Approved with comments

rct229/rules/section21/section21rule10.py
weilixu: I checked the schema and the `loop_or_piping` is required data for pump. I think we dont need `required_field` or `getattr_` in this case. 
Updated.

rct229/rules/section21/section21rule10.py
weilixu: My original intention was that the `loop_or_piping` data is checked in the previous level filter function.
`loop_or_piping` is a required data in the schema so we don't need this comment anymore.

rct229/rules/section21/section21rule10.py
weilixu: This is a good one, thanks.

docs/section21/Rule21-10.md
```diff
@@ -24,7 +24,7 @@
...
   
-  - Check if B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11.2, 12, 1a, 7a, 11.2a, 12a, 1b, 3b, 5b, 6b, 7b, 8b, 9b, 11b, 12b, 13b, 1c, 3c, 7c, 11c, 12c, 13c, continue to rule logic: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-1", "SYS-5", "SYS-7", "SYS-11.2", "SYS-12", "SYS-1A", "SYS-7A", "SYS-11.2A", "SYS-12A", "SYS-1B", "SYS-3B", "SYS-5B", "SYS-6B", "SYS-7B", "SYS-8B", "SYS-9B", "SYS-11B", "SYS-12B", "SYS-13B", "SYS-1C", "SYS-3C", "SYS-7C", "SYS-11C", "SYS-12C", "SYS-13C"]): CHECK_RULE_LOGIC`
+  - Check if B-RMR is modeled with at least one air-side system that is Type-1, 5, 7, 11.2, 12, 1a, 7a, 11.2a, 12a, 1b, 3b, 5b, 6b, 7b, 8b, 9b, 11.1b, 12b, 1c, 3c, 7c, 11.1c, 12c, continue to rule logic: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-1", "SYS-5", "SYS-7", "SYS-11.2", "SYS-12", "SYS-1A", "SYS-7A", "SYS-11.2A", "SYS-12A", "SYS-1B", "SYS-3B", "SYS-5B", "SYS-6B", "SYS-7B", "SYS-8B", "SYS-9B", "SYS-11B", "SYS-12B", "SYS-13B", "SYS-1C", "SYS-3C", "SYS-7C", "SYS-11C", "SYS-12C", "SYS-13C"]): CHECK_RULE_LOGIC`
```
charliepnnl, line 27: I still see "SYS-11B" and "SYS-13C" in the logic

rct229/rules/section21/section21rule10.py
```diff
@@ -0,0 +1,116 @@
...
+    HVAC_SYS.SYS_12B,
+    HVAC_SYS.SYS_1C,
+    HVAC_SYS.SYS_3C,
+    HVAC_SYS.SYS_7C,
+    HVAC_SYS.SYS_11_1C,
+    HVAC_SYS.SYS_12C,
+]
+
+PUMP_SPEED_CONTROL = schema_enums["PumpSpeedControlOptions"]
+PUMP_CONFIGURATION_THRESHOLD = 120000 * ureg("ft2")
```
charliepnnl, line 40: As of python 3.6, we can write a number literal using an underscore as the separator.  So you could write:
```
PUMP_CONFIGURATION_THRESHOLD = 120_000 * ureg("ft2")
```
This has the advantage of being more human readable.

charliepnnl (approved):
Approved with two very minor comments

PR #741, opened 2022-10-28
Update get_primary_secondary_loops.md

claperle:


docs/ruleset_functions/get_primary_secondary_loops.md
```diff
@@ -27,26 +27,14 @@ Logic:
 
   - Check if loop type is cooling: `if fluid_loop.type == "COOLING":`
 
-    - Check if loop is connected to both chiller(s) and non-process cooling coil(s), break logic and return an empty dictionary: `if ( fluid_loop.id in chiller_loop_array ) AND ( fluid_loop in non_process_chw_coil_loop_array ): BREAK and return primary_secondary_loop_dictionary = {}`
+    - Check if loop is connected to both chiller(s) and non-process cooling coil(s), break logic and return an empty dictionary (this would be if the system was not modeled as primary-seconday, a fluid_loop will never be connected to a cooling_system.chilled_water_loop): `if ( fluid_loop.id in chiller_loop_array ) AND ( fluid_loop in non_process_chw_coil_loop_array ): BREAK and return primary_secondary_loop_dictionary = {}`
 
     - Else if loop is connected to chiller(s) only: `else if fluid_loop.id in chiller_loop_array:`
 
       - Check if all child loops of loop serve baseline system Type-7, 8, 11.1, 11.2, 12, 13, 7b, 8b, 11b, 12b, 13b only (to exclude CHW loop served by process chiller(s)): `if child_loop_id in non_process_chw_coil_loop_array for child_loop_id in fluid_loop.child_loops:`
```
weilixu, line 34: I think this shall be updated to:
`if any(child_loop_id in non_process_chw_coil_loop_array for child_loop_id in fluid_loop.child_loops)` ?
This sets the logic to true if at least one child loop id in the fluid_loop.child_loops is in the `non_process_chw_coil_loop_array` then do next.

weilixu (commented):
Thank you for initiating this update.
I think this looks great.
I leave one comment for clarification but approve the changes.

weilixu (approved):

docs/ruleset_functions/get_primary_secondary_loops.md
claperle: Makes sense!

weilixu: I think that can be one of the umbrella check?
The schema did have a note saying the child loops shall be secondary loops.
https://github.com/open229/ruleset-model-description-schema/blob/master/docs229/ASHRAE229.schema.json#L2962-L2968

claperle: @weilixu What about a scenario where a secondary loop is serving process equipment? That is what I am concerned about. Do you not think this could occur?

weilixu: I do not have much experiences in modeling secondary loop serving process equipment. my guess is one primary loop points to a list of secondary loops which could include loops serving cooling coils in the AHU, and loops serving coils for the process equipment? In that case, I think the higher level rule checking or functions shall deal with the scenario. Do you think there could be other different configurations?

PR #742, opened 2022-10-28
RS/CH/Rule22-40

charliepnnl:
I could not test this code since we are missing `check_purchased_chw_hhw()` and the test json.


yunjoonjung-PNNL (approved):
Looks great to me. Also, all of your ```TODO``` comments make sense to me. Thank you!

weilixu (approved):
Looks good! Thanks.

PR #749, opened 2022-10-30
Create Rule_18-2.md

KarenWGard:


docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,87 @@
...
+		- find the actual system type of the system (for example, SYS_5 instead of SYS_5B).  Start by looping through the possible system types: `for possible_type in [SYS_1,SYS_2,SYS_3,SYS_4,SYS_5,SYS_6,SYS_7,SYS_8,SYS_9,SYS_10,SYS_11,SYS_12,SYS_13):`
+			- set system_type to possible_type if the result of using the baseline_system_type_compare function is TRUE: `if baseline_system_type_compare(hvac_system_type,possible_type,FALSE): system_type = possible_type`
+		- see if the hvac_system.id is in the list: `if hvac_system.id in? baseline_system_types_dict[hvac_system_type]:`
+  			- Get the list of zones served by the system from zones_and_terminal_unit_list_dict: `zones_served_by_system = zones_and_terminal_unit_list_dict[hvac_system.id]["ZONE_LIST"]`
+     			- check if the system is a single-zone system type: `if any(system_type == sys_type for sys_type in [SYS_1,SYS_2,SYS_3,SYS_4,SYS_9,SYS_10,SYS_11,SYS_12,SYS_13]):`
+     				- there should only be one zone served by the system.  Check the length of zones_served_by_system: `if len(zones_served_by_system) == 1: result = PASS`
+					- otherwise, FAIL: `else: result = FAIL`
```
mkarpman, line 49: When we categorize a system as one of single zone system types (1,2,3,4, etc.) we check that it only serves single zone, so could skip rows 47-48 and just set result to PASS for such systems.

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,87 @@
...
+- create a list of laboratory zones meeting G3_1_1d: `laboratory_zones_meeting_G3_1_1d = [x for x in target_baseline_systems if target_baseline_systems[x]["SYSTEM_ORIGIN"] == G3_1_1d]`
+- evaluate the rule for each HVAC system in the B_RMI: `for hvac_system in B_RMI...heating_ventilating_air_conditioning_systems:`
+	- set the result to FAIL: `result = FAIL`
+	- find the hvac_system type by looking through the baseline_system_types_dict for the hvac_system.id: `for hvac_system_type in baseline_system_types_dict:`
+		- find the actual system type of the system (for example, SYS_5 instead of SYS_5B).  Start by looping through the possible system types: `for possible_type in [SYS_1,SYS_2,SYS_3,SYS_4,SYS_5,SYS_6,SYS_7,SYS_8,SYS_9,SYS_10,SYS_11,SYS_12,SYS_13):`
+			- set system_type to possible_type if the result of using the baseline_system_type_compare function is TRUE: `if baseline_system_type_compare(hvac_system_type,possible_type,FALSE): system_type = possible_type`
```
mkarpman, line 44: Why do we need to know the actual system type?

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,87 @@
...
+		- see if the hvac_system.id is in the list: `if hvac_system.id in? baseline_system_types_dict[hvac_system_type]:`
+  			- Get the list of zones served by the system from zones_and_terminal_unit_list_dict: `zones_served_by_system = zones_and_terminal_unit_list_dict[hvac_system.id]["ZONE_LIST"]`
+     			- check if the system is a single-zone system type: `if any(system_type == sys_type for sys_type in [SYS_1,SYS_2,SYS_3,SYS_4,SYS_9,SYS_10,SYS_11,SYS_12,SYS_13]):`
+     				- there should only be one zone served by the system.  Check the length of zones_served_by_system: `if len(zones_served_by_system) == 1: result = PASS`
+					- otherwise, FAIL: `else: result = FAIL`
+     			- otherwise it is a multi-zone system:`else:`
+					- for a multi-zone system, in a building without lab zones, the system can pass if:
+						1. all zones are on the same floor
+						2. there are no other systems of the same type that serve zones on the same floor
```
mkarpman, line 53: There is a function get_dict_of_zones_hvac_sys_serving_specific_floor that returns a list of zones on one floor with their associated HVAC systems. Can we call this function for each floor in the building and verify that there is no more than one multizone system per floor is modeled? If this check fails, then the only excuse is that another system is modeled for all labs in the building.

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,87 @@
...
+											- check if the system type is 5 or 7: `if any(system_type == sys_type for sys_type in [SYS_5,SYS_7]):`
+												- check if all of the zones served by hvac_system are lab zones: `if set(laboratory_zones_meeting_G3_1_1d) == set(zones_served_by_system):`
+													- check if the building has greater than 15,000 cfm lab exhaust.  If so, set result to PASS.  If not, set result to UNDETERMINED: `if lab_zone_exhaust > 15000: result = TRUE`
+													- else set result to UNDETERMINED: `result = UNDETERMINED`
+												- else check if all of the zones served by hvac_system2 are lab zones: `if set(laboratory_zones_meeting_G3_1_1d) == set(zones_served_by_system2):`
+													- check if the building has greater than 15,000 cfm lab exhaust.  If so, set result to PASS.  If not, set result to UNDETERMINED: `if lab_zone_exhaust > 15000: result = TRUE`
+													- else set result to UNDETERMINED: `result = UNDETERMINED`
+						- at this point, there is still one case we haven't handled - a system that is not type 5 or 7 and incorrectly serves a lab type zone.  This could be the case if the result is currently PASS: `if result == PASS:`
```
mkarpman, line 70: Wouldn't this scenario be covered by 18-1?

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,87 @@
...
+**Appendix G Section Reference:** Table G3.1.1.3  
+
+**Evaluation Context:** Each HVAC System  
+**Applicability Checks:**  
+
+1. the target string matches, which indicates:  
+	a. the space type is PUBLIC_ASSEMBLY  
+	b. climate zone CZ_0_to_3a  
+	c. area >= 120,000 ft2  
+	d. none of the G3.1.1 exceptions apply
```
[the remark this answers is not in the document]
KarenWGard: does not apply to the new rule structure.  removed

docs/section18/Rule_18-2.md
KarenWGard: I don't think that this function will significantly reduce complexity.  Most of the complexity in this RDS is due to checking for lab zones

docs/section18/Rule_18-2.md
KarenWGard: 18-1 covers the correct system type for lab zones.  But lab zones on multiple floors must be served a multi-zone system (5 or 7).

docs/section18/Rule_18-2.md
KarenWGard: to check for duplicate systems of the same type on the same floor

docs/section18/Rule_18-2.md
KarenWGard: changed

docs/section18/Rule_18-2.md
mkarpman: get_baseline_system_type returns system types including letters (5a, 5b, etc.) So why do we need to look into "possible_type"? Then baseline_system_type_compare(TRUE) can be called.

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,80 @@
...
+     				- only systems that serve single zones can be categorized as single zone systems, so set result to pass: `result = PASS`
+     			- otherwise it is a multi-zone system:`else:`
+					- for a multi-zone system, in a building without lab zones, the system can pass if:
+						1. all zones are on the same floor
+						2. there are no other systems of the same type that serve zones on the same floor
+					- get the list of zones on the same floor by using the function get_zones_on_same_floor: `zones_on_floor = get_zones_on_same_floor(B_RMI,zones_served_by_system[0])`
+					- check if all the zones served by the system are on the same floor: `if all(zone in zones_on_floor for zone in zones_served_by_system):`
+						- now check if there are any other systems of the same system type that serve zones on this floor.  First loop through baseline_system_types_dict again: `for hvac_system_type2 in baseline_system_types_dict:`
+							- check if the hvac_system_type2 is the same type: `if baseline_system_type_compare(hvac_system_type2,system_type,FALSE):`
```
mkarpman, line 50: Should we use baseline_system_type_compare(TRUE) here? Otherwise, we'll not allow systems of the same type but with different letters (e.g., 5 and 5a) on the same floor.

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,80 @@
...
+**Data Lookup:**   
+**Function Call:** 
+
+1. get_baseline_system_types()
+2. get_zones_on_same_floor()
+3. get_dict_of_zones_and_terminal_units_served_by_hvac_sys()
+4. baseline_system_type_compare()
+5. get_zone_target_baseline_system()
+6. get_building_total_lab_exhaust_from_zone_exhaust_fan()
+
```
mkarpman, line 24: Supriya proposed at the last call to have a separate function to check if all lab zones are served by the same HVAC system. Since this is also used in Rule 18-23, perhaps we should do that. It will also simplify verification for this rule.

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,92 @@
...
+                			- if we can tell with 100% certainty that the building has > 15,000 cfm lab exhaust AND the system type is 5 or 7 this system can pass: `if lab_zone_exhaust > 15000:`
+                   				- check if the system type is 5 or 7: `if any(system_type == sys_type for sys_type in [SYS_5,SYS_7]): result = PASS`
+                       				- else, set result to fail: `else: result = FAIL; note = "This system serves all of the lab zones in a building with >15,000 cfm lab exhaust, but it is not system type 5 or 7."
+                   			- otherwise, there *might* be > 15000 cfm exhaust, but we are not sure (the function get_lab_zone_hvac_systems relies on get_zone_target_baseline_system, which does the > 15000 cfm check, so we don't have to do it again): `else:`
+                      				- set result to UNDETERMINED:  `result = UNDETERMINED`
+       						- provide a note expaining the result: `note = "This system serves only lab zones, which is correct if the building has total lab exhaust greater than 15,000 cfm.  However, we could not determine with accuracy the total building exhuast."`
```
mkarpman, line 56: Agreed

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,92 @@
...
+                		- otherwise if the hvac system is in the list (but there is more than one HVAC system): `elif hvac_system.id in? lab_zone_hvac_systems["LAB_ZONES_ONLY]:`
+                			- if we can tell with 100% certainty that the building has > 15,000 cfm lab exhaust, set result to FAIL: `if lab_zone_exhaust > 15000:`
```
mkarpman, line 58: Agreed

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,92 @@
...
+				                - set result to fail: `result = FAIL`
+                      				- provide a note to explain the result: `note = "This HVAC system serves lab zones in a building with > 15,000 cfm of laboratory exhaust.  The baseline system should be type 5 or 7 and should serve ALL laboratory zones."`
+                   			- otherwise, there *might* be > 15000 cfm exhaust, but we are not sure (the function get_lab_zone_hvac_systems relies on get_zone_target_baseline_system, which does the > 15000 cfm check, so we don't have to do it again): `else:`
+                      				- set result to UNDETERMINED:  `result = UNDETERMINED`
+       						- provide a note expaining the result: `note = "This system serves some of the lab zones in a building which may have more than 15,000 cfm.  In buildings with > 15,000 cfm of lab exhaust, ALL lab zones should be served by system type 5 or 7."`
```
mkarpman, line 63: Agreed

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,92 @@
...
+                  		- otherwise, the hvac system might be a system that serves both lab zones and regular zones: `elif hvac_system.id in? lab_zone_havc_systems["LAB_AND_OTHER"]:`
+                    			- if we can tell with 100% certainty that the building has > 15,000 cfm lab exhaust, set result to FAIL: `if lab_zone_exhaust > 15000:`
+				                - set result to fail: `result = FAIL`
```
mkarpman, line 66: Agreed

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,92 @@
...
+                      				- provide a note to explain the result: `note = "This HVAC system serves lab zones in a building with > 15,000 cfm of laboratory exhaust.  The baseline system for laboratory zones should be type 5 or 7 and should serve only and all laboratory zones."`
+                   			- otherwise, there *might* be > 15000 cfm exhaust, but we are not sure (the function get_lab_zone_hvac_systems relies on get_zone_target_baseline_system, which does the > 15000 cfm check, so we don't have to do it again): `else:`
+                      				- set result to UNDETERMINED:  `result = UNDETERMINED`
```
mkarpman, line 69: This "Undetermined" outcome will need to be revisited in rows 72 and down, so perhaps it should be set to "maybe undetermined".  For example, if this building "might" be >15000 but actually isn't, and there is two multi-zone system systems on the same floor, the outcome should be FAIL.

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,92 @@
...
+       						- provide a note expaining the result: `note = "This system serves some lab zones and some non-lab zones in a building which may have more than 15,000 cfm.  In buildings with > 15,000 cfm of lab exhaust, ALL and only lab zones should be served by system type 5 or 7."`
+                      
+                   		 - otherwise, now we know that this system does not serve any laboratory zones.  It is a multi-zone system.  Continue checking whether all the zones it serves are on the same floor: `else:`
+                      			- get the list of zones on the same floor by using the function get_zones_on_same_floor: `zones_on_floor = get_zones_on_same_floor(B_RMI,zones_served_by_system[0])`
+				- check if all the zones served by the system are on the same floor: `if all(zone in zones_on_floor for zone in zones_served_by_system):`
+					- now check if there are any other systems of the same system type that serve zones on this floor.  First loop through baseline_system_types_dict again: `for hvac_system_type2 in baseline_system_types_dict:`
```
mkarpman, line 75: Should this check for any other multizone systems on the same floor?

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,92 @@
...
+						- check if the hvac_system_type2 is the same type: `if baseline_system_type_compare(hvac_system_type2,system_type,FALSE):`
+							- look through all of the hvacs of this system type: `for hvac_system2_id in baseline_system_types_dict[hvac_system_type2]:`
+								- make sure that hvac_system2_id is not the same as hvac_system_id: `if hvac_system2_id != hvac_system_id:`
+									- now check if hvac_system2 serves any of the zones on the same floor as hvac_system.  Get the zones served by hvac_system2: `zones_served_by_system2 = zones_and_terminal_unit_list_dict[hvac_system2_id]["ZONE_LIST"]`
+									- use set.intersection to see if any of these zones are on the same floor: `if(len(set(zones_served_by_system2).intersection(zones_on_floor))) > 0:`
+										- the system fails: `result = FAIL`
+          									- UNLESS system_type is SYS_5 or SYS_7 AND all of the zones in hvac_system2 are lab zones: `if hvac_system2_id in? lab_zone_hvac_systems["LAB_ZONES_ONLY"] && len(lab_zone_hvac_systems["LAB_ZONES_ONLY"]) == 1 && any(system_type == sys_type for sys_type in [SYS_5,SYS_7]): result = PASS`
```
mkarpman, line 82: I don't think this would work. Hvac_system2 may have all lab zones, but if this is the building with exhaust <15000CFM this system shouldn't be there and system 1 should serve all the zones on this floor.

docs/section18/Rule_18-2.md
KarenWGard: discussed on call 7/20/2023, keep as is

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,92 @@
...
+7. get_lab_zone_hvac_systems()
+
+
+**Applicability Checks:**
+
+## Rule Logic:  
+- use the function get_baseline_system_types to get a dictionary of baseline system types and systems: `baseline_system_types_dict = get_baseline_system_types(B-RMI)`
+- use the function get_dict_of_zones_and_terminal_units_served_by_hvac_sys to determine which zones are served by each system system: `zones_and_terminal_unit_list_dict = get_dict_of_zones_and_terminal_units_served_by_hvac_sys(B_RMI)`
+- use the function get_zone_target_baseline_system to get the target baseline system types for each zone.  This will be used when checking for laboratory systems: `target_baseline_systems = get_zone_target_baseline_system(P_RMI, B_RMI)`
+- use the function get_building_total_lab_exhaust_from_zone_exhaust_fan to get the totat lab exhaust air flow from zone exhaust fans.  This will be used when checking for laboratory systems: `lab_zone_exhaust = get_building_total_lab_exhaust_from_zone_exhaust_fan(P_RMI)`
```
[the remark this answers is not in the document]
KarenWGard: no.  will delete

docs/section18/Rule_18-2.md
```diff
@@ -0,0 +1,92 @@
...
+        			- we will deal first with the case of a system serving laboratory zones.  An hvac system serving laboratory zones needs to meet the following:
+             				1.  can only serve lab zones
+             				2.  serves ALL the lab zones
+             				3.  is in a building with > 15,000cfm lab exhaust
+             			- the function get_lab_zone_hvac_systems gets the hvac systems serving the lab zones.  If this system is in the list, AND it is the only one in the list, then this is a lab zone system: `if hvac_system.id in? lab_zone_hvac_systems["LAB_ZONES_ONLY"] && len(lab_zone_hvac_systems["LAB_ZONES_ONLY"]) == 1:`
+                			- if we can tell with 100% certainty that the building has > 15,000 cfm lab exhaust AND the system type is 5 or 7 this system can pass: `if lab_zone_exhaust > 15000:`
+                   				- check if the system type is 5 or 7: `if any(system_type == sys_type for sys_type in [SYS_5,SYS_7]): result = PASS`
+                       				- else, set result to fail: `else: result = FAIL; note = "This system serves all of the lab zones in a building with >15,000 cfm lab exhaust, but it is not system type 5 or 7."
```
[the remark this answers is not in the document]
KarenWGard: ok, changed

docs/section18/Rule_18-2.md
KarenWGard: we agreed on call 7/20/2023 to check only for systems of the same type on the same floor

docs/section18/Rule_18-2.md
KarenWGard: will modify to have the check on line 82 be:

if hvac_system2_id in? lab_zone_hvac_systems["LAB_ZONES_ONLY"] && len(lab_zone_hvac_systems["LAB_ZONES_ONLY"]) == 1 && lab_zone_exhaust > 15000: result = PASS

then if it's the only lab zone system, but we aren't sure about the exhaust air volume, result = UNDETERMINED

otherwise the other system isn't a lab system, result = FAIL

docs/section18/Rule_18-2.md
KarenWGard: the indicated lines were removed and rule logic was modified, adding a boolean do_multi_zone_evaluation, which indicates whether we need to do the multi-zone system evaluation for any systems.

mkarpman (approved):

supriyagoel (approved):
small comment- should the rule assertion be at the HVAC system level rather than the zone level?

yunjoonjung-PNNL:
Could you check how the```lab_zone_exhaust``` is calculated? It seems like it's not explained in RDS.

mkarpman:
This is done using get_building_total_lab_exhaust_from_zone_exhaust_fans.md (PR #1114) [https://github.com/pnnl/ruleset-checking-tool/pull/1114](url)

PR #750, opened 2022-10-30
Create section_18_rule_test.md

KarenWGard:
the section 18 rule tests are very repetitive with the same logic used in ~20 rules.  This function encapsulates the logic so that edits and bug fixes are much easier to do.

### DIFF
--- a/docs/ruleset_functions/section_18_rule_test.md
+++ b/docs/ruleset_functions/section_18_rule_test.md
@@ -3,8 +3,8 @@
 **Description:** uses the function get_zone_target_baseline_system to get the targets for the entire building.  Accepts two enum values as inputs, one is the system type expected by the rule (ex: SYS_3,SYS_7), and the other is the target enum, which is used to determine if the rule calling the function was the one used to set the system designation
 
 **Inputs:**  
-- **P-RMD**: The proposed RMD
-- **B-RMD**: The baseline RMD
+- **P-RMI**: The proposed RMI
+- **B-RMI**: The baseline RMI
 - **zone_id**: the id of the zone to be tested
 - **expected_system_type** - an enum describing the system type expected by the rule calling this function (ex: SYS_2)
 - **target_string** - the target string, which will match if this rule was the one used to designate the system (ex: "PUBLIC_ASSEMBLY CZ_3b_3c_or_4_to_8 < 120,000 ft2")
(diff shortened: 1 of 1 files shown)


docs/ruleset_functions/section_18_rule_test.md
```diff
@@ -0,0 +1,39 @@
...
+
+**Returns:**  
+- **result**: enum indicating PASS, FAIL, NOT_APPLICABLE
+ 
+**Function Call:** 
+
+1. get_zone_target_baseline_system()
+2. is_hvac_system_of_type()
+3. get_list_hvac_systems_associated_with_zone()
+4. get_zone()
```
weilixu, line 21: A typical approach for get_zone in RDS is just zone = get_component_by_id(zone_id).
I think it is referenced in many places, but not sure if it is a convention or not.

But whenever you need to get something by ID in a RMI, that is the function name RDS will use.

docs/ruleset_functions/section_18_rule_test.md
```diff
@@ -0,0 +1,39 @@
...
+5. is_hvac_system_of_type()
+
+## Logic:  
+- get the zone: `get_zone(zone_id)`
+- check whether the description string in zones_expected_system_types matches the target_string: `if zones_expected_system_types[zone][1] == target_string:`
+- now check whether the system(s) serving this zone match the expected system.  Start by getting the list of HVAC systems that serve the zone: `hvac_systems_serving_zone = get_list_hvac_systems_associated_with_zone(B-RMD)`
+	- loop through these systems: `for system in hvac_systems_serving_zone:`
+		- set result to PASS: `result = PASS`
+		- now check that system_b is the expected type by using the is_hvac_system_of_type() function: `if is_hvac_system_of_type(B-RMI,system.id,"SYS-2"):`
```
weilixu, line 30: Why does this need to be "SYS-2"? Is it an example here?

docs/ruleset_functions/section_18_rule_test.md
KarenWGard: updated

docs/ruleset_functions/section_18_rule_test.md
```diff
@@ -0,0 +1,39 @@
...
+
+1. get_zone_target_baseline_system()
+2. is_hvac_system_of_type()
+3. get_list_hvac_systems_associated_with_zone()
+4. get_zone()
+5. is_hvac_system_of_type()
+
+## Logic:  
+- get the zone: `get_zone(zone_id)`
+- check whether the description string in zones_expected_system_types matches the target_string: `if zones_expected_system_types[zone][1] == target_string:`
```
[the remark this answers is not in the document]
KarenWGard: corrected.  it comes from the function `get_zone_target_baseline_system`, but I forgot to include the line for the function call

docs/ruleset_functions/section_18_rule_test.md
KarenWGard: it should have been `expected_system_type`.  It's been corrected.

docs/ruleset_functions/section_18_rule_test.md
```diff
@@ -0,0 +1,38 @@
...
+- get the zone: `zone = get_component_by_id(B-RMI,zone_id)`
+- get the expected system type using `get_zone_target_baseline_system()`: `zones_expected_system_types = get_zone_target_baseline_system(P-RMI,B-RMI)`
+- check whether the description string in zones_expected_system_types matches the target_string: `if zones_expected_system_types[zone]["SYSTEM_ORIGIN"] == target_string:`
+	- now check whether the system(s) serving this zone match the expected system.  Start by getting the list of HVAC systems that serve the zone: `hvac_systems_serving_zone = get_list_hvac_systems_associated_with_zone(B-RMI)`
+		- loop through these systems: `for system_b in hvac_systems_serving_zone:`
+			- set result to PASS: `result = PASS`
+			- loop through the baseline_hvac_system_dict looking for the expected system type - if any of the baseline HVAC systems do not match the expected system type result will be set to FAIL: `for hvac_system_type in baseline_hvac_system_dict:`
+				- check if hvac_system type is the same type (or one of the sub-types of expected_system_type: `if !baseline_system_type_compare(hvac_system_type, expected_system_type, false):`
+					- this system passes: `result = FAIL`
```
mkarpman, line 33: I am confused why it says that the system passes, but result is FAIL?
Also, don't we want an exact match, so the third argument should be true and not false?

KarenWGard:
not necessary with rule re-structuring (ie 18-1)

PR #751, opened 2022-10-30
Create is_hvac_system_of_type.md

KarenWGard:
this function makes it easy to compare a system type that results from Table G3.1.1 ("SYS-5") with the detailed system type given by `get_list_hvac_systems_associated_with_zone`, which might give "SYS-5a"


weilixu:
This function seems to be addressed in #820

weilixu:
Close because the function was addressed in #820

PR #753, opened 2022-10-31
RS+RT/YH/RULE 22-34

yunjoonjung-PNNL:
```primary_secondary_loop_dictionary``` function part should be updated later. I will double-check this part once ```primary_secondary_loop_dictionary```function is available. Thank you!

### CODE REVIEW

**Comment by yunjoonjung-PNNL** [comment]
This code was updated to use ```get_primary_secondary_loops_dict``` function. Thanks!

File: rct229/rules/section22/section22rule34.py [hunk 1]
```diff
@@ -0,0 +1,94 @@
...
+        )
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        primary_secondary_loop_dictionary = get_primary_secondary_loops_dict(rmi_b)
+
+        return {"primary_secondary_loop_dictionary": primary_secondary_loop_dictionary}
+
+    def list_filter(self, context_item, data):
+        fluid_loops_b = context_item.baseline
```
weilixu, line 63: Based on RDS logic, I think the filter shall be using `primary_secondary_loop_dictionary'

```python
return fluid_loops_b["id] in primary_secondary_loop_dictionary.keys()
```

yunjoonjung-PNNL: Right! Thanks for catching this up: ). Fixed.

rct229/rules/section22/section22rule34.py
```diff
@@ -0,0 +1,93 @@
...
+                    "$": ["cooling_or_condensing_design_and_control"],
+                    "cooling_or_condensing_design_and_control": [
+                        "is_sized_using_coincident_load"
+                    ],
+                },
+            )
+
+        def get_calc_vals(self, context, data=None):
+            fluid_loop_b = context.baseline
+            is_sized_using_coincident_load_bool = fluid_loop_b[
```
charliepnnl, line 80: Using `is_` at the beginning of the variable name indicates it holds a boolean value, so there is no need to also use `_bool` at the end.

charliepnnl (approved):
Just one tiny style-type comment

rct229/rules/section22/section22rule34.py
weilixu: Was this issue updated?

yunjoonjung-PNNL: Yes! It was updated.

weilixu (approved):

PR #755, opened 2022-11-01
Rx/wx/update test with primary secondary loops

weilixu:


yunjoonjung-PNNL (approved):
Everything looks great to me except for one error from formatting. Could you run ```black``` before merging?

PR #756, opened 2022-11-01
Create 19-17.md

claperle:


docs/section19/19-17.md
```diff
@@ -0,0 +1,39 @@
...
+    - Check if values are not none for system type key in the dictionary, if it does not equal false then the values are not none and there are applicable system types in the project: `if bool(baseline_system_types_dict.get(sys_type)) != False: is_there_applicable_sys_type = TRUE`
+- Check if is_there_applicable_sys_type == True which means there are applicable system types in the project, if yes then carry on, if no then NOT_APPLICABLE for the project: `if  is_there_applicable_sys_type == True:`    
+    - For each hvac system in the B_RMI: `for hvac in B_RMI...HeatingVentilatingAirConditioningSystem:` 
+        - Check if system type is of an applicable system type, if yes then carry on, if no then NOT_APPLICABLE for the hvac system:  `If any(hvac.id in baseline_hvac_system_dict[sys_type] for sys_type in ["SYS-9","SYS-9b","SYS-10"]): `    
+            - Get the supply temperature setpoint: `supply_air_temp_setpoint = hvac.fan_system.supply_air_setpoint_temperature` 
+
+            - **Rule Assertion:** 
+            - Case 1:If the supply air temperatures setpoint equals 105F : `if supply_air_temp_setpoint == 105: outcome = "PASS"`
```
mkarpman, line 33: I think we need to also check whether fan flow is autosized (need to add schema element for this). Also, should check that FanSystem.minimum_airflow>FanSystem.Minimum_Outdoor_Airflow (FAIL when this doesn't hold, unless AllowMoreLenient flag is set to TRUE). I think we also should add a note to the schema that the minimum_outdoor_airflow should be reported even when the fan flow is autosized, so this value will be based on simulation results for the sizing runs. (Same goes for autosized heating and cooling capacities.)

docs/section19/19-17.md
```diff
@@ -0,0 +1,39 @@
...
+    - For each hvac system in the B_RMI: `for hvac in B_RMI...HeatingVentilatingAirConditioningSystem:` 
+        - Check if system type is of an applicable system type, if yes then carry on, if no then NOT_APPLICABLE for the hvac system:  `If any(hvac.id in baseline_hvac_system_dict[sys_type] for sys_type in ["SYS-9","SYS-9b","SYS-10"]): `    
+            - Get the supply temperature setpoint: `supply_air_temp_setpoint = hvac.fan_system.supply_air_setpoint_temperature` 
+
+            - **Rule Assertion:** 
+            - Case 1:If the supply air temperatures setpoint equals 105F : `if supply_air_temp_setpoint == 105: outcome = "PASS"`  
+            - Case 2: Else, UNDETERMINED: `Else: outcome = "UNDETERMINED and raise_message "The supply air temperature set point does not equal 105°F for [hvac.id]. Recommend a manual check to determine if the supply airflow rates were set based on minimum outdoor airflow rates or the airflow rate required to comply with applicable codes or accreditation standards, whichever is greater rather than a supply air temperature setpoint of 105F."`
```
mkarpman, line 34: I think this should be checked by the RDS that verifies baseline ventilation flow rate and not this one. So I think this RDS should produce PASS or FAIL outcome, no undertermined option..

claperle: I am not sure I understand how this is checked with another RDS. Can you please clarify?

docs/section19/19-17.md
claperle: All done except did not implement the allowmorelenient flag per discussion on the call.

docs/section19/19-17.md
```diff
@@ -0,0 +1,45 @@
...
+            - Get list of terminal units served by the hvac system: `terminal_list_hvac_sys_x = dict_of_zones_and_terminal_units_served_by_hvac_sys_x[hvac.id]["ZONE_LIST"]["Terminal_Unit_List"]`  
+            - For each terminal unit associated with the hvac system (should only be one): `for terminal in terminal_list_hvac_sys_x:`  
+                - Get the design supply temperature setpoint: `design_supply_air_temp_setpoint = terminal.supply_design_heating_setpoint_temperature`  
+                - Check if the design supply air temperature setpoint does not equal 105F, if it doesn't then set all_design_setpoints_105 to false: `if design_supply_air_temp_setpoint != 105: all_design_setpoints_105 = false`  
+            - **Rule Assertion:** 
+            - Case 1: If the design supply air temperatures setpoint equals 105F, the fan is autosized and the FanSystem.minimum_airflow>FanSystem.Minimum_Outdoor_Airflow  : `if all_design_setpoints_105 == true and hvac.fan_system.minimum_airflow >= hvac.fan_system.minimum_outdoor_airflow and are_all_hvac_sys_fan_objs_autosized(B_RMI, hvac) == True : outcome = "PASS"`
```
mkarpman, line 40: Since fans are autosized, we can only get fansystem.minimum_airflow based on simulation outputs. Agreed? If so, I think we need to add a note to this effect to the schema. It would say that this field may be either simulation input or (if airflow is autosized) simulation output.

docs/section19/19-17.md
mkarpman: What I meant is that we have another RDS (or an excluded rule) to verify that the OA modeled in the baseline does not exceed the minimum required by code. So for Case 2, we can check if fansystem.minimum_airflow==fansystem.minimum_outdoor_airflow. If true, the outcome will be undetermined; if false the outcome would be FAIL.

docs/section19/19-17.md
claperle: Discuss on call

docs/section19/19-17.md
claperle: I think there can be other circumstances that do not involve min OA that are requirements of codes and standards. I actually just had a call recently about a healthcare facility that said 6 ACH were required and that only 2 ACH were from OA. I followed up with him to get more details. Discuss this on call.

docs/section19/19-17.md
```diff
@@ -0,0 +1,58 @@
...
+    - Create dictionary of hvac systems and associated zones and terminal units: `dict_of_zones_and_terminal_units_served_by_hvac_sys_x = get_dict_of_zones_and_terminal_units_served_by_hvac_sys(B_RMI)`  
+    - For each hvac system in the B_RMI: `for hvac in B_RMI...HeatingVentilatingAirConditioningSystem:` 
+        - Reset all_design_setpoints_105 boolean to true: `all_design_setpoints_105 = true`  
+        - Get the baseline system type: `sys_type = list(baseline_system_types_dict.keys())[list(baseline_system_types_dict.values()).index(hvac.id)])`
+        - Check if system type is of an applicable system type, if yes then carry on, if no then NOT_APPLICABLE for the hvac system:  `If baseline_system_type_compare(sys_type, HVAC_SYS.SYS_9, false) == TRUE OR baseline_system_type_compare(sys_type, HVAC_SYS.SYS_10, false) == TRUE: `    
+            - Get list of zones served by the hvac system: `zone_list_hvac_sys_x = dict_of_zones_and_terminal_units_served_by_hvac_sys_x[hvac.id]["ZONE_LIST"]`
```
supriyagoel, line 36: System 9 and 10 are single zone systems. Do we expect to get a list or a single zone and a single terminal?

docs/section19/19-17.md
```diff
@@ -0,0 +1,58 @@
...
+                - Check if the design supply air temperature setpoint does not equal 105F, if it doesn't then set all_design_setpoints_105 to false: `if design_supply_air_temp_setpoint != 105: all_design_setpoints_105 = false`  
+            Below sums the terminal unit primary air flow associated with each zone that this baseline hvac system serves in the baseline in the proposed in order to compare baseline and proposed supply cfm in the Case 3 below:
+            - For each zone served by the HVAC system: `for zone in zone_list_hvac_sys_x:`  
+                - Get P_RMI zone object (however the RCT accomplishes this): `zone_p = match_data_element(P_RMI, Zone, zone.id)`  
+                    - For each terminal unit associated with the zone: `for terminal_p in zone_p.terminals:`  
+                        - Add the terminal air flow to the proposed_supply_cfm (I think the sum of terminal unit primary air flow shoould sum to the supply cfm of the hvac system serving all the terminals? Agreed? That is what this assumes): `proposed_supply_cfm = proposed_supply_cfm + terminal_p.primary_airflow`              
+
```
supriyagoel, line 46: i think that is correct.  sum of primary air flow of all terminals should be equal to supply CFM of the HVAC system

supriyagoel (approved):
looks great!

mkarpman (approved):

PR #758, opened 2022-11-02
Rs/wx/rule22 24

weilixu:


yunjoonjung-PNNL (approved):
Looks perfect to me. Thank you for this effort!

rct229/rules/section22/section22rule24.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        primary_secondary_loops_dict = get_primary_secondary_loops_dict(rmi_b)
+        return {"primary_secondary_loops_dict": primary_secondary_loops_dict}
+
+    def list_filter(self, context_item, data):
+        pump_b = context_item.baseline
+        primary_secondary_loops_dict = data["primary_secondary_loops_dict"]
```
charliepnnl, line 69: Is is true that `primary_secondary_loops_dict.keys()` are the primary loop ids?  If so, then lines 69-70 could be:
```
primary_loop_ids = data["primary_secondary_loops_dict"].keys()
return pump_b["loop_or_piping"] in primary_loop_ids
```
which might be slightly more self-documenting.

rct229/rules/section22/section22rule24.py
```diff
@@ -0,0 +1,88 @@
...
+            )
+            and primary_secondary_loop_dict
+        )
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        primary_secondary_loops_dict = get_primary_secondary_loops_dict(rmi_b)
+        return {"primary_secondary_loops_dict": primary_secondary_loops_dict}
+
+    def list_filter(self, context_item, data):
```
charliepnnl, line 67: It might be helpful to provide a comment above this method definition to describe what this method is intended to do – especially since this does not strictly follow the RDS logic (which I think is perfectly fine).

charliepnnl (approved):
Approved with some style-type comments and a question regarding SYS_11B

PR #759, opened 2022-11-02
Rs/wx/rule22 25

weilixu:


rct229/rules/section22/section22rule25.py
```diff
@@ -0,0 +1,112 @@
...
+            )
+
+        def get_calc_vals(self, context, data=None):
+            primary_cooling_loop = context.baseline
+            primary_pump_power_per_flow_rate = primary_cooling_loop[
+                "pump_power_per_flow_rate"
+            ]
+            return {
+                "primary_pump_power_per_flow_rate": primary_pump_power_per_flow_rate,
+                "required_pump_power_per_flow_rate": REQUIRED_PUMP_POWER_PER_FLOW_RATE,
```
yunjoonjung-PNNL, line 100: Could you check whether we have consistently compared ```REQUIRED_PUMP_POWER_PER_FLOW_RATE``` in the ```rule_check``` method? This is correct but, I'm wondering what would be the consistent way to develop the code. Thanks!

yunjoonjung-PNNL (approved):
Looks great to me. I'd appreciate it if you could check the one minor comment.

rct229/rules/section22/section22rule25.py
weilixu: The reason I put it here was because I realized we need the `REQUIRED_PUMP_POWER_PER_FLOW_RATE` in the report. 
To note, only data in the `calc_val` will go into the final reporting. If we dont include this data in the calc_val output, then it missed in the ruleset checking output.
I suggest to follow this way instead of use it in the `rule_check` method.

charliepnnl (approved):
Approved with some comments that need to be addressed.

PR #760, opened 2022-11-02
Rs/wx/rule22 26

weilixu:


yunjoonjung-PNNL (approved):
Looks great to me. Thank you!

rct229/rules/section22/section22rule26.py
```diff
@@ -0,0 +1,100 @@
...
+            description="For chilled-water systems served by chiller(s) and serves baseline System-11, the baseline building constant-volume primary pump power shall be modeled as 12 W/gpm.",
+            rmr_context="ruleset_model_instances/0",
+            list_path="fluid_loops[*]",
+        )
+
+    def is_applicable(self, context, data=None):
+        rmi_b = context.baseline
+        baseline_system_types_dict = get_baseline_system_types(rmi_b)
+        # create a list contains all HVAC systems that are modeled in the rmi_b
+        available_type_lists = [
```
charliepnnl, line 41: I suggest calling this variable `available_types`, `available_types_list`, or `available_sys_types`. The plural on `list` is misplaced since there is only one list here.

rct229/rules/section22/section22rule26.py
```diff
@@ -0,0 +1,100 @@
...
+        primary_secondary_loop_dict = get_primary_secondary_loops_dict(rmi_b)
+
+        return (
+            any(
+                [
+                    available_type in APPLICABLE_SYS_TYPES
+                    for available_type in available_type_lists
+                ]
+            )
+            and primary_secondary_loop_dict
```
charliepnnl, line 55: Would the intent be clearer for line 55 with:
```
and len(primary_secondary_loop_dict) > 0
```

rct229/rules/section22/section22rule26.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        primary_secondary_loops_dict = get_primary_secondary_loops_dict(rmi_b)
+        return {"primary_secondary_loops_dict": primary_secondary_loops_dict}
+
+    def list_filter(self, context_item, data):
+        fluid_loop = context_item.baseline
+        primary_secondary_loops_dict = data["primary_secondary_loops_dict"]
```
charliepnnl, line 65: As I've mentioned in other rules, I wonder if it would be slightly more self-documenting if you used a variable
```
primary_loop_ids = data["primary_secondary_loops_dict"].keys()
```

docs/section22/Rule22-26.md
```diff
@@ -24,7 +24,7 @@
 
 - **Check 1:** Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`
 
-  - Check if B-RMR is modeled with at least one air-side system that is Type-11.1, 11.2 or 11b, i.e. with air-side system served by chiller(s), continue to the next applicability check: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-11.1", "SYS-11.2", "SYS-11B"]): NEXT_APPLICABILITY_CHECK`
+  - Check if B-RMR is modeled with at least one air-side system that is Type-11.1, 11.2 or 11.1b, i.e. with air-side system served by chiller(s), continue to the next applicability check: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-11.1", "SYS-11.2", "SYS-11B"]): NEXT_APPLICABILITY_CHECK`
   
   - Else, rule is not applicable to B-RMR: `else: RULE_NOT_APPLICABLE`
```
charliepnnl, line 30: The RDS Rule Logic says to create a dictionary `loop_pump_dictionary` but it doesn't seem to use it.

rct229/rules/section22/section22rule26.py
```diff
@@ -0,0 +1,100 @@
...
+        def get_calc_vals(self, context, data=None):
+            primary_pump_b = context.baseline
+            primary_pump_power_per_flow_rate = primary_pump_b[
+                "pump_power_per_flow_rate"
+            ]
+            return {
+                "primary_pump_power_per_flow_rate": CalcQ(
+                    "power_per_flow_rate", primary_pump_power_per_flow_rate
+                ),
+                "required_pump_power_per_flow_rate": CalcQ(
```
charliepnnl, line 86: Do we really need `required_pump_power_per_flow_rate` in the output?  This is something that the user can get from the standard itself. There are many other cases like this where we compare to a value in the standard and we have not put those values into the output.

charliepnnl (approved):
Approved with some style comments

rct229/rules/section22/section22rule26.py
weilixu: update to available_sys_types

docs/section22/Rule22-26.md
weilixu: Right, I delete this section in the RDS.

rct229/rules/section22/section22rule26.py
weilixu: We should put it in there as this will be part of the detailed report. - something to work on to update the RCT.

PR #761, opened 2022-11-02
RS+RT/YJ/RULE 22-7

yunjoonjung-PNNL:


weilixu (approved):
Thanks, looks good to me

rct229/rules/section22/section22rule7.py
```diff
@@ -0,0 +1,61 @@
...
+            rmrs_used=UserBaselineProposedVals(False, True, False),
+            id="22-7",
+            description="Baseline chilled water system that does not use purchased chilled water shall be modeled as primary/secondary systems.",
+            rmr_context="ruleset_model_instances/0",
+        )
+
+    def is_applicable(self, context, data=None):
+        rmi_b = context.baseline
+        baseline_system_types_dict = get_baseline_system_types(rmi_b)
+        # create a list contains all HVAC systems that are modeled in the rmi_b
```
charliepnnl, line 37: Should be `containing` instead of `contains`.

rct229/rules/section22/section22rule7.py
```diff
@@ -0,0 +1,61 @@
...
+        available_type_lists = [
```
charliepnnl, line 38: As I've commented in other rules, I think the word `list` should not be plural here. Better options would be `available_types`, `available_types_list`, or `available_sys_types`.

rct229/rules/section22/section22rule7.py
```diff
@@ -0,0 +1,61 @@
...
+        return any(
+            [
+                available_type in APPLICABLE_SYS_TYPES
+                for available_type in available_type_lists
+            ]
+        )
+
+    def get_calc_vals(self, context, data=None):
+        rmi_b = context.baseline
+        primary_secondary_loop_dictionary = get_primary_secondary_loops_dict(rmi_b)
```
charliepnnl, line 52: I hate to see the entire `primary_secondary_loop_dictionary` go into calc_vals since it would then end up in the output. How about putting something like `num_primary_loops` in calc_vals instead.  Unless I am misunderstanding, `len(primary_secondary_loop_dictionary)` is the same as `len(primary_secondary_loop_dictionary.keys())` and `primary_secondary_loop_dictionary.keys()` are the primary loops, so `len(primary_secondary_loop_dictionary)`  is the number of primary loops.

charliepnnl (approved):
Approved with at least one comment that should probably be addressed.

rct229/rules/section22/section22rule7.py
yunjoonjung-PNNL: Thanks for catching this up. I also updated this to all other section 22 rules as well.

rct229/rules/section22/section22rule7.py
yunjoonjung-PNNL: I changed the variable name to ```available_types_list```. Also, I updated this to all other section22 rules.

rct229/rules/section22/section22rule7.py
yunjoonjung-PNNL: It sounds great and this part was updated to have ```len(primary_secondary_loop_dictionary)```.

PR #785, opened 2022-11-10
Rds/kjw/function does zone meet g3 1 1f

KarenWGard:


docs/ruleset_functions/G311_exceptions/does_zone_meet_G3_1_1f.md
```diff
@@ -0,0 +1,26 @@
+# does_zone_meet_G3_1_1f
+**Schema Version:** 0.0.22
+
+**Description:** determines whether a given zone meets the G3_1_1f exception "If the baseline HVAC system type is 9 or 10, use additional system types for all HVAC zones that are mechanically cooled in the proposed design." - this function is only called if the expected baseline system type has already been confirmed to be system type 9 or 10
```
mkarpman, line 4: Doesn't this function also need to determine if the zone is served by systems 9 or 10? Perhaps this will be done in the calling RDS, but as it stands now the description misrepresents what the function actually does, because it only checks whether the corresponding zone in the proposed design has mechanical cooling. Otherwise it's approvable.

KarenWGard: yes, the description has been updated

mkarpman (approved):

docs/ruleset_functions/G311_exceptions/does_zone_meet_G3_1_1f.md
```diff
@@ -0,0 +1,26 @@
+# does_zone_meet_G3_1_1f
+**Schema Version:** 0.0.22
+
+**Description:** determines whether a given zone meets the G3_1_1f exception "If the baseline HVAC system type is 9 or 10, use additional system types for all HVAC zones that are mechanically cooled in the proposed design." - this function is only called if the expected baseline system type has already been confirmed to be system type 9 or 10
+
```
mkarpman, line 5: I approved this function but I wonder if it's needed at all, because it just calls another function and doesn't do anything else. I would consider getting rid of it.

supriyagoel: i agree, lets discuss today.

KarenWGard: discussed on call 230615 - agreed to keep it to encapsulate logic (same structure for each G3.1.1 exception)

supriyagoel (approved):

PR #786, opened 2022-11-10
Rds/kjw/function is zone mechanically heated and not cooled

KarenWGard:


docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
```diff
@@ -0,0 +1,34 @@
+# is_zone_mechanically_heated_and_not_cooled
+**Schema Version:** 0.0.23  
+
+**Description:** determines whether a zone is mechanically heated, but not cooled.  DOES NOT CHECK FOR TRANSFER AIR
+
```
supriyagoel, line 5: @KarenWGard I assume this is to check for the heated-only system  requirement (exception e), correct?

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
```diff
@@ -0,0 +1,34 @@
...
+**Function Call:**
+- **get_list_hvac_systems_associated_with_zone**
+
+## Logic:
+- set result to FALSE: `result = FALSE`
+- get the hvac systems serving the zone by using the get_list_hvac_systems_associated_with_zone function: `list_hvac_systems = get_list_hvac_systems_associated_with_zone(RMD,zone)`
+	- loop through each HeatingVentilatingAirconditionsSystem system: `for system in list_hvac_systems:`
+		- check for a heating system: `if system.heating_system != "Null":`
+			- make sure the heating system type is not equal to None: `if system.heating_system.heating_system_type != NONE:`
+				- look for a cooling system: `if system.cooling_system == "Null":`
```
supriyagoel, line 22: @KarenWGard do we also need to check the terminal heating system type? I believe we're defining PTACs and FPFCUs using the terminal.

supriyagoel: look for a cooling system: `if system.cooling_system == "Null":`
this line could be look for NULL or None.

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
KarenWGard: yes

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
KarenWGard: @supriyagoel - yes, we definitely need to look in terminals.  I updated the function to go through terminals.

KarenWGard: @weilixu - can the following logic handle on one line, or is it better to do the first check first and then the second?

`if system.cooling_system == NULL or system.cooling_system == NONE`

weilixu: In code, it can be handled in one line.
In RDS, I suggest to lay out the logics for public reference.

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
```diff
@@ -0,0 +1,43 @@
...
+- set a boolean has_cooling_system to false: `has_cooling_system = FALSE`
+- set result to FALSE: `result = FALSE`
+- get the hvac systems serving the zone by using the get_list_hvac_systems_associated_with_zone function: `list_hvac_systems = get_list_hvac_systems_associated_with_zone(RMI,zone)`
+	- loop through each HeatingVentilatingAirconditionsSystem system: `for system in list_hvac_systems:`
+		- check for a heating system: `if system.heating_system != NULL or system.heating_system.type != NONE:`
+			- set the boolean has_heating_system to true: `has_heating_system = TRUE`
+		- look for a cooling system: `if system.cooling_system != NULL or system.cooling_system != NONE:`
+			- there is a cooling system: `has_cooling_system = TRUE`
+- at this point, if there is no cooling system, continue looking in the zone terminals for a cooling system (if there is a cooling system at this level, the function should return false regardless of what is in the zone terminals, and we can skip to the end): `if !has_cooling_system:`
+	- look at each terminal in the zone: `for terminal in zone.terminals:`
```
supriyagoel, line 27: since we're looping through each HVAC system, there can be only one terminal for each zone, but there can be multiple zones for each HVAC system. so maybe line 27 should instead be 'for each zone served by the HVAC system' instead of 'for each terminal in the zone'?

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
```diff
@@ -0,0 +1,43 @@
...
+ 
+**Function Call:**
+- **get_list_hvac_systems_associated_with_zone**
+
+## Logic:
+- set a boolean has_heating_system to false: `has_heating_system = FALSE`
+- set a boolean has_cooling_system to false: `has_cooling_system = FALSE`
+- set result to FALSE: `result = FALSE`
+- get the hvac systems serving the zone by using the get_list_hvac_systems_associated_with_zone function: `list_hvac_systems = get_list_hvac_systems_associated_with_zone(RMI,zone)`
+	- loop through each HeatingVentilatingAirconditionsSystem system: `for system in list_hvac_systems:`
```
supriyagoel, line 21: it would help to clarify that line 21 is looping through each zone and for each zone, it is looking through each HVAC system associated with the zone. I incorrectly read this to mean that we're looking through each HVAC system in the RMI.

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
```diff
@@ -0,0 +1,43 @@
...
+	- look at each terminal in the zone: `for terminal in zone.terminals:`
+		- check if there is a terminal.heating_source specified: `if terminal.heating_source != NULL or terminal.heating_source != NONE:`
+			- there is a terminal heat source, set has_heating_system equal to true: `has_heating_system = TRUE`
+		- check if there is a terminal.cooling_source specified: `if terminal.cooling_source != NULL or terminal.cooling_source != NONE:`
+			- there is a terminal cooling source, set has_cooling_system equal to true: `has_cooling_system = TRUE`
+- now, if has_heating_system is true and has_cooling_system is false, set result equal to true: `if has_heating_system and !has_cooling_system: result = TRUE`
+- if result is FALSE, check if the zone receives transfer air from a cooled zone: `if !result:`
+	- if the zone has transfer air: `if zone.transfer_airflow_rate > 0:`
+		- set result equal to the result of is_zone_mechanically_heated_and_not_cooled for the transfer air source zone: `result = is_zone_mechanically_heated_and_not_cooled(RMI, zone.transfer_airflow_source_zone)
```
supriyagoel, line 35: I am not sure I understand what line 35 is doing.

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
```diff
@@ -0,0 +1,43 @@
...
+			- there is a cooling system: `has_cooling_system = TRUE`
+- at this point, if there is no cooling system, continue looking in the zone terminals for a cooling system (if there is a cooling system at this level, the function should return false regardless of what is in the zone terminals, and we can skip to the end): `if !has_cooling_system:`
+	- look at each terminal in the zone: `for terminal in zone.terminals:`
+		- check if there is a terminal.heating_source specified: `if terminal.heating_source != NULL or terminal.heating_source != NONE:`
+			- there is a terminal heat source, set has_heating_system equal to true: `has_heating_system = TRUE`
+		- check if there is a terminal.cooling_source specified: `if terminal.cooling_source != NULL or terminal.cooling_source != NONE:`
+			- there is a terminal cooling source, set has_cooling_system equal to true: `has_cooling_system = TRUE`
+- now, if has_heating_system is true and has_cooling_system is false, set result equal to true: `if has_heating_system and !has_cooling_system: result = TRUE`
+- if result is FALSE, check if the zone receives transfer air from a cooled zone: `if !result:`
```
supriyagoel, line 33: I am confused out here.  If the zone is mechanically heated and not cooled, the  we should be checking if it is exhausting air or receiving transfer air. However, from line 33 it looks like we do this check ONLY f the result = False. Should we instead be doing this if the result = true?

KarenWGard: yes, @supriyagoel - you're right.  Updated

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
mkarpman: I don't think this will work because for the zone.transfer_airflow_source_zone we only need to check if that zone itself is mechanically cooled. (It's irrelevant whether it is heated because we only get to this point if the master zone has heating.). 

Also, I don't think we need to account for "cascading" transfer air (when air is transferred to the master zone from a cooled zone via one or more zones that don't have mechanical cooling). The exception is "Exception: Spaces designed with heating only systems ..... not exhausting or transferring air **from mechanically cooled thermal zones** in the proposed design shall not be modeled with mechanical cooling."

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
KarenWGard: i have changed the logic to have only two booleans:

`is_heated `and `is_cooled`

at the end of the function, we return the result of `is_heated and !is_cooled`

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
KarenWGard: yes, you are right.  I have updated the function to look expressly at the hvac systems and terminals in the source zone

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
```diff
@@ -0,0 +1,58 @@
...
+		- check if there is a terminal.cooling_source specified: `if terminal.cooling_source != NULL:`
+			-  `if terminal.cooling_source != NONE:`
+				- there is a terminal cooling source, set is_cooled equal to true: `is_cooled = TRUE`
+- if the zone is heated, but not cooled, check if the zone receives transfer air from a cooled zone: `if is_heated && !is_cooled:`
+	- if the zone has transfer air: `if zone.transfer_airflow_rate > 0:`
+		- set transfer_source_zone equal to the zone.transfer_airflow_source_zone: `transfer_source_zone = zone.transfer_airflow_source_zone`
+		- get the hvac systems serving the transfer airflow source zone: `transfer_source_zone_list_hvac_systems = get_list_hvac_systems_associated_with_zone(RMI,transfer_source_zone)`
+		- loop through heach HVAC system: `for transfer_system in transfer_source_zone_list_hvac_systems:`
+			- look for a cooling system: `if transfer_system.cooling_system != NULL:`
+				- `if transfer_system.cooling_system != NONE:`
```
mkarpman, line 41: This line seems redundant because it repeats the one right above?

mkarpman (approved):

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
KarenWGard: line 41 should have read `if transfer_system.cooling_system.type != NONE:`

this has been changed

(transfer_system.cooling_system will return NULL if there is no system.  transfer_system.cooling_system.type will return NONE if there is an entry for the cooling system in the RMI, but there are no coils capable of doing cooling (see CoolingSourceOptions)

docs/ruleset_functions/is_zone_mechanically_heated_and_not_cooled.md
```diff
@@ -0,0 +1,58 @@
...
+					- there is a cooling system: `is_cooled = TRUE`
+		- if there is no cooling system, look through the zone terminals of the transfer_source_zone for a cooling system: `if !is_cooled:`
+			- look at each terminal in the transfer_source_zone: `for transfer_terminal in transfer_source_zone.terminals:`
+				- check if there is a transfer_terminal.cooling_source specified: `if transfer_terminal.cooling_source != NULL:`
+					-  `if transfer_terminal.cooling_source != NONE:`
```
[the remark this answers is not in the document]
KarenWGard: also changed

supriyagoel (approved):

PR #796, opened 2022-11-18
Rds/cml/19 18

claperle:


mkarpman (approved):

supriyagoel:
I dont quite agree with the conservative check in this RDS. Should discuss

claperle:
@supriyagoel I suppose the required change is pending the decision by the committees about handling typically more stringent baseline parameters or perhaps we want to exclude this one altogether for having an outcome for the more stringent baseline case? I agree that there could be buildings where less fan power in the baseline could be less stringent due to interactions with HVAC loads but most of the time less fan power in the baseline is more stringent (I think , I am cold climate biased though).

PR #797, opened 2022-11-18
Rds/cml/19 19

claperle:


docs/section19/19-19.md
```diff
@@ -0,0 +1,94 @@
+# Section 19 - Rule 19-19       
+**Schema Version:** 0.0.23    
+**Mandatory Rule:** True    
+**Rule ID:** 19-19     
+**Rule Description:** For baseline systems 3 through 8, and 11, 12, and 13, the system fan electrical power for supply, return, exhaust, and relief shall be Pfan = bhp × 746/fan motor efficiency. Where, bhp = brake horsepower of baseline fan motor from Table G3.1.2.9; fan motor efficiency = the efficiency from Table G3.9.1 for the next motor size greater than the bhp using a totally enclosed fan cooled motor at 1800 rpm.      
+**Rule Assertion:** Options are Pass/UNDETERMINED/NOT_APPLICABLE
```
mkarpman, line 6: Seems like we should also have a FAIL outcome for when the modeled fan power is less than expected without accounting for any pressure drop adjustments.

docs/section19/19-19.md
```diff
@@ -0,0 +1,94 @@
...
+                - Add to the total power: `total_fan_power = total_fan_power + fan_elec_power`       
+            - Cycle through the list of exhaust fans: `for fan_b in fan_sys_exhaust_list:`  
+                - Get the fan power: `fan_elec_power = get_fan_object_electric_power(B_RMI,fan_b)`   
+                - Get the cfm: `total_exhaust_cfm =  total_exhaust_cfm + fan_b.design_airflow`  
+                - Add to the total power: `total_fan_power = total_fan_power + fan_elec_power`     
+            - Cycle through each zone associated with the HVAC system to get the exhaust fan power: `for zone_b in zone_list:`  
+                - Create a zone object from the zone id (not sure how the RCT will do this): `zone_b_obj = zone_b`  
+                - Check if there are any exhaust fan objects associated with the zone: `if zone_b_obj.zonal_exhaust_fan != Null:`  
+                    - Get the fan power: `fan_elec_power = get_fan_object_electric_power(B_RMI,zone_b_obj.zonal_exhaust_fan)`    
+                    - Add to the total power: `total_fan_power = total_fan_power + fan_elec_power`
```
mkarpman, line 63: You ahve this routine in several RDSs... perhaps it should be a function? No strong feelings since you already implemented it.

docs/section19/19-19.md
```diff
@@ -0,0 +1,94 @@
...
+                - Calculate the expected BHP (if no other pressure drop adjustments apply): `expected_BHP = (0.0013 * supply_cfm_b) + A`  
+            - Else, the system is system 11: `Else:`  
+                - Calculate the expected BHP (if no other pressure drop adjustments apply): `expected_BHP = (0.00062 * supply_cfm_b) + A`  
+
+            - Get the motor efficiency from Table G3.9.1 (should be looked up based on the next motor size greater than the bhp (i.e., expected_BHP)): `motor_eff = data_lookup(G3.9.1,expected_BHP)`  
+            - Calculate the expected Wattage (assuming no other pressure drop adjustments apply except energy recovery): `fan_Wattage_expected = expected_BHP * 746 *(1/motor_eff)`  
+
+            - **Rule Assertion:** 
+            - Case 1: If fan wattage expected equals total fan power: `if total_fan_power == fan_Wattage_expected: outcome = "PASS"`  
+            - Case 2: Elif conservative comparison less equal:`elif AHJ_RA_compare == True and total_fan_power <= fan_Wattage_expected: PASS`
```
mkarpman, line 84: We should have FAIL outcome here if "concervative comparision" = false

docs/section19/19-19.md
```diff
@@ -0,0 +1,94 @@
...
+            - Case 3: Else, undetermined: `Else: outcome = "UNDETERMINED" and raise_message "Fan power for <hvac.id> is greater than expected per Section and Table G3.1.2.9 assuming no pressure drop adjustments per Table 6.5.3.1-2 other than energy recovery (if modeled). Conduct manual check."`
```
mkarpman, line 85: Please add other possible pressure drop adjustments to the note (sound attenuation, air filtration, fully ducted return when required by code).

docs/section19/19-19.md
```diff
@@ -0,0 +1,94 @@
...
+            - Get the motor efficiency from Table G3.9.1 (should be looked up based on the next motor size greater than the bhp (i.e., expected_BHP)): `motor_eff = data_lookup(G3.9.1,expected_BHP)`  
+            - Calculate the expected Wattage (assuming no other pressure drop adjustments apply except energy recovery): `fan_Wattage_expected = expected_BHP * 746 *(1/motor_eff)`  
+
+            - **Rule Assertion:** 
+            - Case 1: If fan wattage expected equals total fan power: `if total_fan_power == fan_Wattage_expected: outcome = "PASS"`  
+            - Case 2: Elif conservative comparison less equal:`elif AHJ_RA_compare == True and total_fan_power <= fan_Wattage_expected: PASS`
+            - Case 3: Else, undetermined: `Else: outcome = "UNDETERMINED" and raise_message "Fan power for <hvac.id> is greater than expected per Section and Table G3.1.2.9 assuming no pressure drop adjustments per Table 6.5.3.1-2 other than energy recovery (if modeled). Conduct manual check."`  
+
+**Notes/Questions:**  
+1. I don't need case 1 but I put it here just so the RCT could see the "perfect" result.
```
mkarpman, line 88: I think this is needed and is good.

docs/section19/19-19.md
```diff
@@ -0,0 +1,94 @@
...
+            - Calculate the expected Wattage (assuming no other pressure drop adjustments apply except energy recovery): `fan_Wattage_expected = expected_BHP * 746 *(1/motor_eff)`  
+
+            - **Rule Assertion:** 
+            - Case 1: If fan wattage expected equals total fan power: `if total_fan_power == fan_Wattage_expected: outcome = "PASS"`  
+            - Case 2: Elif conservative comparison less equal:`elif AHJ_RA_compare == True and total_fan_power <= fan_Wattage_expected: PASS`
+            - Case 3: Else, undetermined: `Else: outcome = "UNDETERMINED" and raise_message "Fan power for <hvac.id> is greater than expected per Section and Table G3.1.2.9 assuming no pressure drop adjustments per Table 6.5.3.1-2 other than energy recovery (if modeled). Conduct manual check."`  
+
+**Notes/Questions:**  
+1. I don't need case 1 but I put it here just so the RCT could see the "perfect" result.  
+2. There is not an explicit schema data element for ERV exhaust and OA air flow so I made some assumptions. See above.
```
mkarpman, line 89: Let's discussion on the RDS call to make sure that others agree with the assumptions. They look good to me.

docs/section19/19-19.md
claperle: Done

docs/section19/19-19.md
claperle: But how can we call it fail if we do not know whether other pressure drop adjustments apply? (which is likely the case)

docs/section19/19-19.md
claperle: I would like to discuss this. Since we do not know the other pressure drop adjustments I don't think we should fail any projects (unless we can come up with some sort of high limit ceiling, maybe we assume a MERV 15 filter and sound attenuation and energy recovery if indicated in the schema and fail if greater than the fan power assuming these?)

docs/section19/19-19.md
claperle: Done

docs/section19/19-19.md
claperle: Christina to add MERV filter adjustment, assume supply cfm is cfmd for MERV filter specified in the schema.

docs/section19/19-19.md
claperle: No fails, just undetermined and pass outcome per phone call on 1/12/2023

docs/section19/19-19.md
claperle: Done

docs/section19/19-19.md
```diff
@@ -0,0 +1,96 @@
...
+            - Get list with the hvac system's fan system's supply, return, exhaust and relief fan CFM: `fan_sys_CFM = list(get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM(B_RMI,fan_sys_b)["Fan_Power"]["Fan_CFM"])` 
+                    
+            - Get the hvac system total supply fan kW, need to get the first element in the list of fan powers:`hvac_sys_total_supply_fan_power_b = fan_sys_kW[0]` 
+            - Get the hvac system total return fan kW, need to get the second element in the list of fan powers:`hvac_sys_total_return_fan_power_b = fan_sys_kW[1]` 
+            - Get the hvac system total exhaust fan kW, need to get the third element in the list of fan powers:`hvac_sys_total_exhaust_fan_power_b = fan_sys_kW[2]` 
+            - Get the hvac system total relief fan kW, need to get the fourth element in the list of fan powers:`hvac_sys_total_relief_fan_power_b = fan_sys_kW[3]`     
+
+            - Calculate the total modeled fan power for the hvac system in the baseline: `total_fan_power = hvac_sys_total_supply_fan_power_b + hvac_sys_total_return_fan_power_b + hvac_sys_total_exhaust_fan_power_b + hvac_sys_total_relief_fan_power_b`
```
supriyagoel, line 48: Question: How is like 41 different from line 48? Isnt the function getting thee total kw? and then line 48 is summing up the individual kw to come to the same answer?

supriyagoel (approved):
quite a complicated RDS.

docs/section19/19-19.md
```diff
@@ -0,0 +1,96 @@
...
+    - For each hvac system in the B_RMI: `for hvac in B_RMI...HeatingVentilatingAirConditioningSystem:` 
+        - Get the baseline system type: `sys_type = list(baseline_system_types_dict.keys())[list(baseline_system_types_dict.values()).index(hvac.id)])`
+        - Check if system type is of an applicable system type, if yes then carry on, if no then NOT_APPLICABLE for the hvac system:  `if any(baseline_system_type_compare(sys_type, target_sys_type, false) for target_system_type in target_system_types):`         
+            - Get the list of zones associated with the hvac system: `zone_list = dict_of_zones_and_terminal_units_served_by_hvac_sys[hvac.id]["ZONE_LIST"]`  
+            - Reset supply cfm to zero: `supply_cfm_b = 0`  
+            - Reset total_fan_power to zero: `total_fan_power = 0`   
+            - Reset fan_power_W_CFM to zero: `fan_power_W_CFM = 0`    
+            - Reset total exhaust cfm: `total_exhaust_cfm = 0`       
+            - Reset total relief cfm: `total_relief_cfm = 0`
```
mkarpman, line 38: Doesn't seem we need exhaust and releive CFM?

docs/section19/19-19.md
mkarpman: This does not include terminal fans. Sounds like these should be added even though we are not expecting them for these system types.

docs/section19/19-19.md
```diff
@@ -0,0 +1,96 @@
...
+                - Check if there are any exhaust fan objects associated with the zone: `if zone_b_obj.zonal_exhaust_fan != Null:`  
+                    - Get the fan power: `fan_elec_power = get_fan_object_electric_power(B_RMI,zone_b_obj.zonal_exhaust_fan)`    
+                    - Add to the total power: `total_fan_power = total_fan_power + fan_elec_power`                    
+            
+            Below determines the expected fan power:
+            - Check if there are any energy recovery objects associated with the fan system object: `if fan_sys_b.air_energy_recovery != Null:` 
+                - Get the enthalpy recovery ratio: `enthalpy_rec_ratio = fan_sys_b.air_energy_recovery.enthalpy_recovery_ratio`  
+                - Assume that the outdoor air flow through the ERV is equal to the minimum outdoor airflow: `ERV_OA_air_flow = fan_sys_b.minimum_outdoor_airflow`  
+                - Assume that the exhaust air flow through the ERV is equal to the maximum of either the total exhaust cfm or total relief cfm associated with the fan system: `ERV_Ex_air_flow = max(total_exhaust_cfm,total_relief_cfm)`
```
mkarpman, line 63: I don't think we should include relief CFM, just the exhaust CFM.

docs/section19/19-19.md
```diff
@@ -0,0 +1,96 @@
...
+                - Get the MERV filter rating: `MERV_rating = fan_sys_b.air_filter_merv_rating`  
+            - Else: `Else: MERV_rating = 0`  
+            - Determine the pressure drop adjustment in in. of water if MERV is 9 through 12: `if MERV_rating >= 9 and MERV_rating <=12: MERV_adj = 0.5`  
+            - Else if, MERV is 13 and above (assume 0.9 for all MERVs greater than 12 since we do not know the pressure drop for MERV 16 and greater): `elif MERV_rating >12: MERV_adj = 0.9`  
+            - Else: `Else: MERV_rating = 0`  
+            - Calculate A per Section 6.5.3.1.1 for MERV assuming cfmd equals the supply cfm: `A = A + (MERV_adj * supply_cfm_b/4131)`  
+            
+            - Create a list of the CV system types: `CV_system_types = [HVAC_SYS.SYS_3, HVAC_SYS.SYS_4,HVAC_SYS.SYS_12, HVAC_SYS.SYS_13]`  
+            - Check if system type is constant volume:  `if any(baseline_system_type_compare(sys_type, target_sys_type, false) for target_system_type in CV_system_types):`     
+                - Calculate the expected BHP (if no other pressure drop adjustments apply): `expected_BHP = (0.00094 * supply_cfm_b) + A`
```
mkarpman, line 76: Recommend defining minimum_BHP and minimum_fan_power for each system type which would be based on A=0. Then we can do FAIL outcome if the modeled baseline fan power is lower than this minimum.

docs/section19/19-19.md
mkarpman: If AHJ_RA_Compare==False and total_fan_power<minim_fan_power than the outcome should be FALSE.

docs/section19/19-19.md
claperle: @supriyagoel good Q, 41 is not a total, it creates a list with kW as values for supply, return, exhaust, and relief fans. We need it is total and separate so I think we need this line to create the list rather than just summing them right off the bat.

docs/section19/19-19.md
```diff
@@ -0,0 +1,96 @@
...
+            - Get the list of zones associated with the hvac system: `zone_list = dict_of_zones_and_terminal_units_served_by_hvac_sys[hvac.id]["ZONE_LIST"]`  
+            - Reset supply cfm to zero: `supply_cfm_b = 0`  
+            - Reset total_fan_power to zero: `total_fan_power = 0`   
+            - Reset fan_power_W_CFM to zero: `fan_power_W_CFM = 0`    
+            - Reset total exhaust cfm: `total_exhaust_cfm = 0`       
+            - Reset total relief cfm: `total_relief_cfm = 0`       
+            - Get the fan system object:  `fan_sys_b = hvac.fan_system`  
+            - Get list with the hvac system's fan system's supply, return, exhaust and relief fan powers in kW: `fan_sys_kW = list(get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM(B_RMI,fan_sys_b)["Fan_Power"])`  
+            - Get list with the hvac system's fan system's supply, return, exhaust and relief fan CFM: `fan_sys_CFM = list(get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM(B_RMI,fan_sys_b)["Fan_Power"]["Fan_CFM"])`
```
mkarpman, line 41: I would check to see if there are multiple instandes of any fans (e.g., 2 supply fans) and, if so, bypass all the calculations and throw UNDERTERMINED outcome for that system.

docs/section19/19-19.md
claperle: @mkarpman Oh, maybe I meant no fails if fan power is greater than what we calculate. I can revise,

docs/section19/19-19.md
claperle: @mkarpman terminal fans are covered by G3.1.3.14. Maybe we should add a rule for this and check that they are modeled correctly for 6 and 8 and that no fan power is modeled for the other baseline systems?

docs/section19/19-19.md
claperle: Sounds good. Need to implement.

docs/section19/19-19.md
claperle: Sounds good. Need to implement.

docs/section19/19-19.md
claperle: CML to remove

docs/section19/19-19.md
claperle: Outcome, ignore terminal fans 2/16/2023 because checked by function that does HVAC system mapping.

docs/section19/19-19.md
claperle: Resolved on call. 2/16/2023

docs/section19/19-19.md
claperle: Implemented.

docs/section19/19-19.md
claperle: Implemented.

docs/section19/19-19.md
claperle: Implemented.

docs/section19/19-19.md
claperle: Implemented.

mkarpman: Relief CFM still there in line 37. Not critical, but I think we agreed to remove?

docs/section19/19-19.md
```diff
@@ -0,0 +1,111 @@
...
+            - Case 3: Elif conservative comparison is false and fan is less than expected outcome is Fail:`elif more_than_one_supply_fan == false and AHJ_RA_compare == False and total_fan_power < fan_Wattage_expected: "FAIL" and raise_message "The total fan power for <insert hvac.id> is modeled as <insert total_fan_power> kW which is less than the expected including pressure drop adjustments for exhaust air energy recovery and MERV filters as applicable which was calculated as <insert fan_Wattage_minimim> kW ."`         
+            - Case 4: Elif conservative comparison equals true and fan wattage is less than the minimum possible then Pass: `elif more_than_one_supply_fan == false and AHJ_RA_compare == True and total_fan_power < fan_Wattage_minimum: outcome = "Pass" and raise_message "The total fan power for <insert hvac.id> is modeled as <insert total_fan_power> kW which is less than the minimum possible (i.e., less than the expected fan power assuming no pressure drop adjustments) which was calculated as <insert fan_Wattage_minimim> kW. Pass because this is generally considered more conservative."`
```
mkarpman, line 100: Wouldn't that be covered be Case 2?

docs/section19/19-19.md
```diff
@@ -0,0 +1,111 @@
...
+            - Case 4: Elif conservative comparison equals true and fan wattage is less than the minimum possible then Pass: `elif more_than_one_supply_fan == false and AHJ_RA_compare == True and total_fan_power < fan_Wattage_minimum: outcome = "Pass" and raise_message "The total fan power for <insert hvac.id> is modeled as <insert total_fan_power> kW which is less than the minimum possible (i.e., less than the expected fan power assuming no pressure drop adjustments) which was calculated as <insert fan_Wattage_minimim> kW. Pass because this is generally considered more conservative."`              
+            - Case 4: Elif conservative comparison equals false and fan wattage is less than the minimum possible: `elif more_than_one_supply_fan == false and AHJ_RA_compare == False and total_fan_power < fan_Wattage_minimum: outcome = "FAIL" and raise_message "The total fan power for <insert hvac.id> is modeled as <insert total_fan_power> kW which is less than the minimum possible (i.e., less than the expected fan power assuming no pressure drop adjustments) which was calculated as <insert fan_Wattage_minimim> kW."`
```
mkarpman, line 101: Should be case 5, but then can't it be merged with Case 3?

docs/section19/19-19.md
```diff
@@ -0,0 +1,111 @@
...
+            - Case 2: Elif conservative comparison equals true and fan is less than expected outcome is Pass:`elif more_than_one_supply_fan == false and AHJ_RA_compare == True and total_fan_power < fan_Wattage_expected: PASS and raise_message "The total fan power for <insert hvac.id> is modeled as <insert total_fan_power> kW which is less than the expected including pressure drop adjustments for exhaust air energy recovery and MERV filters as applicable which was calculated as <insert fan_Wattage_minimim> kW. Pass because this is generally considered more conservative."`  `  
+            - Case 3: Elif conservative comparison is false and fan is less than expected outcome is Fail:`elif more_than_one_supply_fan == false and AHJ_RA_compare == False and total_fan_power < fan_Wattage_expected: "FAIL" and raise_message "The total fan power for <insert hvac.id> is modeled as <insert total_fan_power> kW which is less than the expected including pressure drop adjustments for exhaust air energy recovery and MERV filters as applicable which was calculated as <insert fan_Wattage_minimim> kW ."`
```
mkarpman, line 99: Can't it just be that coservative comparison is false and fan is not equal to expected outcome then FAIL?

docs/section19/19-19.md
claperle: @mkarpman Yes but I was not sure why it was suggested that this be calculated (the minimum expected Wattage) during the last round of comments.  Looks like I missed something. Can you please clarify?

docs/section19/19-19.md
claperle: @mkarpman we do not know if there are other relevant pressure drop adjustments so thought it might not be wise to have such a stringent fail outcome.

mkarpman: OK, I see. So if it's more than the outcome would be UNDERTERMINED.

docs/section19/19-19.md
mkarpman: I suggested adding the minimum fan power so that we can produce FAIL if conservative outcome is not allowed and modeled fan power is less than minimum. So I don't think we need case 4 because it's redundant to Case 2.
So I think the logic should be as follows:
1. If no more than 1 supply fan AND no more than one exhaust fan if energy recovery is relevant AND fan power as expected then PASS
2. If no more than 1 supply fan AND conservative comparison allowed AND fan power less than expected then PASS
3. If no more than 1 supply fan AND conservative comparison not allowed AND fan power less than expected then FAIL
4. If no more than 1 supply fan AND conservative comparison allowed AND fan power less than expected then UNDERTERMINED
5. If no more than 1 supply fan AND conservative comparison allowed AND fan power less than minimum possible then PASS.
6. If no more than 1 supply fan AND conservative comparison not allowed AND fan power less than minimum possible then FAIL.
7. If more than 1 supply fan OR more than one exhaust fan if energy recovery is relevant then UNDERTERMINED

claperle: Use this as primary reference for making changes.

claperle: @mkarpman I am not understanding #4. Its the same as 2 but with a different outcome. Can you please explain?  

Also, similar to what you say directly above the list aren't 2 and 5 redundant? If it passes 2 it will pass 5.

docs/section19/19-19.md
```diff
@@ -0,0 +1,111 @@
...
+                - Calculate the total modeled fan power for the hvac system in the baseline: `total_fan_power = hvac_sys_total_supply_fan_power_b + hvac_sys_total_return_fan_power_b + hvac_sys_total_exhaust_fan_power_b + hvac_sys_total_relief_fan_power_b`  
+
+                - Cycle through each zone associated with the HVAC system to get the exhaust fan power: `for zone_b in zone_list:`  
+                    - Create a zone object from the zone id (not sure how the RCT will do this): `zone_b_obj = zone_b`  
+                    - Check if there are any exhaust fan objects associated with the zone: `if zone_b_obj.zonal_exhaust_fan != Null:`  
+                        - Get the fan power: `fan_elec_power = get_fan_object_electric_power(B_RMI,zone_b_obj.zonal_exhaust_fan)`    
+                        - Add to the total power: `total_fan_power = total_fan_power + fan_elec_power`                    
+                
+                Below determines the expected fan power:
+                - Check if there are any energy recovery objects associated with the fan system object: `if fan_sys_b.air_energy_recovery != Null:`
```
claperle, line 63: This will likely need revisions after we modify the energy recovery data elements in the schema.

mkarpman (approved):

weilixu:
This is renumbered to 19-18 and it is likely would be updated again.

PR #799, opened 2022-11-19
Rds/cml/19 20

claperle:


docs/section19/19-20.md
```diff
@@ -0,0 +1,73 @@
...
+                - Get list of hvac systems serving this zone in the proposed design model: `hvac_sys_list_p = Get_list_hvac_systems_associated_with_zone(P_RMI, zone_p)`  
+                - Check if any of the hvac systems serving the zone have non-mechanical cooling as the cooling_system type: `for hvac_p in hvac_sys_list_P:`  
+                    - Check if the system has non-mechanical cooling: `if hvac_p.cooling_system.cooling_system_options == "NON_MECHANICAL": zone_hvac_in_PR_has_non_mech_cooling = True`  
+
+            - Calculate W/CFM: `fan_power_W_CFM = total_fan_power/supply_cfm_b`                 
+            
+            - **Rule Assertion:** 
+            - Case 1: If fan_power_W_CFM equals 0.3: `if zone_hvac_in_PR_has_non_mech_cooling == False and zones_served_by_BL_hvac_has_non_mech_cooling == False and fan_power_W_CFM == 0.3: outcome = "PASS"`  
+            - Case 2: Elif conservative comparison less equal:`elif AHJ_RA_compare == True and fan_power_W_CFM <= 0.3: PASS`
```
mkarpman, line 66: Seems like Case 2 and Case 3 should be swapped, because if project has non-mechanical cooling and fan power is less than 0.3 then it is even more conservative than in the case of no non-mechanical cooling.

docs/section19/19-20.md
```diff
@@ -0,0 +1,73 @@
...
+
+            - Calculate W/CFM: `fan_power_W_CFM = total_fan_power/supply_cfm_b`                 
+            
+            - **Rule Assertion:** 
+            - Case 1: If fan_power_W_CFM equals 0.3: `if zone_hvac_in_PR_has_non_mech_cooling == False and zones_served_by_BL_hvac_has_non_mech_cooling == False and fan_power_W_CFM == 0.3: outcome = "PASS"`  
+            - Case 2: Elif conservative comparison less equal:`elif AHJ_RA_compare == True and fan_power_W_CFM <= 0.3: PASS`  
+            - Case 3: Elif there is non-mechanical cooling associated with the zones served by the baseline HVAC system (should just be one) in the proposed model then the outcome is UNDETERMINED: `if zone_hvac_in_PR_has_non_mech_cooling ==  True or zones_served_by_BL_hvac_has_non_mech_cooling ==  True: outcome = UNDETERMINED and raise_message "<Insert hvac.id> has zone(s) with non-mechanical cooling in the proposed design, conduct a manual check that the baseline building design includes a fan power allowance of Pfan = CFMnmc × 0.054, where, CFMnmc = the baseline non-mechanical cooling fan airflow, cfm for the non-mechanical cooling fan in additional to the 0.3 W/CFM allowance for the HVAC system."`
```
mkarpman, line 67: We have zone.non_mecahnical_cooling_fan_power and zone.non-mecanical_cooling_fan_airflow. Why can't we check that it's 0.054 W/CFM? Or perhaps check that the total baseline fan power is equal to correct total W for all fans including non-mechanical cooling?

docs/section19/19-20.md
claperle: @mkarpman I'm not following, if the fan power is less than 0.3 then it will pass with case 2 regardless of mech cooling so this seems like the right order, no?

docs/section19/19-20.md
claperle: I agree that 19-20 and 21 should be combined. Revised.

docs/section19/19-20.md
```diff
@@ -0,0 +1,73 @@
...
+            - Get list with the hvac system's fan system's supply, return, exhaust and relief fan CFM: `fan_sys_CFM = list(get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM(B_RMI,fan_sys_b)["Fan_Power"]["Fan_CFM"])` 
+                    
+            - Get the hvac system total supply fan kW, need to get the first element in the list of fan powers:`hvac_sys_total_supply_fan_power_b = fan_sys_kW[0]` 
+            - Get the hvac system total return fan kW, need to get the second element in the list of fan powers:`hvac_sys_total_return_fan_power_b = fan_sys_kW[1]` 
+            - Get the hvac system total exhaust fan kW, need to get the third element in the list of fan powers:`hvac_sys_total_exhaust_fan_power_b = fan_sys_kW[2]` 
+            - Get the hvac system total relief fan kW, need to get the fourth element in the list of fan powers:`hvac_sys_total_relief_fan_power_b = fan_sys_kW[3]`     
+
+            - Calculate the total modeled fan power for the hvac system in the baseline: `total_fan_power = hvac_sys_total_supply_fan_power_b + hvac_sys_total_return_fan_power_b + hvac_sys_total_exhaust_fan_power_b + hvac_sys_total_relief_fan_power_b`  
+
```
supriyagoel, line 49: same as #797 confused on why we first break down the fan_sys_kw and then sum it up again. wont it give the same answer?

docs/section19/19-20.md
supriyagoel: on a separate note, i'm not sure if the conservative outcome is correct. System 9 has a fossil fuel furnace for heating. With lower fan power, there will be higher heating energy consumption through the furnace which will have a lower efficiency than an electric fan where 100% of the fan heat is added to the air stream..

docs/section19/19-20.md
```diff
@@ -0,0 +1,73 @@
...
+        - Check if system type is of an applicable system type:  `If (baseline_system_type_compare(sys_type, HVAC_SYS.SYS_9, false) == TRUE OR baseline_system_type_compare(sys_type, HVAC_SYS.SYS_10, false) == TRUE):`     
+            - Get the list of zones associated with the hvac system: `zone_list = dict_of_zones_and_terminal_units_served_by_hvac_sys[hvac.id]["ZONE_LIST"]`  
+            - Reset supply cfm to zero: `supply_cfm_b = 0`  
+            - Reset total_fan_power to zero: `total_fan_power = 0`   
+            - Reset fan_power_W_CFM to zero: `fan_power_W_CFM = 0`                 
+            - Get the fan system object:  `fan_sys_b = hvac.fan_system`  
+            - Get list with the hvac system's fan system's supply, return, exhaust and relief fan powers in kW: `fan_sys_kW = list(get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM(B_RMI,fan_sys_b)["Fan_Power"])`  
+            - Get list with the hvac system's fan system's supply, return, exhaust and relief fan CFM: `fan_sys_CFM = list(get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM(B_RMI,fan_sys_b)["Fan_Power"]["Fan_CFM"])`
```
mkarpman, line 41: I think we agreed that if there is more than one fan of any type we'd do UNDERTERMINED outcome for this system.

docs/section19/19-20.md
claperle: Discussed and resolved on call on 2/16/2023.

docs/section19/19-20.md
claperle: Remove this per discussion on 2/16/2023.

docs/section19/19-20.md
claperle: OK, I am going to go with this then.

claperle: Updated

supriyagoel (approved):

docs/section19/19-20.md
```diff
@@ -0,0 +1,78 @@
...
+            - Case 4: Elif more_than_one_supply_fan equals true: `elif more_than_one_supply_fan == true: outcome = "UNDETERMINED" and raise_message "<insert hvac.id> has more than one supply fan associated with the HVAC system in the baseline and therefore this check could not be conducted for this HVAC sytem. Conduct manual check for compliance with G3.1.2.9."`  
+            - Case 5: Elif there is non-mechanical cooling associated with the zones served by the baseline HVAC system (should just be one) in the proposed model then the outcome is UNDETERMINED: `elif zone_hvac_in_PR_has_non_mech_cooling ==  True or zones_served_by_BL_hvac_has_non_mech_cooling ==  True: outcome = UNDETERMINED and raise_message "<Insert hvac.id> has zone(s) with non-mechanical cooling in the proposed design, conduct a manual check that the baseline building design includes a fan power allowance of Pfan = CFMnmc × 0.054, where, CFMnmc = the baseline non-mechanical cooling fan airflow, cfm for the non-mechanical cooling fan in additional to the 0.3 W/CFM allowance for the HVAC system."`  
+            - Case 4: Else, fail: `Else: outcome = "Fail"`
```
mkarpman, line 73: Sould be Case 6

mkarpman (approved):

PR #801, opened 2022-11-20
Create get_largest_exhaust_source_cfm.md

claperle:


docs/ruleset_functions/get_largest_exhaust_source_cfm.md
```diff
@@ -0,0 +1,39 @@
+# get_largest_exhaust_source_cfm  
+
+**Description:** Returns the largest exhaust source cfm in the RMI sent to this function. This is for evaluating G3.1.2.10 exception 6 which states "Where the largest exhaust source is less than 75% of the design outdoor airflow. This exception shall only be used if exhaust air energy recovery is not used in the proposed design. "
```
mkarpman, line 3: I think the 75% criteria should be checked based on the zones served by a given baseline system, not for the entire RMI.
Also, I think we only need to be looking at the baseline design, not the proposed design.

docs/ruleset_functions/get_largest_exhaust_source_cfm.md
```diff
@@ -0,0 +1,39 @@
...
+- For each hvac system in the P_RMI: `for hvac_p in P_RMI...HeatingVentilatingAirConditioningSystem`  
+    - Create a fan system object: `fan_system_p = hvac_p.fan_system`  
+    - Get list of relief fan fan objects: `fan_sys_relief_list = list(fan_system_p.relief_fans)`  
+    - Get list of exhaust fan fan objects: `fan_sys_exhaust_list = list(fan_system_p.exhaust_fans)`  
+    - Reset the total fan cfm variable: `total_fan_cfm = 0` 
+    - Reset the fan cfm variable: `fan_cfm = 0` 
+    - Cycle through the list of relief fans: `for fan_p in fan_sys_relief_list:`  
+        - Get the fan cfm: `fan_cfm = fan_p.design_airflow`  
+        - Total the fan cfm across the multiple fans: `total_fan_cfm = total_fan_cfm + fan_cfm`  
+    - Check if the total fan cfm is greater than the current value of Get_largest_exhaust_source_cfm, if it is then set Get_largest_exhaust_source_cfm equal to the cfm: `if total_fan_cfm > Get_largest_exhaust_source_cfm: Get_largest_exhaust_source_cfm = total_fan_cfm`
```
mkarpman, line 24: If the goal of this function is to determine the maximum single exhaust flow, why are we summing flows from multiple fans?

claperle: For the purposes of finding the largest exhaust source, even if an hvac system has multiple fans I think we would consider that one exhaust source right? The exhaust source would be the sum of the cfm (assuming they are in parallel).

docs/ruleset_functions/get_largest_exhaust_source_cfm.md
```diff
@@ -0,0 +1,39 @@
+# get_largest_exhaust_source_cfm
```
claperle, line 1: Archived because a different approach was used for this

PR #802, opened 2022-11-21
Rds/cml/19 24

claperle:


docs/section19/19-24.md
```diff
@@ -0,0 +1,100 @@
...
+    - Reset ER_modeled boolean variable to false: `ER_modeld = False`  
+    - Create a fan system object: `fan_system_b = hvac.fan_system`  
+    - Get list of supply fan fan objects: `fan_sys_supply_list = list(fan_system_p.supply_fans)`  
+    - Cycle through the list of supply fans: `for fan_b in fan_sys_supply_list:`  
+        - Get the supply cfm and add to cumulative supply cfm as the logic cycles through the supply fans: `supply_cfm_b = supply_cfm_b + fan_b.design_airflow`  
+    - Get the minimum OA cfm (use min or max OA cfm?): `OA_cfm = fan_system_b.minimum_outdoor_airflow`  
+    - Calculate the OA fraction: `OA_fraction = OA_cfm/supply_cfm_b`  
+    - Check if there are any energy recovery objects associated with the fan system object and that the type is not equal to None: `if fan_sys_b.air_energy_recovery != Null and fan_sys_b.air_energy_recovery.energy_recovery_type != "NONE":` 
+        - Set energy recovery modeled boolean variable to true: `ER_modeled = True`  
+        - Get the enthalpy recovery ratio: `enthalpy_rec_ratio = fan_sys_b.air_energy_recovery.enthalpy_recovery_ratio`
```
mkarpman, line 46: To verify the 50% enthalpy recovery ratio we also need to check the percentage of total exhaust and supply go through the ERV device. I think we need to add AirEnergyRecovery.outdoor_air_flow_rate and AirEnergyRecovery.exhaust_flow_rate

docs/section19/19-24.md
```diff
@@ -0,0 +1,100 @@
+# Section 19 - Rule 19-24         
+**Schema Version:** 0.0.23    
+**Mandatory Rule:** True    
+**Rule ID:** 19-24       
+**Rule Description:** Baseline systems with >= 5,000 CFM supply air and >= 70 %OA shall have energy recovery with 50% enthalpy recovery ratio. The following exceptions apply:
+1. Systems serving spaces that are not cooled and that are heated to less than 60°F. 
+2. Systems exhausting toxic, flammable, or corrosive fumes or paint or dust. This exception shall only be used if exhaust air energy recovery is not used in the proposed design. 
+3. Commercial kitchen hoods (grease) classified as Type 1 by NFPA 96. This exception shall only be used if exhaust air energy recovery is not used in the proposed design. 
+4. Heating systems in Climate Zones 0 through 3. 
+5. Cooling systems in Climate Zones 3C, 4C, 5B, 5C, 6B, 7, and 8.
```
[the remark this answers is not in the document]
claperle: 4, only 9&10 don't require in CZ 0 through 3
5, 3C never requires ER

Pending...

docs/section19/19-24.md
claperle: exhaust flow = OA flow then enthalpy ratio at the input value

claperle: @claperle Create an issue to add these to the schema.

claperle: I am still not sure how we would verify it without enthalpy data for the air streams at rated conditions. I would like to add these for the sake of 19-19 calculations but I am not sure we can verify this here.

docs/section19/19-24.md
```diff
@@ -0,0 +1,100 @@
...
+**Schema Version:** 0.0.23    
+**Mandatory Rule:** True    
+**Rule ID:** 19-24       
+**Rule Description:** Baseline systems with >= 5,000 CFM supply air and >= 70 %OA shall have energy recovery with 50% enthalpy recovery ratio. The following exceptions apply:
+1. Systems serving spaces that are not cooled and that are heated to less than 60°F. 
+2. Systems exhausting toxic, flammable, or corrosive fumes or paint or dust. This exception shall only be used if exhaust air energy recovery is not used in the proposed design. 
+3. Commercial kitchen hoods (grease) classified as Type 1 by NFPA 96. This exception shall only be used if exhaust air energy recovery is not used in the proposed design. 
+4. Heating systems in Climate Zones 0 through 3. 
+5. Cooling systems in Climate Zones 3C, 4C, 5B, 5C, 6B, 7, and 8. 
+6. Where the largest exhaust source is less than 75% of the design outdoor airflow. This exception shall only be used if exhaust air energy recovery is not used in the proposed design.
```
mkarpman, line 11: I think this exception should be applied based on exhaust rates of the zones served by a given baseline system, not based on the zones in the entire building.

docs/section19/19-24.md
mkarpman: If we have these additional fields, we could calculate the recovery effectiveness as AirEnergyRecovery.enthalpy_recovery_ratio*AirEnergyRecovery.exhaust_flow_rate/(Total central and zonal exhaust flow for all zones served by the system)* AirEnergyRecovery.outdoor_air_flow_rate/(Total OA flow rates supplies to all zones served by this syste).

docs/section19/19-24.md
```diff
@@ -0,0 +1,100 @@
+# Section 19 - Rule 19-24         
+**Schema Version:** 0.0.23    
+**Mandatory Rule:** True    
+**Rule ID:** 19-24       
+**Rule Description:** Baseline systems with >= 5,000 CFM supply air and >= 70 %OA shall have energy recovery with 50% enthalpy recovery ratio. The following exceptions apply:
+1. Systems serving spaces that are not cooled and that are heated to less than 60°F.
```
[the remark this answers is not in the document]
claperle: Done

docs/section19/19-24.md
claperle: This is still in progress.

docs/section19/19-24.md
claperle: Function updated with question: For the purposes of finding the largest exhaust source, even if an hvac system has multiple fans I think we would consider that one exhaust source right? The exhaust source would be the sum of the cfm (assuming they are in parallel).

docs/section19/19-24.md
```diff
@@ -0,0 +1,100 @@
...
+    - Reset serves zones with hvac systems with series heat recovery for dehumidification (for assessing exception 7): `serves_zones_that_have_dehumid_heat_recovery = FALSE` 
+
+    - Get the list of zones associated with the hvac system: `zone_list = dict_of_zones_and_terminal_units_served_by_hvac_sys[hvac.id]["ZONE_LIST"]`     
+    - for each zone in zone_list: `for zone in zone_list`  
+        - Get the matching zone in the P_RMI (Weili: can we do this with an object instead of ID like I did here?): `zone_p = match_data_element(P_RMI,Zone,zone)`  
+        - For each space in zone: `for space_p in zone_p.spaces:`
+            - Check if space is of type that is likely to have systems exhausting toxic, flammable, or corrosive fumes or paint or dust, if yes then set zone is likely to have systems exhausting toxic, flammable, or corrosive fumes or paint or dust boolean variable to true based on lighting_space_type: `if space_p.lighting_space_type in ["LABORATORY_EXCEPT_IN_OR_AS_A_CLASSROOM"]: serves_zones_with_systems_likely_exhausting_toxic_etc = TRUE`
```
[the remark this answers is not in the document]
claperle: I need to add logic to address this

docs/section19/19-24.md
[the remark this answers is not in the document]
claperle: I think this is covered in the RDS

docs/section19/19-24.md
claperle: we decided that checking the enthalpy recovery ratio would be undetermined. This means that none of the outcomes can be pass for this rule.

How about we split this into two rules. One that just checks that energy recovery was modeled at all if required and another that checks that the enthalpy recovery ratio was modeled per the 50%. This way we can separate checking the enthalpy ratio part of G3.1.2.10. We would get some passes with the first rule but the second rule would always be undetermined and would just show all the hvac systems where energy recovery was modeled in the baseline where the enthalpy recovery ratio needs to be checked.

claperle: Group on board with this. CML to implement.

mkarpman (approved):

supriyagoel (approved):

PR #803, opened 2022-11-22
RT/YJ/SECTION 22 RMD

yunjoonjung-PNNL:
While I was updating 22-33(b), I realized our recent secondary loop addition task caused errors for some of the rules. So, I examined all the section 22 RMDs and made them pass. The Excel master file on OneDrive was updated as well.


weilixu (approved):
Thanks Yun Joon, I think this PR looks good to me.

However, please fix the conflict before merging this PR.

PR #804, opened 2022-11-22
Rs/wx/rule22 36

weilixu:


yunjoonjung-PNNL (approved):
@weilixu I made some cosmetic updates (e.g., variable name, comment) to be consistent with other rules. Also, I tested this rule with the RMD and the rule passed with the RMD. However, I didn't add the RMD in this branch b/c it's already included in #803 PR.

weilixu:
Updated to schema 0.23

PR #808, opened 2022-11-29
YJ/Section 5 and 6 RMD updates based on schema 0.0.23

yunjoonjung-PNNL:
This PR includes RMD updates for the schema 0.0.23 version.

A few notes are
1) the 0.0.23 schema doesn't affect any of section 6 RMDs.
2) Rule dependency key is removed in section 5 RMDs because the dependency table is hidden in the envelope master file and other sections don't also use the rule dependency key. Please let me know if this has to be included in the RMD.

All the rules passed with the updated RMDs.


weilixu:
Looks good! Thank you Yun Joon. I do not recall the rule dependency is ever in the rule table anymore. @jugonzal07 could you confirm if it can be removed?

weilixu (approved):

PR #817, opened 2022-11-30
Rct/wx/rule definition unit test

weilixu:
Add rule tests to unit test


yunjoonjung-PNNL (approved):
This is a great addition and thanks for this effort Weili. 
I believe the failed unit tests should be fine  when we check this again after the schema update PRs are approved.

weilixu:
> and thanks for this effort Weili.
> I believe the failed unit tests should be fine when we check this again after the schem
Thanks Yun Joon. Yes, the failed cases were due to some errors in the chillers.json or boilers.json. That is also the reason we need this to be part of the unit test.

yunjoonjung-PNNL:
> > and thanks for this effort Weili.
> > I believe the failed unit tests should be fine when we check this again after the schem
> > Thanks Yun Joon. Yes, the failed cases were due to some errors in the chillers.json or boilers.json. That is also the reason we need this to be part of the unit test.

Make sense! Thanks for addressing this Weili.

charliepnnl (approved):
I think the proof is in the pudding here. If it runs and the tests all pass, I am good with this.  Thanks Weili

yunjoonjung-PNNL:
@weilixu @jugonzal07 The 5-19 and 5-42 RMDs were just updated. 

This issue occurred because these RMDs were deleted in the envelope master Excel file for some reason. Hence, when the RMDs in the Excel file were updated for the schema 0.0.23 version, these two RMDs remained as they were in the Excel file instead of getting updated.

New Pytest passed as expected now!

weilixu:
@yunjoonjung-PNNL @jugonzal07 @charliepnnl 
Thanks all. I think this looks really good. I will merge it today.

PR #818, opened 2022-12-02
Change the behavior of required_fields

weilixu:
to be same as missing data

### DIFF
--- a/rct229/rule_engine/rule_base.py
+++ b/rct229/rule_engine/rule_base.py
@@ -106,7 +106,7 @@ class RuleDefinitionBase:
         if self.rmr_context:
             outcome["rmr_context"] = self.rmr_context
 
-        # context will be a string if the context does not exist for any of the RMR used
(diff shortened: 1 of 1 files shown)


yunjoonjung-PNNL (approved):
Great addition! Thank you Weili.

charliepnnl (approved):
While we're here, should we also handle a missing context in a similar way?
Starting with what is now line 172:
```
else: 
    # context_or_string should be a string indicating the RMDs that are missing
    # such as "MISSING_BASELINE"
    outcome["result"] = "UNDETERMINED"
    outcome["message"] = context_or_string
```

charliepnnl:
Should we create a class called `OutcomeResult` that enumerates the allowed values of `outcome["result"]`?

weilixu:
> Should we create a class called `OutcomeResult` that enumerates the allowed values of `outcome["result"]`?

This is addressed in the reporting PR

PR #823, opened 2022-12-06
Section 12 to 0.23 schema

jugonzal07:
updated receptacles ruletest JSONs to latest schema


rct229/ruletest_engine/ruletest_jsons/section12/rule_12_4.json
```diff
@@ -0,0 +1,190 @@
+{
```
yunjoonjung-PNNL, line 1: Could you verify whether we have rules 12- 4, 5, 6, and 7? This is because we don't have these Python codes.

yunjoonjung-PNNL (approved):
Looks great to me! I'd appreciate it if you could verify whether we need 12-4, 5, 6, and 7's RMDs or not.

rct229/ruletest_engine/ruletest_jsons/section12/rule_12_4.json
weilixu: We may need to clean up section 12 since some of the rules are changed to secondary rules. For now, we are ok with having 4, 5, 6, 7 but need to address them in the next spring.

weilixu (approved):

PR #825, opened 2022-12-06
Section 22 rule tests to 0.23

jugonzal07:
Updated chiller rule test JSONs to work with 0.23 schema update


yunjoonjung-PNNL (approved):
Looks great to me Juan. Thank you!

weilixu (approved):
I added YJ schema updates to this branch so it passes the pytest.
Also, two rule definition bugs are fixed, one in 21 and the other is in 31.
This PR is good to merge now.

PR #829, opened 2022-12-08
RS/YJ/is baseline system 12

yunjoonjung-PNNL:
This PR includes both the baseline system 12 function and testing developments. Thank you!


weilixu (approved):
Great work, thanks

PR #835, opened 2022-12-13
Disaggregate master RMD file script

jugonzal07:
Added a script that breaks out a master JSON file that has no test in…formation. An example of this is the system_types.xlsx file that simply has examples of every system type without any tie to rules or tests.


weilixu (approved):
Nice, thanks to update the function

PR #839, opened 2022-12-13
Create 19-21.md

claperle:


docs/section19/19-21.md
```diff
@@ -0,0 +1,52 @@
+# Section 19 - Rule 19-21             
+**Schema Version:** 0.0.23    
+**Mandatory Rule:** False    
+**Rule ID:** 19-21             
+**Rule Description:** For Systems 9 and 10, if modeling a non-mechanical cooling fan is required by Section G3.1.2.8.2, there is a fan power allowance of Pfan = CFMnmc × 0.054, where, CFMnmc = the baseline non-mechanical cooling fan airflow, cfm for the non-mechanical cooling.
```
mkarpman, line 5: I think we can get rid of this RDS because the logic seems to be covered by 19-20 case 3.

claperle: Agreed. Marked archived.

weilixu:
Closed because:
I think we can get rid of this RDS because the logic seems to be covered by 19-20 case 3.

PR #841, opened 2022-12-13
RS/YJ/is baseline system 10

yunjoonjung-PNNL:
This PR includes both ```is_baseline_system_10``` development and its pytest files. Thank you!


rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
```diff
@@ -0,0 +1,137 @@
...
+     The function returns either Sys-10 or Not_Sys_10 string output which indicates whether the HVAC system is ASHRAE 90.1 2019 Appendix G system 10 (Heating and Ventilation with electric heating).
+    """
+
+    is_baseline_system_10 = HVAC_SYS.UNMATCHED
+
+    hvac_b = find_exactly_one_hvac_system(rmi_b, hvac_b_id)
+
+    # check if the hvac system has the required sub systems for system type 10
+    preheat_system = hvac_b.get("preheat_system")
+    has_required_preheat_sys = (
```
weilixu, line 72: Since we have many of these checks in every is_baseline_system function, I was thinking we can condense them into a utility function.
So there will be three utility functions: preheat, heating and cooling systems.
@charliepnnl what do you think?

rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
```diff
@@ -0,0 +1,137 @@
...
+    )
+
+    heating_system = hvac_b.get("heating_system")
+    has_required_heating_sys = (
+        heating_system is None
+        or heating_system.get("heating_system_type") is None
+        or heating_system["heating_system_type"] == HEATING_SYSTEM.NONE
+    )
+
+    has_required_fan_sys = hvac_b.get("fan_system") is None
```
weilixu, line 85: We might need to also find out the whether there is a data under fan system shows NONE. Probably worth developing a utility function for checking has required fan sys,

rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
```diff
@@ -0,0 +1,137 @@
...
+    # check if the hvac system has the required sub systems for system type 10
+    preheat_system = hvac_b.get("preheat_system")
+    has_required_preheat_sys = (
+        preheat_system is None
+        or preheat_system.get("heating_system_type") is None
+        or preheat_system["heating_system_type"] == HEATING_SYSTEM.NONE
+    )
+
+    heating_system = hvac_b.get("heating_system")
+    has_required_heating_sys = (
```
weilixu, line 79: The check returns True if no required heating sys. Seems we have rename it to:

has_heating_sys?
I suggest use Not instead of repeating the same logic twice.

rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
```diff
@@ -0,0 +1,137 @@
...
+        return is_baseline_system_10
+
+    # When the first logic of are_sys_10_data_matched is false
+    has_required_heating_sys = (
+        hvac_b.get("heating_system") is not None
+        or hvac_b["heating_system"].get("heating_system_type") is not None
+        or hvac_b["heating_system"]["heating_system_type"] != HEATING_SYSTEM.NONE
+    )
+
+    has_required_preheat_sys = (
```
weilixu, line 113: This should be not Null

I recommend to just use one parameter:
has_preheat_sys.
True means has preheat system
False means does not have required system.

weilixu (changes requested):
Some logics need to update

rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
yunjoonjung-PNNL: For ```fan_system``` key, there is no ```NONE``` type. In this case, does it make sense to just check whether the ```fan_system``` key exists or not?

rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
yunjoonjung-PNNL: Based on our discussion, I'll create this part's function in the ```utils``` file and make a separate PR to get all the ```is_baseline_system``` functions.

rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
charliepnnl: I think the big point here is that the name `has_required_fan_sys` is misleading. One would think that if it is True, then a fan system exists.

Creating a utility function `has_fan_sys` that parallels the other utility functions mentioned above would clean this up nicely.

rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
charliepnnl: Presumably, you'll use the new utility function you will create for this.

charliepnnl (approved):
Approved with comments that should be addressed

rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
yunjoonjung-PNNL: Yes. These functions were in PR #891. I'd appreciate if you could review this PR as well!

rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
yunjoonjung-PNNL: Yes this is in #891.

rct229/ruleset_functions/baseline_systems/is_baseline_system_10.py
yunjoonjung-PNNL: I agree! ```has_fan_sys``` function is part of #891.

weilixu (approved):

PR #843, opened 2022-12-15
Rds/kjw/rule 11 1

KarenWGard:


docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
...
+
+**Rule ID:** 11-1
+
+**Rule Description:** "Where a complete service water-heating system exists, the proposed design shall reflect the actual system type. Where a service water-heating system has been designed the service waterheating type shall be consistent with design documents."
+
+**Rule Assertion:** Options are PASS/FAIL
+
+**Appendix G Section Reference:** Table G3.1 #11, proposed column, a & b
+
+**Evaluation Context:** P-RMD each SHW type
```
claperle, line 14: Should this have a different context? Maybe P-RMD each ServiceWaterHeatingEquipment. I am not following why it would be evaluated for each type as opposed to system.

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
...
+- **get_component_by_id**
+- **compare_context_pair** - there is no RDS for this function, but it is a function developed for Rule 1-6 that compares two elements
+- **get_SHW_types_and_spaces**
+- **get_SHW_equipment_connected_to_use_type**  
+
+**Applicability Checks:**
+- check that the SHW type in the U_RMD has SHW loads
+
+## Applicability Checks:
+- only projects with SHW for the SHW space type in the user model are expected to have a SHW system in the user model.  If there is no SHW in the user model, we assume there is no SHW system designed, and instead rule 11-3 Applies and P_RMD matches B_RMD.
```
claperle, line 26: Not following what is meant by "with SHW for the SHW space type". Maybe it will become cleaerer as I review this.

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
...
+- use the function get_SHW_types_and_spaces to get a list of spaces for each SHW type: `shw_and_spaces_dict = get_SHW_types_and_spaces(U_RMD)`
```
claperle, line 27: Suggest modifying the language to be more specific about what "SHW type" means. Suggest  "get a list of spaces associated with each SWH building area type". Using SHW type makes you think it is a type of SWH.

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
...
+- look at each SHW space type: `for shw_space_type in shw_and_spaces_dict:`
```
claperle, line 28: suggest "look at each building area type"

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
+# Service_Water_Heating - Rule 11-1
+**Schema Version:** 0.0.38
```
JacksonJ-KC, line 2: current version is 0.0.36 right?

docs/section11/Rule_11-1.md
JacksonJ-KC: I am struggling with the same thing

docs/section11/Rule_11-1.md
JacksonJ-KC: I think this is referring to how we won't expect garages to have SHW but agree it can be clarified

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
...
+**Applicability Checks:**
+- check that the SHW type in the U_RMD has SHW loads
+
+## Applicability Checks:
+- only projects with SHW for the SHW space type in the user model are expected to have a SHW system in the user model.  If there is no SHW in the user model, we assume there is no SHW system designed, and instead rule 11-3 Applies and P_RMD matches B_RMD.
+- use the function get_SHW_types_and_spaces to get a list of spaces for each SHW type: `shw_and_spaces_dict = get_SHW_types_and_spaces(U_RMD)`
+- look at each SHW space type: `for shw_space_type in shw_and_spaces_dict:`
+  - look at each space: `for space_id in shw_and_spaces_dict[shw_space_type]:`
+    - get the space using get_component_by_id: `space = get_component_by_id(U_RMD, space_id)`
+    - look for the ServiceWaterHeatingUse in the space.  If even one space has a ServiceWaterHeatingUse, continue to rule logic: `if len(u.space.service_water_heating_uses) > 0: CONTINUE TO RULE LOGIC`
```
JacksonJ-KC, line 31: Iteration is already through each space so I think this should say "If the space has at least 1 ServiceWaterHeatingUse, continue to rule logic"

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
...
+    - continue if all_match is still true: `if all_match:`
+      - check if this equipment type matches between the proposed and user models.  First check whether there are the same number equipment: `if len(p_shw_equipment_dict[shw_equipment_type]) == len(u_shw_equipment_dict[shw_equipment_type]):`
+          - look at each equipment type system in the proposed model and see if there is one that is the same in the user model.  This check relies on the understanding that the RCT team has a method for comparing all elements within an object match (compare_context_pair): `for p_SHW_equip_id in p_shw_equipment_dict[shw_equipment_type]:`
+              - if this system is in U_RMD.service_water_heating_distribution_systems, compare the systems: `if p_SHW_equip_id in u_shw_equipment_dict[shw_equipment_type]:`
+                  - use compare_context_pair to compare systems and set all_match to FALSE if the systems don't compare: `if !compare_context_pair(p_SHW_equip_id, p_SHW_equip_id, $, extra_schema_for_SHW_comparison.json, true, compare_context_str, error_str): all_match = FALSE`
+              - otherwise, set all_match to false: `else: all_match = FALSE`
```
JacksonJ-KC, line 47: I think all_match = TRUE here? It is currently being set to FALSE in all conditions.

docs/section11/Rule_11-1.md
JacksonJ-KC: Can SWH systems serve more than one SWH building area type? Is that accounted for already?

JacksonJ-KC:

docs/section11/Rule_11-1.md
claperle: I think it could certainly be the case that a SHW system could serve more than one building area type in the proposed design.

docs/section11/Rule_11-1.md
claperle: This is clearer now. Recommend modifying nomenclature for clarity.

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
...
+    - get the space using get_component_by_id: `space = get_component_by_id(U_RMD, space_id)`
+    - look for the ServiceWaterHeatingUse in the space.  If even one space has a ServiceWaterHeatingUse, continue to rule logic: `if len(u.space.service_water_heating_uses) > 0: CONTINUE TO RULE LOGIC`
+  - if the program reaches this line without going to the rule logic, the project is not applicable for this SHW heating use: `NOT_APPLICABLE`
+
+    ## Rule Logic: 
+    - create a boolean to keep track of whether everything matches: `all_match = TRUE`
+  - create an error string: `error_str = ""`
+  - create the compare context string: `compare_context_str = "AppG 11-1 P_RMD Equals U_RMD"`
+  - get a dictionary of all SHW equipment connected to the SHW space type in the proposed model: `p_shw_equipment_dict = get_SHW_equipment_connected_to_use_type(P_RMD)[shw_space_type]`
+  - get a dictionary of all SHW equipment connected to the SHW space type in the user model: `u_shw_equipment_dict = get_SHW_equipment_connected_to_use_type(U_RMD)[shw_space_type]`
```
claperle, line 39: This description is a little confusing because we have the ServiceWaterHeatingEquipment object so I was not sure if you meant that or ALL the equipment. Suggest just adding what you mean in paraentheses (pumps, tasks, distribution systems, etc.).

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
...
+  - create a list of all of the SHW equipment types that exist in both the proposed and user models: `shw_equipment_types = set(p_shw_equipment_dict.keys()) | set(u_shw_equipment_dict.keys())`
```
claperle, line 40: I am not sure what the | is doing exactly. Can you confirm this is creating a list of the the unique keys included across the two dictionaries?

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+    tanks:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    is_central_system:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    service_water_piping:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    distribution_compactness:
```
claperle, line 27: does anyone know what this is?

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+      AppG 11-1 P_RMD Equals U_RMD: false
+      AppG 11-3 P_RMD Equals B_RMD: false
+    control_type:
+      AppG 11-1 P_RMD Equals U_RMD: false
```
claperle, line 31: Not following why the control type woudl not be the same

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+      AppG 11-3 P_RMD Equals B_RMD: false
+    notes:
+      AppG 11-1 P_RMD Equals U_RMD: false
+      AppG 11-3 P_RMD Equals B_RMD: false
+    is_recirculation_loop:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    insulation_thickness:
+      AppG 11-1 P_RMD Equals U_RMD: false
```
claperle, line 71: why false for P and U comparison but not P and B?

claperle, line 71: Seems inconsistent to have length, diameter, etc. as set to true but this set to false. What was the thinking behind this choice?

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+      AppG 11-3 P_RMD Equals B_RMD: true
+    service_water_piping:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    distribution_compactness:
+      AppG 11-1 P_RMD Equals U_RMD: false
+      AppG 11-3 P_RMD Equals B_RMD: false
+    control_type:
+      AppG 11-1 P_RMD Equals U_RMD: false
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 32: Not following why the control type would not be the same

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+    configuration_type:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    is_recovered_heat_from_drain_used_by_water_heater:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 38: This seems like a measure that would not be implemented by a tenant because it is intimately connected with the SWH system itself.  I am thinking maybe this one should be true. Let's discuss.

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+    drain_heat_recovery_efficiency:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 41: Same as above.

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+    drain_heat_recovery_type:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 44: Same as above.

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+      AppG 11-3 P_RMD Equals B_RMD: true
+    tank:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    status_type:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    solar_thermal_systems:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 209: Same question as above.

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+    drain_heat_recovery_efficiency:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
+    drain_heat_recovery_type:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
+    flow_multiplier_schedule:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 47: I am not remembering why we thougth this could change. If it is a tenant fitout then a SWH system would exist and that would be modeled so I can't think of a scenario where this would be false. Can you please remind me?

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+      AppG 11-3 P_RMD Equals B_RMD: true
+    is_central_system:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    service_water_piping:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    distribution_compactness:
+      AppG 11-1 P_RMD Equals U_RMD: false
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 29: why are both these false?

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+      AppG 11-3 P_RMD Equals B_RMD: true
+    solar_thermal_systems:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
+    hot_water_loop:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    draw_pattern:
+      AppG 11-1 P_RMD Equals U_RMD: false
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 215: Can we discuss this on the call. If it is a tenant fitout then a SWH system would exist and that would be modeled so I can't think of a scenario where this would be false. I must be blanking on the scenario where this could be false.

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+    loop_or_piping:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    specification_method:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    design_electric_power:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 237: why false?

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+    motor_nameplate_power:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    design_head:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 243: Discuss as part of the tentant fitout discussion.

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+      AppG 11-3 P_RMD Equals B_RMD: true
+    motor_efficiency:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    speed_control:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    design_flow:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
```
claperle, line 255: Discuss as part of the tentant fitout discussion.

docs/section11/SHW comparison fields.yaml
supriyagoel: @claperle and @KarenWGard could this be the case when there are low flow fixtures in the P RMD?

claperle: @supriyagoel I think that is the reason I think Karen chose fail here but I was trying to remember when there could be low flow fixtures in the proposed but no (1) existing SWH or (2)  no water heater as part of the design. I was thinking if it was a tenent fit-out there would either be an existing water heater or a water heater as part of the design and if it was core and shell then the low flow fixtures were not be relevant. I guess maybe there is a a rare scenario where someone could be doing an interior fitout where the central SWH systems are yet to be designed?

docs/section11/Rule_11-1.md
KarenWGard: converting the lists to sets and using "|" allows me to get a set of all of the keys in both `p_shw_equipment_dict `and `u_shw_equipment_dict`.  In this case, I'm expecting `p_shw_equipment_dict ` to equal `u_shw_equipment_dict `, so if the combination of the two has a different number of elements, I know that at least one dictionary has keys that are not present in the other

docs/section11/SHW comparison fields.yaml
KarenWGard: I think it's a Title 24 element.

docs/section11/SHW comparison fields.yaml
KarenWGard: agree that length, diameter, etc should maybe not be evaluated.  For insulation thickness - the SWH are not supposed to model piping losses, and there is another data element for that, so the insulation thickness, even if it exists in the models should have no effect on the simulation.

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    solar_thermal_systems:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: false
+    hot_water_loop:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    draw_pattern:
+      AppG 11-1 P_RMD Equals U_RMD: false
```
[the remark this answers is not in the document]
KarenWGard: this is another one that would be good to discuss - my thoughts here are that that draw pattern is based on SWH use and use patterns and that SWH use could differ between user and proposed models if the user model reflects. SWH equipment savings that are not allowed in the proposed model.

docs/section11/SHW comparison fields.yaml
KarenWGard: it's a title 24 element.  Keep in mind that "false" just means that we don't evaluate whether they match, not that they shouldn't match.

docs/section11/SHW comparison fields.yaml
KarenWGard: we discussed on the call 5/15/2024, and we expect everything to be the same, unless there is a specific reason the two would reasonably be different.

docs/section11/SHW comparison fields.yaml
KarenWGard: draw pattern is calculated based on flow and a couple other variables.  There are situations in which the proposed water use and user water use might differ - the Marcel Hotel is one example - where the User model reflected lower SWH use because of the ozone system for the laundry, however, we were not able to take 100% credit for this in the proposed model.

docs/section11/SHW comparison fields.yaml
KarenWGard: > we discussed on the call 5/15/2024, and we expect everything to be the same, unless there is a specific reason the two would reasonably be different.

Discussed with Christina on 5/20/2024 & she agrees with this approach

docs/section11/SHW comparison fields.yaml
KarenWGard: 5/20/2024 discussion with Christaina - all of these should be false

docs/section11/SHW comparison fields.yaml
KarenWGard: 5/20/24 this should be the same

docs/section11/SHW comparison fields.yaml
KarenWGard: 5/24/2024 - agreed this is OK

docs/section11/SHW comparison fields.yaml
KarenWGard: 5/24/2024 - agreed this is OK

docs/section11/SHW comparison fields.yaml
KarenWGard: 5/24/2024 - agreed this is OK

docs/section11/SHW comparison fields.yaml
KarenWGard: 5/20/2024 - change to true - while there could be a case where a tenant installs Solar hot water panels without a complete SWH system, this is an extremely rare edge-case that should trigger a manual review anyway

docs/section11/SHW comparison fields.yaml
KarenWGard: 5/20/2024 - change to true

docs/section11/SHW comparison fields.yaml
KarenWGard: 5/20/2024 - change to true

docs/section11/SHW comparison fields.yaml
KarenWGard: 5/20/2024 - change to true

docs/section11/SHW comparison fields.yaml
KarenWGard: change to true - we expect the flow rate to be adjusted by the design_flow_rate, not by the schedule

docs/section11/Rule_11-1.md
KarenWGard: the evaluation context for this rule was changed to each SWH distribution, so there is no longer "type" in the applicability check

docs/section11/Rule_11-1.md
KarenWGard: evaluation context changed to each SWH Distribution system

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,65 @@
...
+**Rule Assertion:** Options are PASS/FAIL
+
+**Appendix G Section Reference:** Table G3.1 #11, proposed column, a & b
+
+**Evaluation Context:** P-RMD each SWH Distribution
+**Data Lookup:**   
+**Function Call:** 
+- **get_component_by_id**
+- **compare_context_pair** - there is no RDS for this function, but it is a function developed for Rule 1-6 that compares two elements
+- **get_SWH_types_and_spaces**
```
claperle, line 19: Should this now be "get_spaces_associated_with_each_SWH_bat"?

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,65 @@
...
+- use the function get_SWH_equipment_associated_with_each_swh_distribution_system to get a dictionary of SWH equipment associated with each Distribution System for the User model (pumps, tanks, SWH use, distribution system, etc): `u_swh_system_and_equip_dict = get_SWH_equipment_associated_with_each_swh_distribution_system(U_RMD)`
+- for each distribution system: `for distribution_id in p_swh_system_and_equip_dict:`
+    - get the distribution system: `p_distribution = get_component_by_id(P_RMD, distribution_id)`
+    - for each swh use in the dictionary, check to see that the swh use exists in the user model: `for swh_use_id in p_swh_system_and_equip_dict[distribution_id]["SPACE_USES"]:`
+        - get the user swh_use: `u_swh_use = get_component_by_id(U_RMD, swh_use_id)`
+        - if the u_swh_use exists, continue to rule logic as we only need one SWH use to have this rule be applicable: `if(u_swh_use): CONTINUE TO RULE LOGIC`
```
claperle, line 33: So this will keep repeating the logic below for all uses?

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,65 @@
...
+      - continue if all_match is still true: `if all_match:`
+        - check if this equipment type matches between the proposed and user models.  First check whether there are the same number equipment: `if len(p_shw_equipment_dict[shw_equipment_type]) == len(u_shw_equipment_dict[shw_equipment_type]):`
+            - look at each equipment type system in the proposed model and see if there is one that is the same in the user model.  This check relies on the understanding that the RCT team has a method for comparing all elements within an object match (compare_context_pair): `for p_SWH_equip_id in p_shw_equipment_dict[shw_equipment_type]:`
+                - if this system is in U_RMD.service_water_heating_distribution_systems, compare the systems: `if p_SWH_equip_id in u_shw_equipment_dict[shw_equipment_type]:`
+                    - use compare_context_pair to compare systems and set all_match to FALSE if the systems don't compare: `if !compare_context_pair(p_SWH_equip_id, p_SWH_equip_id, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, error_str): all_match = FALSE`
```
claperle, line 52: I do not know where to see this function but trust that this works. Maybe Weili should check out this line.

docs/section11/SHW comparison fields.yaml
```diff
@@ -0,0 +1,396 @@
...
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    child:
+      AppG 11-1 P_RMD Equals U_RMD: true
+      AppG 11-3 P_RMD Equals B_RMD: true
+    piping_losses:
+      AppG 11-1 P_RMD Equals U_RMD: false
+      AppG 11-3 P_RMD Equals B_RMD: true
+
+SolarThermal:
```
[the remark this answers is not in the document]
KarenWGard: this document refers both to 11-1 and 11-3, and for 11-1, there shouldn't be SolarThermal, but we are agreed that if 11-1 applies, everything in the SWH system should be equal, so I held that true for SolarThermal even though it shouldn't be in a baseline model

docs/section11/Rule_11-1.md
KarenWGard: removed - not used

docs/section11/Rule_11-1.md
KarenWGard: yes, for all object types in included in [SHW comparison fields.yaml](https://github.com/pnnl/ruleset-checking-tool/pull/843/files#diff-e7ef3fbae75b2279d8581158c48c728c60850fa1e3b3101d3b4c613fad81ef4e)

JacksonJ-KC (changes requested):

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,64 @@
...
+- **get_SWH_equipment_associated_with_each_swh_distribution_system**  
+
+**Applicability Checks:**
+- check that the U_RMD has SWH Uses with loads
+
+## Applicability Checks:
+- only projects with SWH in the user model are expected to have a SWH system in the proposed model.  If there is no SWH in the user model, we assume there is no SWH system designed, and instead rule 11-3 Applies and P_RMD matches B_RMD.
+- use the function get_SWH_equipment_associated_with_each_swh_distribution_system to get a dictionary of SWH equipment associated with each Distribution System for the Proposed model (pumps, tanks, SWH use, distribution system, etc): `p_swh_system_and_equip_dict = get_SWH_equipment_associated_with_each_swh_distribution_system(P_RMD)`
+- use the function get_SWH_equipment_associated_with_each_swh_distribution_system to get a dictionary of SWH equipment associated with each Distribution System for the User model (pumps, tanks, SWH use, distribution system, etc): `u_swh_system_and_equip_dict = get_SWH_equipment_associated_with_each_swh_distribution_system(U_RMD)`
+- for each distribution system: `for distribution_id in p_swh_system_and_equip_dict:`
```
JacksonJ-KC, line 28: If we are verifying that all equipment in the user model has matching equipment in the proposed model, should we be iterating through the user model instead of the proposed model?

docs/section11/Rule_11-1.md
JacksonJ-KC: Agreed, would like Weili to confirm that the recursive nature of the compare_context_pair function has been communicated properly in this RDS

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,64 @@
...
+  - otherwise, the distribution equipment doesn't exist in the user model, set all_match to false: `all_match = FALSE; error_str += distribution_id + " not found in the user model"`
+  - continue if all_match is still true: `if all_match`
+    - create a list of all of the SWH equipment types that exist in both the proposed and user models: `shw_equipment_types = set(p_shw_equipment_dict.keys()) | set(u_shw_equipment_dict.keys())`
+    - loop through each of the equipment types: `for shw_equipment_type in shw_equipment_types:`
+      - continue if all_match is still true: `if all_match:`
+        - check if this equipment type matches between the proposed and user models.  First check whether there are the same number equipment: `if len(p_shw_equipment_dict[shw_equipment_type]) == len(u_shw_equipment_dict[shw_equipment_type]):`
```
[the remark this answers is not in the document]
KarenWGard: @weilixu - does the length comparison happen inside compare_context_pair?

@JacksonJ-KC - this isn't recursing inside each element, but recursing a list of different types of swh equipment (Tanks, piping, pumps, etc) where the links aren't simple enough to happen directly within compare_context_pair

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,64 @@
...
+
+**Evaluation Context:** P-RMD each SWH Distribution
+**Data Lookup:**   
+**Function Call:** 
+- **get_component_by_id**
+- **compare_context_pair** - there is no RDS for this function, but it is a function developed for Rule 1-6 that compares two elements
+- **get_SWH_equipment_associated_with_each_swh_distribution_system**  
+
+**Applicability Checks:**
+- check that the U_RMD has SWH Uses with loads
```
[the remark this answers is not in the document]
claperle: I think this rule is needed because where there is no SWH designed P_RMD would not equal U_RMD. I wonder how this is handeled for analogous situations for HVAC...In the schema these HVAC components are set to TRUE for P_RMD equalling U_RMD. I think this is a good call discussion topic.

docs/section11/Rule_11-1.md
claperle: I am not following why this all should be repeated for each use

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
...
+      - compare the two SWH equipment using compare_context_pair, if the result is false, set all_match equal to false.  We won't exit early if all_match is false as we allow the function to keep running so errors is fully populated and available to the user: `if !compare_context_pair(swh_eq_id, swh_eq_id, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, error_str): all_match = false`
+  - next, check Pumps - this will also recursively check PumpOutputValidationPointPumpOutputValidationPoint: `if len(p_swh_equipment_dict["Pumps"]) == len(u_swh_equipment_dict["Pumps"]):`
+    - look at each SWHEquipment in the proposed model: `for pump_id in p_swh_equipment_dict["Pumps"]:`
+      - compare the two pumps using compare_context_pair, if the result is false, set all_match equal to false.  We won't exit early if all_match is false as we allow the function to keep running so errors is fully populated and available to the user: `if !compare_context_pair(pump_id, pump_id, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, error_str): all_match = false`
+
```
claperle, line 49: I defer to Weili on this one since it uses a complex, novel approach.

docs/section11/Rule_11-1.md
KarenWGard: @JacksonJ-KC - I think we discussed this on a call, but I never responded formally.  There are advantages and disadvantages for both modes.  The way this rule and 11-3 are written, they both look at the **DistributionSytems** in the proposed model to confirm that each one is modeled correctly.

However, looking over this again, this misses the case where a distribution system is included in the user model but not the proposed model, so I've added a check for (and give a FAIL result) for this case

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,59 @@
...
+      - get the user swh_use: `u_swh_use = get_component_by_id(U_RMD, swh_use_id)`
+      - if the u_swh_use exists, and is greater than zero: `if u_swh_use != NULL and u_swh_use.use > 0:`
+        - compare the proposed and user models using the function compare_swh_dist_systems_and_components.  This function returns a list of errors.  If the list of errors has a length of 0, then the comparison encountered no issues: `user_proposed_comparison = compare_swh_dist_systems_and_components(P_RMD,U_RMD,user_proposed_str,swh_dist_system_id)`
+        - from here, go directly to the rule assertion: `GO TO RULE ASSERTION`
+
+  - if we got to this line without going to the rule assertion, it means that there is no equivalent system in the user model, or that the user model does not have SWH use.  Compare the proposed and baseline models using the function compare_swh_dist_systems_and_components.  This function returns a list of errors.  If the list of errors has a length of 0, then the comparison encountered no issues: `user_baseline_comparison = compare_swh_dist_systems_and_components(P_RMD,B_RMD,user_proposed_str,swh_dist_system_id)`
```
claperle, line 48: i may be missing something but couldn't it be an error and there was supposed to be a corresponding system in the user model when there is not one? Like in the case that there are SWH load? It is not clear to me that this would be caught.

KarenWGard: this is caught in lines 29 and 32 of the function:  compare_swh_dist_systems_and_components, but the error returned is not as descriptive as it could be.  A check and a specific error is added here.

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Applicability Checks:
+- in this case, all of the SWH Distribution IDs identified when the custom context was created are applicable to this rule.
+- Look at each swh distribution system id: `for swh_dist_system_id in context: CONTINUE TO RULE LOGIC`
+
+  ## Rule Logic #
+  - there are a couple cases here:
+    1.  If the swh distribution exists in the user model, AND has swh use, then expect user and proposed systems to be match
+    2.  Otherwise, expect the distribution system to match between proposed and baseline
+    3.  Systems that don't do 1 or 2 fail
```
JacksonJ-KC, line 34: even though you are writing these as a numbered list in the "code", the markdown displays them as a) b) c)

I suggest changing to lettering in the "code" too so it is consistent

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,63 @@
...
+- **compare_context_pair** - there is no RDS for this function, but it is a function developed for Rule 1-6 that compares two elements  
+- **get_swh_components_associated_with_each_swh_distribution_system**  
+
+
+**Applicability Checks:**
+- check that the U_RMD has SWH Uses with loads
+
+## Applicability Checks:
+- in this case, all of the SWH Distribution IDs identified when the custom context was created are applicable to this rule.
+- Look at each swh distribution system id: `for swh_dist_system_id in context: CONTINUE TO RULE LOGIC`
```
JacksonJ-KC, line 28: This line being in the applicability checks section is confusing me

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,63 @@
...
+  - set a list user_baseline_comparison to [].  If the end of the rule is reached and this list (and the corresponding proposed list) still have zero length, then the rule passes.  Otherwise the list will include a series of messages for the user about why it didn't pass: `user_baseline_comparison = []`
+  
+  - Check whether the system exists in the user model: `if u_swh_system_and_equip_dict[swh_dist_system_id]:`
+    - check whether the system has SWH uses in the user model: `if u_swh_system_and_equip_dict[swh_dist_system_id]["USES"]:`
+      - get the user swh_use: `u_swh_use = get_component_by_id(U_RMD, swh_use_id)`
+      - if the u_swh_use exists, and is greater than zero: `if u_swh_use != NULL and u_swh_use.use > 0:`
+        - compare the proposed and user models using the function compare_swh_dist_systems_and_components.  This function returns a list of errors.  If the list of errors has a length of 0, then the comparison encountered no issues: `user_proposed_comparison = compare_swh_dist_systems_and_components(P_RMD,U_RMD,user_proposed_str,swh_dist_system_id)`
+        - from here, go directly to the rule assertion: `GO TO RULE ASSERTION`
```
JacksonJ-KC, line 46: This is implying that a rule outcome is given for each SWH Distribution System. Is that correct?

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,63 @@
...
+
+**Rule ID:** 11-1
+
+**Rule Description:** "Where a complete service water-heating system exists, the proposed design shall reflect the actual system type. Where a service water-heating system has been designed the service waterheating type shall be consistent with design documents.  Where no service water-heating system exists or has been designed and submitted with design documents but the building will have service water-heating loads, a service water-heating system shall be modeled that matches the system type in the baseline building design"
+
+**Rule Assertion:** Options are PASS/FAIL
+
+**Appendix G Section Reference:** Table G3.1 #11, proposed column, a & b
+
+**Evaluation Context:** custom evaluation context created for this rule of all SWH Distribution system ids located in the proposed, baseline and user RMDs
```
JacksonJ-KC, line 14: Evaluation context should make it clear if this rule is evaluated once for all SWH Distribution Systems or once for each SWH Distribution System.

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,63 @@
...
+  
+  - Check whether the system exists in the user model: `if u_swh_system_and_equip_dict[swh_dist_system_id]:`
+    - check whether the system has SWH uses in the user model: `if u_swh_system_and_equip_dict[swh_dist_system_id]["USES"]:`
+      - get the user swh_use: `u_swh_use = get_component_by_id(U_RMD, swh_use_id)`
+      - if the u_swh_use exists, and is greater than zero: `if u_swh_use != NULL and u_swh_use.use > 0:`
+        - compare the proposed and user models using the function compare_swh_dist_systems_and_components.  This function returns a list of errors.  If the list of errors has a length of 0, then the comparison encountered no issues: `user_proposed_comparison = compare_swh_dist_systems_and_components(P_RMD,U_RMD,user_proposed_str,swh_dist_system_id)`
+        - from here, go directly to the rule assertion: `GO TO RULE ASSERTION`
+
+  -  If the system doesn't exist in the proposed model, we need to return an error: `if swh_dist_system_id not in p_swh_system_and_equip_dict: errors << swh_dist_system_id + " was not found in the Proposed model.
```
JacksonJ-KC, line 48: << is not python syntax

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,63 @@
...
+      - get the user swh_use: `u_swh_use = get_component_by_id(U_RMD, swh_use_id)`
+      - if the u_swh_use exists, and is greater than zero: `if u_swh_use != NULL and u_swh_use.use > 0:`
+        - compare the proposed and user models using the function compare_swh_dist_systems_and_components.  This function returns a list of errors.  If the list of errors has a length of 0, then the comparison encountered no issues: `user_proposed_comparison = compare_swh_dist_systems_and_components(P_RMD,U_RMD,user_proposed_str,swh_dist_system_id)`
+        - from here, go directly to the rule assertion: `GO TO RULE ASSERTION`
+
+  -  If the system doesn't exist in the proposed model, we need to return an error: `if swh_dist_system_id not in p_swh_system_and_equip_dict: errors << swh_dist_system_id + " was not found in the Proposed model. 
+ Because there are no SWH loads in the User model, we are expecting the Proposed and Baseline systems to match."`
+  -  If the system doesn't exist in the baseline model, we need to return an error: `if swh_dist_system_id not in b_swh_system_and_equip_dict: errors << swh_dist_system_id + " was not found in the Baseline model.
```
JacksonJ-KC, line 50: << is not python syntax

docs/section11/Rule_11-1.md
```diff
@@ -0,0 +1,63 @@
...
+  -  If the system doesn't exist in the baseline model, we need to return an error: `if swh_dist_system_id not in b_swh_system_and_equip_dict: errors << swh_dist_system_id + " was not found in the Baseline model. 
+ Because there are no SWH loads in the User model, we are expecting the Proposed and Baseline systems to match."`
+  -  if we got to this line without going to the rule assertion, it means that there is no equivalent system in the user model, or that the user model does not have SWH use.  Compare the proposed and baseline models using the function compare_swh_dist_systems_and_components.  This function returns a list of errors.  If the list of errors has a length of 0, then the comparison encountered no issues: `user_baseline_comparison = compare_swh_dist_systems_and_components(P_RMD,B_RMD,user_proposed_str,swh_dist_system_id)`
+  
+  - now go to rule assertion: `GO TO RULE ASSERTION`
+    
+
+## Rule Assertion: 
+- Case1: if the length of either user_proposed_comparison or user_baseline comparison is 0, then the rule passes: `if (len(user_proposed_comparison) == 0) || (len(user_baseline_comparison) == 0): PASS`
```
JacksonJ-KC, line 58: wouldn't we want the length of both (not either) of these lists to be 0?

docs/section11/Rule_11-1.md
KarenWGard: for each SWH Distribution system ID, yes.

docs/section11/Rule_11-1.md
KarenWGard: yes! good catch

docs/section11/Rule_11-1.md
KarenWGard: can you elaborate on why it's confusing?

JacksonJ-KC (approved):

claperle (approved):

PR #867, opened 2022-12-18
Create get_fan_system_object_supply_return_exhaust_relief_total_kW_CF…

claperle:
…M.md


docs/ruleset_functions/get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM.md
```diff
@@ -0,0 +1,76 @@
+# get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM
+
+**Description:** Get the supply, return, exhaust, and relief total fan power and CFM associated with a fan system object.   The function returns a dictionary that saves the supply, return, exhaust, and relief fan power has a list and the saves the supply, return, exhaust, and relief cfm as a list {"Fan_Power": [supply fan power kW, return fan power kW, exhaust fan power kW, relief fan power kW], "Fan_CFM": [supply fan power CFM, return fan power CFM, exhaust fan power CFM, relief fan power CFM]}. Values will be equal to zero where not defined for a fan system. For CFM values this function assumes that all fans associated with the supply_fans object are in parallel (i.e., if multiple fans the cfm is additive.).
```
mkarpman, line 3: Should we add a note to schema to aggregate fans of the same type so that there is always single supply, return, exhaust and relief fan per fan system?

mkarpman: Shouldn't we also account for Terminal.fan flow and power?

mkarpman (approved):

docs/ruleset_functions/get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM.md
claperle: If more than 1 fan then outcome for pass or fail include note that it was assumed that the fans are in parallel and that the cf can be added.

claperle: I added a check on the qty of fans and for pressure drop consistency. Would like to discuss on the call.

docs/ruleset_functions/get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM.md
```diff
@@ -0,0 +1,140 @@
...
+
```
mkarpman, line 4: Haven't we decided to have "undertermined" outcome if there is more than one supply, return or relief fan in the baseline? I think we should also assume that when there are multiple exhaust fans they are zonal fans.  This function may stay as is though because the corresponding logic may be implemented in the RDS.

docs/ruleset_functions/get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM.md
```diff
@@ -0,0 +1,140 @@
...
+- Reset total fan power to zero: `total_fan_power = 0`  
+- Reset total cfm to zero: `total_fan_cfm = 0`  
+- Reset fan_qty equal to 0 (counts the quantity of fans): `fan_qty = 0`  
+- Reset pressure drop identical across fans: `pressure_drop_identical = true`  
+- Reset pressure drop defined for all fans: `pressure_drop_defined_for_all_fans = true`  
+- Check if there are any fan objects associated with the fan system's supply fans:`if len(fan_system_obj.supply_fans) != 0 and fan_system_obj.supply_fans != Null:`   
+  - Loop through the supply fan: `for fan_obj in fan_system_obj.supply_fans:`  
+    - Set the fan_elec_power equal to zero: `fan_elec_power = 0`
+    - Set the fan_cfm equal to zero: `fan_cfm = 0`
+    - Get the fan objects total kW: `fan_elec_power = get_fan_object_electric_power(P_RMI, fan_obj)`
```
mkarpman, line 27: Why is it referencing P_RMI?

claperle: @mkarpman Revised to make it non specific. Thanks.

PR #870, opened 2022-12-19
Rds/cml/19 23

claperle:


supriyagoel (approved):
very clear and easy to understand.
Since there are 3 ways zone fan power could be calculated, it  would help to add a short description of the 3 methods to the description section
'This function first identifies if a HVAC system is serving more than one zone, in which case the fan power is apportioned to the zone based on the fraction of airflow to that zone. For single zone systems, the function sums up the airflow for all HVAC systems that might be serving the zone. For systems defined at the terminal, such as FPFCU, it sums up the fan power specified at the terminal' 
Please feel free to correct

docs/ruleset_functions/get_zone_supply_return_exhaust_relief_terminal_fan_power_dict.md
```diff
@@ -0,0 +1,85 @@
...
+        - Check if there is a fan system associated with the hvac system (if not then assume the fan is defined at the terminal unit like it is for a four pipe fan coil units and go to Else): `if hvac.fan_system != Null:`  
+            - Get list of zones served by the hvac system: `zone_list_hvac_sys_x = dict_of_zones_and_terminal_units_served_by_hvac_sys_x[hvac.id]["ZONE_LIST"]`  
+            - Check if the hvac system serves more than one zone, if it does then carry on, if not then go to Else: `if len(zone_list_hvac_sys_x) >1:`  
+                - Get list of terminal units served by the hvac system: `terminal_list_hvac_sys_x = dict_of_zones_and_terminal_units_served_by_hvac_sys_x[hvac.id]["ZONE_LIST"]["Terminal_Unit_List"]`  
+                - For each terminal unit served by the HVAC system: `for terminal in terminal_list_hvac_sys_x:`  
+                    - Get the primary air flow: `primary_air_flow_terminal = terminal.primary_airflow`  
+                    - Add to the total cfm across all terminals served by the HVAC system: `total_terminal_air_flow = total_terminal_air_flow + primary_air_flow_terminal`
```
supriyagoel, line 37: @claperle just to confirm, the total terminal airflow is the total airflow for the HVAC system, correct?

docs/section19/19-23.md
```diff
@@ -0,0 +1,77 @@
...
+    - Case 2: Else, then fail: `Else: outcome = "Fail" and raise_message "The calculated system fan power does appear to be distributed to the supply, return, exhaust, and relief fans in the same proportion as the proposed design for <insert hvac.id>. The expected modeled baseline supply, return, exhaust, and relief kW is <insert expected_baseline_fan_power_supply, expected_baseline_fan_power_return, expected_baseline_fan_power_exhaust, expected_baseline_fan_power_relief>, respectively."`
+
+
+**Notes/Questions:**  
+1. Should I be including Zone.non_mechanical_cooling_fan_power as supply fan power when figuring out the fan power proportions in the proposed?  This is calculated in the get_zone_supply_return_exhaust_relief_terminal_fan_power_dict function, not in this RDS directly.
```
mkarpman, line 74: I don't think so. See also comment up above regarding fan power boxes fan power.

docs/section19/19-23.md
```diff
@@ -0,0 +1,77 @@
...
+
+**Applicability Checks:** None   
+
+**Function Calls:**  
+1. get_dict_of_zones_and_terminal_units_served_by_hvac_sys()    
+2. get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM() 
+3. get_zone_supply_return_exhaust_relief_terminal_fan_power_dict()  
+4. match_data_element()  
+5. get_component_by_id()  
+
```
mkarpman, line 23: Fan power of the baseline VAV fan-powered boxes in systems 6 & 8 is prescribed and I think should not participate in the "allocation". Seems like the logic below does not exclude them since line 43 looks at all fans at zone level. I wonder then if we should also exclude zonal fans in the proposed design from the target percent allocation to have apples to apples comparision?

supriyagoel: good point. we need to include terminal fans as FPFCU use that nomenclature. To exclude PIUs, we might need to check afn configuration and exclude when it is series or parallel?

claperle: @mkarpman @supriyagoel Shouldn't we exclude terminal fans in the proposed since they have a separate fan power allocation in the baseline?

docs/ruleset_functions/get_zone_supply_return_exhaust_relief_terminal_fan_power_dict.md
claperle: @supriyagoel yes, exactly

docs/ruleset_functions/get_zone_supply_return_exhaust_relief_terminal_fan_power_dict.md
```diff
@@ -0,0 +1,85 @@
...
+                - Apportion the return kW to this zone based on the terminal unit primary flow for the zone and the total primary terminal unit flow across all zones that this HVAC system serves: `zone_total_return_fan_power = zone_total_return_fan_power + (hvac_sys_total_return_fan_power * (zone_primary_air_flow/total_terminal_air_flow))`  
+                - Apportion the exhaust kW to this zone based on the terminal unit primary flow for the zone and the total primary terminal unit flow across all zones that this HVAC system serves: `zone_total_exhaust_fan_power = zone_total_exhaust_fan_power + (hvac_sys_total_exhaust_fan_power * (zone_primary_air_flow/total_terminal_air_flow))`  
+                - Apportion the relief kW to this zone based on the terminal unit primary flow for the zone and the total primary terminal unit flow across all zones that this HVAC system serves: `zone_total_relief_fan_power = zone_total_relief_fan_power + (hvac_sys_total_relief_fan_power * (zone_primary_air_flow/total_terminal_air_flow))`  
+
+            - Else, the HVAC system serves a single zone: `Else:`
```
mkarpman, line 53: Wouldn't the logic above work for both single zone and multi-zone systems? Why do we need a separate routine for this case?

docs/ruleset_functions/get_zone_supply_return_exhaust_relief_terminal_fan_power_dict.md
```diff
@@ -0,0 +1,85 @@
...
+                - Add the exhaust kW to this zone: `zone_total_exhaust_fan_power = zone_total_exhaust_fan_power + hvac_sys_total_exhaust_fan_power`  
+                - Add the relief kW to this zone: `zone_total_relief_fan_power = zone_total_relief_fan_power + hvac_sys_total_relief_fan_power` 
+
+                - Get list of terminal units served by the hvac system: `terminal_list_hvac_sys_x = dict_of_zones_and_terminal_units_served_by_hvac_sys_x[hvac.id]["ZONE_LIST"]["Terminal_Unit_List"]`  
+                - For each terminal unit served by the HVAC system: `for terminal in terminal_list_hvac_sys_x:`  
+                    - Get the terminal unit fan power: `terminal_fan_power = get_fan_object_electric_power(RMI,terminal.fan)`  
+                    - Add to the total terminal kW associated with the zone: `zone_total_terminal_fan_power = zone_total_terminal_fan_power + terminal_fan_power`  
+
+        - Else, the hvac fan system equals Null meaning that the fan is defined at the terminal unit level like it is for a four pipe fan coil unit: `Else:`
```
mkarpman, line 70: Wouldn't we have fan systems both at the HVAC system level and terminal level for PVAV HVAC system?

docs/section19/19-23.md
mkarpman: OK, I realized that my original comment wasn't accurate as it was referring to fans at the zone level. So to correct myself, consider scenario when both the baseline and proposed designs have PVAV system. In that case, the total fan energy associated with a zone would include the prorated energy of the FanSystem of the associated HVAC zone level, and also energy from Terminal.Fan that would represent the fan in the VAV box.

claperle: @mkarpman OK, so you think the VAV terminal fans should be included for apportioning the baseline fan power allowance? This does not make sense to me since there is a separate fan power allowance for the VAV fan powered boxes in the baseline. How would that terminal fan power from the proposed be apportioned in the baseline then since the allowance is set? Perhaps the two of us should speak about this RDS outside of the group call to get on the same page.

docs/ruleset_functions/get_zone_supply_return_exhaust_relief_terminal_fan_power_dict.md
claperle: @mkarpman It would but to save run time (reduce the number of lines of code that need to be run) I separated out the scenarios.

docs/section19/19-23.md
claperle: We are on the same page. Resolved.

mkarpman (approved):

PR #873, opened 2022-12-20
Rds/cml/19 13

claperle:


docs/section19/19-13.md
```diff
@@ -0,0 +1,82 @@
...
+                    - Set the design_thermostat_cooling_setpoint equal to this zone's value: `design_thermostat_cooling_setpoint = zone_design_thermostat_cooling_setpoint`
+                    - Set the design_thermostat_heating_setpoint equal to this zone's value: `design_thermostat_heating_setpoint = zone_design_thermostat_heating_setpoint`     
+                - Else, means that this is not the first zone so need to compare this zone's values to the current values: `Else:`  
+                    - Check if the zone_design_thermostat_cooling_setpoint is less than the design_thermostat_cooling_setpoint: `if zone_design_thermostat_cooling_setpoint < design_thermostat_cooling_setpoint: design_thermostat_cooling_setpoint =  zone_design_thermostat_cooling_setpoint`                 
+                    - Check if the zone_design_thermostat_heating_setpoint is greater than the design_thermostat_heating_setpoint: `if zone_design_thermostat_heating_setpoint > design_thermostat_heating_setpoint: design_thermostat_heating_setpoint =  zone_design_thermostat_heating_setpoint`
```
weilixu, line 55: The above logic can be simplified to:
```python
design_thermostat_heating_setpoint = -infinity
design_thermostat_cooling_setpoint = +infinity
for zone in zone_list_hvac_sys_b:
    if zone.zone_design_thermostat_cooling_setpoint < design_thermostat_cooling_setpoint:
        design_thermostat_cooling_setpoint = zone.zone_design_thermostat_cooling
    if zone.zone_design_thermostat_heating_setpoint > design_thermostat_heating_setpoint:
        design_thermostat_heating_setpoint = zone.zone_design_thermostat_heating
```
This will guarantee to get the lowest cooling setpoint and highest heating setpoint in the list of zones.

claperle: Addressed

docs/section19/19-13.md
```diff
@@ -0,0 +1,75 @@
...
+**Rule ID:** 19-13             
+**Rule Description:** For baseline system types 1-8 and 11-13, system design supply airflow rates shall be based on a supply-air-to-room temperature set-point difference of 20°F or the minimum outdoor airflow rate, or the airflow rate required to comply with applicable codes or accreditation standards, whichever is greater. For systems with multiple zone thermostat setpoints, use the design set point that will result in the lowest supply air cooling set point or highest supply air heating set point.  
+
+Exceptions:
+1. For baseline system types 1-8 and 11-13, for systems serving laboratory spaces, airflow rate shall be based on a supply-air-to-room temperature set-point difference of 17°F or the required ventilation air or makeup air, whichever is greater.
+2. For baseline system types 1-8, 11-13, if the proposed design HVAC system airflow rate based on latent loads is greater than the design airflow rate based on sensible loads, then the same supply-air-to-room-air humidity ratio difference (gr/lb) used to calculate the proposed design airflow shall be used to calculate design airflow rates for the baseline building design.
```
mkarpman, line 9: I think exception #2 should be excluded from the RCT scope and notes as such in the body of 229P. Also, I recommend excluding references to system types 1-8 and 11-13 from the text of the exceptions. (E.g., it's weird to think that Sys 1 may serve a lab.) This would also align with 90.1 language.

docs/section19/19-13.md
```diff
@@ -0,0 +1,75 @@
...
+                    - Set the design setpoint equal to the zone design setpoint: `design_thermostat_cooling_setpoint = zone.zone_design_thermostat_cooling`  
+                - Check if the zone_design_thermostat_heating_setpoint is greater than the design_thermostat_heating_setpoint: `if zone.zone_design_thermostat_heating_setpoint > design_thermostat_heating_setpoint:`  
+                    - Set the design setpoint equal to the zone design setpoint: `design_thermostat_heating_setpoint = zone.zone_design_thermostat_heating`  
+            
+            Below checks for a delta T of 20F:
+            - For each zone in the list of zones served by the hvac system: `for zone in zones_list_hvac_sys_b:`  
+                - Reset zone has lab space boolean variable: `zone_has_lab_space = false`  
+                - For each space in zone: `for space in zone.spaces:`
+                    - Check if space is a lab based on lighting_space_type: `if space.lighting_space_type in ["LABORATORY_EXCEPT_IN_OR_AS_A_CLASSROOM"]: zone_has_lab_space = true`
```
mkarpman, line 56: So are we assuming that if there is at least one space that is a lab, the system design flow rate must be based on 17F instead of 20? 
We are excluding "Preschool through 12th grade, laboratory, and shop classrooms" space type, but I think it's OK.

docs/section19/19-13.md
```diff
@@ -0,0 +1,75 @@
...
+                        - Check if the zone serves at least one lab space: `if zone_has_lab_space == true:`  
+                            - Check if the delta temperature is NOT 17F for heating: `if design_heating_supply_air_temp_setpoint-zone_design_thermostat_heating_setpoint != 17:all_design_setpoints_delta_Ts_are_per_reqs = false`  
+                            - Check if the delta temperature is NOT 17F for cooling: `if zone_design_thermostat_cooling_setpoint - design_cooling_supply_air_temp_setpoin != 17:all_design_setpoints_delta_Ts_are_per_reqs = false`  
+                        - Else, zone does not serve a lab space: `Else:`  
+                            - Check if the delta temperature is NOT 20F for heating: `if design_heating_supply_air_temp_setpoint-zone_design_thermostat_heating_setpoint != 20:all_design_setpoints_delta_Ts_are_per_reqs = false`  
+                            - Check if the delta temperature is NOT 20F for cooling: `if zone_design_thermostat_cooling_setpoint - design_cooling_supply_air_temp_setpoin != 20:all_design_setpoints_delta_Ts_are_per_reqs = false`  
+
```
mkarpman, line 67: I think we also need to check for Fan.is_airflow_autosized=true

docs/section19/19-13.md
```diff
@@ -0,0 +1,75 @@
...
+        - **Rule Assertion:** 
+        - Case 1: If the supply-air-to-room temperature set-point difference equals 20°F or 17F per G3.1.2.8.1 including Exceptions 1, the fan is autosized and the FanSystem.minimum_airflow>FanSystem.Minimum_Outdoor_Airflow: `if all_design_setpoints_delta_Ts_are_per_reqs == true and are_all_hvac_sys_fan_objs_autosized(B_RMI, hvac) == True : outcome = "PASS"`
```
mkarpman, line 69: For labs, we are not checking the part that "or the required ventilation air or makeup air, whichever is greater". We have FanSystem.minimum_outdoor_airflow attribute that can potentially be used for this.

claperle: Assume what they specified in the proposed design is the ventilation required. if does not meet 20 or 17 (if lab) then see if the unit is modeled as min OA = supply CFM based on proposed OA assuming to meet code requirements.

docs/section19/19-13.md
claperle: That is the what I assumed here. Its not clear from the language what the correct answer is.

Also, if no space types are defined it defaults to no lab. Maybe I should add a caveat if no all space types are defined, "Pass unless the hvac system serves a lab"?

docs/section19/19-13.md
claperle: This is checked in Case 1 via a function.

docs/section19/19-13.md
claperle: This does not apply to just labs. Should we include this for all units?

docs/section19/19-13.md
claperle: Okay, removed from this rule and is a rule 19-16 which is marked as N for evaluated.

supriyagoel (approved):
this looks good to me

docs/section19/19-13.md
```diff
@@ -0,0 +1,75 @@
...
+                - Check if the zone_design_thermostat_cooling_setpoint is less than the design_thermostat_cooling_setpoint: `if zone.zone_design_thermostat_cooling_setpoint < design_thermostat_cooling_setpoint:` 
+                    - Set the design setpoint equal to the zone design setpoint: `design_thermostat_cooling_setpoint = zone.zone_design_thermostat_cooling`  
+                - Check if the zone_design_thermostat_heating_setpoint is greater than the design_thermostat_heating_setpoint: `if zone.zone_design_thermostat_heating_setpoint > design_thermostat_heating_setpoint:`  
+                    - Set the design setpoint equal to the zone design setpoint: `design_thermostat_heating_setpoint = zone.zone_design_thermostat_heating`  
+            
+            Below checks for a delta T of 20F:
+            - For each zone in the list of zones served by the hvac system: `for zone in zones_list_hvac_sys_b:`  
+                - Reset zone has lab space boolean variable: `zone_has_lab_space = false`
```
supriyagoel, line 53: @claperle question: the section above determines the lowest cooling and the highest heating thermostat setpoint in the scenario where zones might have differing setpoints
The section line 51 onwards again cycles through each terminal of an HVAC system and calculates the delta. Shouldn't we be using the lowest cooling and the highest heating that was established through lines 42-49 for this calculation? If there are differing setpoints, then this would be correct. If there are not differing setpoints then also this would be correct.

docs/section19/19-13.md
mkarpman: Sounds good on counting a zone with at least one lab space as a lab zone - this is consistent with the analogous computer room interpretation. Also OK with excluding lab classrooms. 

I am also OK with assuming that space is not a lab if space type is missing. Alternatively, we could make the opposite assumption for office building types (and perhaps also health-care clinic or hospital), but that will result in "undeterimined" flag for a lot of projects which is undesirable. Plus modeling higher delta T would result in more efficient baseline. I like your idea of adding note to Pass outcome.

claperle: All set, no changes needed.

docs/section19/19-13.md
claperle: Updated per discuss for 19-17 which is analogous.

mkarpman (approved):

PR #874, opened 2022-12-21
Rds/cml/19 14

claperle:


docs/section19/19-14.md
```diff
@@ -0,0 +1,71 @@
...
+                - Get list of the zone's supply, return, exhaust, and relief fan power modeled in the proposed: `modeled_fan_power_list_p = list(zone_supply_return_exhaust_relief_terminal_fan_power_dict[zone_p.id])`  
+                - Get the zone's total return fan kW, need to get the second element in the list of fan powers:`modeled_return_fan_power_p = modeled_fan_power_list_p[1]` 
+                - Get the zone's total relief fan kW, need to get the fourth element in the list of fan powers:`modeled_relief_fan_power_p = modeled_fan_power_list_p[3]` 
+                - Check if the total return fan power is greater than 0 in the proposed, if yes then set is_modeled_with_return_fan_in_proposed to true: `if modeled_return_fan_power_p > 0: is_modeled_with_return_fan_in_proposed = true`  
+                - Check if the total relief fan power is greater than 0 in the proposed, if yes then set is_modeled_with_relief_fan_in_proposed to true: `if modeled_relief_fan_power_p > 0: is_modeled_with_relief_fan_in_proposed = true`
```
mkarpman, line 49: I think we need to account for scenarios when, based on the above check, the baseline system must be modeled with both releif and return fans (i.e., some of the zones that it serves have relief fan in the proposed design, and some have return fan). I don't think the App G language is unclear about what to do in this case, so I'll start an exchange with the ECB. I see that you are interpreting it in Case 2 as requiring that both return and relief fans must be specified in the baseline. Perhaps this is reasonable.

docs/section19/19-14.md
```diff
@@ -0,0 +1,71 @@
...
+            - Check if is_modeled_with_return_fan_in_proposed == true, if so then set the modeled_cfm equal to the return cfm (second element in list of baseline hvac fan cfms): `if is_modeled_with_return_fan_in_proposed == true: modeled_cfm = fan_sys_CFM[1]`  
+            - Else, is_modeled_with_relief_fan_in_proposed == true, then set the modeled_cfm equal to the relief cfm (fourth element in list of baseline hvac fan cfms): `Else: modeled_cfm = fan_sys_CFM[3]`   
+
+            - **Rule Assertion:** 
+            - Case 1: If return or relief fans are specified in the proposed design (but not both) and the baseline building design is modeled with fans serving the same functions and sized for the baseline system supply fan air quantity less the minimum outdoor air, or 90% of the supply fan air quantity, whichever is larger then pass: `if (all(is_modeled_with_return_fan_in_proposed, is_modeled_with_relief_fan_in_proposed) == false) and (is_modeled_with_return_fan_in_proposed == true or is_modeled_with_relief_fan_in_proposed == true) and (modeled_cfm == max(supply_minus_OA_cfm,90_percent_supply_cfm)): outcome = "PASS"`
```
mkarpman, line 61: What does "all" means in this context? That it is true for ALL zones served by that system?

I don't follow the meaning of the first two conditions in the if clause. Can you please explain?

The last condition (is_modeled_with_return_fan_in_proposed == true or is_modeled_with_relief_fan_in_proposed == true) allows for schenarios when both are true, which I don't think the intent.

docs/section19/19-14.md
```diff
@@ -0,0 +1,71 @@
+# Section 19 - Rule 19-14             
+**Schema Version:** 0.0.23    
+**Mandatory Rule:** True    
+**Rule ID:** 19-14               
+**Rule Description:**  For baseline system types 1-8 and 11-13, if return or relief fans are specified in the proposed design, the baseline building design shall also be modeled with fans serving the same functions and sized for the baseline system supply fan air quantity less the minimum outdoor air, or 90% of the supply fan air quantity, whichever is larger.
+
```
mkarpman, line 6: I don't think eQUEST has "relief" fan type. Does E+? Can releif fan be modeled as exhaust fan?

docs/section19/19-14.md
```diff
@@ -0,0 +1,71 @@
...
+            - Create fan system object: `fan_sys_b = hvac.fan_system`  
+            - Get list from dictionary of the hvac system's fan system's supply, return, exhaust and relief fan CFM: `fan_sys_CFM = list(get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM(B_RMI,fan_sys_b)["Fan_Power"]["Fan_CFM"])`  
+            - Get the hvac system total supply fan cfm, need to get the first element in the list of fan cfms (this assumes that all fans associated with the supply_fans object are in parallel (i.e., if multiple fans the cfm is additive.):`hvac_sys_total_supply_fan_cfm_b = fan_sys_CFM[0]` 
+            - Calculate 90% of the supply air quantity: `90_percent_supply_cfm = 90% * hvac_sys_total_supply_fan_cfm_b`   
+            - Calculate the supply CFM minus the minimum OA CFM for the fan system: `supply_minus_OA_cfm = hvac_sys_total_supply_fan_cfm_b - fan_sys_b.minimum_outdoor_airflow`  
+            - Check if is_modeled_with_return_fan_in_proposed == true, if so then set the modeled_cfm equal to the return cfm (second element in list of baseline hvac fan cfms): `if is_modeled_with_return_fan_in_proposed == true: modeled_cfm = fan_sys_CFM[1]`
```
mkarpman, line 57: Perhaps we need modeled_releif_CFM and modeled_return_CFM in case both are present.

docs/section19/19-14.md
```diff
@@ -0,0 +1,71 @@
...
+            - Case 2: Else if both return or relief fans are specified in the proposed design then outcome is UNDETERMINED: `elif all(is_modeled_with_return_fan_in_proposed, is_modeled_with_relief_fan_in_proposed) == true: outome = "UNDETERMINED" and raise_message "Return and relief fans both appear to be modeled in the proposed design, conduct manual check that the baseline building design is modeled with fans serving the same functions and sized for the baseline system supply fan air quantity less the minimum outdoor air, or 90% of the supply fan air quantity, whichever is larger per Section G3.1.2.8.1."`
```
mkarpman, line 62: In this case, should we check that baseline has both return and relief fans and that their total airflow meets the requirement?

docs/section19/19-14.md
claperle: Discuss next step on call.

docs/section19/19-14.md
claperle: @mkarpman 
all() only returns true if both conditions are true so ensures that the last condition you mention above is not the case. 

This is the code: (all(is_modeled_with_return_fan_in_proposed, is_modeled_with_relief_fan_in_proposed) == false) so if this equals false then both of these are not true. Does this work or am I missing something?

docs/section19/19-14.md
claperle: eQuest has options for all 3. Discuss on call.
image

docs/section19/19-14.md
claperle: Do projects ever have both?

docs/section19/19-14.md
claperle: Adjust depending on above discussions.

claperle: For the purpose of this RDS, whenever fans are included in the proposed should also be included in the baseline. 

Total cfm (return + relief) is the 90%. Check for total and not distribution. 

Exhaust cfm same as in proposed (ask Mike's about this).

docs/section19/19-14.md
claperle: See below, waiting.

docs/section19/19-14.md
claperle: Updated per everything except exhaust because Mike R agreed. This rule does not discuss exhaust, shall I add another rule? If so, not sure which section of App G to reference since its not in the Standard?

mkarpman (approved):

supriyagoel (approved):

PR #656, opened 2022-09-21
Create 19-12.md

claperle:

PR #875, opened 2022-12-21
Rds/cml/19 7

claperle:


docs/section19/19-7.md
```diff
@@ -0,0 +1,87 @@
...
+    
+    - For each zone associated with the HVAC system: `for zone in zones_list_hvac_sys_b:`  
+        - For each space associated with the zone, check if any spaces are NOT labs: `for space in zone.spaces:`
+            - Check if the lighting space type is NOT a lab: `if space.lighting_space_type not in [LABORATORY_EXCEPT_IN_OR_AS_A_CLASSROOM]: hvac_system_serves_only_labs = false`    
+        
+        - Add each zone's OA CFM schedule to the list of zone OA CFM schedules for the baseline: `zone_OA_CFM_list_of_schedules_b[counter] = get_min_OA_CFM_sch_zone(B_RMI, zone.id)`  
+        - Get the analogous zone object from the proposed RMI (if this is not strictly the correct syntax please advise Weili): `zone_p = get_component_by_id(P_RMI,match_data_element(P_RMI,Zone,zone.id))`    
+        - Add each zone's OA CFM schedule to the list of zone OA CFM schedules for the proposed: `zone_OA_CFM_list_of_schedules_p[counter] = get_min_OA_CFM_sch_zone(P_RMI, zone_p.id)`  
+
+        - For each terminal unit serving the zone, check if DCV was modeled: `for terminal in zone.terminals:`
```
supriyagoel, line 52: @claperle this might already be be the case but just confirming- in some cases a zone might be served by 2 HVAC systems, example- DOOAS + HP. in this case, the terrminals associated with the DOAS would have DCV but not the heat pump. 
I think this is already addressed by looking through each terminal serving a zone, but line 44- _For each zone associated with the HVAC system: `for zone in zones_list_hvac_sys_b_:`  makes me think we're looping through HVAC systems and not through zones.

claperle: @supriyagoel I think the way it is setup works because the logic is cycling through the **baseline** HVAC systems, identifying all zones served, and then looking at ALL of the terminals associated with each of these zones in the proposed design (so it would catch all HVACs serving the zones in the proposed). If any terminals or their associated HVAC systems have DCV in the proposed design then the RDS considers DCV to have been modeled in the proposed.  Please let me know if I am missing something.

supriyagoel: yes, that makes sense. No, I dont think you're missing anything.

docs/section19/19-7.md
```diff
@@ -0,0 +1,87 @@
...
+3. Where the minimum outdoor air intake flow in the proposed design is provided in excess of the amount required by the building code or the rating authority, the baseline building design shall be modeled to reflect the greater of that required by either the rating authority or the building code and will be less than the proposed design. 
+4. For baseline systems serving only laboratory spaces that are prohibited from recirculating return air by code or accreditation standards, the baseline system shall be modeled as 100% outdoor air.
```
mkarpman, line 10: It seems that this too should be a separate RMD?

docs/section19/19-7.md
```diff
@@ -0,0 +1,87 @@
...
+    - Check if schedules mismatch for any hours of the year, if so set boolean to false: `for p to Range(8760):`  
+        - Check if the baseline and proposed schedules mismatch: `if aggregated_min_OA_schedule_across_zones_b[p] != aggregated_min_OA_schedule_across_zones_p[p]: OA_CFM_schedule_match = false`   
+
+    - Sum OA CFM across baseline schedule to get the total for comparison purposes in the rule assertions: `modeled_baseline_total_zone_min_OA_CFM = sum(aggregated_min_OA_schedule_across_zones_b)`  
+    - Sum OA CFM across proposed schedule to get the total for comparison purposes in the rule assertions: `modeled_proposed_total_zone_min_OA_CFM = sum(aggregated_min_OA_schedule_across_zones_p)`  
+
+
+
+    - **Rule Assertion:** 
+    - Case 1: If the modeled baseline and proposed minimum OA cfm values are identical then pass: `if OA_CFM_schedule_match == true: outcome = "PASS"`
```
mkarpman, line 76: What if spaces are over-ventilated and proposed should actually be less than baseline?

docs/section19/19-7.md
```diff
@@ -0,0 +1,87 @@
...
+    - Case 4: Else if the modeled baseline minimum OA cfm are greater than the modeled proposed minimum OA cfm and the hvac system only serves lab spaces (Exception 4) the outcome is UNDETERMINED: `elif modeled_baseline_total_zone_min_OA_CFM > modeled_proposed_total_zone_min_OA_CFM and hvac_system_serves_only_labs == true: outome = "UNDETERMINED" and raise_message "For <insert hvac.id> the modeled baseline minimum ventilation system outdoor air intake flow CFM appears higher than the minimum ventilation system outdoor air intake flow CFM modeled in the proposed design. It appears as though G3.1.2.5 Exception 4 may be applicable. This exception states that for baseline systems serving only laboratory spaces that are prohibited from recirculating return air by code or accreditation standards, the baseline system shall be modeled as 100% outdoor air. A manual check for this exception is recommended otherwise fail."`
```
mkarpman, line 79: I think we need another RDS to check that baseline is 100% OA

docs/section19/19-7.md
```diff
@@ -0,0 +1,87 @@
...
+    - Case 5: Else if the modeled baseline minimum OA cfm are less than the modeled proposed minimum OA cfm the outcome is FAIL: `elif modeled_baseline_total_zone_min_OA_CFM < modeled_proposed_total_zone_min_OA_CFM: outome = "FAIL" and raise_message "For <insert hvac.id> the modeled minimum ventilation system outdoor air intake flow CFM appears lower than the minimum ventilation system outdoor air intake flow CFM modeled in the proposed design. Fail unless G3.1.2.5 Exception 3 is applicable. This exception states that where the minimum outdoor air intake flow in the proposed design is provided in excess of the amount required by the building code or the rating authority, the baseline building design shall be modeled to reflect the greater of that required by either the rating authority or the building code and will be less than the proposed design. "`
```
mkarpman, line 80: If we decide to differentiate mistakes that are likely to lead to more conservative outcomes, this one may be a candidate.

docs/section19/19-7.md
```diff
@@ -0,0 +1,87 @@
+# Section 19 - Rule 19-7               
+**Schema Version:** 0.0.23    
+**Mandatory Rule:** True    
+**Rule ID:** 19-7                 
+**Rule Description:**  Minimum ventilation system outdoor air intake flow shall be the same for the proposed design and baseline building design except when any of the 4 exceptions defined in Section G3.1.2.5 are met.  
+Exceptions
+1. When modeling demand control ventilation in the proposed design in systems with outdoor air capacity less than or equal to 3000 cfm serving areas with an average design capacity of 100 people per 1000 ft^2 or less.
```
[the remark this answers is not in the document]
claperle: This is already a separate RDS, 19-10, https://github.com/pnnl/ruleset-checking-tool/pull/656

docs/section19/19-7.md
claperle: Proposed less than baseline?

docs/section19/19-7.md
```diff
@@ -0,0 +1,87 @@
...
+    - Sum OA CFM across proposed schedule to get the total for comparison purposes in the rule assertions: `modeled_proposed_total_zone_min_OA_CFM = sum(aggregated_min_OA_schedule_across_zones_p)`  
+
+
+
+    - **Rule Assertion:** 
+    - Case 1: If the modeled baseline and proposed minimum OA cfm values are identical then pass: `if OA_CFM_schedule_match == true: outcome = "PASS"`  
+    - Case 2: Else if the modeled baseline minimum OA cfm are greater than the modeled proposed minimum OA cfm and a zone served by the HVAC in the baseline has demand controlled ventilation modeled in the proposed (Exception 1) then the outcome is UNDETERMINED: `elif modeled_baseline_total_zone_min_OA_CFM > modeled_proposed_total_zone_min_OA_CFM and was_DCV_modeled_baseline == false and was_DCV_modeled_proposed == true: outome = "UNDETERMINED" and raise_message "For <insert hvac.id> the baseline modeled minimum ventilation system outdoor air intake flow CFM appears higher than the minimum ventilation system outdoor air intake flow CFM modeled in the proposed design. It appears as though G3.1.2.5 Exception 1 may be applicable. A manual check for this exception is recommended otherwise fail."`
```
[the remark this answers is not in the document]
claperle: We discussed this previously, sometimes modelers model DCV with the schedule. Up to group.

docs/section19/19-7.md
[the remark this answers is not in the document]
claperle: Flag if rates and schedules are not equal between baseline and proposed even when there is DCV. Should be doing both, modeling DCV and different rates and schedules.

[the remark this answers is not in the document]
claperle: Done

docs/section19/19-7.md
claperle: Need to implement

docs/section19/19-7.md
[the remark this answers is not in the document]
claperle: Addressed

docs/section19/19-7.md
claperle: Need to address

docs/section19/19-7.md
claperle: Done

docs/section19/19-7.md
```diff
@@ -0,0 +1,84 @@
...
+    - Reset was_DCV_modeled_baseline boolean variable to false: `was_DCV_modeled_baseline = false`  
+    - Reset was_DCV_modeled_proposed boolean variable to false: `was_DCV_modeled_proposed = false`  
+    - Reset zone_air_distribution_effectiveness_greater_than_1 boolean to false: `zone_air_distribution_effectiveness_greater_than_1 = false`  
+    - Reset hvac_system_serves_only_labs boolean variable to true: `hvac_system_serves_only_labs = true`  
+    - Check if demand controlled ventilation was modeled at the fan_system: `if hvac.fan_system.demand_control_ventilation_control == true: was_DCV_modeled_baseline= true`  
+    - **Applicability Check 1** - Check that the HVAC system does not only serve labs (covered by 19-39)
+    - Get list of zones served by the hvac system: `zones_list_hvac_sys_b = dict_of_zones_and_terminal_units_served_by_hvac_sys_b[hvac.id]["ZONE_LIST"]`  
+        - For each space associated with the zone, check if any spaces are NOT labs: `for space in zone.spaces:`
+            - Check if the lighting space type is NOT a lab: `if space.lighting_space_type not in [LABORATORY_EXCEPT_IN_OR_AS_A_CLASSROOM]: hvac_system_serves_only_labs = false`    
+
```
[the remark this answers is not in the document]
claperle: My only concern with that is if it happens often then all outcomes in a project could be undetermined. Maybe we could have separate pass and fail outcomes for those scenarios saying fail unless lab and pass with caveats about labs??

docs/section19/19-7.md
```diff
@@ -0,0 +1,84 @@
+# Section 19 - Rule 19-7               
+**Schema Version:** 0.0.24      
+**Mandatory Rule:** True    
+**Rule ID:** 19-7                 
+**Rule Description:**  Minimum ventilation system outdoor air intake flow shall be the same for the proposed design and baseline building design except when any of the 4 exceptions defined in Section G3.1.2.5 are met.  
+Exceptions included in this RDS:
+2. When designing systems in accordance with Standard 62.1, Section 6.2, “Ventilation Rate Procedure,” reduced ventilation airflow rates may be calculated for each HVAC zone in the proposed design with a zone air distribution effectiveness (Ez) > 1.0 as defined by Standard 62.1, Table 6-2. Baseline ventilation airflow rates in those zones shall be calcu-lated using the proposed design Ventilation Rate Procedure calculation with the following change only. Zone air distribution effectiveness shall be changed to (Ez) = 1.0 in each zone having a zone air distribution effectiveness (Ez) > 1.0. Proposed design and baseline build-ing design Ventilation Rate Procedure calculations, as described in Standard 62.1, shall be submitted to the rating authority to claim credit for this exception. M
+
```
mkarpman, line 8: I think rule description should be uupdated to say that it also covers exception 1 (DCV) since in this case we exect min. OA rates in the baseline and proposed design to still be the same but the DCV control to be modeled.

docs/section19/19-7.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+        - **Rule Assertion:** 
+        - Case 1: If the modeled baseline and proposed minimum OA cfm values are identical then pass: `if OA_CFM_schedule_match == true: outcome = "PASS"`  
+        - Case 2: Else if the modeled baseline minimum OA cfm are greater than the modeled proposed minimum OA cfm and a zone served by the HVAC in the baseline has demand controlled ventilation modeled in the proposed (Exception 1) then the outcome is Fail: `elif modeled_baseline_total_zone_min_OA_CFM > modeled_proposed_total_zone_min_OA_CFM and was_DCV_modeled_baseline == false and was_DCV_modeled_proposed == true: outome = "FAIL" and raise_message "For <insert hvac.id> the baseline modeled minimum ventilation system outdoor air intake flow CFM appears higher than the minimum ventilation system outdoor air intake flow CFM modeled in the proposed design. It appears that demand-controlled ventilation was modeled in the proposed and not the baseline model and that demand-controlled ventilation may be double accounted for in the model (per the HVAC controls and via reduced OA CFM rates in the proposed)."`
```
mkarpman, line 75: I think you should add condition "and zone_air_distribution_effectiveness_greater_than_1 == false" before ":outcome="FAIL""

docs/section19/19-7.md
```diff
@@ -0,0 +1,84 @@
...
+        - Case 4: Else if the modeled baseline minimum OA cfm are less than the modeled proposed minimum OA cfm the outcome is FAIL: `elif modeled_baseline_total_zone_min_OA_CFM < modeled_proposed_total_zone_min_OA_CFM: outome = "FAIL" and raise_message "For <insert hvac.id> the modeled minimum ventilation system outdoor air intake flow CFM appears lower than the minimum ventilation system outdoor air intake flow CFM modeled in the proposed design. Fail unless G3.1.2.5 Exception 3 is applicable. This exception states that where the minimum outdoor air intake flow in the proposed design is provided in excess of the amount required by the building code or the rating authority, the baseline building design shall be modeled to reflect the greater of that required by either the rating authority or the building code and will be less than the proposed design. "`
```
mkarpman, line 77: I would make it UNDERTERMINED instead of FAIL

docs/section19/19-7.md
```diff
@@ -0,0 +1,84 @@
...
+        - Case 5: Else, fail: `Else: outcome = "Fail" and raise_message "For <insert hvac.id> the modeled baseline minimum ventilation system outdoor air intake flow CFM appears higher than the minimum ventilation system outdoor air intake flow CFM modeled in the proposed design which does not meet the requirements of Section G3.1.2.5."`
```
mkarpman, line 78: I would not use "appear" in the notes because we have established it to be higher, right? But no strong feelings.

docs/section19/19-7.md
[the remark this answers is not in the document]
claperle: Go with CML suggestion

docs/section19/19-7.md
claperle: Leave it as is per call

docs/section19/19-7.md
claperle: Change to UNDETERMINED

docs/section19/19-7.md
claperle: removed

docs/section19/19-7.md
claperle: Done

docs/section19/19-7.md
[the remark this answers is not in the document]
claperle: I may have made some mistakes in implementing this. But I think I captured what we decided on.

supriyagoel (approved):
looks good!

PR #878, opened 2022-12-30
Rs/wx/report module updates

weilixu:


yunjoonjung-PNNL (approved):
Thank you for making these updates! Looks great to me. I'll also make the corresponding updates on the summary report module (next week).

PR #880, opened 2023-01-01
Rds/kjw/rule 23 6

KarenWGard:
updated to use `get_dict_of_zones_and_terminal_units_served_by_hvac_sys `


docs/section23/Rule23-6.md
```diff
@@ -37,19 +39,17 @@
     - add the hvac_ids for this system type to the list of eligible_systems: `eligible_systems += baseline_hvac_system_dict[baseline_system_type]`
 
 
-- For each zone in B_RMR: `for zone_b in B_RMI...zones:`
+- get a dictionary of hvac systems, their zones and terminal units using the function get_dict_of_zones_and_terminal_units_served_by_hvac_sys : `dict_of_zones_and_terminal_units_served_by_hvac_sys = get_dict_of_zones_and_terminal_units_served_by_hvac_sys(B_RMI)`
+- loop through the hvac systems in the dictionary: `for hvac_b_id in dict_of_zones_and_terminal_units_served_by_hvac_sys:`
+  - Check if HVAC system is type 6, 8, 8a, 6b, 8b: `if hvac_b_id in eligible_systems:`
+    - we do the rule check for each terminal: `for terminal_b_id in dict_of_zones_and_terminal_units_served_by_hvac_sys[hvac_b_id]["Terminal_Unit_List"]:`
+      - get the terminal unit associated with the id: `terminal_b = match_data_element(B_RMI,Terminal,terminal_b_id)`
```
weilixu, line 47: I may misunderstood, if the applicability check already filtered out the applicable HVAC systems, why do we need to check for eligibility again at line 44?
We can do:
`dict_of_zones_and_terminal_units_served_by_hvac_sys[hvac_b_id]`?

KarenWGard: you're right - eliminated

docs/section23/Rule23-6.md
```diff
@@ -37,19 +39,16 @@
...
 
-    - Get HVAC system serving terminal: `hvac_b = terminal_b.served_by_heating_ventilating_air_conditioning_system`
-  
-      - Check if HVAC system is type 6, 8, 8a, 6b, 8b: `if hvac_b.id in eligible_systems:`
+      - Case 1: For each terminal that is served by HVAC system that is Type-6, 8, 8a, 6b, 8b, if fan in parallel VAV-powered box is sized for 50% of the peak design primary air (from the VAV air-handling unit) flow rate (CFM) and is modeled with 0.35W/cfm fan power: `if ( terminal_b.fan.design_airflow == terminal_b.primary_airflow * 0.5 ) AND ( terminal_b.fan.design_electric_power == 0.35 * terminal_b.fan.design_airflow ): PASS`
 
-        **Rule Assertion:**
-
-        - Case 1: For each terminal that is served by HVAC system that is Type-6, 8, 8a, 6b, 8b, if fan in parallel VAV-powered box is sized for 50% of the peak design primary air (from the VAV air-handling unit) flow rate (CFM) and is modeled with 0.35W/cfm fan power: `if ( terminal_b.fan.design_airflow == terminal_b.primary_airflow * 0.5 ) AND ( terminal_b.fan.design_electric_power == 0.35 * terminal_b.fan.design_airflow ): PASS`
-
```
claperle, line 51: This does not appear to account for the fact that the power of the fan may be capture in the Fan object other than via design_electric_power. Is it reasonable to expect that the Fan.specification_method will always be Simple for the baseline?

supriyagoel: discussed during 06.04.2025 RDS Call and @KarenWGard will update this to use the function to calculate fan power when detailed approach is used (instead of W/CFM)

KarenWGard: this is looking at individual fans, so will use `get_fan_object_electric_power` function and fan flow

docs/section23/Rule23-6.md
```diff
@@ -37,19 +39,16 @@
...
+  - we do the rule check for each terminal: `for terminal_b_id in dict_of_zones_and_terminal_units_served_by_hvac_sys[hvac_b_id]["Terminal_Unit_List"]:`
+    - get the terminal unit associated with the id: `terminal_b = match_data_element(B_RMI,Terminal,terminal_b_id)`
 
-  - For each terminal in zone: `for terminal_b in zone_b.terminals:`
+**Rule Assertion:**
 
-    - Get HVAC system serving terminal: `hvac_b = terminal_b.served_by_heating_ventilating_air_conditioning_system`
-  
-      - Check if HVAC system is type 6, 8, 8a, 6b, 8b: `if hvac_b.id in eligible_systems:`
+      - Case 1: For each terminal that is served by HVAC system that is Type-6, 8, 8a, 6b, 8b, if fan in parallel VAV-powered box is sized for 50% of the peak design primary air (from the VAV air-handling unit) flow rate (CFM) and is modeled with 0.35W/cfm fan power: `if ( terminal_b.fan.design_airflow == terminal_b.primary_airflow * 0.5 ) AND ( get_fan_object_electric_power(B_RMD, terminal_b.fan.design_electric_power) == 0.35 * terminal_b.fan.design_airflow ): PASS`
```
claperle, line 49: get_fan_object_electric_power should be sent terminal_b.fan (i.e., remove .design_electric_power)

docs/section23/Rule23-6.md
```diff
@@ -37,19 +39,16 @@
     - add the hvac_ids for this system type to the list of eligible_systems: `eligible_systems += baseline_hvac_system_dict[baseline_system_type]`
 
 
-- For each zone in B_RMR: `for zone_b in B_RMI...zones:`
+- get a dictionary of hvac systems, their zones and terminal units using the function get_dict_of_zones_and_terminal_units_served_by_hvac_sys : `dict_of_zones_and_terminal_units_served_by_hvac_sys = get_dict_of_zones_and_terminal_units_served_by_hvac_sys(B_RMI)`
+- loop through the hvac systems in the dictionary: `for hvac_b_id in dict_of_zones_and_terminal_units_served_by_hvac_sys:`
+  - we do the rule check for each terminal: `for terminal_b_id in dict_of_zones_and_terminal_units_served_by_hvac_sys[hvac_b_id]["Terminal_Unit_List"]:`
+    - get the terminal unit associated with the id: `terminal_b = match_data_element(B_RMI,Terminal,terminal_b_id)`
```
claperle, line 45: My understanding of match_data_element was to retrieve matching data elements across P_RMI and B_RMI. I think you mean to use the function that takes and ID and return the object.

supriyagoel: @KarenWGard based on @claperle comment, this should probably be 'get_component_by_id'

supriyagoel (changes requested):
small edits as requested and Case 1 needs to be updated to use get_fan_object_electric_power

claperle (approved):

supriyagoel (approved):

PR #885, opened 2023-01-05
Rds/cml/function get most used weekday hourly schedule

claperle:


docs/ruleset_functions/get_most_used_weekday_hourly_schedule.md
```diff
@@ -0,0 +1,23 @@
...
+
+**Description:** Get the most used weekday hourly schedule from an annual 8760 schedule as list of hourly values for a 24 hour period.
+
+**Inputs:**  
+- **B-RMI,P-RMI**: To calculate the the most used weekday hourly schedule from an annual 8760 schedule.   
+- **RMD**: To obtain the data elements associated with the Calendar object (day of week for Jan 1st, is leap year, and has daylight savings time). 
+- **schedule_obj**: The schedule object sent to this function.  
+
+**Returns:**  
+- **get_most_used_weekday_hourly_schedule**: The function calculates and returns the most used weekday hourly schedule from an annual 8760 schedule as list of hourly values for a 24 hour period.
```
mkarpman, line 11: Did we agree on the logic for this? Would it based on the 24 matcing hourly values? For example, if the sum of fractions is the same between the two schedules but profiles are different, the schecules will be considered different for the purpose of this function.

claperle: Add "Based on 24 matching hourly values (not just the sum across the day)."

mkarpman (approved):

supriyagoel (approved):

PR #888, opened 2023-01-06
Complete the implementation

weilixu:


yunjoonjung-PNNL (approved):
Looks great to me. Thank you!

rct229/ruleset_functions/baseline_system_type_compare.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def baseline_system_type_compare(system_type, target_system_type, exact_match=True):
+    """
+
+    Parameters
+    ----------
+    system_type: String the enum indicating the hvac system type example: SYS_5 or SYS_8c.  This will usually be the value given by the function get_baseline_system_types
+    target_system_type: String the enum indicating the target system type, will usually be SYS_# without a further number or letter modifier
+    exact_match: Bool TRUE or FALSE.  if exact_match is TRUE, then the system type must match exactly the enum given by hvac_system_type.  If false, an approximate match will return true.  This would be used in the case where the user wants any hvac system ot type 8, without having to type in 8a, 8b, 8c.  In this case, SYS_8 would be passed to the function and the function would return TRUE for any system 8, regardless of it's 8a, 8b, or 8c
```
charliepnnl, line 14: Should `hvac_system_type` be `target_system_type`?
Also, should the phrase "the system type" be replaced with `system_type`?

rct229/ruleset_functions/baseline_systems/baseline_system_util.py
```diff
@@ -44,6 +44,22 @@ class HVAC_SYS:
     SYS_13A = "Sys-13a"
     UNMATCHED = "Not_Sys"
 
+    HVAC_SYSTEM_TYPE_DICTIONARY = {
```
charliepnnl, line 47: It might be cleaner to make this a separate value outside of the HVAC_SYS class.

rct229/ruleset_functions/baseline_system_type_compare.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Returns
+    -------
+    True or False, indicating whether the hvac system type matches the hvac_system_type.
+    """
+
+    # A list of the attribute values from the HVAC_SYS class
```
charliepnnl, line 21: Should this be 
```
# A list of the HVAC systems in HVAC_SYS
```

rct229/ruleset_functions/baseline_system_type_compare.py
```diff
@@ -0,0 +1,40 @@
...
+    hvac_sys_list = [
```
charliepnnl, line 22: I think this can be simplified slightly if `HVAC_SYSTEM_TYPE_DICTIONARY` is pulled out of `HVAC_SYS` as described below.

rct229/ruleset_functions/baseline_system_type_compare.py
```diff
@@ -0,0 +1,40 @@
...
+        i[1]
+        for i in inspect.getmembers(HVAC_SYS)
+        if type(i[0]) is str and i[0].startswith("SYS")
+    ]
+    if system_type not in hvac_sys_list:
```
charliepnnl, line 27: It might be a little cleaner to make these two if statements assertions such as
```
assert system_type in hvac_sys_list, f"{system_type} does not match any baseline HVAC system type"
```

Notice that I also removed the word "to" in the error message.

rct229/ruleset_functions/baseline_system_type_compare.py
```diff
@@ -0,0 +1,40 @@
...
+        )
+    if target_system_type not in hvac_sys_list:
+        raise RCTException(
+            f"{target_system_type} does not match to any baseline HVAC system type"
+        )
+
+    return (
+        system_type == target_system_type
+        if exact_match
+        else target_system_type in HVAC_SYS.HVAC_SYSTEM_TYPE_DICTIONARY[system_type]
```
charliepnnl, line 39: I don't think there is any guarantee that `system_type` is a key of `HVAC_SYSTEM_TYPE_DICTIONARY`, nor does the documentation indicate that is should be.

How should this function behave?  If `exact_match` is False, should `SYS_1A` and `SYS_1B` match? If that is the desired behavior, then using regular expressions might be the easiest way.

charliepnnl (approved):
Approved with a few comments that need to be addressed.

PR #894, opened 2023-01-11
Ruletest 23-6 Test JSON

jugonzal07:
first cut at 23-6 test


yunjoonjung-PNNL (commented):
For system 6 and 6b, we need ```fan_configuration``` key under the ```terminals```. Could you add it? FYI, the ```sys_type``` Excel template on Onedrive was updated to include the ```fan_configuration``` so, I believe you can copy it from the ```sys_type``` template.

jugonzal07:
> For system 6 and 6b, we need `fan_configuration` key under the `terminals`. Could you add it? FYI, the `sys_type` Excel template on Onedrive was updated to include the `fan_configuration` so, I believe you can copy it from the `sys_type` template.

Addressed.

yunjoonjung-PNNL (approved):

weilixu (approved):

PR #907, opened 2023-01-16
RS/YJ/Rule 23-6 and 7

yunjoonjung-PNNL:
This PR addresses Rule 23-6 and 7 development. Thank you!


rct229/rules/section23/section23rule6.py
```diff
@@ -0,0 +1,105 @@
...
+        applicable_hvac_sys_ids = data["applicable_hvac_sys_ids"]
+
+        return context_item.baseline["id"] in applicable_hvac_sys_ids
+
+    class TerminalRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section23Rule6.TerminalRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                required_fields={
+                    "$": ["primary_airflow"],
```
charliepnnl, line 76: Do we also need to require the "fan" field?

rct229/rules/section23/section23rule6.py
```diff
@@ -0,0 +1,105 @@
...
+                ),
+            }
+
+        def rule_check(self, context, calc_vals=None, data=None):
+            design_airflow_b = calc_vals["design_airflow_b"]
+            primary_airflow_b = calc_vals["primary_airflow_b"]
+            design_electric_power_b = calc_vals["design_electric_power_b"]
+
+            return std_equal(design_airflow_b, 0.5 * primary_airflow_b) and std_equal(
+                design_electric_power_b.m, 0.35 * design_airflow_b.to("cfm").m
```
charliepnnl, line 104: I don't think you need to use `.m` for convert to cfm.  `std_equal()` should be able to handle the Pint Quantites directly.

yunjoonjung-PNNL: Great point! However, the reason why I put ```.m``` was because the two variables don't have the same type of units. More specifically, we compared ```electric_power``` in watts with ```air_flow``` in cfm. So, we may need to keep the code as is.

rct229/rules/section23/section23rule6.py
yunjoonjung-PNNL: Great catch! Fixed.

rct229/rules/section23/section23rule6.py
```diff
@@ -0,0 +1,111 @@
...
+        # baseline_system_types_dict = get_baseline_system_types(rmi_b)
+        baseline_system_types_dict = {
+            "Sys-6": ["System 6"],
+            "Sys-8": [],
+        }  # This will be removed once PR #899 is merged
+        applicable_hvac_sys_ids = [
+            hvac_id
+            for sys_type in APPLICABLE_SYS_TYPES
+            for hvac_id in baseline_system_types_dict[sys_type]
+        ]
```
weilixu, line 63: I also have the same comment here as the 23-2 and 23-3

This will miss sub-baseline HVAC system types.

rct229/rules/section23/section23rule6.py
```diff
@@ -0,0 +1,111 @@
...
+
+        hvac_id = find_exactly_one(
+            "$..heating_ventilating_air_conditioning_systems[*].id", rmi_b
+        )
+
+        return {"hvac_id": hvac_id, "applicable_hvac_sys_ids": applicable_hvac_sys_ids}
+
+    def list_filter(self, context_item, data):
+        hvac_id = data["hvac_id"]
```
weilixu, line 72: I likes the 23-2 and 23-3 solution better.
In this case, the next level is terminal, then you can get the hvac_id from
terminal["served_by_heating_ventilating_air_conditioning_system"]

In that case, you don't need to take out the hvac_id from the create_data function.

rct229/rules/section23/section23rule7.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        # baseline_system_types_dict = get_baseline_system_types(rmi_b)
+        baseline_system_types_dict = {
+            "Sys-6": ["System 6"],
+            "Sys-8": [],
+        }  # This will be removed once PR #899 is merged
+        applicable_hvac_sys_ids = [
```
weilixu, line 60: Same comment as 23-2

weilixu (approved):
Approved with condition of addressing comments

rct229/rules/section23/section23rule6.py
yunjoonjung-PNNL: Great point. I just updated the way you mentioned (assign ```W/cfm``` to the 0.35) because this is more elegant.

rct229/rules/section23/section23rule6.py
yunjoonjung-PNNL: Fixed.

rct229/rules/section23/section23rule7.py
yunjoonjung-PNNL: Fixed.

rct229/rules/section23/section23rule6.py
yunjoonjung-PNNL: Great point! Fixed.

PR #914, opened 2023-01-17
RS/YJ/Rule 23-8

yunjoonjung-PNNL:
This PR includes Rule 23-8 development. Thank you!


rct229/rules/section23/section23rule8.py
```diff
@@ -0,0 +1,139 @@
...
+                        )
+                    )
+                    and list(
+                        map(
+                            lambda x, y: std_equal(x[1], y[1]),
+                            output_validation_points,
+                            target_validation_points,
+                        )
+                    )
+                )
```
weilixu, line 139: I like the solution written here - it is a good use of map function.
To further simply it, we could use zip function as well:
```python
all([std_equal(x[0], y[0] and std_equal(x[1], y[1]) for x, y in zip(output_validation_points, target_validation_points)
```
One issue is the length matching. Using map, we will need to ensure two have the same length, otherwise, it will crash.
Using zip, the unmatched portion will be silently skipped.

So we will need to add:
```python
len(output_validation_points) == len(target_validation_points)
```

weilixu (approved):
Approved with some comments need to be addressed

rct229/rules/section23/section23rule8.py
```diff
@@ -0,0 +1,139 @@
...
+                ]
+
+                target_validation_points = [
+                    [
+                        SUPPLY_AIRFLOW_COEFF[idx] * design_airflow_b,
+                        DESIGN_POWER_COEFF[idx] * design_electric_power_b,
+                    ]
+                    for idx in range(11)
+                ]
+                place = 1
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Fixed.

rct229/rules/section23/section23rule8.py
```diff
@@ -0,0 +1,139 @@
...
+                    for output in output_validation_points_b
+                ]
+
+                target_validation_points = [
+                    [
+                        SUPPLY_AIRFLOW_COEFF[idx] * design_airflow_b,
+                        DESIGN_POWER_COEFF[idx] * design_electric_power_b,
+                    ]
+                    for idx in range(11)
+                ]
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Fixed.

rct229/rules/section23/section23rule8.py
yunjoonjung-PNNL: Great point! Thanks for letting me know about this!

rct229/rules/section23/section23rule8.py
```diff
@@ -120,20 +109,27 @@ def rule_check(self, context, calc_vals=None, data=None):
                     ]
                     for idx in range(11)
```
weilixu, line 110: I don't really like the idea of having a "magic number" in the code.
First of all, `SUPPLY_AIRFLOW_COEFF` used this 11 to create a list.
In here `target_validation_points` also used 11 to create a 2-D list.
I think we can create a global variable called `VALIDATION_POINT_LENGTH = 11`  and use it to replace the magic numbers in the code.

To further stretch the logic, in the rule check condition, instead of 
'''python
len(output_validation_points) == len(
                    target_validation_points
                )
'''
as we are certain about the length of the target validation points, we could update this to 
'''python
len(output_validation_points) == VALIDATION_POINTS_LENGTH
'''

weilixu (approved):
Approved with one stylist suggestion needs to addressed

rct229/rules/section23/section23rule8.py
yunjoonjung-PNNL: Great point! Fixed.

charliepnnl:
I believe some of my comments for Rule23-16 also apply here. If it works for you  @yunjoonjung-PNNL  I let you makes those changes here before I review this.

yunjoonjung-PNNL:
@Vh

> 

@charliepnnl I believe this PR is ready to be merged. I'd appreciate it if you could review it! Thank you!

rct229/rules/section23/section23rule8.py
```diff
@@ -0,0 +1,137 @@
...
+            def rule_check(self, context, calc_vals=None, data=None):
+                output_validation_points = calc_vals["output_validation_points"]
+                target_validation_points = calc_vals["target_validation_points"]
+
+                return len(
+                    output_validation_points
+                ) == VALIDATION_POINTS_LENGTH and all(
+                    [
+                        std_equal(x[0], y[0]) and std_equal(x[1], y[1])
+                        for x, y in zip(
```
charliepnnl, line 133: Maybe use variables `ovp` and `tvp` instead of the more generic `x` and `y`?

charliepnnl (approved):
Looks good to me. I just have a couple of very small style-type comments.

I agree that the RDS needs some cleanup. Particularly regarding the `eligivle_hvac_systems`.

rct229/rules/section23/section23rule8.py
```diff
@@ -0,0 +1,137 @@
...
+APPLICABLE_SYS_TYPES = [
+    HVAC_SYS.SYS_5,
+    HVAC_SYS.SYS_6,
+    HVAC_SYS.SYS_7,
+    HVAC_SYS.SYS_8,
+    HVAC_SYS.SYS_11_1,
+    HVAC_SYS.SYS_11_2,
+]
+VALIDATION_POINTS_LENGTH = 11
+SUPPLY_AIRFLOW_COEFF = [0.1 * i for i in range(VALIDATION_POINTS_LENGTH)]
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Good point! Fixed.

rct229/rules/section23/section23rule8.py
yunjoonjung-PNNL: Fixed.

PR #916, opened 2023-01-20
Updates to .coveragerc file

charliepnnl:
Added `show_missing = True` and `omit `**/rules/**` to pytest coverage config file.


yunjoonjung-PNNL (approved):
Looks great! Thank you!

PR #917, opened 2023-01-24
RCT/YJ/branch_coverage_check_purchased_chw_hhw_status_dict

yunjoonjung-PNNL:
This PR addresses the ```check_purchased_chw_hhw_status_dict``` function branch coverage.

**Result**: `36 statements   36 run 0 missing 0 excluded 6 partial`
**Coverage rate change**: 69% -> 90% 

The reason why the function didn't achieve 100% is there were many ```break``` used in the function. Further, when I looked closely into the ```.html``` output, it turned out the messages pattern from 6 _partial_ parts is like ```line 53 didn't jump to line 50```, which makes sense to me because ```break``` blocks getting back to line 50. Therefore, considering the code structure, (I think) 90% coverage rate is the best we can reach.

After this PR is approved, I'll pick up 2-3 functions next time. Thank you!


weilixu:
Great work adding additional tests for this function!.

I think in that case, we should rethink whether there is at least one reason to keep the break in the for loop.
If not, I think we can safely remove the break and make sure the tests go through.
If yes, we should keep the break and moving on.

As I was poking around the code, I think 53->50 is not caused by break -> we should double check if our test cases cover this branch.

yunjoonjung-PNNL:
> 

Great point! Let me try to re-structure the function in a more efficient way and in a not-use ```break```.  

Once I'm done, I'll reach out to you again!

charliepnnl:
It looks like the structure would work instead of the `break`s: 
```
if expr1 {
} elif expt2 {
}
```

yunjoonjung-PNNL:
@weilixu @charliepnnl 

After I updated the function structure and RMDs, I was able to reach a 100% coverage rate.

I'd appreciate it if you could review this PR one more time. Thanks!

weilixu (approved):

PR #921, opened 2023-01-31
RCT/CH/fix-list_filter

charliepnnl:
I realized that the base implementation for list_path should return True instead of context_item. This worked since context_item was always Truthy, but it was still wrong.  I also fixed the methods documentation to match.

### DIFF
--- a/rct229/rule_engine/rule_list_base.py
+++ b/rct229/rule_engine/rule_list_base.py
@@ -130,10 +130,9 @@ class RuleDefinitionListBase(RuleDefinitionBase):
 
         Returns
         -------
-        list of UserBaselineProposedVals
-            A filtered list of context trios
+        bool True iff context_item should passed through the filter.
         """
-        return context_item
+        return True
 
     def rule_check(self, context, calc_vals=None, data={}):
         """Overrides the base implementation to apply a rule to each entry in


yunjoonjung-PNNL:
Charlie, Could you change the merging branch to ```develop```?

charliepnnl:
I changed the base branch to develop. Thanks Yun Joon.

yunjoonjung-PNNL (approved):
Thanks for finding/fixing this. Looks great to me!

weilixu (approved):

rct229/rule_engine/rule_list_base.py
```diff
@@ -130,10 +130,9 @@ def list_filter(self, context_item, data):
 
         Returns
         -------
-        list of UserBaselineProposedVals
-            A filtered list of context trios
+        bool True iff context_item should passed through the filter.
```
weilixu, line 133: change `iff` to `if`

weilixu (commented):
Approved with a minor comment

rct229/rule_engine/rule_list_base.py
charliepnnl: The `iff` was intentional but probably a bad idea. In mathematical-speak, if means "if and only if".  I will write it out.

PR #927, opened 2023-02-06
RCT/YJ/branch_coverage_compare_standard_val

yunjoonjung-PNNL:
This PR makes the ```compare_standard_val``` function achieve 100% branch coverage.  Thanks!

### DIFF
--- a/rct229/ruleset_functions/compare_standard_val_test.py
+++ b/rct229/ruleset_functions/compare_standard_val_test.py
@@ -26,7 +26,7 @@ def test__compare_standard_val_non_stringent_code_compare():
     assert compare_standard_val(False, 0.9999 * _M2, 1.0 * _M2)
 
 
-def test__compare_standard_val_strict_compare_gt():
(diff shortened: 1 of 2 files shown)


weilixu (approved):

rct229/ruleset_functions/compare_standard_val_test.py
```diff
@@ -23,3 +24,11 @@ def test__compare_standard_val_stringent_compare_gt_equal():
 
 def test__compare_standard_val_non_stringent_code_compare():
     assert compare_standard_val(False, 0.9999 * _M2, 1.0 * _M2)
+
+
+def test__compare_standard_val_strict_compare_gt():
```
charliepnnl, line 29: Our convention for naming tests is supposed to be of the form:
test + two underscores + name of function to be tested [+ two underscores + case descriptor]

So here you would just add another underscore before "gt".

charliepnnl (approved):

PR #930, opened 2023-02-09
RCT/YJ/branch_coverage_compare_schedules

yunjoonjung-PNNL:
This PR addresses to achieve 100% branch coverage for the ```compare_schedules``` function. Thanks!

### DIFF
--- a/rct229/ruleset_functions/compare_schedule_test.py
+++ b/rct229/ruleset_functions/compare_schedule_test.py
@@ -38,7 +38,7 @@ def test__Test_Schedule_Compare_Success_2():
 
 def test__Test_Schedule_Compare_Success_3():
     """
-    Test that will not compare 100 hours between two schedules.
+    Test when two schedules are not identical.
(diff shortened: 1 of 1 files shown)


rct229/ruleset_functions/compare_schedule_test.py
```diff
@@ -36,6 +36,23 @@ def test__Test_Schedule_Compare_Success_2():
     )
 
 
+def test__Test_Schedule_Compare_Success_3():
```
charliepnnl, line 39: I looks like the `compare_schedules` function is being tested. So I think the name of this test should be something like `test__compare_schedules__not_identical`.

charliepnnl (approved):

rct229/ruleset_functions/compare_schedule_test.py
yunjoonjung-PNNL: Great point! Let me get the function name fixed. I'll also fix the other function names as well.

weilixu (approved):

PR #932, opened 2023-02-10
Add baseline system type fixes caught by KJW

jugonzal07:
Karen caught issues with system types when checking boilers. This addresses those fixes for our baseline system JSONs.

Also, addresses an issue where the master system_types JSON incorrectly referenced section 21.


yunjoonjung-PNNL:
Great work! One question) Was the master system type Excel file on OneDrive updated as well?

jugonzal07:
> 

Yup! I updated it this morning.

yunjoonjung-PNNL:
> 

I appreciate it, Juan!

yunjoonjung-PNNL (approved):

weilixu (commented):
I am surprised that these issues were not picked up by the unit test.
I checked system 1a and it seems the logic does not cover the data tracing to the fuild_loops. For example, `are_all_terminal_cool_sources_chilled_water` only checks if the `terminal.cooling_source == CHILLED_WATER`. it does not trace to `fluid_loop` to verify the terminal is actually connected with a `COOLING` type fluid loop.

Though this part of logic may not be covered in the RDS, I wonder if it should be part of RMD test - @mkarpman

yunjoonjung-PNNL:
> I am surprised that these issues are not picked up by the unit test. I checked system 1a and it seems the logic does not cover the data tracing to the fuild_loops. For example, `are_all_terminal_cool_sources_chilled_water` only checks if the `terminal.cooling_source == CHILLED_WATER`. it does not trace to `fluid_loop` to verify the terminal is actually connected with a `COOLING` type fluid loop.
> 
> Though this part of logic may not be covered in the RDS, I wonder if it should be part of RMD test - @mkarpman

@weilixu 

I think one reason why this was not detected in the unit tests was for example, ```is_baseline_system_X``` == True/False was not set for the unit tests using system type RMDs. I also identified/was trying to fix this issue last week in the ```RCT/YJ/Fix_baseline_system_test``` branch and plan to make a PR sometime this week.

mkarpman:
I think that checking for that shouldn't be part of the baseline system RDSs because that same issue (a terminal having chilled water or hot water cooling/heating source but missing a reference to the fluid loop) may affect proposed design. Perhaps there should be a family of RDSs checking for data consistency? Copying @claperle and @supriyagoel to see if they have any thoughts.

weilixu (approved):

PR #935, opened 2023-02-14
RCT/YJ/Fix_is_baseline_system_uni_test_syst_type_RMDs

yunjoonjung-PNNL:
While working on the baseline system, I found the ```is_baseline_system``` functions' unit test didn't have ```==HVAC_SYS.SYS_X``` when tested with JSON sys type RMDs. In the past, the reason why all the sys type tests passed was even if the return value is ```Not_Sys```, the assert statement still passes this because this isn't still False.

I fixed the system type RMDs if they have any issues. The ```system_types``` Excel file was updated on OneDrive.


charliepnnl (approved):
Nice work!  I love your attention to detail!

weilixu (approved):
Looks good. Please fix the conflicts and then merge.

PR #940, opened 2023-02-16
Rds/kjw/rule 21 8

KarenWGard:


docs/section21/Rule21-8.md
```diff
@@ -0,0 +1,52 @@
...
+    - `AND boiler_fluid_loop.heating_design_and_control.outdoor_high_for_loop_supply_reset_temperature equals == 50`
+    - `AND boiler_fluid_loop.heating_design_and_control.outdoor_low_for_loop_supply_reset_temperature equals == 20:`
+      - the rule passes: `PASS`
+    - otherwise: `ELSE:`
+      - the rule fails: `FAIL`
+
+
+**Questions:**  
+1.  Does this rule apply to any boiler loop, or only those serving heating systems?
+2.  if it applies only to boiler loops serving heating systemers, is there any
```
supriyagoel, line 50: On question 1, since we start with hw_loop_zone_list_w_area to see if there are zone terminal units connected to hot water loops- we should be avoiding this scenario, correct?

supriyagoel (approved):

docs/section21/Rule21-8.md
mkarpman: I agree that it should apply only HW loop serving heating systems. As Supriya noted, I think this is how it currently works because get_hw_loop_zone_list_w_area(B-RMI) function is called.

KarenWGard: yes, this is how it currently works.

do we need to also check that the heating system to which the boiler is connected is a valid baseline HVAC system type (ie does the rule apply to boilers that also serve process loads)?

docs/section21/Rule21-8.md
```diff
@@ -0,0 +1,52 @@
...
+- iterate through all the boilers the building looking for heating fluid loops: `for boiler in B_RMI.boilers:`
+  - check if the boiler loop is in the hw_loop_zone_list_w_area_dictionary: `if boiler.loop.id in hw_loop_zone_list_w_area_dictionary:`
+    - add the boiler loop to the boiler_fluid_loops_set: `boiler_fluid_loops_set.add(boiler.loop)`
+- Check if there are heating loops connected to boilers by checking the length of the boiler_fluid_loops_set: `if len(boiler_fluid_loops_set) > 0: CHECK RULE LOGIC`
+
+  - Else, rule is not applicable to B-RMR: `else: RULE_NOT_APPLICABLE`
+
+## Rule Logic:  
+
+- iterate through each fluid loop: `for boiler_fluid_loop in boiler_fluid_loops_set:`
```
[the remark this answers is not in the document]
KarenWGard: thank you @mkarpman for the clarification - yes, this is how it currently works - only hot water loops connected to boilers (primary loops) AND serving baseline systems are checked.

mkarpman (approved):

weilixu:
@supriyagoel 21-8 seems ready for merge, could you confirm?

PR #948, opened 2023-02-21
RCT/YJ/branch_coverage_get_hw_loop_zone_list_w_area

yunjoonjung-PNNL:
This PR addresses achieving 100% branch coverage for the ```get_hw_loop_zone_list_w_area``` function.

Note)
1) I deleted ```if type(hhw_loop_id) is str:``` condition since this is validated against the schema.
2) When ```hhw_loop_id``` is ```None```, I edited not to include ```None``` type in the ```hw_loop_zone_list_w_area_dict```. Please let me know if the ```None``` type should be included.


weilixu (approved):
Approved with stylist comment

PR #950, opened 2023-02-21
RCT/CH/profiler-2

charliepnnl:
This PR replaces my earlier PR RCT'CH/profiler and is built on top of Weili's restructure.


rct229/rulesets/ashrae9012019/ruleset_functions/get_baseline_system_types.py
```diff
@@ -99,33 +99,30 @@ def get_baseline_system_types(rmi_b):
...
+        rmi_b,
+    ):
         hvac_b_id = hvac_b["id"]
         terminal_unit_id_list = dict_of_zones_and_terminal_units_served_by_hvac_sys[
             hvac_b_id
         ]["terminal_unit_list"]
         zone_id_list = dict_of_zones_and_terminal_units_served_by_hvac_sys[hvac_b_id][
             "zone_list"
         ]
-        # one terminal per zone for all baseline systems
```
charliepnnl, line 110: This `if` statement is always true because each terminal _lives_ inside a single zone.  So I was able to simplify the logic.  The test file was also simplified since there is one fewer branch to test.

yunjoonjung-PNNL (approved):
Looks great to me! Thanks for doing this Charlie!

weilixu (approved):

PR #955, opened 2023-02-23
update so files are created in the correct locations

KarenWGard:


rct229/ruletest_engine/ruletest_jsons/scripts/excel_to_test_json.py
```diff
@@ -328,7 +329,7 @@ def create_dictionary_from_excel(spreadsheet_name, sheet_name):
     return json_dict
 
 
-def create_test_json_from_excel(spreadsheet_name, sheet_name, json_name):
```
weilixu, line 331: If we move all the master jsons to the ruleset, we will need to update the function in disaggregate_master_ruletest_json to take the same path.

KarenWGard:
Juan - this sounds good to me.

Thanks!

weilixu:
@jugonzal07 @KarenWGard do we still need this PR?

KarenWGard:
no - I think it duplicated a fix that Juan did.

PR #956, opened 2023-02-24
RCT/YJ/branch_coverage_get_primary_secondary_loops_dict

yunjoonjung-PNNL:
This PR addresses achieving 100% branch coverage for ```get_primary_secondary_loops_dict``` function.


rct229/rulesets/ashrae9012019/ruleset_functions/get_primary_secondary_loops_dict.py
```diff
@@ -64,25 +64,20 @@ def get_primary_secondary_loops_dict(rmi_b):
...
+    tmp_primary_secondary_loops_dict = {}
 
     # Iterate through cooling type fluid loops
     for chilled_fluid_loop in find_all(
         f'fluid_loops[*][?(@.type="{FLUID_LOOP.COOLING}")]', rmi_b
     ):
         cfl_id = chilled_fluid_loop["id"]
-        if cfl_id in chiller_loop_ids and cfl_id in non_process_chw_coil_loop_ids:
-            # No loop in baseline shall be
-            tmp_primary_secondary_loops_dict = dict()
```
weilixu, line 76: After read the RDS again, I think the implementation is missing/misinterpretating one logic.
The RDS says to `break logic and return an empty dictionary (this would be if the system was not modeled as primary-seconday, a fluid_loop will never be connected to a cooling_system.chilled_water_loop)` 

So `if cfl_id in chiller_loop_ids and cfl_id in non_process_chw_coil_loop_ids`, we shall set the `tmp_primary_secondary_loops_dict` to an empty dictionary, and also clear the `primary_loops`, then escape the loop.

We will need to update the current logic and add a test case for that.

weilixu (commented):
Add one comment to address

rct229/rulesets/ashrae9012019/ruleset_functions/get_primary_secondary_loops_dict.py
```diff
@@ -64,25 +64,20 @@ def get_primary_secondary_loops_dict(rmi_b):
     ]
     # Initialize variables
     primary_loops = []
-    tmp_primary_secondary_loops_dict = dict()
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: This is because of consistency and it has nothing to do with branch coverage. Please let me know if this change is not acceptable.

.coveragerc
```diff
@@ -1,6 +1,6 @@
 [run]
 branch = True
-omit = **/test_*, **/*_test*.*, **/__init__.py, **/rules/**
+omit = **/test_*, **/*_test*.*, **/__init__.py, **/section/**
```
charliepnnl, line 3: I'm not sure this will do what you intend. I could be wrong, but I don't think it will match folders like _section21_. I think you need **/section*/**.

rct229/rulesets/ashrae9012019/ruleset_functions/get_primary_secondary_loops_dict.py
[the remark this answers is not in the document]
charliepnnl: I have noticed the inconsistency but haven't brought it up.  However, I think this change goes in the wrong direction.  Using the literal syntax `{}` is more pythonic.  It turns out to be more time efficient too, though that doesn't really matter in our situation.

charliepnnl (approved):
Approved assuming Weili's comment about missing logic is resolved.

rct229/rulesets/ashrae9012019/ruleset_functions/get_primary_secondary_loops_dict.py
yunjoonjung-PNNL: Addressed.

yunjoonjung-PNNL (commented):
@weilixu Thanks for your patience in this PR. This PR is ready to be reviewed again.

.coveragerc
yunjoonjung-PNNL: addressed.

rct229/rulesets/ashrae9012019/ruleset_functions/get_primary_secondary_loops_dict.py
```diff
@@ -73,16 +73,12 @@ def get_primary_secondary_loops_dict(rmi_b):
...
-            break
-        elif cfl_id in chiller_loop_ids:
-            if all(
-                child_loop["id"] in non_process_chw_coil_loop_ids
-                for child_loop in getattr_(
-                    chilled_fluid_loop, "FluidLoop", "child_loops"
-                )
-            ):
-                primary_loops.append(chilled_fluid_loop)
+            tmp_primary_secondary_loops_dict = {}
```
weilixu, line 76: You will have to break here.
So in this scenario,
```python
# reset the primary loop
primary_loop = []
break
```
You don't need to reset the `tmp_primary_secondary_loop_dict` here because it does nothing in this loop.

weilixu (changes requested):

weilixu:
This is no longer relevant. Close the PR.

PR #963, opened 2023-02-28
Update the secondary rule workflow

weilixu:
Secondary rule check

I think the best option could be following exactly what we have right now.
In this proposed workflow, we will still use `calc_val` to do calculations and `rule_check` for applicability checking.

Rule_check returns `False`, then the engine returns `NOT_APPLICABLE`. otherwise, the engine returns `UNDETERMINED`.
In this way, we can also report the calculated values.

By using this implementation, Section5Rule3 does not need to modify the implementation at all. It naturally fits the workflow and produces secondary rule results. 

Let me know if this is a good approach.

### DIFF
--- a/rct229/rule_engine/rule_base.py
+++ b/rct229/rule_engine/rule_base.py
@@ -158,6 +158,19 @@ class RuleDefinitionBase:
                                 # The result is a list of outcomes
                                 outcome["result"] = result
                             # Assume result type is bool
+                            # using is False to include the None case.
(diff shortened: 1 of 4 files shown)


weilixu:
The rule test json check would fail in this case because 5-3 is producing a secondary rule outcome, but the rule test is expecting PASS and Failed outcome.

weilixu:
#964 is implemented instead of this approach. Close this PR.

PR #964, opened 2023-03-01
Rct/wx/secondary rule applicability check second

weilixu:
@charliepnnl 

Second approach to implement applicability check
A new class, `PartialRuleDefinition` is created by inherenting the RuleDefinitionBase class.
A new function called `applicability_check` is created as the template function in the `rule_check` function in this new class.

Also, the `is_primary_rule` in the `PartialRuleDefinition` is set to False - in that case, the returning value from `applicability_check` will only return `UNDETERMINED` or `NOT_APPLICABLE`


rct229/rule_engine/partial_rule_definition.py
```diff
@@ -0,0 +1,99 @@
+from rct229.rule_engine.rule_base import RuleDefinitionBase
+from rct229.utils.json_utils import slash_prefix_guarantee
```
charliepnnl, line 2: This import doesn't seem to be used in this file.

rct229/rule_engine/partial_rule_definition.py
```diff
@@ -0,0 +1,99 @@
...
+        self,
+        rmrs_used,
+        id=None,
+        description=None,
+        ruleset_section_title=None,
+        standard_section=None,
+        is_primary_rule=False,
+        rmr_context="",
+        required_fields=None,
+        must_match_by_ids=[],
```
charliepnnl, line 16: I don't think `must_match_by_ids` is used anywhere in the rule engine.  I think it should be removed here and from rule_base.py.  I don't recall when or why this was added.

rct229/rule_engine/partial_rule_definition.py
```diff
@@ -0,0 +1,99 @@
...
+        manual_check_required_msg="",
+        not_applicable_msg="",
+    ):
+        """Base class for all Partial Rule definitions (secondary)
+
+        Parameters
+        ----------
```
charliepnnl, line 23: there are other parameters that are missing documentation here.

rct229/rule_engine/partial_rule_definition.py
```diff
@@ -0,0 +1,99 @@
...
+            required_fields=required_fields,
+            manual_check_required_msg=manual_check_required_msg,
+            not_applicable_msg=not_applicable_msg,
+            ruleset_section_title=ruleset_section_title,
+            standard_section=standard_section,
+            is_primary_rule=is_primary_rule,
+        )
+
+    def rule_check(self, context, calc_vals=None, data={}):
+        """Overrides the base implementation to apply a rule to each entry in
```
charliepnnl, line 59: This documentation needs to be updated.

rct229/rule_engine/rule_base_test.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 
+from rct229.rule_engine.partial_rule_definition import PartialRuleDefinition
```
charliepnnl, line 3: The tests for PartialRuleDefinition should be put into its own test file.

charliepnnl (approved):
I think this is a very good design.  I made several comments that should be addressed.

yunjoonjung-PNNL (approved):
Very thoughtful design! I appreciate this addition Weili.

PR #965, opened 2023-03-01
updated rule 5-22 RDS

JacksonJ-KC:


docs/section5/Rule5-22_nonprimary.md
```diff
@@ -0,0 +1,30 @@
...
+- For each building segment in the Baseline model: ```for building_segment_b in B_RMR.building.building_segments:```
+
+  - For each zone in building segment: ```for zone_b in building_segment_b.zones:``` 
+
+    - For each space in zone: ```for space_b in zone_b.spaces:```
+      
+      **Rule Assertion**
+    
+      - Case 1: Check if space is existing or altered, outcome is UNDETERMINED: ```if ( space_b.status_type == EXISTING ) OR ( space_b.status_type == ALTERED ): outcome = UNDETERMINED```
+
```
supriyagoel, line 27: @JacksonJ-KC since surfaces are associated with zones, there could be a scenario that a zone contains a mix of spaces, where some spaces are existing and some are new. The rule evaluation would be provides for each exterior wall, which is defined at a zone level. So we could instead check if ANY space in a zone is 'existing' or 'altered' and return an undetermined outcome for each exterior wall associated with that zone.

supriyagoel (changes requested):
see comment on rule evaluation

docs/section5/Rule5-22_nonprimary.md
JacksonJ-KC: @supriyagoel Ok, I made adjustments according to this feedback. Did I address this similarly to how you are thinking?

PR #973, opened 2023-03-07
get_most_used_weekday_hourly_schedule function development

weilixu:
Note:
I moved Leap year hours to rulesets.py.
I think maybe I need to update the this file name to something like ruleset_common_enum? 

Any suggestions?


yunjoonjung-PNNL (approved):
Looks great to me. Thanks for this effort Weili.

rct229/rulesets/ashrae9012019/ruleset_functions/get_most_used_weekday_hourly_schedule.py
```diff
@@ -0,0 +1,61 @@
+import datetime
+from collections import Counter
+
+from rct229.rule_engine.rulesets import LeapYear
+from rct229.utils.assertions import RCTFailureException, assert_
+
+HOURS_IN_DAY = 24
+
+
+def get_most_used_weekday_hourly_schedule(hourly_data: list, year: int):
```
charliepnnl, line 10: I am very confused.

## Questions:
- Why do we have a `year` input?  I don't find a `year` field in the schema, so how would we get this value?
  - It looks like the year is required so the `datetime` library can be used. That seems like overkill for this problem. Once the day of the week of January 1st is known, the rest should be fairly straightforward list manipulations. There are probably some functions from `pydash` that will be helpful.
- I don't understand the significance whether the hourly_data represents a leap year or non-leap year. The point seems to be "what day of the week is January 1st" – and that is a value in the schema.  Besides, if we wanted to know if it was a leap year, couldn't we just check how many hours are in `hourly_data`?

charliepnnl (commented):
See my confused comment.

weilixu:
Updated the function to simply the inputs - this function now requires the hourly data list and the the day of week on jan 1st from Calendar object.

yunjoonjung-PNNL (approved):
Looks great to me. Thank you Weili.

PR #974, opened 2023-03-11
Section 5 rule 29 development

weilixu:
Add section 5 rule 29


yunjoonjung-PNNL (approved):
Looks great to me! 

Quick question) it seems like there is no RMD for rule 5-29. Would you want me to create one?

rct229/rulesets/ashrae9012019/section5/section5rule29.py
```diff
@@ -0,0 +1,100 @@
...
+        def __init__(self):
+            super(Section5Rule29.BuildingRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                each_rule=Section5Rule29.BuildingRule.AboveGradeWallRule(),
+                index_rmr="baseline",
+                list_path="$.building_segments[*].zones[*].surfaces[*]",
+            )
+
+        def create_data(self, context, data):
+            building_b = context.baseline
```
charliepnnl, line 52: The `data_items` are put into `data` by the base implementation of this method. Since the method is overridden here, the base implementation won't run and  `climate_zone` won't be in `data`.  To over come this, add the following line of code:
```
super().create_data(self, context, data)
```

rct229/rulesets/ashrae9012019/section5/section5rule29.py
charliepnnl, line 1: The RDS should be updated to remove the thermal blocks.

charliepnnl (approved):
Approved with a couple of comments that should be addressed.

PR #980, opened 2023-03-12
RS+RMD/YS/RULE19-9

yunjoonjung-PNNL:
This PR is for section 19-9 RS and RMD development.


weilixu (approved):
Approved

rct229/rulesets/ashrae9012019/section19/section19rule9.py
```diff
@@ -0,0 +1,82 @@
...
+    class HVACRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section19Rule9.HVACRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                required_fields={
+                    "$": [
+                        "fan_system",
+                    ],
+                },
+            )
```
weilixu, line 59: In the RDS, part of the logic is a component-based applicability check.
We should add another is_applicable here to check whether this HVAC is applicable for evaluation.

rct229/rulesets/ashrae9012019/section19/section19rule9.py
```diff
@@ -0,0 +1,82 @@
...
+
+        def get_calc_vals(self, context, data=None):
+            hvac_b = context.baseline
+            fan_system_b = hvac_b["fan_system"]
+
+            air_economizer_b = fan_system_b.get("air_economizer")
+            air_economizer_type_b = (
+                air_economizer_b.get("type") if air_economizer_b else None
+            )
```
weilixu, line 68: I think you can use `find_one` here
```python
air_economizer_type_b = find_one("$.air_economizer.type", fan_system)
```
This will return None if there is no `air_economizer` and if there is no `type` in an `air_economizer`.

weilixu (approved):
Approved with some suggestions

rct229/rulesets/ashrae9012019/section19/section19rule9.py
yunjoonjung-PNNL: I totally agree with the code. 

However, I think when the ```air_economizer``` doesn't exist, the RDS logic should be either ```not_applicable``` or ```undetermined```. Or. we can put the  ```air_economizer``` and its ```type``` keys as a mandatory keys .What do you think?

rct229/rulesets/ashrae9012019/section19/section19rule9.py
yunjoonjung-PNNL: From my understanding, isn't HVAC system applicability already checked in the ```is_applicable``` method above?

rct229/rulesets/ashrae9012019/section19/section19rule9.py
yunjoonjung-PNNL: Fixed.

rct229/rulesets/ashrae9012019/section19/section19rule9.py
yunjoonjung-PNNL: Fixed.

rct229/rulesets/ashrae9012019/section19/section19rule9.py
```diff
@@ -58,25 +70,30 @@ def __init__(self):
...
 
+        def is_applicable(self, context, data=None):
+            hvac_b = context.baseline
+            hvac_id_b = hvac_b["id"]
+            baseline_system_types_dict = data["baseline_system_types_dict"]
+
+            for key, value in baseline_system_types_dict.items():
+                if hvac_id_b in value:
+                    return True
+            return False
```
weilixu, line 81: The goal seems to be finding an HVAC system in the  baseline_system_types_dict

We typically use one return statement in the end instead of multiple return statements unless we are in a complex scenario (nested if for example).

```python
return any(hvac_id_b in baseline_system_types_dict[system_type] for system_type in baseline_system_types_dict.keys())
```
This shall do the trick.

weilixu (approved):
Approved with some stylist suggestions

rct229/rulesets/ashrae9012019/section19/section19rule9.py
yunjoonjung-PNNL: Good point! Addressed.

PR #983, opened 2023-03-16
Rds/cml/function get hvac sys and assoc zones largest exhaust source

claperle:


docs/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.md
```diff
@@ -0,0 +1,41 @@
...
+- Set largest hvac exhaust cfm equal to zero: `maximum_hvac_exhaust = 0`  
+- Create a fan system object: `fan_system_p = sys.fan_system`   
+- Get list of exhaust fan fan objects: `fan_sys_exhaust_list = list(fan_system_p.exhaust_fans)`  
+- Get the number of exhaust fans associated with the hvac system: `num_hvac_exhaust_fans = len(fan_sys_exhaust_list)`  
+- Reset the total fan cfm variable: `total_fan_cfm = 0` 
+- Reset the fan cfm variable: `fan_cfm = 0` 
+- Check if the num_hvac_exhaust_fans is greater than 0: `if num_hvac_exhaust_fans > 0:`  
+    - Cycle through the list of exhaust fans: `for fan_p in fan_sys_exhaust_list:`  
+        - Get the fan cfm: `fan_cfm = fan_p.design_airflow`  
+        - Total the fan cfm across the multiple fans: `total_fan_cfm = total_fan_cfm + fan_cfm`
```
mkarpman, line 29: Why do we need the total? It was included in my original comment, but you questioned it in the email and I think we agreed that you were right and we don't need it? The energy recovery exception refers to a single fan accounting for 75% of the total design outdoor air flow, not the total design exhaust flow.

docs/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.md
```diff
@@ -0,0 +1,41 @@
...
+**Function Call:**   
+1. get_dict_of_zones_and_terminal_units_served_by_hvac_sys()   
+
+
+## Logic:  
+- Create dictionary of hvac systems, zones, and terminal units: `dict_of_zones_and_terminal_units_served_by_hvac_sys = get_dict_of_zones_and_terminal_units_served_by_hvac_sys(B_RMI)`  
+- Get the list of zones associated with the hvac system: `zone_list = dict_of_zones_and_terminal_units_served_by_hvac_sys[hvac_sys.id]["ZONE_LIST"]`  
+- Set largest zone exhaust source cfm equal to zero: `maximum_zone_exhaust = 0`  
+- Set largest hvac exhaust cfm equal to zero: `maximum_hvac_exhaust = 0`  
+- Create a fan system object: `fan_system_p = sys.fan_system`
```
mkarpman, line 21: Why is there "p" suffix here and in other places? Shouldn't this apply to the baseline?

docs/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.md
```diff
@@ -0,0 +1,41 @@
...
+- Get list of exhaust fan fan objects: `fan_sys_exhaust_list = list(fan_system_p.exhaust_fans)`  
+- Get the number of exhaust fans associated with the hvac system: `num_hvac_exhaust_fans = len(fan_sys_exhaust_list)`  
+- Reset the total fan cfm variable: `total_fan_cfm = 0` 
+- Reset the fan cfm variable: `fan_cfm = 0` 
+- Check if the num_hvac_exhaust_fans is greater than 0: `if num_hvac_exhaust_fans > 0:`  
+    - Cycle through the list of exhaust fans: `for fan_p in fan_sys_exhaust_list:`  
+        - Get the fan cfm: `fan_cfm = fan_p.design_airflow`  
+        - Total the fan cfm across the multiple fans: `total_fan_cfm = total_fan_cfm + fan_cfm`  
+        - Check if the fan cfm is greater than maximum_hvac_exhaust, if it is then set the maximum_hvac_exhaust equal to the fan cfm: `if fan_cfm > maximum_hvac_exhaust: maximum_hvac_exhaust = fan_cfm`  
+- Set hvac_sys_exhaust_cfm_sum equal to the sum of hvac system exhaust fan cfm (this assumes the hvac system exhaust fans are collectively considered one source and sums them as if they were in paraellel): `hvac_sys_exhaust_cfm_sum = total_fan_cfm`
```
mkarpman, line 31: I am not sure I understand why this is needed.

docs/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.md
```diff
@@ -0,0 +1,41 @@
...
+- Set hvac_sys_exhaust_cfm_sum equal to the sum of hvac system exhaust fan cfm (this assumes the hvac system exhaust fans are collectively considered one source and sums them as if they were in paraellel): `hvac_sys_exhaust_cfm_sum = total_fan_cfm`  
+- Reset the fan cfm variable: `fan_cfm = 0` 
+- Cycle through each zone associated with the HVAC system in the B_RMI: `for zone_b in zone_list:`  
+    - Check if there are any exhaust fan objects associated with the zone: `if zone_b_obj.zonal_exhaust_fan != Null:`  
+        - Get the fan cfm: `fan_cfm = zone_b_obj.zonal_exhaust_fan.design_airflow`    
+        - Check if the cfm is greater than the current value of maximum_zone_exhaust, if it is then set maximum_zone_exhaust equal to the cfm: `if fan_cfm > maximum_zone_exhaust: maximum_zone_exhaust = fan_cfm`  
+- Create list to be returned by function: `get_hvac_sys_and_assoc_zones_largest_exhaust_source  = [hvac_sys_exhaust_cfm_sum, maximum_zone_exhaust,num_hvac_exhaust_fans, maximum_hvac_exhaust]`
```
mkarpman, line 37: I wonder if we just need one maximum exhaust, not separate maximum exhausts for HVAC system vs zone.  Perhaps you deal with it at the RDS level.

docs/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.md
claperle: @mkarpman I think its clear in the RDS

docs/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.md
claperle: What if there was an array of exhaust fans? I think the way this function and 19-24 is setup works. 

Basically what it does is it checks whether the max of the zone level exhaust and the sum of the exhaust associated with the hvac system is less 75% of the OA cfm. If yes then the exception applies. I think this makes sense because if there is more than 1 exhaust fan associated with the hvac system and it sums to less than 75% of the OA cfm then this exception would apply still. If there is only 1 exhaust fan associated with the hvac system then the sum would equal the cfm of the exhaust fan which is what we are interested in comparing.

The RDS then looks at the scenario where there is more than 1 exhaust fan associated with the HVAC system. It checks whether the max of the max of the zone level exhaust and the max exhaust cfm associated with the hvac system exhaust fans is less 75% of the OA cfm. If yes then it returns an UNDETERMINED outcome explaining that the exception may apply. I think this makes sense. Am I missing something?

docs/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.md
claperle: This is a  mistake. Fixed.

mkarpman (approved):

PR #986, opened 2023-03-16
Rds/jdj/rule5 53

JacksonJ-KC:


supriyagoel (approved):
looks great!

docs/section5/Rule5-53_nonprimary.md
```diff
@@ -0,0 +1,60 @@
+# Envelope - Rule 5-53  
+**Schema Version** 0.0.23
+**Primary Rule** False
```
supriyagoel, line 3: changee flag to true

docs/section5/Rule5-53_nonprimary.md
```diff
@@ -0,0 +1,60 @@
...
+
+  - For each zone in building_segment: ```for zone_b in building_segment_b.zones:```
+
+    - For each surface in zone: ```for surface_b in zone_b.surfaces:```
+
+      - Get surface conditioning category: ```scc_b = scc_dictionary_b[surface_b.id]```
+
+      - For each subsurface in surface: ```for subsurface_b in surface_b.subsurfaces:```
+
+        - Check if subsurface is opaque door: ```if ( subsurface_b.classification == "DOOR" ) AND ( subsurface_b.glazed_area < subsurface.opaque_area )```
```
mkarpman, line 35: I beleive it should be "subsurface_b.glazed_area <= subsurface.opaque_area" based on Section 3 deifinition.

docs/section5/Rule5-53_nonprimary.md
```diff
@@ -0,0 +1,60 @@
...
+
+            - If residential and non-residential type door construction requirements are the same, save as baseline construction: ```if target_u_factor_res == target_u_factor_nonres: target_u_factor = target_u_factor_res```
+
+            - Else, outcome is UNDETERMINED: ```manual_review_flag = TRUE```
+
+            **Rule Assertion:**  
+
+            - Case 1: For each opaque door, if the door's parent zone has both residential and non-residential spaces and the construction requirements for door are different, outcome is UNDETERMINED: ```if manual_review_flag == TRUE:
+              outcome = UNDETERMINED and raise_message "ZONE HAS BOTH RESIDENTIAL AND NON-RESIDENTIAL TYPE SPACES AND THE REQUIREMENT FOR U-FACTOR FOR DOORS ARE DIFFERENT. VERIFY DOOR U-FACTOR IS MODELED CORRECTLY."```
```
mkarpman, line 52: Please add the modeled U-factor to the message.
Also, if the modeled U-factor is not equal to the values prescribed for either residential or non-residential, then it should be FAIL instead of undetermined. The message would be the same except the last sentence would say that the modeled door U-factor doesn't match the prescribe value for either conditioning category.

docs/section5/Rule5-53_nonprimary.md
```diff
@@ -0,0 +1,60 @@
...
+
```
mkarpman, line 53: What if subsurface type is door, but the subtype doesn't map to either winging or non-swinging? I think this should also produce an undetermined outcome.

docs/section5/Rule5-53.md
```diff
@@ -19,40 +17,40 @@
...
 
-          **Rule Assertion:**  
+            - If residential and non-residential type door construction requirements are the same, save as baseline construction: ```if target_u_factor_res == target_u_factor_nonres: target_u_factor = target_u_factor_res```
 
-          - Case 1: For each opaque door, if the door's parent zone has both residential and non-residential spaces and the construction requirements for door are different, request manual review: `if manual_review_flag == TRUE: UNDETERMINED and raise_message "ZONE HAS BOTH RESIDENTIAL AND NON-RESIDENTIAL TYPE SPACES AND THE REQUIREMENT FOR U-FACTOR FOR DOORS ARE DIFFERENT. VERIFY DOOR U-FACTOR IS MODELED CORRECTLY."`
+            **Rule Assertion:**  
+            - Case 1: If the door's subclassification is not SWINGING_DOOR or NONSWINGING_DOOR, outcome is UNDETERMINED: ```if subclassification_b not in ["SWINGING_DOOR","NONSWINGING_DOOR"]: outcome = UNDETERMINED```
```
mkarpman, line 45: I suggest that we provide FAIL outcome if door U-factor doesn't match prescribed values for either door types, similar to what we do in Case 3 for doors with unclear residential vs non-residential conditioning category status. Also, in some cases U-factor requirements are the same for both types of doors (e.g., residential in CZ6)- for these cases, we are able to provide a definitinve pass/fail outcome. Also, for the Fail outcome, please provide a message indicating whether the modeled U-factor is lower than prescribed (a conservative FAIL) or higher. @supriyagoel , please share the latest on how to handle this. (I think the plan is to drop the AHJ Conservative flag, right?)

supriyagoel: @mkarpman and @JacksonJ-KC in the case the U factor is lower than the prescribed value, please simply include a text in the outcome message that states, 'rule evaluation fails with a conservative outcome'. The RCT will be able to filter based on this outcome message and provide a count  of rules evaluations that include this message

mkarpman (approved):

PR #997, opened 2023-03-21
Rds/kjw/rule 22 12

KarenWGard:


docs/section22/Rule22-12.md
```diff
@@ -0,0 +1,47 @@
...
+**Data Lookup:** None  
+**Evaluation Context:** Each Heat Rejection Loop  
+**Applicability Checks:**  
+
+1. B-RMR is modeled with at least one air-side system that is Type-7, 8, 12, 13, 7b, 8b, 12b.
+2. B-RMR is modeled with heat rejection loop
+
+**Function Call:**  
+
+1. get_heat_rejection_loops_connected_to_baseline_systems()
```
supriyagoel, line 20: @KarenWGard would this function be appropriate to use instead? https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/ruleset_functions/get_primary_secondary_loops_dict.md

supriyagoel (approved):
the function 'get_heat_rejection_loops_connected_to_baseline_systems' might need to be revised but this function looks good.

docs/section22/Rule22-12.md
```diff
@@ -0,0 +1,47 @@
...
+
+- check applicability for each loop in the model: `for fluid_loop in B_RMI.fluid_loops:`
+
+  - check if heat_rejection is one of the applicable loops: `if fluid_loop.id in heat_rejection_loop_ids: CHECK_RULE_LOGIC`
+
+  - Else, rule is not applicable to the heat rejection: `else: RULE_NOT_APPLICABLE`
+
+  ## Rule Logic:  
+
+  - create a heat_rejections_count variable to count the number of heat_rejections that are connected to the fluid_loop: `heat_rejections_count = 0`
```
mkarpman, line 34: What if there are several heat rejection loops in the project? I don't think we have an RDS to confirm that the baseline chw plant has a single heat rejection loop, so perhaps it should be part of this RDS in addition to checking that there is a single cooling tower on the loop.

docs/section22/Rule22-12.md
```diff
@@ -0,0 +1,47 @@
...
+
+  - loop through the heat rejections in the model: `for heat_rejection in B_RMI.heat_rejections:`
+
+    - check if the heat_rejection is connected to the fluid_loop, and increment the heat_rejections_count if it is: `if heat_rejection.fluid == fluid_loop: heat_rejections_count += 1`
+
+
+  **Rule Assertion - FluidLoop:**
+
+  - Case 1: If there is exactly one heat rejection connected to the fluid loop, PASS: `if heat_rejections_count == 1: PASS`
```
mkarpman, line 43: What if there are multiple heat rejection loops? It seems that with its current scope the RDS should be verifying that there is one cooling tower per loop, not the total number of cooling towers for all heat rejection loops.

docs/section22/Rule22-12.md
KarenWGard: `get_primariy_seciondary_loops_dict` returns only loops that are correctly formatted as primary / secondary loops.  They way I've written the code, the user will get feedback about heat rejection equipment regardless of whether they modeled the primary / secondary loops correctly.  As a modeler, I prefer to resolve as many issues a possible in each iteration of modeling, so it was a personal choice to do it this way.

docs/section22/Rule22-12.md
KarenWGard: line 26 looks each each fluid loop.  The rule logic is indented under line 26, so applies continues the logic from the applicability check and is looking at each individual loop.

docs/section22/Rule22-12.md
mkarpman: I am not sure what you mean by "correctly formatted". I think the function returns all primary and all secondary loops associated with the baseline HVAC. So for example, if there is only primary but no secondary loop, that would be an incorrect arrangement but the RDS would still be verifying information based on the primary loop.

KarenWGard: @mkarpman - if a cooling loop has no secondary loop, then the function `get_primary_secondary_loops_dict` will not include a key for the primary loop.  So if I rely on this function for the rule, then the rule would return "not applicable"  

There is another rule that checks specifically for correct primary-secondary configuration, so I think we can give feedback here about the heat rejection system regardless of the chilled water loop configuration.

docs/section22/Rule22-12.md
```diff
@@ -0,0 +1,65 @@
...
+
+  ## Rule Logic:  
+
+  - create a heat_rejections_count variable to count the number of heat_rejections that are connected to the fluid_loop: `heat_rejections_count = 0`
+
+  - loop through the heat rejections in the model: `for heat_rejection in B_RMI.heat_rejections:`
+
+    - check if the heat_rejection is connected to the fluid_loop, and increment the heat_rejections_count if it is: `if heat_rejection.fluid == fluid_loop: heat_rejections_count += 1`
+
+  - check whether there is more than one heat rejection loop for the chiller plant.  To do this, we need to find a chiller that is connected to the heat rejection loop.  Look at all chillers: `for chiller B_RMI.chillers:`
```
mkarpman, line 40: Wouldn't the "get_heat_rejection_loops_connected_to_baseline_systems()" return all heat rejection loops we need? So if there is more than two loops, then we'll know that's an error.

docs/section22/Rule22-12.md
```diff
@@ -0,0 +1,65 @@
...
+      - check if the other_chiller is not connected to the same condensing_loop: `if other_chiller.condensing_loop != fluid_loop:`
+
+        - this means that there is more than one fluid loop for the chilled water plant, set one_heat_rejection_loop equal to false and break out of the loop: `one_heat_rejection_loop = false; break`
+
+
+  **Rule Assertion - FluidLoop:**
+
+  - Case 1: If there is exactly one heat rejection connected to the fluid loop, AND there is exactly one heat rejection loop, PASS: `if heat_rejections_count == 1 AND one_heat_rejection_loop: PASS`
+
+  - Case 2: Elsif there is exactly one heat rejection connected to the fluid loop, but there is more than one fluid loop, fail and include note: `elsif heat_rejections_count == 1 AND !one_heat_rejection_loop: FAIL; note = "there is more than one condensing loop for this chiller plant.  There should only be one condensing loop attached to all chillers in the chiller plant"`
```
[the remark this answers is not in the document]
KarenWGard: done

docs/section22/Rule22-12.md
KarenWGard: @mkarpman - I have re-written the logic of this rule to apply at the building level instead of at the loop level.  I think this is a simpler approach and addresses most of your comments.

docs/section22/Rule22-12.md
```diff
@@ -0,0 +1,65 @@
...
+
+  - loop through the heat rejections in the model: `for heat_rejection in B_RMI.heat_rejections:`
+
+    - check if the heat_rejection is connected to the fluid_loop, and increment the heat_rejections_count if it is: `if heat_rejection.fluid == fluid_loop: heat_rejections_count += 1`
+
+  - check whether there is more than one heat rejection loop for the chiller plant.  To do this, we need to find a chiller that is connected to the heat rejection loop.  Look at all chillers: `for chiller B_RMI.chillers:`
+  
+    - check if the chiller is connected to the heat rejection loop, and if so, break out of the loop: `if chiller.condensing_loop == fluid_loop: break`
+
+  - create variable one_heat_rejection_loop to keep track of whether there are one or more heat rejection loops connected to the chiller plant.  Set this variable equal to true (at the moment, we know of only one heat rejection loop): `one_heat_rejection_loop = true`
```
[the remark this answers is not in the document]
KarenWGard: rule logic re-written to apply at the building level, not loop level, so I am now using len(heat_rejection_loop_ids) instead of counting the number of heat rejection loops connected to the chiller plant

docs/section22/Rule22-12.md
```diff
@@ -0,0 +1,65 @@
...
+
+  - find all the other chillers connected to the same cooling loop: `for other_chiller in B_RMI.chillers:`
+
+    - check if other_chiller is connected to the same cooling loop: `if other_chiller.cooling_loop = chiller.cooling_loop :`
+
+      - check if the other_chiller is not connected to the same condensing_loop: `if other_chiller.condensing_loop != fluid_loop:`
+
+        - this means that there is more than one fluid loop for the chilled water plant, set one_heat_rejection_loop equal to false and break out of the loop: `one_heat_rejection_loop = false; break`
```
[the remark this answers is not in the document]
KarenWGard: function logic has been changed so that this is possible

docs/section22/Rule22-12.md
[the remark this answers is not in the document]
supriyagoel: @KarenWGard i dont see the change of 'cooling tower' instead of 'heat rejection' and 'condenser loop' instead of 'condensing loop'. Not sure if I'm doing something wrong on GitHub

mkarpman (approved):

supriyagoel (approved):

PR #998, opened 2023-03-21
Create get_heat_rejection_loops_connected_to_baseline_systems.md

KarenWGard:


docs/ruleset_functions/get_heat_rejection_loops_connected_to_baseline_systems.md
```diff
@@ -0,0 +1,46 @@
...
+
+- For each building segment in RMI: `for building_segment in RMI...building_segments:`
+
+  - For each HVAC system in building segment: `for hvac in building_segment.heating_ventilating_air_conditioning_systems`
+
+    - Check if HVAC system is baseline system Type-7, 8, 11.1, 11.2, 12, 13, 7b, 8b, 11.1b, 12b: `if any(hvac.id in baseline_hvac_system_dict[sys_type] for sys_type in ["SYS-7", "SYS-8", "SYS-11.1", "SYS-11.2", "SYS-12", "SYS-13", "SYS-7B", "SYS-8B", "SYS-11B", "SYS-12B", "SYS-13B"]):`
+
+      - Save chilled water loop that serves cooling systems to non-process CHW coil loop array (array for all loops connected to cooling coils): `non_process_chw_coil_loop_array.append(hvac.cooling_system.chilled_water_loop)`
+
+- For each fluid loop in RMI: `for fluid_loop in RMI.RulesetModelInstance.fluid_loops:`
```
supriyagoel, line 26: does this need to be 'for each unique fluid loop'. more than one HVAC system might be served by the same fluid loop sot the array might have thr same CHW loop multiple times.

docs/ruleset_functions/get_heat_rejection_loops_connected_to_baseline_systems.md
```diff
@@ -0,0 +1,46 @@
...
+
+  - Check if loop type is cooling: `if fluid_loop.type == "COOLING":`
+
+    - Check if loop is connected to chillers, but not cooling coils: `if ( fluid_loop.id in chiller_loop_array ) AND ( !fluid_loop in non_process_chw_coil_loop_array ):
+`
+      - Check if all child loops of loop serve baseline system Type-7, 8, 11.1, 11.2, 12, 13, 7b, 8b, 11.1b, 12b only (to exclude CHW loop served by process chiller(s)): `if all(child_loop_id in non_process_chw_coil_loop_array for child_loop_id in fluid_loop.child_loops):`
+
```
supriyagoel, line 33: instead of if all(child_loop_id in non_process_chw_coil_loop_array for child_loop_id in fluid_loop.child_loops):, would it make sense to search for EACH child loop in non process CHW coil loop array 
If I understand correctly, Line 24 identified CHW loops which are serving HVAC systems.
Line 32 identifies the primary CHW loop.

docs/ruleset_functions/get_heat_rejection_loops_connected_to_baseline_systems.md
```diff
@@ -0,0 +1,46 @@
...
+## get_heat_rejection_loops_connected_to_baseline_systems
+
+Description: Get a list of all heat rejection loops in an RMI that are connected to a baseline HVAC System (Type-7, 8, 11.1, 11.2, 12, 13, 7b, 8b, 11.1b, 12b)
+
+Inputs:  
+- **RMI**: RMI that needs to get the list of primary and secondary loops (usually baseline RMI).
+
+Returns: 
+- **heat_rejection_loop_list**: A list that saves the ids of heat rejection loops in the model that serve baseline system types, e.g. [heat_rejection_loop1.id, heat_rejection_loop2.id]. If RMI does not have any qualifying heat rejection loops, it will return an empty list ([]).
+
```
supriyagoel, line 11: @KarenWGard would it make sense to use https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/ruleset_functions/get_primary_secondary_loops_dict.md to identify P/S loops and then identify the heat rejection system serving the primary loop?

docs/ruleset_functions/get_heat_rejection_loops_connected_to_baseline_systems.md
supriyagoel: also, this comment is irrelevant if it is appropriate to use https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/ruleset_functions/get_primary_secondary_loops_dict.md

docs/ruleset_functions/get_heat_rejection_loops_connected_to_baseline_systems.md
KarenWGard: Having multiples is not an issue because later (lines 34 and 36) all we're checking for is whether the chiller.cooling_loop.id is in the list.  However, I've changed the lists from type `list` to type `set` for clarity - sets can have only one copy of each item

docs/ruleset_functions/get_heat_rejection_loops_connected_to_baseline_systems.md
KarenWGard: Supriya - I'm pretty sure that `all` is correct here.  It's the same logic as in `get_primary_secondary_loops_dict.md` line 34.

The intent is to eliminate chillers that serve process loads, even if they also serve baseline systems.

docs/ruleset_functions/get_heat_rejection_loops_connected_to_baseline_systems.md
KarenWGard: `get_primariy_seciondary_loops_dict` returns only loops that are correctly formatted as primary / secondary loops.  The way I've written this code, the user will get feedback about heat rejection equipment regardless of whether they modeled the primary / secondary loops correctly.  As a modeler, I prefer to resolve as many issues a possible in each iteration of modeling, so it was a personal choice to do it this way.

KarenWGard: updated the function per our notes on the call.  The function now relies on `get_primary_secondary_loops_dict` to maintain consistency with other rules

docs/ruleset_functions/get_heat_rejection_loops_connected_to_baseline_systems.md
KarenWGard: due to the update to using `get_primary_secondary_loop_dict` this comment thread is no longer applicable

docs/ruleset_functions/get_heat_rejection_loops_connected_to_baseline_systems.md
KarenWGard: due to the update to using `get_primary_secondary_loop_dict` this comment thread is no longer applicable

mkarpman (approved):

supriyagoel (approved):

PR #1000, opened 2023-03-21
Create Rule22-15.md

KarenWGard:


docs/section22/Rule22-15.md
```diff
@@ -0,0 +1,52 @@
...
+    - check if the design wetbulb temperature is between 55 and 90F (inclusive): `if ( heat_rejection.design_wetbulb_temperature >= 55 ) AND ( heat_rejection.design_wetbulb_temperature <= 90 ): CHECK_RULE_LOGIC`
+
+    - Else, rule is not applicable to the heat rejection: `else: RULE_NOT_APPLICABLE`
+
+  - Else, rule is not applicable to the heat rejection: `else: RULE_NOT_APPLICABLE`
+
+  ## Rule Logic:  
+
+  - get the approach: `approach = heat_rejection.approach`
+  - calculate the target approach: `target_approach = 25.72 - (0.24 * heat_rejection.design_wetbulb_temperature)`
```
supriyagoel, line 39: should clarify expected units here. I am assuming the wetbulb here is in F and the target approach would be in F. Might be good to mention it so that there is no confusion while implementing this

supriyagoel (commented):
small suggestion to clarify units

docs/section22/Rule22-15.md
```diff
@@ -0,0 +1,52 @@
...
+
+- check applicability for each HeatRejection in the model: `for heat_rejection in B_RMI.heat_rejections:`
+
+  - check if heat_rejection fluid_loop is one of the applicable loops: `if heat_rejection.loop.id in heat_rejection_loop_ids:`
+  
+    - check if the design wetbulb temperature is between 55 and 90F (inclusive): `if ( heat_rejection.design_wetbulb_temperature >= 55 ) AND ( heat_rejection.design_wetbulb_temperature <= 90 ): CHECK_RULE_LOGIC`
+
+    - Else, rule is not applicable to the heat rejection: `else: RULE_NOT_APPLICABLE`
+
+  - Else, rule is not applicable to the heat rejection: `else: RULE_NOT_APPLICABLE`
```
mkarpman, line 34: Should we account for multiple heat rejection loops? Otherwise looks good.

docs/section22/Rule22-15.md
KarenWGard: yes, thank you.  done

docs/section22/Rule22-15.md
KarenWGard: line 26 looks at each heat rejection loop.  You'll notice that the rule logic is indented so that it is within the loop created on line 26, so we are looking at every heat rejection in the RMI

mkarpman (approved):

supriyagoel (approved):

PR #1001, opened 2023-03-21
Create Rule22-17.md

KarenWGard:


docs/section22/Rule22-17.md
```diff
@@ -0,0 +1,52 @@
...
+  - get the fan_electric_input_power for the heat rjection: `fan_W = heat_rejection.fan_nameplate_power`
+  
+  - get the design water flow rate (assumed to be gpm): `design_water_flowrate = heat_rejection.design_water_flowrate`
+  
+  - calculate the efficiency in gpm/hp, the following equation includes a conversion from Watts to hp: `heat_rejection_efficiency = design_water_flowrate / (fan_W * 0.00134102)`
+
+  **Rule Assertion - HeatRejection:**
+
+  - Case 1: If the heat_rejection_efficiency equals 38.2 gpm/hp, PASS: `if ( heat_rejection_efficiency == 38.2 ): PASS`
+
```
supriyagoel, line 43: @mkarpman do we need to include a conservative outcome here?

supriyagoel (commented):
comment for MK

docs/section22/Rule22-17.md
mkarpman: I think so. If the value is higher, we would consider it conservative.

docs/section22/Rule22-17.md
```diff
@@ -0,0 +1,52 @@
...
+# CHW&CW - Rule 22-17  
+
+**Schema Version:** 0.0.25    
+**Mandatory Rule:** True  
+**Rule ID:** 22-17  
+**Rule Description:** The baseline heat rejection device shall have an efficiency of 38.2 gpm/hp
```
mkarpman, line 7: We are not checking that this rating is at conditions specified in Table 6.8.1-7. I think this is OK, but perhaps worth including in a note accompanying the outcome.

docs/section22/Rule22-17.md
```diff
@@ -0,0 +1,52 @@
...
+
+  - check if heat_rejection fluid_loop is one of the applicable loops, if so, continue to the rule logic: `if heat_rejection.loop.id in heat_rejection_loop_ids: CONTINUE TO RULE LOGIC`
+  
+  - Else, rule is not applicable to the heat rejection: `else: RULE_NOT_APPLICABLE`
+
+  ## Rule Logic:  
+
+  - get the fan_electric_input_power for the heat rjection: `fan_W = heat_rejection.fan_nameplate_power`
+  
+  - get the design water flow rate (assumed to be gpm): `design_water_flowrate = heat_rejection.design_water_flowrate`
```
mkarpman, line 36: I think this should be the rated flow, not design flow, since the prescribed efficiency is "...at the conditions specified
in Table 6.8.1-7". We should also add a note to this effect to the schema to indicate that the rated flow rate is used for App G, not design flow rate.

docs/section22/Rule22-17.md
```diff
@@ -0,0 +1,52 @@
...
+
+- check applicability for each HeatRejection in the model: `for heat_rejection in B_RMI.heat_rejections:`
+
+  - check if heat_rejection fluid_loop is one of the applicable loops, if so, continue to the rule logic: `if heat_rejection.loop.id in heat_rejection_loop_ids: CONTINUE TO RULE LOGIC`
+  
+  - Else, rule is not applicable to the heat rejection: `else: RULE_NOT_APPLICABLE`
+
+  ## Rule Logic:  
+
+  - get the fan_electric_input_power for the heat rjection: `fan_W = heat_rejection.fan_nameplate_power`
```
mkarpman, line 34: I am not sure nameplate power is the right term here because efficiency should be at the rated conditions at which the HP may not necessarily be equal to the nameplate HP. I know we discussed this at the last call, but I am having second thoughts. Perhaps ask Reid, Mike R and Mike T?

docs/section22/Rule22-17.md
```diff
@@ -0,0 +1,52 @@
...
+
+1. get_heat_rejection_loops_connected_to_baseline_systems()
+
+**Applicability Checks:**  
+
+- The function get_heat_rejection_loops_connected_to_baseline_systems() returns a list of loops that are connected to Type 7,8,12,13,7b,8b,12b.  Get the list of applicable heat rejection loops: `heat_rejection_loop_ids = get_heat_rejection_loops_connected_to_baseline_systems(B_RMI)`
+
+- check applicability for each HeatRejection in the model: `for heat_rejection in B_RMI.heat_rejections:`
+
+  - check if heat_rejection fluid_loop is one of the applicable loops, if so, continue to the rule logic: `if heat_rejection.loop.id in heat_rejection_loop_ids: CONTINUE TO RULE LOGIC`
```
mkarpman, line 28: Should we check it for each heat rejection object attached to baseline CW loops? (Same comment as on 2-13, etc.)

docs/section22/Rule22-17.md
KarenWGard: is this the ahj_compare flag?

docs/section22/Rule22-17.md
KarenWGard: I don't think that we have the data to check whether this is at the specified rating conditions.  I added a note to the PASS result

docs/section22/Rule22-17.md
KarenWGard: ok, changed.

docs/section22/Rule22-17.md
KarenWGard: noted.  How should we resolve this?

docs/section22/Rule22-17.md
KarenWGard: yes, this does check for each.  (same response as for the others)

docs/section22/Rule22-17.md
```diff
@@ -0,0 +1,61 @@
...
+  
+  - check if the heat rejection has fan_shaft_power (not sure the exact python command for this): `if heat_rejection.fan_shaft_power:
+  
+    - set fan_shaft_power equal to heat_rejection.fan_shaft_power: `fan_shaft_power = heat_rejection.fan_shaft_power`
+    
+  - otherwise, we need to calculate shaft power based on the motor nameplate power: `else:`
+  
+    - set fan_shaft_power equal to the heat_rejection.fan_motor_nameplate_power * 0.9 (0.9 is a rule of thumb suggested by Michael Rosenberg in April 2023): `fan_shaft_power = heat_rejection.fan_motor_nameplate_power * 0.9`
+    
+  - calculate fan brake horsepower by multiplying the fan shaft power by the fan efficiency and 0.00134102 (the conversion from Watts to HP): `fan_BHP = fan_shaft_power * heat_rejection.fan_motor_efficiency * 0.00134102`
```
mkarpman, line 42: Since fan_shaft_power is equal to BHP, we shouldn't be multiplying it by motor efficiency here. 
Also, I am not sure whether we need to conver to HP since that wuold be done by RCT as it parses the RMD riles. @weilixu please confirm.

So if fan_shaft_power is entered, than fan_BHP = fan_shaft_power; else if fan_motor_nameplate is 
entered then fan_BHP = fan_motor_nameplate_power * fan_motor_efficiency *0.9

docs/section22/Rule22-17.md
```diff
@@ -0,0 +1,61 @@
...
+
+  ## Rule Logic:  
+  
+  - check if the heat rejection has fan_shaft_power (not sure the exact python command for this): `if heat_rejection.fan_shaft_power:
+  
+    - set fan_shaft_power equal to heat_rejection.fan_shaft_power: `fan_shaft_power = heat_rejection.fan_shaft_power`
+    
+  - otherwise, we need to calculate shaft power based on the motor nameplate power: `else:`
+  
+    - set fan_shaft_power equal to the heat_rejection.fan_motor_nameplate_power * 0.9 (0.9 is a rule of thumb suggested by Michael Rosenberg in April 2023): `fan_shaft_power = heat_rejection.fan_motor_nameplate_power * 0.9`
```
mkarpman, line 40: fan_shaft_power is equal to BHP, so we should calculate it as heat_rejection.fan_motor_nameplate_power * 0.9*fan_motor_efficiency

docs/section22/Rule22-17.md
```diff
@@ -0,0 +1,61 @@
...
+    
+  - calculate fan brake horsepower by multiplying the fan shaft power by the fan efficiency and 0.00134102 (the conversion from Watts to HP): `fan_BHP = fan_shaft_power * heat_rejection.fan_motor_efficiency * 0.00134102`
+ 
+  - get the rated water flow rate (assumed to be gpm): `rated_water_flowrate = heat_rejection.rated_water_flowrate`
+  
+  - calculate the efficiency in gpm/hp: `heat_rejection_efficiency = rated_water_flowrate / (fan_BHP)`
+
+  **Rule Assertion - HeatRejection:**
+
+  - Case 1: check if the efficiency is greater than the rated efficiency: `if ( ahj_flag AND ( heat_rejection_efficiency > 38.2 )): UNDETERMINED; note = "The project includes a cooling tower.  We calculated the cooling tower efficiency to be greater than the required efficiency of 38.2 gpm / hp.  Please review the efficiency and confirm that it is correct at the rating conditions as specified in the Table 6.8.1-7."`
```
mkarpman, line 50: Perhaps change the message as follows: "Based on simulation inputs, the baseline cooling tower efficiency is equal to "heat_rejectionj_efficiency" which is better than the required efficiency of 38.2 gpm / hp, resulting in a more stringent baseline.  However, it was not possible to verify that the modeling inputs correspond to the rating conditions in Table 6.8.1-7."

docs/section22/Rule22-17.md
```diff
@@ -0,0 +1,61 @@
...
+  
+  - Case 2: elsif  the heat_rejection_efficiency equals 38.2 gpm/hp, PASS, but include note: `if ( heat_rejection_efficiency == 38.2 ): UNDETERMINED; note = "The project includes a cooling tower.  We calculated the cooling tower efficiency to be correct at 38.2 gpm / hp.  Ensure that the efficiency is specified at the design rating conditions per table 6.8.1-7"`
```
mkarpman, line 52: Perhpas for both Case 1 and Case 2 we should provide different outcomes for cases when fan shaft power was directly entered vs cases when we estimated it using rule of thumb 0.9 multiplier. The outcome will be undetermined in all cases, but messages may point to the additional approximation when fan shaft power was not directly entered. I have no strong feelings either way.

KarenWGard: I added a variable `additional_note_for_no_shaft_power` - which gives the details about how the `fan_shaft_power `is calculated if the `heat_rejection.fan_shaft_power` data element does not exist.

docs/section22/Rule22-17.md
KarenWGard: changed.

docs/section22/Rule22-17.md
KarenWGard: thank you

docs/section22/Rule22-17.md
KarenWGard: thanks, @mkarpman.  I've updated the fan_shaft_power (W) calculation per your notes.

I'm leaving the conversion to BHP for the moment until I have a chance to review it with @weilixu.

docs/section22/Rule22-17.md
supriyagoel: @KarenWGard and @mkarpman - why do we need 3 cases here. Can it simply be one case, which always results in an undetermined outcome and the message simple states 
The project includes a cooling tower.  We calculated the cooling tower efficiency to be " + heat_rejection_efficiency + ", and the required efficiency is 38.2 gpm / hp.  However, it was not possible to verify that the modeling inputs correspond to the rating conditions in Table 6.8.1-7.  Please review the efficiency and ensure that it is correct at the rating conditions as specified in the Table 6.8.1-7." + addtitional_note_for_no_shaft_power`

mkarpman: I don't have strong feelings one way or the other. I thought it may be worth mentioning that our calculation includes rule of thumb hp to bhp conversion in cases when fan_shaft_power is not entered, but I may be overthinking it.

mkarpman (approved):

supriyagoel (approved):

PR #1003, opened 2023-03-22
Rule test directory restructure

jugonzal07:
I restructured the rule test directory to better reflect the notion that RCT can support multiple rulesets. I updated scripts to account for this new directory structure. I also took this opportunity to decouple the excel_to_test_json.py script from functions utilized to perform the excel to test JSON workflow.


weilixu (approved):

jugonzal07:
> unless you've eliminated the need for it, I believe that `disaggregate_master_ruletest_json.py` will need to be updated as well.

Thanks Karen. As far as I can tell, this is already handled in the [json_genreation_utilities](https://github.com/pnnl/ruleset-checking-tool/blob/develop/rct229/ruletest_engine/ruletest_jsons/scripts/json_generation_utilities.py#L346). I was surprised to see the ruletest_doc parameter already there, it looks like Weili added it back in February. Perhaps we should update the documentation in the docstrings, though.

PR #1023, opened 2023-04-03
RS/YJ/aggregate_min_OA_schedule_across_zones

yunjoonjung-PNNL:
The PR includes the ```aggregate_min_OA_schedule_across_zones``` function development. Its unit test also achieves 100% branch coverage.

Questions) (I'd appreciate it if @KarenWGard and @JacksonJ-KC could also consider the question.)
Do we need the RMI argument in this function? I don't see it's used anywhere and it seems like it's not needed considering the logic.


rct229/rulesets/ashrae9012019/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/aggregate_min_OA_schedule_across_zones.py
```diff
@@ -0,0 +1,26 @@
...
+    list
+        This function receives an RMR (B, U, or P) and a list of zone OA min cfm schedules (generated by the get_min_OA_CFM_sch_zone function) and loops through each zone to create an aggregated 8760 for OA CFM for the HVAC system.
+    """
+
+    for zone_OA_CFM_list_of_schedule in zone_OA_CFM_list_of_schedules:
+        try:
+            hourly_zone_running_OA_CFM_total = list(
+                map(add, hourly_zone_running_OA_CFM_total, zone_OA_CFM_list_of_schedule)
+            )
+        except NameError:  # when this is the first iteration of the loop
```
weilixu, line 23: I prefer to initialize the list - say: `hourly_zone_running_OA_CMF_total = [0] * 8760` unless there is a particular reason that we want to return None for example.

rct229/rulesets/ashrae9012019/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/aggregate_min_OA_schedule_across_zones.py
```diff
@@ -0,0 +1,26 @@
...
+    Returns
+    -------
+    list
+        This function receives an RMR (B, U, or P) and a list of zone OA min cfm schedules (generated by the get_min_OA_CFM_sch_zone function) and loops through each zone to create an aggregated 8760 for OA CFM for the HVAC system.
+    """
+
+    for zone_OA_CFM_list_of_schedule in zone_OA_CFM_list_of_schedules:
+        try:
+            hourly_zone_running_OA_CFM_total = list(
+                map(add, hourly_zone_running_OA_CFM_total, zone_OA_CFM_list_of_schedule)
```
weilixu, line 21: I am not sure if we want to have `add` from operator.
My main concern is from the `init` in the rulesets folder which may need to exclude these modules when filtering rule definition classes from the module.

rct229/rulesets/ashrae9012019/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/aggregate_min_OA_schedule_across_zones.py
yunjoonjung-PNNL: One question. Do we also have a case when we have a leap year? Or would we always have regular year (8760 hrs)?

yunjoonjung-PNNL: As we discussed during the group meeting, I added the```assert``` and ```hourly_zone_running_OA_CMF_total = [0] * 8760```.

rct229/rulesets/ashrae9012019/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/aggregate_min_OA_schedule_across_zones.py
yunjoonjung-PNNL: Great point! I just changed the code to use for loops to handle this.

rct229/rulesets/ashrae9012019/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/aggregate_min_OA_schedule_across_zones.py
```diff
@@ -0,0 +1,31 @@
...
+    ----------
+    zone_OA_CFM_list_of_schedules: list of list
+        A list which contains each zone's 8760 OA CFM schedule as a list value.
+
+    Returns
+    -------
+    list
+        This function receives an RMR (B, U, or P) and a list of zone OA min cfm schedules (generated by the get_min_OA_CFM_sch_zone function) and loops through each zone to create an aggregated 8760 for OA CFM for the HVAC system.
+    """
+
```
weilixu, line 17: line 25 takes out the first element from `zone_OA_CFM_list_of_schedules`. It is actually a bit dangerous to do so unless you are 100% sure it is not an empty list. To gain the confident on this operation, we shall first, in the argument specify the type:
```python
aggregate_min_OA_schedule_across_zones(zone_OA_CFM_list_of_schedules: list)
```
second, make sure it is not an empty list
```python
assert_(len(zone_OA_CFM_list_of_schedules)>0, 'The provided list of schedules is an empty list.'
```

rct229/rulesets/ashrae9012019/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/aggregate_min_OA_schedule_across_zones.py
```diff
@@ -0,0 +1,31 @@
...
+    assert all(
+        [
+            len(zone_OA_CFM_list_of_schedule) in (8760, 8784)
+            for zone_OA_CFM_list_of_schedule in zone_OA_CFM_list_of_schedules
+        ]
+    ), f"Not all schedules are either length of 8760 or 8784."
```
weilixu, line 23: Let's use `assert_` instead of `assert` because `assert_` raise to an exception that can be handled by engine.

In this assertion, we shall check two things: 1. len of a list is either 8760 or 8784, 2. all lists shall have same length.

You can use iter to do the check
```python
schedule_iter = iter(zone_OA_CFM_list_of_schedule)
schedule_len = len(next(schedule_iter))
assert_(all(len(schedule) == schedule_len and len(schedule) in (8760, 8784) for schedule in schedule_iter)), "Not all schedules...")

```

rct229/rulesets/ashrae9012019/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/aggregate_min_OA_schedule_across_zones.py
```diff
@@ -0,0 +1,31 @@
...
+
+    hourly_zone_running_OA_CFM_total = [0] * len(zone_OA_CFM_list_of_schedules[0])
+
+    for zone_OA_CFM_list_of_schedule in zone_OA_CFM_list_of_schedules:
+        for i in range(len(zone_OA_CFM_list_of_schedule)):
+            hourly_zone_running_OA_CFM_total[i] += zone_OA_CFM_list_of_schedule[i]
+
```
weilixu, line 30: One quick way to remove a nested for loop is through zip
```python
hourly_zone_running_OA_CFM_total = [sum(x) for x in zip(*zone_OA_CFM_list_of_schedules)]
```

JacksonJ-KC:
> The PR includes the `aggregate_min_OA_schedule_across_zones` function development. Its unit test also achieves 100% branch coverage.
> 
> Questions) (I'd appreciate it if @KarenWGard and @JacksonJ-KC could also consider the question.) Do we need the RMI argument in this function? I don't see it's used anywhere and it seems like it's not needed considering the logic.

@yunjoonjung-PNNL I agree, it seems the context for user/baseline/proposed would have already been established prior to calling this function in order to create the list of schedules that gets passed as an argument.

rct229/rulesets/ashrae9012019/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/aggregate_min_OA_schedule_across_zones.py
yunjoonjung-PNNL: Great point! Addressed.

rct229/rulesets/ashrae9012019/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/aggregate_min_OA_schedule_across_zones.py
yunjoonjung-PNNL: I assumed ```zone_OA_CFM_list_of_schedules``` to be a list of lists. So, I specified ```List[list]``` type hint.

rct229/rulesets/ashrae9012019/ruleset_functions/baseline_systems/baseline_hvac_sub_functions/aggregate_min_OA_schedule_across_zones.py
yunjoonjung-PNNL: Good point! addressed.

weilixu (approved):

PR #1030, opened 2023-04-10
non-primary rule

KarenWGard:


weilixu:
It seems black test failed. Could you run `pipenv run black .` and make another push?

KarenWGard:
yes, I ran black on all of them.  Thanks!

Karen

rct229/rulesets/ashrae9012019/section22/section22rule35.py
```diff
@@ -0,0 +1,29 @@
...
+            standard_section="Section G3.1.3.10 Chilled-Water Pumps (Systems 7, 8, 11, 12, and 13)",
+            is_primary_rule=False,
+            rmr_context="ruleset_model_instances/0",
+            manual_check_required_msg="Manual Check Required - Baseline is modeled with purchased chilled water.  Make sure "
+            "baseline systems served by purchased chilled water are not modeled with chilled water reset.",
+            not_applicable_msg="Rule 22-35 Not Applicable - the baseline is not modeled with Purchased Chilled Water",
+        )
+
+    def applicability_check(self, context, calc_vals, data):
+        rmi_b = context.baseline
```
weilixu, line 27: Under the PartialRuleDefinition, the level is still at the top level instead of rule model instance level.

```python
class Section22Rule35(RuleDefinitionListIndexedBase):
....
    class RuleSetModelInstanceRule(PartialRuleDefinition):
    ....
    def applicability_check(self, context, calc_vals, data):
        rmi_b = context.baseline
...
```

section21rule5 can be a good example for data processing at RMI level.

weilixu (changes requested):

rct229/rulesets/ashrae9012019/section22/section22rule35.py
KarenWGard: Thank you.  I made an update.  If this is correct, I'll need to make the same changes to the other non-primary rules I created as part of this batch.

rct229/rulesets/ashrae9012019/section22/section22rule35.py
```diff
@@ -0,0 +1,38 @@
...
+            rmrs_used=UserBaselineProposedVals(False, True, False),
+            each_rule=Section22Rule35.RulesetModelInstanceRule(),
+            index_rmr="baseline",
+            id="22-35",
+            description="Baseline systems served by purchased chilled water shall not be modeled with"
+            " chilled water reset",
+            ruleset_section_title="HVAC - Chiller",
+            standard_section="Section G3.1.3.10 Chilled-Water Pumps (Systems 7, 8, 11, 12, and 13)",
+            is_primary_rule=False,
+            rmr_context="ruleset_model_instances/0",
```
weilixu, line 23: We typically don't use `rmr_context` in the `RuleDefinitionListIndexedBase` just because it is not used by this sub-class.
Alternatively, you can replace it with a `list_path="ruleset_model_instances[0]"`. It helps loop into the ruleset_model_instance in the nested class.

weilixu (approved):
Approved with one comment need to address

weilixu:
Nice work, approved and will merge it today

KarenWGard:
Great.  Thank you.  I updated the other non-primary RT's for section 21 and
22 following your comments form this rule test.

yunjoonjung-PNNL (approved):
Great work! Thank you!

PR #1039, opened 2023-04-12
Update Rule19-13.md

claperle:
@yunjoonjung-PNNL can you please make sure I updated this correctly as discussed? I would like you to check before I revise all of the others. Thanks!


yunjoonjung-PNNL (approved):
Looks great to me. Thank you for making this PR.

docs/section19/Rule19-13.md
```diff
@@ -72,9 +72,9 @@ Exceptions:
...
-        - Case 3: Else if the supply-air-to-room temperature set-point difference does not equal 20°F or 17F per G3.1.2.8.1 including Exceptions 1, and the baseline fan system supply cfm equals the proposed (i.e., airflow rate could be dictated by applicable codes or accreditation standard) then UNDETERMINED: `elif all_design_setpoints_delta_Ts_are_per_reqs == false and list(get_fan_system_object_supply_return_exhaust_relief_total_kW_CFM(B_RMI, hvac.fan_system)["Fan_CFM"])[0] == proposed_supply_cfm: outcome = "UNDETERMINED" and raise_message "<insert hvac.id> was not modeled based on a supply-air-to-room temperature set-point difference of 20°F (or 17°F, if lab). The baseline and proposed supply cfm was modeled identically at <insert proposed_supply_cfm> CFM. Manual review is required to determine if the airflow rate was modeled to comply with applicable codes or accreditation standards. If not, fail."`
```
claperle, line 77: @yunjoonjung-PNNL I removed the list() around the code because it is no longer a list (its a single value) so I don't think its needed. Please let me know if I am mistaken.

yunjoonjung-PNNL: Looks great to me!

yunjoonjung-PNNL:
I've one general comment.

I found that some variables have a specific unit in their name (e.g., ```proposed_supply_cfm ```). I believe it'd be better not to have unit in variable names because we use the ```pint``` library which allows us to compare two quantities automatically converting the units. For example, if we compare 10 ft with 3.0 m, we can directly compare two values without converting them to the same unit because the ```pint``` library does it for us.

weilixu (approved):

PR #1041, opened 2023-04-12
Update get_zones_computer_rooms.md

claperle:


mkarpman (approved):

weilixu:
@claperle I do not see this function called in section 19. Is it used in the section 18?

PR #1048, opened 2023-04-17
RS/YJ/RULE19-13

yunjoonjung-PNNL:
This PR hasn't been tested with RMD yet. I'll work with @JacksonJ-KC to test this rule.


JacksonJ-KC:
> This PR hasn't been tested with RMD yet. I'll work with @JacksonJ-KC to test this rule.

@yunjoonjung-PNNL I now have a PR open to merge the test rmd for this rule.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,255 @@
...
+    def create_data(self, context, data):
+        rmi_b = context.baseline
+        rmi_p = context.proposed
+
+        dict_of_zones_and_terminal_units_served_by_hvac_sys_b = (
+            get_dict_of_zones_and_terminal_units_served_by_hvac_sys(rmi_b)
+        )
+
+        zone_info = {}
+        for hvac_id_b in dict_of_zones_and_terminal_units_served_by_hvac_sys_b:
```
weilixu, line 74: Do you need to add keys()? 'dict_of_zones_and_terminal_units.served_by_hvac_sys_b.keys()'

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,255 @@
...
+                zone_info[zone_id_b]["proposed_supply_airflow"] = ZERO.FLOW
+                zone_info[zone_id_b][
+                    "all_design_setpoints_delta_Ts_are_per_reqs"
+                ] = ZERO.TEMPERATURE
+
+                # get `design_thermostat_cooling_setpoint` and `design_thermostat_heating_setpoint`
+                for zone_b in find_all(
+                    '$.buildings[*].building_segments[*].zones[?(@.id == "{zone_id_b}")]',
+                    rmi_b,
+                ):
```
weilixu, line 91: The logic seems wrong here.

The RDS is trying to calculate the HVAC design thermostate setpoint for heating and cooling based highest and lowest zone_design_thermostat_setpoints respectively. Based on the implementation, I will just include the zone_design_thermostat_cooling_setpoint & zone_design_thermostat_heating_setpoint in the `create_data` function.

```python
zone_info[zone_id_b]["design_thermostat_cooling_setpoint"] = min(find_one(f'$.buildings[*].building_segments[*].zones[(@.id == "{zone_id_b}")].design_thermostat_cooling_setpoint',rmi_b),
100*ureg("C")) 
# set to some very large temperature. - maybe create an infinity, I am not sure how pint could handle infinity, you can try this out.
zone_info[zone_id_b]["design_thermostat_heating_setpoint"] = min(find_one(f'$.buildings[*].building_segments[*].zones[(@.id == "{zone_id_b}")].design_thermostat_heating_setpoint',rmi_b),
ZERO.temperature)
```
You should also remove the for loop - instead, use `find_one` function.
The above should do the trick to extract the design thermostat.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,255 @@
...
+                        )
+                    )
+                    zone_info[zone_id_b]["design_thermostat_heating_setpoint"] = max(
+                        find_all(
+                            f'$.buildings[*].building_segments[*].zones[?(@.id == "{zone_id_b}")].design_thermostat_heating_setpoint',
+                            rmi_b,
+                        )
+                    )
+
+                    for space_b in getattr_(zone_b, "Zone", "spaces"):
```
weilixu, line 105: can be simplified to 
```python
zone_has_lab_space = any(space_b.get("lighting_space_type") == LIGHTING_SPACE.LABORATORY_EXCEPT_IN_OR_AS_A_CLASSROOM) for space in find_all("$.spaces[*]", zone_b)]
```
Note:
1. We don't want to fail the missing `lighting_space_type` because some space may not have `lighting_space_type` - e.g., attic. In that case, we shall handle it as `False`.
2. We do not want to fail a zone with no space in here -> the any([]) returns `False`

The simplified code handles both cases with `False` outcome.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,255 @@
...
+                        )
+
+                    # calculate proposed_supply_airflow
+                    for terminal_p in find_all(
+                        '$.buildings[*].building_segments[*].zones[?(@.id == "{zone_id_b}")].terminals[*]',
+                        rmi_p,
+                    ):
+                        zone_info[zone_id_b]["proposed_supply_airflow"] += getattr_(
+                            terminal_p, "Terminal", "primary_airflow"
+                        )
```
weilixu, line 122: The above logic can also simplified to:
```python
zone_info[zone_id_b]["proposed_supply_airflow"] = pint_sum([terminal_p.get("primary_air_flow", ZERO.FLOW) for terminal_p in find_all('$.buildings[*].building_segments[*].zones[?(@.id == "{zone_id_b}")].terminals[*]',rmi_p,)]
```
Again, we do not want to raise exception if a zone has no `primary_airflow` data. Instead, we assign it with `ZERO.FLOW`

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,255 @@
...
+
+        return any(
+            [
+                baseline_system_type_compare(system_type, applicable_sys_type, False)
+                for system_type in baseline_system_types_dict.keys()
+                for applicable_sys_type in APPLICABLE_SYS_TYPES
+            ]
+        )
+
+    class HVACRule(RuleDefinitionBase):
```
weilixu, line 180: After reading the RDS, I think the rule shall be evaluated at zone level instead of HVAC level.
The rule check is `supply_air_to_room` differences. So at room level, we shall compare the highest `heating_setpoint` to the `heating_supply_air_temp` or the lowest `cooling_setpoint` to the `cooling_supply_air_temp`

I suggest to revise the rule definition logic.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,255 @@
...
+                    # calculate proposed_supply_airflow
+                    for terminal_p in find_all(
+                        '$.buildings[*].building_segments[*].zones[?(@.id == "{zone_id_b}")].terminals[*]',
+                        rmi_p,
+                    ):
+                        zone_info[zone_id_b]["proposed_supply_airflow"] += getattr_(
+                            terminal_p, "Terminal", "primary_airflow"
+                        )
+
+                for terminal_b in dict_of_zones_and_terminal_units_served_by_hvac_sys_b[
```
weilixu, line 124: I prefer to have below logics happen in the `calc_val` function so it can be part of reporting values.

weilixu (changes requested):
Need to address the comments

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,255 @@
...
+                    "fan_system": ["minimum_outdoor_airflow"],
+                },
+            )
+
+        def get_calc_vals(self, context, data=None):
+            hvac_b = context.baseline
+            hvac_id_b = hvac_b["id"]
+            zone_info = data["zone_info"][hvac_id_b]
+            all_design_setpoints_delta_Ts_are_per_reqs_b = zone_info[
+                "all_design_setpoints_delta_Ts_are_per_reqs"
```
weilixu, line 195: I think this shall be calculated in here, not in `create_data`

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,255 @@
...
+            are_all_hvac_sys_fan_objs_autosized_b = zone_info[
+                "are_all_hvac_sys_fan_objs_autosized"
+            ]
+
+            fan_sys_b = hvac_b["fan_system"]
+            supply_fans_airflow_b = (
+                get_fan_system_object_supply_return_exhaust_relief_total_power_flow(
+                    fan_sys_b
+                )
+            )
```
weilixu, line 206: We should either update the variable name to:
`fans_total_power_flow_b` or get the supply fans calculated values from the return value:
```python
fans_total_power_flow_b = get_fan_system_object_supply_return_exhaust_relief_total_power_flow(fan_sys_b)
supply_fan_qty = fans_total_power_flow_b["supply_fans_qty"]
``` 

For reporting purpose, I prefer the second approach which avoid having a nested dict in the calc_val value.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,255 @@
...
+                "supply_fans_airflow_b": supply_fans_airflow_b,
+                "fan_minimum_outdoor_airflow_b": fan_minimum_outdoor_airflow_b,
+            }
+
+        def manual_check_required(self, context, calc_vals=None, data=None):
+            hvac_b = context.baseline
+            hvac_id_b = hvac_b["id"]
+            zone_info = data["zone_info"][hvac_id_b]
+            supply_fans_airflow_b = calc_vals["supply_fans_airflow_b"]
+
```
weilixu, line 221: This is a dictionary, not a value - address the comments above.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,255 @@
...
+            return not all_design_setpoints_delta_Ts_are_per_reqs_b and std_equal(
+                zone_info["proposed_supply_airflow"], supply_fans_airflow_b
+            )
+
+        def get_manual_check_required_msg(self, context, calc_vals=None, data=None):
+            hvac_b = context.baseline
+            hvac_id_b = hvac_b["id"]
+
+            return f"{hvac_id_b} was not modeled based on a supply-air-to-room temperature set-point difference of 20°F (or 17°F, if lab). The baseline and proposed supply cfm was modeled identically at <insert proposed_supply_airflow> CFM. Manual review is required to determine if the airflow rate was modeled to comply with applicable codes or accreditation standards. If not, fail."
+
```
weilixu, line 235: You will need to insert an airflow value here.
Add the variable to CalcQ and then call calcq_to_str with a unit system to convert the value.
You can remove the CFM.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
yunjoonjung-PNNL: I think if there is no ``` .keys()``` in Python, a dict is iterated through keys. I'll double-check this when tested with TCD.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
yunjoonjung-PNNL: Good point! Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
JacksonJ-KC: the keys() method is available if you need to convert to a list of keys, but iterating through a dict will iterate through the keys as you said Yun Joon

rct229/rulesets/ashrae9012019/section19/section19rule13.py
yunjoonjung-PNNL: I believe this can't be done in the ```HVACRule``` level. Please let me know if I am missing something.

yunjoonjung-PNNL: However, I found that ```terminal_b``` should be ```terminal_id_b``` and made some updates this part.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
yunjoonjung-PNNL: Good point! Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,261 @@
...
+        dict_of_zones_and_terminal_units_served_by_hvac_sys_b = (
+            get_dict_of_zones_and_terminal_units_served_by_hvac_sys(rmi_b)
+        )
+
+        zone_info = {}
+        for hvac_id_b in dict_of_zones_and_terminal_units_served_by_hvac_sys_b:
+            for zone_id_b in dict_of_zones_and_terminal_units_served_by_hvac_sys_b[
+                hvac_id_b
+            ]["zone_list"]:
+                zone_info[zone_id_b] = {}
```
charliepnnl, line 78: I worry that this logic may not quite work.  The issue I _think_ I am seeing is that an hvac system can server more than one zone (the schema seems to allow that at least).  That means that the same `zone_id_b` can occur in more than one zone_list.  Any `zone_info[zone_id_b]` after the first one will overwrite what was already there.

I _think_ the solution is to put `dict_of_zones_and_terminal_units_served_by_hvac_sys_b` in `data` and then put the rest of the logic in `calc_vals` method.

Wherever a function here uses `rmi_b` you'll need to create a dictionary holding all the calls for that function.  For example:
```
are_all_hvac_sys_fan_objs_autosized_dict = {
    hvac_id_b: are_all_hvac_sys_fan_objs_autosized(rmi_b, hvac_id_b) for
    hvac_id_b in dict_of_zones_and_terminal_units_served_by_hvac_sys_b
}
```

charliepnnl (changes requested):
Most importantly, see my comment just above line 79.  I am worried that there is a subtle flaw in the logic that will require significant refactoring.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,261 @@
...
+            id="19-13",
+            description="For baseline system types 1-8 and 11-13, system design supply airflow rates shall be based on a supply-air-to-room temperature set-point difference of 20°F or the minimum outdoor airflow rate, or the airflow rate required to comply with applicable codes or accreditation standards, whichever is greater. For systems with multiple zone thermostat setpoints, use the design set point that will result in the lowest supply air cooling set point or highest supply air heating set point.",
+            ruleset_section_title="HVAC - General",
+            standard_section="Section G3.1.2.8.1 and Exception 1",
+            is_primary_rule=True,
+            rmr_context="ruleset_model_instances/0",
+            list_path="$.buildings[*].building_segments[*].heating_ventilating_air_conditioning_systems[*]",
+        )
+
+    def create_data(self, context, data):
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
yunjoonjung-PNNL: Addressed!

rct229/rulesets/ashrae9012019/section19/section19rule13.py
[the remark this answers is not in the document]
yunjoonjung-PNNL: Good point. Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,261 @@
...
+                        "Terminal",
+                        "supply_design_heating_setpoint_temperature",
+                    )
+                    design_cooling_supply_air_temp_setpoint = getattr_(
+                        terminal_b,
+                        "Terminal",
+                        "supply_design_cooling_setpoint_temperature",
+                    )
+
+                    if zone_has_lab_space:
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
yunjoonjung-PNNL: Great point! The fundamental issue was the ```zone_info``` dict had the ```zone_id```. Now, it was changed to have ```hvac_id``` instead. Thanks for this cooment!

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,261 @@
...
+    HVAC_SYS.SYS_7,
+    HVAC_SYS.SYS_8,
+    HVAC_SYS.SYS_11,
+    HVAC_SYS.SYS_12,
+    HVAC_SYS.SYS_13,
+]
+
+LIGHTING_SPACE = schema_enums["LightingSpaceOptions2019ASHRAE901TG37"]
+TEMP_DELTA_17F = 17.0 * ureg("R")
+TEMP_DELTA_20F = 20.0 * ureg("R")
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

weilixu (commented):
Some comments need to be addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,260 @@
...
+        rmi_b = context.baseline
+        rmi_p = context.proposed
+
+        dict_of_zones_and_terminal_units_served_by_hvac_sys_b = (
+            get_dict_of_zones_and_terminal_units_served_by_hvac_sys(rmi_b)
+        )
+
+        zone_info = {}
+        for hvac_id_b in dict_of_zones_and_terminal_units_served_by_hvac_sys_b:
+            zone_info[hvac_id_b] = {}
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,260 @@
...
+                "all_design_setpoints_delta_Ts_are_per_reqs": False,
+                "supply_flow_p": ZERO.TEMPERATURE,
+            }
+
+            for zone_id_b in dict_of_zones_and_terminal_units_served_by_hvac_sys_b[
+                hvac_id_b
+            ]["zone_list"]:
+                zone_info[hvac_id_b][
+                    "are_all_hvac_sys_fan_objs_autosized"
+                ] = are_all_hvac_sys_fan_objs_autosized(rmi_b, hvac_id_b)
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,260 @@
...
+        zone_info = {}
+        for hvac_id_b in dict_of_zones_and_terminal_units_served_by_hvac_sys_b:
+            zone_info[hvac_id_b] = {}
+            zone_info[hvac_id_b] = {
+                "are_all_hvac_sys_fan_objs_autosized": False,
+                "design_thermostat_cooling_setpoint": [],
+                "design_thermostat_heating_setpoint": [],
+                "zone_has_lab_space": False,
+                "all_design_setpoints_delta_Ts_are_per_reqs": False,
+                "supply_flow_p": ZERO.TEMPERATURE,
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,260 @@
...
+                zone_info[hvac_id_b]["design_thermostat_heating_setpoint"].append(
+                    zone_b.get("design_thermostat_heating_setpoint", ZERO.TEMPERATURE)
+                )
+
+                for space_b in find_all("$.spaces[*]", zone_b):
+                    zone_info[hvac_id_b]["zone_has_lab_space"] = (
+                        getattr_(space_b, "Space", "lighting_space_type")
+                        == LIGHTING_SPACE.LABORATORY_EXCEPT_IN_OR_AS_A_CLASSROOM
+                    )
+
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Good point! Thank you Weili! Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,260 @@
...
+                    "are_all_hvac_sys_fan_objs_autosized"
+                ] = are_all_hvac_sys_fan_objs_autosized(rmi_b, hvac_id_b)
+                zone_b = find_one(
+                    f'$.buildings[*].building_segments[*].zones[*][?(@.id = "{zone_id_b}")]',
+                    rmi_b,
+                )
+
+                zone_info[hvac_id_b]["design_thermostat_cooling_setpoint"].append(
+                    zone_b.get("design_thermostat_cooling_setpoint", ZERO.TEMPERATURE)
+                )
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule13.py
```diff
@@ -0,0 +1,260 @@
...
+        for hvac_id_b in dict_of_zones_and_terminal_units_served_by_hvac_sys_b:
+            zone_info[hvac_id_b] = {}
+            zone_info[hvac_id_b] = {
+                "are_all_hvac_sys_fan_objs_autosized": False,
+                "design_thermostat_cooling_setpoint": [],
+                "design_thermostat_heating_setpoint": [],
+                "zone_has_lab_space": False,
+                "all_design_setpoints_delta_Ts_are_per_reqs": False,
+                "supply_flow_p": ZERO.TEMPERATURE,
+            }
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Good point! Addressed.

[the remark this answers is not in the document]
yunjoonjung-PNNL: Good point! Addressed.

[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

weilixu (approved):
Good work, this is ready to merge now.

PR #1049, opened 2023-04-17
Wx/rd/get zone supply return exhaust relief terminal fan power dict

weilixu:


charliepnnl (approved):
This function is pretty complex.  To me honest, I'm not 100% sure I followed everything. But I'm pretty sure it is correct.

I just had a few comments.

PR #1050, opened 2023-04-17
RS/YJ/Rule19-22

yunjoonjung-PNNL:
This rule hasn't been tested with RMD. I'll test it once RMD is available.


JacksonJ-KC:
> This rule hasn't been tested with RMD. I'll test it once RMD is available.
New PR with test json for this rule!
https://github.com/pnnl/ruleset-checking-tool/pull/1053

yunjoonjung-PNNL:
Update: This rule was tested with the TCDs and all passed as expected.

rct229/rulesets/ashrae9012019/section19/section19rule22.py
```diff
@@ -0,0 +1,56 @@
...
+        def get_calc_vals(self, context, data=None):
+            hvac_b = context.baseline
+            fan_sys_b = hvac_b["fan_system"]
+
+            ER_operation = getattr_(
+                fan_sys_b,
+                "Fan System",
+                "air_energy_recovery",
+                "energy_recovery_operation",
+            )
```
weilixu, line 50: I don't think we should use `getattr_` here because we do not want to raise exception to undetermined results due to missing `air_energy_recovery` or `air_energy_recovery.energy_recovery_operation`
Instead I suggest to use `find_one("$.air_energy_recovery.energy_recovery_operation", fan_sys_b)`

the `er_operation` equals to `None` if data is missing, which in the `rule_check` funciton, it will return `False`.

weilixu (changes requested):
One comment needs to be addressed.

weilixu (approved):
Approved, good work

PR #1051, opened 2023-04-17
Complete the get hvac sys and assoc zones largest exhaust source

weilixu:


docs/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.md
```diff
@@ -7,7 +7,7 @@
...
+- **get_hvac_sys_and_assoc_zones_largest_exhaust_source**: Returns a dict with the sum of the hvac fan system exhaust fan cfm values, the maximum zone level exhaust source across the zones associated with the HVAC system, the number of exhaust fans associated with the hvac fan system, and the maximum cfm of all of the exhaust fans associated with the hvac system fan system {"hvac_sys_exhaust_cfm_sum": 3000, "maximum_zone_exhaust": 500, "num_hvac_exhaust_fans": 2, "maximum_hvac_exhaust": 1000}. This is for evaluating G3.1.2.10 exception 6 which states "Where the largest exhaust source is less than 75% of the design outdoor airflow. This exception shall only be used if exhaust air energy recovery is not used in the proposed design. "
```
charliepnnl, line 10: The units should not appear in the keys of the returned dictionary.  This must serve both IP and SI unit systems. So, `hvac_sys_exhaust_cfm_sum` should be `hvac_sys_exhaust_flow_sum` (or if you want to be more verbose, you could use `flow_rate` instead of `flow`.  Also, "maximum cfm" should be replaced with "maximum flow rate" in the accompanying text.

Should `maximum_zone_exhaust` be `maximum_zone_exhaust_flow`? and similarly for `maximum_hvac_exhuast`?

I don't think it is a good idea to include example values in the dictionary here. First, there is no mention that this is just an example.  Taken literally, this indicates that this particular dictionary is always returned. Second, the values are not numbers, they are Pint Quantities. I suggest wording something like:
```
Returns a dict with the following form:
{
   "hvac_sys_exhaust_flow_sum": the sum of the hvac fan system exhaust fan flow rates,
   "maximum_zone_exhaust_flow": the maximum zone level exhaust source across the zones associated with the HVAC system,
   "num_hvac_exhaust_fans": the number of exhaust fans associated with the hvac fan system,
   "maximum_hvac_exhaust_flow": the maximum flow rate of all of the exhaust fans associated with the hvac system fan system
}
```

rct229/rulesets/ashrae9012019/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.py
```diff
@@ -0,0 +1,91 @@
...
+    This is for evaluating G3.1.2.10 exception 6 which states "Where the largest exhaust source is less than 75% of the
+    design outdoor airflow. This exception shall only be used if exhaust air energy recovery is not used in the proposed
+    design.
+    """
+    zones_and_terminal_units_served_by_hvac_sys_dict = (
+        get_dict_of_zones_and_terminal_units_served_by_hvac_sys(rmi)
+    )
+    assert_(
+        hvac_sys_id in zones_and_terminal_units_served_by_hvac_sys_dict.keys(),
+        f"HVAC {hvac_sys_id} does not "
```
charliepnnl, line 52: To make this error message more clear:
```
f"HVAC system {hvac_sys_id} does not have associated terminals and zones. Check inputs!"
```

rct229/rulesets/ashrae9012019/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.py
```diff
@@ -0,0 +1,91 @@
...
+                    find_exactly_one_zone(rmi, zone_id),
+                )
+                for zone_id in zone_id_list
+            ],
+        )
+    )
+
+    return {
+        "hvac_fan_sys_exhaust_sum": sum(hvac_fan_sys_exhaust_flow_list, ZERO.FLOW),
+        "maximum_zone_exhaust": max(zone_fan_exhaust_flow_list)
```
charliepnnl, line 84: I believe you could shorten this to: `max(*zone_fan_exhaust_flow_list, ZERO.FLOW)`
And similarly for `maximum_hvac_exhaust` below.

rct229/rulesets/ashrae9012019/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source.py
```diff
@@ -0,0 +1,91 @@
...
+                find_one(
+                    "$.zonal_exhaust_fan.design_airflow",
+                    find_exactly_one_zone(rmi, zone_id),
+                )
+                for zone_id in zone_id_list
+            ],
+        )
+    )
+
+    return {
```
charliepnnl, line 82: I see that you removed the `cfm` from the return dict key.  That is good.  The RDS should be changed to match.

rct229/rulesets/ashrae9012019/ruleset_functions/get_hvac_sys_and_assoc_zones_largest_exhaust_source_test.py
```diff
@@ -0,0 +1,242 @@
+from rct229.rulesets.ashrae9012019.ruleset_functions.get_hvac_sys_and_assoc_zones_largest_exhaust_source import (
+    get_hvac_sys_and_assoc_zones_largest_exhaust_source,
+)
+from rct229.schema.schema_utils import quantify_rmr
+from rct229.schema.validate import schema_validate_rmr
+from rct229.schema.config import ureg
+
+TEST_RMD = {
```
charliepnnl, line 8: Isn't this a `TEST_RMI`? We should keep the distinction between RMI and RMD straight in our test code. The two are confused throughout this test code.

charliepnnl (approved):
Approved with comments that need to be addressed.

PR #1062, opened 2023-04-23
RS/YJ/RULE19-29

yunjoonjung-PNNL:
This rule hasn't been tested. I'll test it out when the TCD is available.


weilixu (approved):
This is the baseline version of rule 19-28.
Same comments applies to here.
Must address

JacksonJ-KC:
> This rule hasn't been tested. I'll test it out when the TCD is available.

https://github.com/pnnl/ruleset-checking-tool/pull/1090

rct229/rulesets/ashrae9012019/section19/section19rule29.py
```diff
@@ -0,0 +1,78 @@
...
+        return {"inapplicable_hvac_sys_list_b": inapplicable_hvac_sys_list_b}
+
+    class HVACRule(RuleDefinitionBase):
+        def __init__(self):
+            super(Section19Rule29.HVACRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+                required_fields={
+                    "$": ["fan_system"],
+                    "fan_system": [
+                        "operation_during_occupied",
```
JacksonJ-KC, line 54: @yunjoonjung-PNNL same comment as 19-28. Should this instead be `operation_during_unoccupied`?

yunjoonjung-PNNL:
This Rule was tested with TCD and passed as expected.

rct229/rulesets/ashrae9012019/section19/section19rule29.py
yunjoonjung-PNNL: Good catch. Addressed.

PR #1064, opened 2023-04-23
RS/YJ/RULE19-32

yunjoonjung-PNNL:
This rule hasn't been tested. I'll test it out when the TCD is available.


weilixu (approved):
This is baseline version of rule 31.
Same comments apply here as well.
Must address

JacksonJ-KC:
> This rule hasn't been tested. I'll test it out when the TCD is available.

https://github.com/pnnl/ruleset-checking-tool/pull/1092

yunjoonjung-PNNL:
This Rule was tested with TCD and passed as expected.

PR #1096, opened 2023-05-24
Updated scheduling to allow for 24 hour schedule

jugonzal07:
Updated scheduling to allow for 24 hour schedule by utilizing enumeration "CONSTANT_DAY" after SCHEDULE.

This value feature is valuable when setting a schedule's hourly values. For example, "SCHEDULE:CONSTANT_DAY-1.0" will set a 24 length array with a constant value of 1.0.


yunjoonjung-PNNL (approved):
Looks great to me. Thank you Juan!

Before merging, could you format (black) the ```is_economizer_modeled_in_proposed``` function? I think this function was modified without formatting. Thanks!

jugonzal07:
I'm unsure how a file I didn't change fails black formatting but I can push the fix for the black formatting if it helps get the check approved.

PR #1102, opened 2023-05-26
add get_zone_peak_internal_load_floor_area

weilixu:
Function for section 18.

### DIFF
--- a/rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_peak_internal_load_floor_area_dict.py
+++ b/rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_peak_internal_load_floor_area_dict.py
@@ -1,0 +1,89 @@
+from rct229.rulesets.ashrae9012019.ruleset_functions.baseline_systems.baseline_system_util import (
+    find_exactly_one_zone,
+    find_exactly_one_schedule,
+)
+from rct229.utils.assertions import getattr_
+from rct229.utils.jsonpath_utils import find_all
+from rct229.utils.pint_utils import ZERO
+
+
+def get_zone_peak_internal_load_floor_area_dict(rmi, zone_id):
+    """
+    Finds the peak coincident internal loads of a zone and returns the value in btu/h/ft2
+    The function returns a dict giving 2 values: {"PEAK":total peak btu/h/ft2 in the zone, "AREA":total zone area} the
+    total peak btu/sf is the internal non-coincident peak loads in all spaces in the zone
+
+    The function does not raise exception for missing values rather it applies default values.
+
+    Parameters
+    ----------
+    rmi: dict
+    A dictionary representing a RuleModelInstance object as defined by the ASHRAE229 schema
+    zone_id: string
+    zone id
+
+    Returns
+    -------
+    result: dict
+    a dictionary that contains two keys, peak, and area.
+    """
+    zone = find_exactly_one_zone(rmi, zone_id)
+    zone_area = ZERO.AREA
+    zone_load = ZERO.POWER
+
+    for space in find_all("$.spaces[*]", zone):
+        space_area = space.get("floor_area", ZERO.AREA)
+        zone_area += space_area
+        for light in find_all("$.interior_lighting[*]", space):
+            lighting_design_schedule = find_exactly_one_schedule(
+                rmi,
+                getattr_(light, "interior_lighting", "lighting_multiplier_schedule"),
+            )
+            lighting_max_schedule_fraction = max(
+                getattr_(
+                    lighting_design_schedule,
+                    "lighting_multiplier_schedule",
+                    "hourly_cooling_design_day",
+                )
+            )
+            zone_load += (
+                light.get("power_per_area", ZERO.POWER_PER_AREA)
+                * space_area
+                * lighting_max_schedule_fraction
+            )
+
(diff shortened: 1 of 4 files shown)


rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_peak_internal_load_floor_area_dict.py
```diff
@@ -0,0 +1,89 @@
...
+)
+from rct229.utils.assertions import getattr_
+from rct229.utils.jsonpath_utils import find_all
+from rct229.utils.pint_utils import ZERO
+
+
+def get_zone_peak_internal_load_floor_area_dict(rmi, zone_id):
+    """
+    Finds the peak coincident internal loads of a zone and returns the value in btu/h/ft2
+    The function returns a dict giving 2 values: {"PEAK":total peak btu/h/ft2 in the zone, "AREA":total zone area} the
```
charliepnnl, line 13: `PEAK` should be `peak` and `AREA` should be `area` in this documentation and in the RDS.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_peak_internal_load_floor_area_dict.py
```diff
@@ -0,0 +1,89 @@
...
+    zone_id: string
+    zone id
+
+    Returns
+    -------
+    result: dict
+    a dictionary that contains two keys, peak, and area.
+    """
+    zone = find_exactly_one_zone(rmi, zone_id)
+    zone_area = ZERO.AREA
```
charliepnnl, line 31: This could be replace with 
```
zone_area = sum(find_all("$.spaces.floor_area", zone), ZERO.AREA)
```
Not that it's a big deal.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_peak_internal_load_floor_area_dict.py
```diff
@@ -0,0 +1,89 @@
...
+    zone_load = ZERO.POWER
+
+    for space in find_all("$.spaces[*]", zone):
+        space_area = space.get("floor_area", ZERO.AREA)
+        zone_area += space_area
+        for light in find_all("$.interior_lighting[*]", space):
+            lighting_design_schedule = find_exactly_one_schedule(
+                rmi,
+                getattr_(light, "interior_lighting", "lighting_multiplier_schedule"),
```
charliepnnl, line 40: The notes for lighting_multiplier_schedule says:
"If the schedule is not present, the multiplier is always one. Constraint to use when implemented :Schedule:"

Does this mean that instead of throwing an exception if _lighting_multiplier_schedule_ is missing we should instead  take it to be a schedule with all ones?

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_peak_internal_load_floor_area_dict.py
```diff
@@ -0,0 +1,89 @@
...
+            )
+            lighting_max_schedule_fraction = max(
+                getattr_(
+                    lighting_design_schedule,
```
charliepnnl, line 44: I'm lost. In the schema, I don't see that a Schedule has a _llighting_multiplier_schedule_.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_peak_internal_load_floor_area_dict.py
```diff
@@ -0,0 +1,89 @@
...
+    find_exactly_one_schedule,
+)
+from rct229.utils.assertions import getattr_
+from rct229.utils.jsonpath_utils import find_all
+from rct229.utils.pint_utils import ZERO
+
+
+def get_zone_peak_internal_load_floor_area_dict(rmi, zone_id):
+    """
+    Finds the peak coincident internal loads of a zone and returns the value in btu/h/ft2
```
charliepnnl, line 12: It actually doesn't return the result in `btu/h/ft2`.  How about adding "or similar units" after that.

charliepnnl (approved):
Approved with comments that should be addressed and others that are merely style.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_peak_internal_load_floor_area_dict.py
weilixu: Since we will need space_area anyway in the loop, I will skip using this line for summation.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_peak_internal_load_floor_area_dict.py
weilixu: You are correct. Updated this code

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_peak_internal_load_floor_area_dict.py
weilixu: This is for reporting only. I updated to `Schedule`

PR #1103, opened 2023-06-06
TCD/JDJ/RULE19-17

JacksonJ-KC:


rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section19/rule_19_17.json
```diff
@@ -66,30 +66,30 @@
                                                     "supply_fans": [
                                                         {
                                                             "id": "Supply Fan 1",
-                                                            "design_airflow": 534.7164531455998,
                                                             "specification_method": "SIMPLE",
-                                                            "design_electric_power": 35
+                                                            "design_electric_power": 60,
+                                                            "design_airflow": 471.94744319999984
```
yunjoonjung-PNNL, line 71: To pass rule 17-17-a, the sum of fan powers /  supply airflow <= 0.3 (W/cfm). 

Could you slightly increase the ```design_airflow``` value? This is because currently, what you did is correct, however because of the conversion factor, the ```fan_power_W_CFM``` becomes 0.3000000000000002 watt / cfm, which becomes false.

JacksonJ-KC: While this is an easy edit to make to the TCD, I think it would be better for this to be corrected in a way that does not result in failed outcomes due to insignificant rounding of digits in conversion calculations. Can we implement rounding of the results, or an allowable margin of error?

yunjoonjung-PNNL: There is a ```std_equal``` function to compare if two values are the same with margins (default 5%). However, we don't consider the margin when we compare if one value is greater/less than the other. I believe this is the first time to face this issue.

As you said, we can update our code to allow the margin or error (we also may need to dive more into the ```pint``` package to see whether there is still this issue in the newer version). However, the right procedure would be to bring up this topic during the meeting and discuss this with other team members. Could you do this during the TCD meeting? If accepted, we should update all the parts in the code that used ```>```, ```<```, ```>=```, and ```<=```.

While doing this, could we update the value to move forward?

JacksonJ-KC: The TCD meeting was cancelled because it no longer seemed necessary after Karen and I were up to speed but I will email the team to get feedback, and I will update this value so that it passes testing. Thanks!

yunjoonjung-PNNL: Thank you Jackson!

yunjoonjung-PNNL (approved):

JacksonJ-KC:
> @JacksonJ-KC Could you check this TCD and resolve the conflict?

All set!

yunjoonjung-PNNL:
@JacksonJ-KC 
This TCD used to pass/fail correctly. However, rule (a) doesn't pass this time for some reason. It's because ```total fan power/supply airflow``` equal 0.30000088235553635 watt/cfm. I believe you did it correctly in Excel and it may've happened during the conversion. Could you make the ```total fan power/supply airflow``` less than 0.30 (maybe 0.28 or so to be safe)?

JacksonJ-KC:
> @JacksonJ-KC This TCD used to pass/fail correctly. However, rule (a) doesn't pass this time for some reason. It's because `total fan power/supply airflow` equal 0.30000088235553635 watt/cfm. I believe you did it correctly in Excel and it may've happened during the conversion. Could you make the `total fan power/supply airflow` less than 0.30 (maybe 0.28 or so to be safe)?

@yunjoonjung-PNNL This confused me a bit. A couple of weeks ago (I think there was a deadline for PNNL) when the unmerged TCDs were moved onto a separate tab of the workbook,  I noticed that rule 12a had been changed and I assumed that it was you making some adjustments to get the TCD to pass. 

When I used that change in [this previous commit](https://github.com/pnnl/ruleset-checking-tool/pull/1103/commits/3a4d4a52bf4b6dd2d4c8f85fade94fedcdd22177), the automated 'Run unit tests' check performed by github resulted in all Pass but then you provided this comment that rule 12a is not passing. 

I have now reverted to the way I originally wrote the TCD and also provided an additional TCD with even lower fan power (0.28 as you suggested). But now the 'Run unit tests' is no longer passing.

PR #1110, opened 2023-06-20
Software Test Report

jugonzal07:
First cut at adding a software test report.


jugonzal07:
Unsure why black formatting required so many changes-- I would think they would've been caught before. Perhaps I'm using an old version?

rct229/ruletest_engine/ruletest_engine.py
```diff
@@ -337,6 +343,208 @@ def run_section_tests(test_json_name: str, ruleset_doc: str):
...
+        "23": "HVAC-Airside?",
+    }
+
+    # Maps excel enumerations for pass/fail etc. to RCTOutcomeLabel. Unfortunately there's a disconnect.
+    ruletest_outcome_dict = {
+        "pass": RCTOutcomeLabel.PASS,
+        "fail": RCTOutcomeLabel.FAILED,
+        "undetermined": RCTOutcomeLabel.UNDETERMINED,
+        "not_applicable": RCTOutcomeLabel.NOT_APPLICABLE,
+        "manual_check": RCTOutcomeLabel.NOT_APPLICABLE,
```
jugonzal07, line 435: We should address this to be consistent. Perhaps these were supposed to be undetermined come to think of it. Addressing this would require updating the ruletest spreadsheets and ruletests. Not a major lift.

rct229/ruletest_engine/ruletest_engine.py
```diff
@@ -337,6 +343,208 @@ def run_section_tests(test_json_name: str, ruleset_doc: str):
...
+        ashrae901_2019_software_test_report
+
+    """
+
+    # Master list of RCT engine outcomes and invalid RMR messages used to populate starting point for an RCTReport.
+    # Initialize them here
+    rct_outcomes_list = []
+    invalid_rmr_messages = []
+
+    # Maps section lists to their titles TODO: resolve 19 vs 23
```
jugonzal07, line 417: This is something I could use help with.

rct229/reports/ashrae901_2019_software_test_report.py
```diff
@@ -0,0 +1,76 @@
...
+        self.title = "ASHRAE STD 229P RULESET CHECKING TOOL"
+        self.purpose = "RCT Ruleset Software Testing Report"
+        self.ruleset = "ASHRAE 90.1-2019 Performance Rating Method (Appendix G)"
+        self.schema_version = "0.0.23"
+        self.date_run = str(datetime.now())
+        self.ruleset_report_file = "ashrae901_2019_software_testing_report.json"
+
+    def initialize_ruleset_report(self):
+        report_json = dict()
+        report_json["title"] = self.title
```
yunjoonjung-PNNL, line 22: This is a great way to set the values. Another way could be (and can delete line 21 - just as a reference)
```python
report_json = {
"title":  self.title,
"purpose": self.purpose,
 "ruleset": self.ruleset,
"date_run": self.date_run,
"schema_version": self.schema_version,
"rule_tests": [],}
```
Also, depending on your coding style, I'll leave this up to you!

rct229/ruletest_engine/ruletest_engine.py
```diff
@@ -337,6 +343,208 @@ def run_section_tests(test_json_name: str, ruleset_doc: str):
...
+        List of test JSON directorys in 'test_jsons/[MY_STANDARD]' directory. (e.g., ['section5', 'section6'])
+
+    output_json_path: str
+
+        output_dir: str - directory in which you want the ashrae901_2019_software_testing_report.json to appear
+
+
+    """
+
+    # Initialize report dictionary in which to
```
yunjoonjung-PNNL, line 366: Could we have a ```TO-DO``` comment here to add more versions of the ruleset in the future?

yunjoonjung-PNNL (approved):
Great work! I left a few (optional) comments. I'd appreciate it if you looked into these.

rct229/reports/ashrae901_2019_software_test_report.py
```diff
@@ -0,0 +1,76 @@
...
+        super(ASHRAE9012019SoftwareTestReport, self).__init__()
+        self.title = "ASHRAE STD 229P RULESET CHECKING TOOL"
+        self.purpose = "RCT Ruleset Software Testing Report"
+        self.ruleset = "ASHRAE 90.1-2019 Performance Rating Method (Appendix G)"
+        self.schema_version = "0.0.23"
+        self.date_run = str(datetime.now())
+        self.ruleset_report_file = "ashrae901_2019_software_testing_report.json"
+
+    def initialize_ruleset_report(self):
+        report_json = dict()
```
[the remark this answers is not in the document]
jugonzal07: Great to know! I don't have a strong attachment one way or another. I can update this.

rct229/ruletest_engine/ruletest_engine.py
jugonzal07: Sure it's not a problem to add that. If another ruletest is is ran, you'll note that it raises an exception to let you know it hasn't been set up yet.

yunjoonjung-PNNL:
Juan, For some reason, the ```section6_mater.json``` file only includes 6-1 TCD. Could you check this out?

rct229/ruletest_engine/ruletest_engine.py
jugonzal07: Rough categories for each section per Weili on call:

18 - HVAC-BaselineHVAC
19 - HVAC-Airside
23 - HVAC-SystemSpecificRequirements

jugonzal07:
> Juan, For some reason, the `section6_mater.json` file only includes 6-1 TCD. Could you check this out?

Thank you YJ, I addressed this and the other issues! I hope that resolves things. Let me know if you are unsuccessful when running this test report.

yunjoonjung-PNNL:
@jugonzal07 I looked into why the unit tests failed. 
For errors from section 19, I believe this is because the ```section19_master.json``` file includes the RMDS whose rules haven't been merged (e.g., 19-4, etc.) or has some errors (Jackson and I have worked together to get these corrected).
For the errors from section 5 (e.g., 5-3), this is because the expected outcome in the RMD was wrong, as we discussed yesterday.

jugonzal07:
> @jugonzal07 I looked into why the unit tests failed. For errors from section 19, I believe this is because the `section19_master.json` file includes the RMDS whose rules haven't been merged (e.g., 19-4, etc.) or has some errors (Jackson and I have worked together to get these corrected). For the errors from section 5 (e.g., 5-3), this is because the expected outcome in the RMD was wrong, as we discussed yesterday.

Thank you! Yes, I forgot the software pytest reads every JSON in the directory, including this new master JSON. I suppose there's two solutions.

1) Exclude the master JSON from testing. This would be ok once we resolve discrepancies between master JSON and the disaggregated JSONs but it's a bit janky to do so.
2) Remove the master JSONs and instead have the software test report read individual rule JSONs but run slower. I will do a test to see how much slower it is to do so later. It'd help solve the redundancy of having a master JSON and individual rule JSONs. If it's not much slower, than perhaps it simplifies the process.

weilixu (approved):
Tested with master jsons, works great!

There would be changes to the software testing report. We can leave them to another PR after the report schema is finalized.

weilixu:
Some of rules are failing in this branch, we need to fix them first before merging.

PR #1118, opened 2023-07-03
RCT/YJ/Remove .keys()

yunjoonjung-PNNL:
This PR removes all the ```.keys()``` in the code.


charliepnnl (approved):
Approved with a couple of comments here to consider, and ignore if you like.

I'm not sure I agree with removing `.keys()` from the python code, but I'm ok with it.  To me, it makes the code slightly easier to read.

I do not think removing `.keys()` from the md files is a good idea. It requires the reader of the RDS to know the intricacies of python.

yunjoonjung-PNNL:
> 

Thanks for sharing your insight!

1) remove ```.keys()``` in Python: I believe we'd be able to have slightly more neat code by implementing this.  I think this is why Weili assigned this task. 

2) keep ```.keys()``` in RDS: I agree with you. I'll leave it as is until @weilixu shares how he thinks about this. If Weili also agrees, I'll revert the RDS files to have ```.keys()```.

weilixu:
> Approved with a couple of comments here to consider, and ignore if you like.
> 
> I'm not sure I agree with removing `.keys()` from the python code, but I'm ok with it. To me, it makes the code slightly easier to read.
> 
> I do not think removing `.keys()` from the md files is a good idea. It requires the reader of the RDS to know the intricacies of python.

1. It seems removing the .keys() is more pythonic in a way that it turned out to be slightly faster than calling .keys().
2. Agree, we do not remove .keys() from the MD files.

yunjoonjung-PNNL:
> 

Good point! I reverted the RDS files back to original. I'd appreciate it if you could review this PR again!

weilixu (approved):

PR #1122, opened 2023-07-17
Create is_hvac_sys_cooling_type_none_or_non_mechanical.md

claperle:


weilixu:
This is exactly the same as the `is_hvac_sys_cooling_type_none` but with one more condition `non_mechanical`. Approved for merge.

PR #1129, opened 2023-07-19
Complete the development of is_zone_mechanically_cooled

weilixu:


rct229/rulesets/ashrae9012019/ruleset_functions/g311_exceptions/g311_sub_functions/is_zone_mechanically_cooled.py
```diff
@@ -0,0 +1,88 @@
+import pydash
+from pydash import flat_map, chain, flow, is_equal
```
yunjoonjung-PNNL, line 2: It seems like the ```chain``` and ```is_equal``` functions aren't used.  Also ```import pydash``` isn't used. Could we remove them?

yunjoonjung-PNNL (approved):
Great work! Thank you Weili!

rct229/rulesets/ashrae9012019/ruleset_functions/g311_exceptions/g311_sub_functions/is_zone_mechanically_heated_and_not_cooled.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # function to check all terminals to check at least one terminal has a cooling source
+    hvac_heating_terminal_func = flow(
+        lambda z_id: find_exactly_one_zone(rmi, z_id),
+        lambda z: find_all("$.terminals[*]", z),
+        lambda terminal_list: flat_map(terminal_list, is_terminal_non_heating),
+        any,
+    )
+
+    is_heated = any(
```
charliepnnl, line 63: I believe `hvac_heating_sys_func` returns true if the heating system is None. So I think the over all here logic is backwards.

I think what happened is that the function names are not descriptive enough and so the logic got confused.
I suggest:
- Use function definitions to wrap the flow logic so that the arguments are clearer.
- Do not separate the `is_hvac_system_non_heating` and `is_terminal_non_heating` logic into their own functions. Instead, put that logic where it is used.
- Change the names of `hvac_heating_sys_func` and `hvac_heating_terminal_func ` to better match their intent.

It looks like a similar logic reversal happened for the is_zone_mechanically_coooled_and_not_heated case also.

charliepnnl (changes requested):
I believe the logic is reversed in these functions. See my note.

PR #1135, opened 2023-07-28
Rule 19-12 corrections

jugonzal07:
Required high limit temperatures were in the wrong units. Also updated std_equal call to reference units in absolute temperature scales (e.g., Kelvin in this case).


weilixu (approved):
I approved the changes however, I vaguely remember there was a complication with the temperature unit conversion. @yunjoonjung-PNNL should review the change and make sure it is correct.

yunjoonjung-PNNL (approved):
Thank you so much for finding/correcting this Juan. Looks great to me!

PR #1136, opened 2023-07-31
Rd/wx/get zone eflh

weilixu:


yunjoonjung-PNNL (approved):
Great work! Thank you Weili. It seems like using the ```flow``` function makes the code much more readable :).

PR #1139, opened 2023-08-14
RCT/YJ/Replace pint_sum with sum function

yunjoonjung-PNNL:
This PR addresses replacing the ```pint_sum``` function with the ```sum``` function.

### DIFF
--- a/rct229/utils/pint_utils.py
+++ b/rct229/utils/pint_utils.py
@@ -147,7 +147,7 @@ def calcq_to_str(unit_system, obj) -> str:
     return retval
 
 
-def pint_sum(qty_list, default=None):
+def sum(qty_list, default=None):
     if len(qty_list) == 0:
         assert default is not None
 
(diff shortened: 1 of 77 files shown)


rct229/utils/pint_utils.py
```diff
@@ -147,7 +147,7 @@ def calcq_to_str(unit_system, obj) -> str:
     return retval
 
 
-def pint_sum(qty_list, default=None):
+def sum(qty_list, default=None):
```
weilixu, line 150: we do not need to change this function name from `pint_sum` to `sum`.
The intention was to replace the `pint_sum` with the python native `sum` function

yunjoonjung-PNNL: Good point! Addressed.

weilixu (approved):

charliepnnl (approved):
This looks good. 
I did not test it. If you have, I am happy.

charliepnnl:
I did check that there are no more instances of pint_sum in use.

yunjoonjung-PNNL:
> This looks good. 
> 
> I did not test it. If you have, I am happy.

Thank you Charlie. All the rules passed and I believe this is a good sign to merge this PR.

PR #1142, opened 2023-08-24
Bug fixes and update for system to zone assignment 

jugonzal07:
- Updated references to ruleset_model_instances to ruleset_model_descriptions to update to 0.29 schema.
- Fixed issue where HVAC assignments were only being applied to the first Test ID in a ruletest excel file
- Fixed issue where setting multizone air loop names broke code if no HVAC systems had been set in a building segment yet

### DIFF
--- a/rct229/ruletest_engine/ruletest_jsons/scripts/excel_to_test_json_utilities.py
+++ b/rct229/ruletest_engine/ruletest_jsons/scripts/excel_to_test_json_utilities.py
@@ -605,16 +605,17 @@ def set_systems_to_zones(json_dict, system_to_zone_dict, rule_set):
         add_plant_loop_equipment(json_dict, system_rmd)
 
 
-def get_rmr_triplet_dicts_from_ruletest_json_dict(ruletest_json_dict):
-    """Reads in a ruletest JSON dictionary and returns list of RMRs for any triplet flagged in the ruletest JSON
-    structure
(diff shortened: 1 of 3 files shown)


rct229/ruletest_engine/ruletest_jsons/scripts/excel_to_test_json_utilities.py
```diff
@@ -980,15 +980,15 @@ def add_plant_loop_equipment(template_rmd, system_rmd):
         "heat_rejections",
     ]
 
-    sys_ruleset_model_instance = system_rmd["ruleset_model_instances"][0]
+    sys_ruleset_model_instance = system_rmd["ruleset_model_descriptions"][0]
```
yunjoonjung-PNNL, line 983: Would it be more clear if we changed the ```sys_ruleset_model_instance``` variable name to ```sys_ruleset_model_descriptions```?

yunjoonjung-PNNL (approved):
Thank you Juan. My comment is just a suggestion (optional) and I'd appreciate it if you could consider it before merging.

rct229/ruletest_engine/ruletest_jsons/scripts/excel_to_test_json_utilities.py
jugonzal07: Yes it would! Thanks for catching that.

weilixu (approved):

PR #1144, opened 2023-08-31
Excel to Test JSON - Multiple MZ Air Loop Fan Name Fix

jugonzal07:
Fixed issue where multiple MZ air loops in the same building segment would not always have unique fan IDs and system names when doing excel system to zone assignment. This update addresses that and now let's the latest section 18 spreadsheet create a valid 229 JSON file per the schema.


yunjoonjung-PNNL (approved):
Excellent work! Thank you Juan!

PR #1148, opened 2023-09-10
RS/YJ/Rule 18-2

yunjoonjung-PNNL:
1. This rule hasn't been tested out yet. I'll do it once the TCD is available.
2. I couldn't modify the RDS since it hasn't been merged to develop yet. I'll do this after it's available in the develop branch.
Thank you!


rct229/rulesets/ashrae9012019/section18/section18rule2.py
```diff
@@ -0,0 +1,290 @@
...
+            if baseline_system_type_compare(sys_type, target_sys_type, False)
+            for hvac_id in baseline_system_types_dict_b[sys_type]
+        ]
+
+        zones_and_terminal_unit_list_dict_b = (
+            get_dict_of_zones_and_terminal_units_served_by_hvac_sys(rmd_b)
+        )
+        lab_zone_hvac_systems = get_lab_zone_hvac_systems(
+            rmd_b, rmd_p, climate_zone_b, False
+        )  # TODO the last arg is leap_year
```
weilixu, line 89: Need to complete this TODO

rct229/rulesets/ashrae9012019/section18/section18rule2.py
```diff
@@ -0,0 +1,290 @@
...
+                        ]
+                    ),
+                    "does_sys_only_serve_lab_b": (
+                        hvac_id_b in lab_zone_hvac_systems["lab_zones_only"]
+                        and lab_zone_exhaust_b > AIRFLOW_15000_CFM
+                        and len(lab_zone_hvac_systems["lab_zones_only"]) == 1
+                    ),
+                    "does_sys_serve_lab_b": (
+                        hvac_id_b in lab_zone_hvac_systems["lab_zones_only"]
+                        and lab_zone_exhaust_b <= AIRFLOW_15000_CFM
```
weilixu, line 115: I am not sure if this logic needs `lab_zone_exhaust_b <= AIRFLOW_15000_CFM`

rct229/rulesets/ashrae9012019/section18/section18rule2.py
```diff
@@ -0,0 +1,290 @@
...
+
+        zones_and_terminal_unit_list_dict_b = (
+            get_dict_of_zones_and_terminal_units_served_by_hvac_sys(rmd_b)
+        )
+        lab_zone_hvac_systems = get_lab_zone_hvac_systems(
+            rmd_b, rmd_p, climate_zone_b, False
+        )  # TODO the last arg is leap_year
+        lab_zone_exhaust_b = get_building_total_lab_exhaust_from_zone_exhaust_fans(
+            rmd_b
+        )
```
weilixu, line 92: Could you update the name to `building_total_lab_zone_exhaust`? This value is for entire building, not a zone. The name is a bit confusing.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
```diff
@@ -0,0 +1,290 @@
...
+                hvac_data_b[hvac_id_b] = {
+                    "is_zone_single_zone_sys_b": any(
+                        [
+                            sys_type in APPLICABLE_SYS_TYPES
+                            for sys_type in APPLICABLE_SYS_TYPES
+                        ]
+                    ),
+                    "does_sys_only_serve_lab_b": (
+                        hvac_id_b in lab_zone_hvac_systems["lab_zones_only"]
+                        and lab_zone_exhaust_b > AIRFLOW_15000_CFM
```
weilixu, line 110: the `lab_zone_exhaust_b` may not need to add here since the variable name does not suggest the flag include the `lab_zone_exhaust` logic.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
```diff
@@ -0,0 +1,290 @@
...
+            rmd_b,
+        ):
+            if hvac_id_b in applicable_hvac_sys_ids_b:
+                hvac_data_b[hvac_id_b] = {
+                    "is_zone_single_zone_sys_b": any(
+                        [
+                            sys_type in APPLICABLE_SYS_TYPES
+                            for sys_type in APPLICABLE_SYS_TYPES
+                        ]
+                    ),
```
weilixu, line 107: I dont quite understand this logic. isn't it always True? because `sys_type` is always in `APPLICABLE_SYS_TYPES`?
or do you mean `hvac_id_b in APPLICABLE_SYS_TYPES[sys_type]`?

rct229/rulesets/ashrae9012019/section18/section18rule2.py
```diff
@@ -0,0 +1,290 @@
...
+                        and lab_zone_exhaust_b <= AIRFLOW_15000_CFM
+                    ),
+                }
+
+                do_multi_zone_evaluation = (
+                    hvac_data_b[hvac_id_b]["is_zone_single_zone_sys_b"]
+                    or not hvac_data_b[hvac_id_b]["does_sys_only_serve_lab_b"]
+                    or not hvac_data_b[hvac_id_b]["does_sys_serve_lab_b"]
+                    or not hvac_data_b[hvac_id_b]["does_sys_serve_lab_and_other_b"]
+                )
```
weilixu, line 128: Will need to add the `lab_zone_exhaust_b` in here. 
In RDS, the `do_multi_zone_evaluation` is initialized as a True flag. It turned to False if any one of the conditions met. So it will be `not any([hvac_data_b[hvac_id_b]["does_sys_only_serve_lab_b"], hvac_data_b[hvac_id_b]["does_sys_serve_lab_b"] ....])`

rct229/rulesets/ashrae9012019/section18/section18rule2.py
```diff
@@ -0,0 +1,290 @@
...
+                            # check if hvac_id_b and hvac_sys2_id_b are the same system
+                            if hvac_sys2_id_b != hvac_id_b:
+                                zones_served_by_system2 = (
+                                    zones_and_terminal_unit_list_dict_b[hvac_sys2_id_b][
+                                        "zone_list"
+                                    ]
+                                )
+                                if set(zones_served_by_system2).intersection(
+                                    set(zones_on_floor)
+                                ):
```
weilixu, line 159: The condition is whether the length of set will be equal to 0 after intersection.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
```diff
@@ -0,0 +1,290 @@
...
+                                    and lab_zone_exhaust_b <= AIRFLOW_15000_CFM
+                                ):
+                                    hvac_data_b[hvac_id_b][
+                                        "does_two_sys_exist_on_same_fl_b"
+                                    ] = "undetermined"
+
+                                else:
+                                    hvac_data_b[hvac_id_b][
+                                        "does_two_sys_exist_on_same_fl_b"
+                                    ] = False
```
weilixu, line 188: If there are three labels, I think we should set all of them to a string (true, false and undetermined).
Also, I think this shall be an any logic for a list, but in this case, the logic is just keep overriding the flag with new booleans.

weilixu (commented):
There are a few issues need to be addressed.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
yunjoonjung-PNNL: I believe we still need ```lab_zone_exhaust_b``` because of [this line in RDS](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section18/Rule18-2.md#:~:text=do_multi_zone_evaluation%20to%20FALSE%3A-,if%20lab_zone_exhaust%20%3E%2015000,-%3A%20result%20%3D%20PASS%3B%20do_multi_zone_evaluation)?

rct229/rulesets/ashrae9012019/section18/section18rule2.py
yunjoonjung-PNNL: Good point. I updated the logic.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
yunjoonjung-PNNL: I believe ```lab_zone_exhaust_b``` is needed because when result's pass/faill is determined there is ```if lab_zone_exhaust  > 15000``` condition. Please let me know if I am misunderstanding something.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section18/section18rule2.py
weilixu: I will reply one for both instances.
It took me a while to recall my memory but here is my thought process on these two comments:
The name of the variable suggest whether the system (only) serves labs so it is not obvious to add `lab_zone_exhaust_b` in the logic.
Another point is `lab_zone_exhaust_b` is at building level, so it does not need to combine with any pre-determined flag. 
However, I do think the logic here are correct. So I would not force the change in these places.
Apologize for the confusion.

yunjoonjung-PNNL: I see! Thank you for explaining Weili. It's very helpful.

rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section18/rule_18_3.json
```diff
@@ -273,7 +273,7 @@
                                                 "spaces": [
                                                     {
                                                         "id": "Space 3",
-                                                        "floor_area": 40000
+                                                        "floor_area": 3716.1215999999995
```
yunjoonjung-PNNL, line 276: Revised 18-3 TCD was added here b/c this is a very minor update.

PR #1150, opened 2023-09-18
Account for heating and cooling only systems in excel to test JSON framework

jugonzal07:
Added small check to ensure heating/cooling only baseline systems can be use system to zone assignment logic in the excel_to_test_json logic.


yunjoonjung-PNNL (approved):
Great work! Thank you Juan.

weilixu (approved):

PR #1155, opened 2023-10-06
Chiller master JSON with heat rejection TCDs

jugonzal07:
Updated chiller master JSON with heat rejection TCDs for Section 22 rules12,13, 15,17,18

PR #1156, opened 2023-10-06
RS/YJ/Rule 22-12

yunjoonjung-PNNL:
This rule was tested with Juan's TCD from PR #1155. I included the TCD here. Thanks!

### DIFF
--- a/rct229/rulesets/ashrae9012019/section22/section22rule12.py
+++ b/rct229/rulesets/ashrae9012019/section22/section22rule12.py
@@ -42,10 +42,8 @@ class Section22Rule12(RuleDefinitionListIndexedBase):
         number_of_baseline_heat_rejections_b = sum(
             [
                 1
-                for heat_rejection_b in find_all(
-                    "$.buildings[*.].heat_rejections[*]", rmd_b
-                )
-                if getattr_("$.loop", "loop", heat_rejection_b)
+                for heat_rejection_b in find_all("$.heat_rejections[*]", rmd_b)
+                if getattr_(heat_rejection_b, "heat_rejections", "loop")
                 in heat_rejection_loop_ids_b
             ]
         )
@@ -62,11 +60,13 @@ class Section22Rule12(RuleDefinitionListIndexedBase):
             )
 
         def get_calc_vals(self, context, data=None):
+            heat_rejection_loop_ids_b = data["heat_rejection_loop_ids_b"]
             number_of_baseline_heat_rejections_b = data[
                 "number_of_baseline_heat_rejections_b"
             ]
             return {
-                "number_of_baseline_heat_rejections_b": number_of_baseline_heat_rejections_b
+                "heat_rejection_loop_ids_b": heat_rejection_loop_ids_b,
+                "number_of_baseline_heat_rejections_b": number_of_baseline_heat_rejections_b,
             }
 
         def rule_check(self, context, calc_vals=None, data=None):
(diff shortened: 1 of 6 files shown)


rct229/rulesets/ashrae9012019/section22/section22rule12.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+
+        number_of_baseline_heat_rejections_b = sum(
+            [
+                1
+                for heat_rejection_b in find_all("$.heat_rejections[*]", rmd_b)
+                if getattr_(heat_rejection_b, "heat_rejections", "loop")
+                in heat_rejection_loop_ids_b
+            ]
+        )
```
weilixu, line 49: Another way to do it could be:

```python
number_of_baseline_heat_rejections_b = len(
            [
                heat_rejection_b
                for heat_rejection_b in find_all("$.heat_rejections[*]", rmd_b)
                if getattr_(heat_rejection_b, "heat_rejections", "loop")
                in heat_rejection_loop_ids_b
            ]
        )
```

rct229/rulesets/ashrae9012019/section22/section22rule12.py
```diff
@@ -0,0 +1,106 @@
...
+
+        return {
+            "heat_rejection_loop_ids_b": heat_rejection_loop_ids_b,
+            "number_of_baseline_heat_rejections_b": number_of_baseline_heat_rejections_b,
+        }
+
+    class HeatRejectionRule(RuleDefinitionBase):
```
weilixu, line 56: I don't think you need this nested class.
This can be done at the rule level.

rct229/rulesets/ashrae9012019/section22/section22rule12.py
```diff
@@ -0,0 +1,106 @@
...
+            number_of_baseline_heat_rejections_b = calc_vals[
+                "number_of_baseline_heat_rejections_b"
+            ]
+            len_heat_rejection_loop_ids_b = len(heat_rejection_loop_ids_b)
+
+            if (
+                number_of_baseline_heat_rejections_b == 1
+                and len_heat_rejection_loop_ids_b > 1
+            ):
+                FAIL_MSG = "There is more than one condenser loop for this building. There should only be one condenser loop attached to all chillers in the baseline chiller plant."
```
weilixu, line 94: This should not be the building, but the project.

Let's update the message to:
"There is more than one condenser loop for this project. There should only be one condenser loop attached to all chillers in the baseline chiller plant."

weilixu (commented):
I think we do not need child class here, we will need to rewrite it.

rct229/rulesets/ashrae9012019/section22/section22rule12.py
yunjoonjung-PNNL: Good point! Addressed.

rct229/rulesets/ashrae9012019/section22/section22rule12.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section22/section22rule12.py
yunjoonjung-PNNL: Good point! Addressed.

weilixu (approved):

PR #1157, opened 2023-10-06
RS/YJ/Rule 22-13

yunjoonjung-PNNL:
This rule was tested with Juan's TCD from PR https://github.com/pnnl/ruleset-checking-tool/pull/1155. I included the TCD here. Thanks!

### DIFF
--- a/rct229/rulesets/ashrae9012019/section22/section22rule13.py
+++ b/rct229/rulesets/ashrae9012019/section22/section22rule13.py
@@ -28,14 +28,6 @@ class Section22Rule13(RuleDefinitionListIndexedBase):
             list_path="$.heat_rejections[*]",
         )
 
-    def is_applicable(self, context, data=None):
-        rmd_b = context.baseline
-        heat_rejection_loop_ids_b = (
-            get_heat_rejection_loops_connected_to_baseline_systems(rmd_b)
-        )
-
-        return heat_rejection_loop_ids_b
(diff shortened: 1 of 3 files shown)


rct229/rulesets/ashrae9012019/section22/section22rule13.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+        return {"heat_rejection_loop_ids_b": heat_rejection_loop_ids_b}
+
+    def list_filter(self, context_item, data):
+        heat_rejection_b = context_item.baseline
+        heat_rejection_loop_ids_b = data["heat_rejection_loop_ids_b"]
+        heat_rejection_loop_b = getattr_(heat_rejection_b, "heat_rejections", "loop")
+
+        return heat_rejection_loop_b in heat_rejection_loop_ids_b
```
weilixu, line 52: Ii think we will need to move this to `is_applicable` function in the `HeatRejectionRule` class follows the applicability checks section in the RDS.

weilixu (approved):
Approved with one comment must address

rct229/rulesets/ashrae9012019/section22/section22rule13.py
weilixu: Remove this `list_filter` since it is already being moved to is_applicable.

PR #1162, opened 2023-10-10
Rd/wx/modify structure

weilixu:
First attempt modifying the structure.
This modification accomplished a list of items:
1. Remove dependency on 90.1 based ruleset model types.
2. Support future schema extensions (sub-schemas)
3. Support multiple input approaches such as multi-RPD files and single-RPD file with multiple RMDs


weilixu:
I have most of the structure completed.
The things left here is test module @jugonzal07, we will need to discuss how to update it.

weilixu:
@jugonzal07 test module is updated.
Some are still hard-coded but will removing them once every spreadsheets can be regenerated.

jugonzal07:
> @jugonzal07 test module is updated. Some are still hard-coded but will removing them once every spreadsheets can be regenerated.

Looks good to me! Only minor comment is that `add_ruleset_model_types` is added in both the excel_to_test_json.py and excel_to_test_json_utilities.py. 

Great work! That was a ton of refactoring.

jugonzal07 (approved):

weilixu:
> `add_ruleset_model_types`

Thanks Juan, I removed the add_ruleset_model_types in the excel_to_test_json.py. We just need one.

PR #1165, opened 2023-10-26
RS/YJ/Rule 19-36

yunjoonjung-PNNL:
This rule hasn't been tested yet. I'll work with Juan to test out this rule. Thanks!

PR #1167, opened 2023-10-27
RS/YJ/Rule 19-30

yunjoonjung-PNNL:
This rule hasn't been tested yet. I'll work with Juan to test this out. Thanks!

### DIFF
--- a/rct229/rulesets/ashrae9012019/section19/section19rule30.py
+++ b/rct229/rulesets/ashrae9012019/section19/section19rule30.py
@@ -42,13 +42,7 @@ class Section19Rule30(RuleDefinitionListIndexedBase):
                 rmrs_used=UserBaselineProposedVals(False, True, False),
             )
 
-        def get_manual_check_required_msg(self, context, calc_vals=None, data=None):
-            hvac_b = context.baseline
-            hvac_id_b = hvac_b["id"]
-
-            return f"FOR {hvac_id_b} perform manual check that only the terminal-unit fan and reheat coil are being energized to meet the heating set point during unoccupied hours."
-
-        def rule_check(self, context, calc_vals, data):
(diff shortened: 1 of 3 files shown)


rct229/rulesets/ashrae9012019/section19/section19rule30.py
```diff
@@ -0,0 +1,63 @@
...
+        return {"baseline_system_types_dict_b": get_baseline_system_types(rmd_b)}
+
+    class HVACRule(PartialRuleDefinition):
+        def __init__(self):
+            super(Section19Rule30.HVACRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(False, True, False),
+            )
+
+        def applicability_check(self, context, calc_vals, data):
+            rmd_b = context.baseline
```
weilixu, line 46: Should use `hvac_b` here

rct229/rulesets/ashrae9012019/section19/section19rule30.py
```diff
@@ -0,0 +1,63 @@
...
+            baseline_system_types_dict_b = get_baseline_system_types(rmd_b)
+
```
weilixu, line 48: I think this has the same issue as the #1165 
In HVAC level, you do not need to loop a list to get the result. This is specific in an HVAC, just need to check whether this particular HVAC applicability.

weilixu (commented):
A comment needs to be addressed

rct229/rulesets/ashrae9012019/section19/section19rule30.py
yunjoonjung-PNNL: addressed.

weilixu (approved):

PR #1170, opened 2023-10-27
RS/YJ/Rule 19-34

yunjoonjung-PNNL:
This rule hasn't been tested out yet. I'll work with Juan to test this out. Thanks!


rct229/rulesets/ashrae9012019/section19/section19rule34.py
```diff
@@ -0,0 +1,111 @@
...
+                + HVAC_systems_virtual_heating_list_p
+            )
+        )
+
+        hvac_sys_zones_served_dict_p = get_hvac_zone_list_w_area_by_rmi_dict(rmd_p)
+
+        zones_virtual_heating_cooling_list = list(
+            set(
+                [
+                    hvac_sys_zones_served_dict_p[hvac_id_p]["zone_list"].value()
```
weilixu, line 61: List object does not have an attribute of value(). I think this would fail.
What this list comprehensive did was to extract the `zone_list` from the `hvac_sys_zones_served_dict_p`.
One issue is if we leave it as `hvac_sys_zones_served_dict_p[hvac_id_p]["zone_list"]`, then the array would become a 2D array.
A good way to resolve the issue is to use `flat_map` from pydash package. This function will take in a lambda function, and then flatten a 2D array to 1D array. Then you can apply list(set()) to remove the duplicates.

rct229/rulesets/ashrae9012019/section19/section19rule34.py
```diff
@@ -0,0 +1,111 @@
...
+                    for hvac_id_p in HVAC_systems_virtual_list_p
+                ]
+            )
+        )
+
+        applicable_hvac_with_virtual_heating_cooling_b = list(
+            set(
+                [
+                    get_list_hvac_systems_associated_with_zone(rmd_b, zone_id_p)
```
weilixu, line 70: This will also result in a 2D array.
You would need to flatten it.

rct229/rulesets/ashrae9012019/section19/section19rule34.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+
+        return {
+            "applicable_hvac_with_virtual_heating_cooling_b": applicable_hvac_with_virtual_heating_cooling_b
+        }
+
+    class HVACRule(PartialRuleDefinition):
+        def __init__(self):
+            super(Section19Rule34.HVACRule, self).__init__(
+                rmrs_used=UserBaselineProposedVals(True, True, True),
```
weilixu, line 83: You can remove user in here.

rct229/rulesets/ashrae9012019/section19/section19rule34.py
```diff
@@ -0,0 +1,111 @@
...
+            operation_during_occupied_b = getattr_(
+                hvac_b, "HVAC", "fan_system", "operation_during_occupied"
+            )
+
+            return (
+                f"t appears that {hvac_id_p} is only being simulated in the P_RMI to meet the requirements described in Section G3.1-10 HVAC Systems proposed column c and d for heating and/or cooling. "
+                f"Check that the hvac system fan is simulated to be cycled ON and OFF to meet heating and/or cooling loads during occupied hours as applicable. "
+                f"Note that per the RMD the fan associated with {hvac_id_b} is operating as {operation_during_occupied_b} during occupied hours. "
+                f"This may require further investigation if only heating or cooling is being simulated to meet Section G3.1-10 HVAC Systems proposed column c or d because different fan operation will be required depending on whether the system is operating in heating or cooling mode."
+            )
```
weilixu, line 111: I have a hard time understand the assertion in the RDS.
1. The rule description is "heating and/or cooling system fans shall not be simulated as running continuously during occupied hours but shall be cycled ON and OFF to meet heating and cooling loads during all hours in the baseline design." But the rule assertion is not checking this. Should it be if operation_during_occupied_b is not None?
2. A big assumption here is `hvac_id_p` matches `hvac_id_b`. From various implementation, we know that these two ids wont match in many cases. Although Table 3-1 suggest in this scenario `hvac_id_p` and `hvac_id_b` could match, but that is a dangerous assumption programmatically. I would alter this text a bit to not include the message of `hvac_id_p`

weilixu (commented):
Comments need to address and waiting for RDS response.

rct229/rulesets/ashrae9012019/section19/section19rule34.py
weilixu: @yunjoonjung-PNNL #1184 is approved with a new message.
We should remove the needs of proposed model in the rule.

rct229/rulesets/ashrae9012019/section19/section19rule34.py
yunjoonjung-PNNL: Addressed

rct229/rulesets/ashrae9012019/section19/section19rule34.py
yunjoonjung-PNNL: Great point. I used ```flatten``` function in pydash instead because it has already the feature we exactly needed.

rct229/rulesets/ashrae9012019/section19/section19rule34.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section19/section19rule34.py
yunjoonjung-PNNL: Thank you Weili. Addressed.

yunjoonjung-PNNL:
For this rule, I need to ask Juan one question about TCD creation. I'll add the TCD after asking the question.

weilixu (approved):

PR #1172, opened 2023-10-30
Create Rule23-12.md

KarenWGard:


docs/section23/Rule23-12.md
```diff
@@ -0,0 +1,33 @@
...
+**Function Calls:**  
+
+1. get_baseline_system_types()
+2. baseline_system_type_compare()
+
+**Applicability Checks:**  
+- create a list of the target system types: `APPLICABLE_SYS_TYPES = [HVAC_SYS.SYS_11]`
+- Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`
+
+  - Check if B-RMR is modeled with at least one air-side system that is Type-11, continue to rule logic: `if any(baseline_system_type_compare(system_type, target_sys_type, false) for system_type in baseline_system_types_dict.keys() for applicable_sys_type in APPLICABLE_SYS_TYPES): APPLICABLE`
```
mkarpman, line 27: Should be "undetermined" outcome

KarenWGard:
Per @mkarpman 's request - I researched changing this rule to mandatory - we do not currently have the data elements required to evaluate this rule.

claperle:
> Per @mkarpman 's request - I researched changing this rule to mandatory - we do not currently have the data elements required to evaluate this rule.

@KarenWGard @mkarpman @supriyagoel  I see that the schema only has one data element for <FanSystemTemperatureControlOptions> per FanSystem and also only one data element for <FanSystemSupplyFanControlOptions> so in order to make this mandatory we would need to have separate data elements for heating and cooling. Seems like this would be quite a big change that would also likely effect other RDSs. Since system 11 is only used for computer rooms probably okay then to leave as non mandatory.

docs/section23/Rule23-12.md
```diff
@@ -0,0 +1,33 @@
+
+# Airside System - Rule 23-12 
+
+**Schema Version:** 0.0.34  
+**Mandatory Rule:** FALSE  
+**Rule ID:** 23-12  
+**Rule Description:** System 11 in heating mode SAT shall be modulated to maintain space temp and airflow shall be fixed at minimum airflow.
```
claperle, line 7: I am not sure if this matters but in case it does suggest not using abbreviations in the rule description (SAT and temp).

claperle (approved):

docs/section23/Rule23-12.md
KarenWGard: sounds good.  Changed

supriyagoel (approved):

PR #1179, opened 2023-10-30
Rds/kjw/rule 23 1

KarenWGard:


docs/section23/Rule23-1.md
```diff
@@ -0,0 +1,34 @@
+
+# Airside System - Rule 23-1 
+
+**Schema Version:** 0.0.34  
+**Mandatory Rule:** True
```
supriyagoel, line 5: this should be 'false'

docs/section23/Rule23-1.md
```diff
@@ -0,0 +1,34 @@
...
+
+1. get_baseline_system_types()
+2. baseline_system_type_compare()
+3. get_component_by_id()
+
+**Applicability Checks:**  
+- create a list of the target system types: `APPLICABLE_SYS_TYPES = [HVAC_SYS.SYS_2,HVAC_SYS.SYS_4]`
+- Get B-RMR system types: `baseline_hvac_system_dict = get_baseline_system_types(B-RMR)`
+
+  - Check if B-RMR is modeled with at least one air-side system that is Type-2, or 4, continue to rule logic: `if any(baseline_system_type_compare(system_type, target_sys_type, false) for system_type in baseline_system_types_dict.keys() for applicable_sys_type in APPLICABLE_SYS_TYPES): APPLICABLE`
```
supriyagoel, line 28: should be 'undetermined' if rule is applicable

docs/section23/Rule23-1.md
```diff
@@ -0,0 +1,34 @@
+
+# Airside System - Rule 23-1 
+
+**Schema Version:** 0.0.34  
+**Mandatory Rule:** False   
+**Rule ID:** 23-1  
+**Rule Description:** System 2 and 4 - Electric air-source heat pumps shall be modeled with electric auxiliary heat and an outdoor air thermostat. The systems shall be controlled to energize auxiliary heat only when the outdoor air temperature is less than 40°F. The air-source heat pump shall be modeled to continue to operate while auxiliary heat is energized.
```
mkarpman, line 7: I wonder why we decided to make these rule non-primary. It will trigger an undertermined outcome for all projects with baseline systems 2 & 4 (e.g., all residential projects in cooling-dominated climates). Plus, it seems that there is an easy way to verify it by checking HeatingSystem.h e a t p u m p _ a u x i l l i a r y _ h e a t _ t y p e and h e a t p u m p _ l o w _ s h u t o f f _ t e m p e r a t u r e. The only missing piece in the schema is for verifying the part that "The air-source heat pump shall be modeled to continue to operate while auxiliary heat is energized." Perhaps add HeatingSystem.heatpump_operate_when_auxilliary_heat (true/false).

KarenWGard: @mkarpman - I think you're right that we can easily change this to a primary rule.

For the requirement for heat pumps to operate while the auxiliary heat is also running, there is `heatpump_low_shutoff_temperature` - can we require that this temperature be lower than a certain number?  Perhaps a lookup table based on climate zone?

claperle: Do we know what the heatpump_low_shutoff_temperature should be? What would we be checking for in both Karen and Maria's suggestion I am not sure what the correct value would be? Does anyone have an answer?

KarenWGard: the `heatpump_auxilliary_heat_high_shutoff_temperature` should be 40F.

suggestion is to check that `heatpump_low_shutoff_temperature` is less than `heatpump_auxilliary_heat_high_shutoff_temperature`, results in PASS

KarenWGard: update based on Maria's suggested logic

image

docs/section23/Rule23-1.md
```diff
@@ -0,0 +1,49 @@
...
+            - otherwise, rule not applicable: `else: RULE_NOT_APPLICABLE`
+         
+**Rule Logic:**  
+- get the hvac system from the hvac_system_id: `hvac_system = get_component_by_id(hvac_system_id)`
+- get the heating system: `heating_system = hvac_system.heating_system`
+- get the heatpump_low_shutoff_temperature = `heatpump_low_shutoff = heating_system.heatpump_low_shutoff_temperature`
+
+**Rule Assertion:**
+- Case 1: The system type is 2 and the heatpump_low_shutoff is <= 17F, PASS: `if system_type == HVAC_SYS.2: if heatpump_low_shutoff <= 17: PASS`
+- Case 2: The system type is 2 and the heatpump_low_shutoff is >25F, FAIL: `if system_type == HVAC_SYS.2: if heatpump_low_shutoff > 25: FAIL`
```
claperle, line 42: I think these should be elifs

docs/section23/Rule23-1.md
```diff
@@ -0,0 +1,49 @@
...
+- Case 3: the system type is 2 and the heatpump_low_shutoff is between 17F and 25: UNDETERMINED: `if system_type == HVAC_SYS.2: UNDETERMINED`
```
claperle, line 43: shoudl be elif

docs/section23/Rule23-1.md
```diff
@@ -0,0 +1,49 @@
...
+- Case 4: the system type is 4 (we already know it's 4, because we've checked the 3 SYS 2 cases above), and the low shutoff temperature is <=10F: PASS: `if heatpump_low_shutoff <= 10F: PASS`
```
claperle, line 44: elif

KarenWGard: updated

claperle (approved):

PR #1187, opened 2023-11-09
RDS/JDJ/Rule5-15

JacksonJ-KC:
I just noticed this mistake as I was familiarizing with similar rules to the RDS I am currently working on. I am sure it was implemented correctly and this is just a correction to documentation - but I did not verify that.


claperle:
> I just noticed this mistake as I was familiarizing with similar rules to the RDS I am currently working on. I am sure it was implemented correctly and this is just a correction to documentation - but I did not verify that.

@JacksonJ-KC Interesting, I agree but not sure why this was not picked up by the TCDs or the RCT implementation team. Just an FYI, I was not involved in the Section 5 RDS or rule development so I don't have any historical knowledge but perhaps @supriyagoel or @weilixu  does on this one.

weilixu (approved):
@JacksonJ-KC your modification is correct. Rule definition is implemented following the modified RDS logic.

PR #1188, opened 2023-11-09
create rule RDS

JacksonJ-KC:


docs/section5/Rule5-41.md
```diff
@@ -0,0 +1,46 @@
+
+# Envelope - Rule 5-41
+
+**Rule ID:** 5-41
+**Rule Description:** Opaque roof surfaces that are not regulated (not part of opaque building envelope) must be modeled with the same thermal emittance and solar reflectance in the baseline as in the proposed design. 
+**Appendix G Section:** Section G3.1-5 Building Envelope Modeling Requirements for the Baseline building  
+**Appendix G Section Reference:** None  
+
+**Applicability:** All required data elements exist for B_RMR
```
claperle, line 9: I think the terminology is B_RMD now. @weilixu can you please confirm.

docs/section5/Rule5-41.md
```diff
@@ -0,0 +1,46 @@
...
+
+  - get_surface_conditioning_category()
+  - get_opaque_surface_type()
+  - match_data_element()
+
+## Rule Logic:  
+
+- Get surface conditioning category dictionary for B_RMR: `scc_dictionary_b = get_surface_conditioning_category(B_RMR)`  
+
+- For each building segment in the Proposed model: `for building_segment_b in B_RMR.building.building_segments:`
```
claperle, line 25: 1. This refers to the proposed model but is looping through the building segments in the baseline model.

2. The syntax we have been using is `for building_segment_b in B_RMD...building.building_segments:`  to indicate that the RDS should loop through the buildings in the RMD.

docs/section5/Rule5-41.md
```diff
@@ -0,0 +1,46 @@
...
+
+- Get surface conditioning category dictionary for B_RMR: `scc_dictionary_b = get_surface_conditioning_category(B_RMR)`  
+
+- For each building segment in the Proposed model: `for building_segment_b in B_RMR.building.building_segments:`  
+
+  - For each zone in building segment: `for zone_b in building_segment_b.zones:`  
+
+    - For each surface in zone: `for surface_b in zone_b.surfaces:`  
+
+      - Check if surface is an unregulated roof: `if (scc_dictionary_b[surface_b.id] == UNREGULATED) AND (get_opaque_surface_type(surface_b) == "ROOF"):`
```
claperle, line 31: FYI, I assumed these functions were previously approved so did not look at them.

docs/section5/Rule5-41.md
```diff
@@ -0,0 +1,46 @@
...
+
+        - Get matching surface from P_RMR: `surface_p = match_data_element(P_RMR, surfaces, surface_b.id)`
```
claperle, line 33: small typo, I think the middle term should be "surface"

docs/section5/Rule5-41.md
```diff
@@ -0,0 +1,46 @@
...
+
+          **Rule Assertion:**  
+
+          - Case 1: If thermal emittance and solar reflectance in B_RMR matches P_RMR; PASS: `if (surface_b.construction.optical_properties.absorptance_thermal_exterior == surface_p.construction.optical_properties.absorptance_thermal_exterior) AND (surface_b.construction.optical_properties.absorptance_solar_exterior == surface_p.construction.optical_properties.absorptance_solar_exterior): PASS`
```
claperle, line 37: I don't think the .construction is necessary. I think to get to the optical properties it can be surface_b.optical_properties.absorptance_thermal_exterior (applies to all the paths in this case).

docs/section5/Rule5-41.md
```diff
@@ -0,0 +1,46 @@
...
+
+          - Case 2: Else: `else: FAIL`
```
claperle, line 39: General question, what happens if these properties are undefined? Is this taken care of by a global check? If not, do we want a case that covers where the properties are undefined so we return UNDETERMINED? I can't quite remember how we can handled this in other RDSs. Perhaps we assumed these crucial parameters would be defined (I mean this would have to be defined in the model so I don't see why this would not be populated in the RMD).

docs/section5/Rule5-41.md
weilixu: yes, that is correct,

docs/section5/Rule5-41.md
supriyagoel: @claperle and @JacksonJ-KC if NULL then undetermined would make sense .. I agree that the only reason this might be missing from the RPD would be if the RPD generation software doesnt support it

JacksonJ-KC:
@supriyagoel @claperle All comments addressed!

claperle (approved):

supriyagoel (approved):

PR #1191, opened 2023-11-10
RDS/CML/Rule19-3

JacksonJ-KC:


docs/section19/19-3.md
```diff
@@ -0,0 +1,25 @@
...
+
+**Evaluation Context:** Each RulesetProjectDescription
+
+**Applicability Checks:** None  
+
+**Function Calls:**  None
+
+## Rule Logic:
+- **Rule Assertion:** 
+- All cases: Conduct manual check that rule is met:  `outcome = "UNDETERMINED" and raise_message "Check that the weather conditions used in sizing runs to determine baseline equipment capacities is based either on hourly historical weather files containing typical peak conditions or on design days developed using 99.6% heating design temperatures and 1% dry-bulb and 1% wet-bulb cooling design temperatures."      `
```
claperle, line 20: I wonder if we should have the outcome be FAIL or UNDETERMINED.
1. Weather-->cooling_design_day_type and heating_design_day_type are selected correctly and we raise_message to tell the reviewer the check in the model that the correct values were used for the location. Outcome == UNDETERMINED
2. Weather-->cooling_design_day_type and heating_design_day_type are NOT selected correctly based on the rules in 90.1 and the outcome ==  FAIL with a raise_message explaining the reason.

JacksonJ-KC: @claperle This was the line of thinking that I was headed on originally as well! Where I got turned around was the rule definition "Weather conditions used in sizing runs to determine baseline equipment capacities shall be based _**either**_ on hourly historical weather files containing typical peak conditions or on design days developed using 99.6%..." so I agreed with the way you wrote it to return an UNDETERMINED outcome for all projects, though I did not find out why the rule differs from language in 90.1 here.

claperle: @JacksonJ-KC Where do you see it says in 90.1 that it will be based on either of those? I am not seeing where the language in the rule comes from. Actually reading the language in 90.1 2019 I retract my suggestion. I think this RDS should be what you have now since it does not specify the required design day conditions. 
"Weather conditions used in sizing runs to determine baseline equipment capacities shall be based on design days developed using heating design temperatures, cooling design temperature, and cooling design wet-bulb temperature."

docs/section19/19-3.md
```diff
@@ -2,7 +2,7 @@
 **Schema Version:** 0.0.29
 **Mandatory Rule:** False    
 **Rule ID:** 19-3     
-**Rule Description:** Weather conditions used in sizing runs to determine baseline equipment capacities shall be based either on hourly historical weather files containing typical peak conditions or on design days developed using 99.6% heating design temperatures and 1% dry-bulb and 1% wet-bulb cooling design temperatures.
```
JacksonJ-KC: @supriyagoel 
Supriya, @claperle  and I discussed this rule and saw that "hourly historical weather files containing typical peak conditions" is no longer an accepted option for weather in sizing runs in 90.1-2019. This was allowed in 90.1-2016 but seems was intentionally removed in 90.1-2019. We are proposing modifying this rule description and adding the FAIL case to the RDS as below.

supriyagoel (approved):

docs/section19/19-3.md
supriyagoel: @JacksonJ-KC Yes, that makes sense to me. I have approved this RDS

docs/section19/19-3.md
```diff
@@ -0,0 +1,28 @@
...
+
+**Applicability Checks:** None  
+
+**Function Calls:**  None
+
+## Rule Logic:
+- Get the weather for the ruleset project description: `weather = RulesetProjectDescription.weather`
+
+- **Rule Assertion:**
+- Case 1: Heating/cooling design day types do not match the definitions in 90.1-2019; outcome=FAIL: `if (weather.cooling_design_day_type != COOLING_1_0) or (weather.heating_design_day_type != HEATING_99_6): FAIL`
```
claperle, line 22: @JacksonJ-KC I think this should also check that these are not equal to null as well. If they are equal to null then i recommend triggering case 2. I am just thinking that software tools like eQuest would have no way of populating this data element in the schema and so would always result in a FAIL with the current logic.

docs/section19/19-3.md
```diff
@@ -19,8 +19,9 @@
 - Get the weather for the ruleset project description: `weather = RulesetProjectDescription.weather`
 
 - **Rule Assertion:**
-- Case 1: Heating/cooling design day types do not match the definitions in 90.1-2019; outcome=FAIL: `if (weather.cooling_design_day_type != COOLING_1_0) or (weather.heating_design_day_type != HEATING_99_6): FAIL`
-- Case 2: Otherwise, conduct manual check that rule is met; outcome=UNDETERMINED:  `else: UNDETERMINED and raise_message "Check that the weather conditions used in sizing runs to determine baseline equipment capacities is based on design days developed using 99.6% heating design temperatures and 1% dry-bulb and 1% wet-bulb cooling design temperatures."`  
+- Case 1: Heating/cooling design day types do not match the definitions in 90.1-2019; outcome=FAIL: `if (weather.cooling_design_day_type != COOLING_1_0 or weather.heating_design_day_type != HEATING_99_6): FAIL`
```
claperle, line 22: The logic of this won't work because if they equal null it will return a fail since this is the first case. I would put case 2 first, then case 1 second, and then case 3

JacksonJ-KC: woops!

PR #1196, opened 2023-11-21
RCT/JX/Rule5-39

Jiarongx-Xie:
This ruleset has not been tested.


Jiarongx-Xie:
> Good work and you are getting close to the finish line. Some comments need to address.

Thanks Weili for the comments and clarification! I thought the required_fields`            required_fields={
                "$": ["weather", "ruleset_model_descriptions"],
                "$.weather": ["ground_temperature_schedule"],
            }` could prevent the None exception. If it does not, do we still keep this field?

weilixu:
> > Good work and you are getting close to the finish line. Some comments need to address.
> 
> Thanks Weili for the comments and clarification! I thought the required_fields` required_fields={ "$": ["weather", "ruleset_model_descriptions"], "$.weather": ["ground_temperature_schedule"], }` could prevent the None exception. If it does not, do we still keep this field? Can you confirm this? Thank you! @weilixu

You are correct.
If that is the case, you can directly do
`ground_temperature_schedule = rpd["weather"]["ground_temperature_schedule"]` Because the `required_fields` guaranteed the existence of these data. Using `.get` is just confusing because `.get` return type is `None | any`
Similiarly, you should be safe to get the rmds by `rmds = rpd["ruleset_model_instances"]

Jiarongx-Xie:
Great! Thank you for the clarification!

weilixu:
@Jiarongx-Xie when updating this rule, please change the rule id to 5-38, and rebase the rule to [RT/JG/section_5_numbering_update](https://github.com/pnnl/ruleset-checking-tool/tree/RT/JG/section_5_numbering_update)

rct229/rulesets/ashrae9012019/section5/section5rule38.py
```diff
@@ -1,163 +1,51 @@
...
-                        zone["infiltration"], "infiltration", "flow_rate"
-                    )
-
-                measured_air_leakage_rate = zone["infiltration"].get(
-                    "measured_air_leakage_rate"
+    def is_applicable(self, context, data=None):
+        rpd = context.BASELINE_0
+        return any(
+            [
+                surface_b
```
weilixu, line 34: I think you can do more explicit logic here:
```python
any([get_opaque_surface_type(surface_b) == OST.BELOW_GRADE_WALL for surface_b in find_all(..., rpd)])
```

rct229/rulesets/ashrae9012019/section5/section5rule38.py
```diff
@@ -1,163 +1,51 @@
...
-                != TOTAL_AIR_LEAKAGE_COEFF * target_air_leakage_rate_75pa_p
-                and empty_measured_air_leakage_rate_flow_flag == False
-                and std_equal(
-                    building_total_air_leakage_rate,
-                    TOTAL_AIR_LEAKAGE_COEFF * building_total_measured_air_leakage_rate,
-                )
-            )
+    def applicability_check(self, context, calc_vals, data):
+        ground_temperature_schedule = calc_vals["ground_temperature_schedule"]
+        return ground_temperature_schedule
```
weilixu, line 51: I think here you miss the `get_manual_check_required_msg` function to determine the message according to the RDS.
if `ground_temperature_schedule` exist, then return `It cannot be determined if the ground temperature schedule for the project is representative of the project climate.`
else:
`A ground temperature schedule was not found for the project.`

rct229/rulesets/ashrae9012019/section5/section5rule38.py
```diff
@@ -1,163 +1,51 @@
...
             ruleset_section_title="Envelope",
-            standard_section="Section G3.1-5(b) Building Envelope Modeling Requirements for the Proposed design",
-            is_primary_rule=True,
-            list_path="ruleset_model_descriptions[0].buildings[*]",
-            data_items={"climate_zone": (PROPOSED, "weather/climate_zone")},
-            manual_check_required_msg=MANUAL_CHECK_MSG,
+            standard_section="Section G3.1-14(b) Building Envelope Modeling Requirements for the Proposed design and Baseline",
+            is_primary_rule=False,
+            required_fields={
+                "$": ["weather", "ruleset_model_descriptions"],
```
weilixu, line 26: We do not need to include `ruleset_model_descriptions` in the `required_fields` since in the latest schema, this field is [required ](https://github.com/open229/ruleset-model-description-schema/blob/main/docs229/ASHRAE229.schema.json#L76)by schema.

weilixu (commented):
There are some stylist comments but one of them needs to be addressed.

rct229/rulesets/ashrae9012019/section5/section5rule38.py
Jiarongx-Xie: Thank you for the clarification!

rct229/rulesets/ashrae9012019/section5/section5rule38.py
Jiarongx-Xie: Addressed!

rct229/rulesets/ashrae9012019/section5/section5rule38.py
Jiarongx-Xie: Good catch!

Jiarongx-Xie:
> @Jiarongx-Xie There are some stylist comments but one of them needs to be addressed
Thank you Weili for the review. I have addressed these comments.

weilixu:
> > @Jiarongx-Xie There are some stylist comments but one of them needs to be addressed
> 
> Thank you Weili for the review. I have addressed these comments. @weilixu

I did not see updates, have you pushed your commits?

weilixu (approved):

PR #1198, opened 2023-11-28
add support for software output schema

weilixu:


rct229/report_engine/rct_report.py
```diff
@@ -8,6 +8,8 @@
 class RCTReport:
     def __init__(self):
         self.title = "Ruleset Checking Tool"
+        self.tool = "Ruleset Checking Tool"
+        self.version = "0.1"
```
yunjoonjung-PNNL, line 12: Would this version be different from the version (```from rct229 import __version__ as version```)?

yunjoonjung-PNNL (approved):
All looks great to me. I'd appreciate it if you could double-check one comment. Thanks!

rct229/report_engine/rct_report.py
weilixu: It is a placeholder and it will be override by the child classes. See example [here](https://github.com/pnnl/ruleset-checking-tool/blob/bb48b32f747fbae265e935b4de73821e3e994f97/rct229/reports/ashrae901_2019_software_test_report.py#L16).

yunjoonjung-PNNL: This is clear! Thank you Weili.

PR #1206, opened 2023-12-01
RT/JDJ/Rule5-22

JacksonJ-KC:


yunjoonjung-PNNL (commented):
In the[ rule-check in RDS](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section5/Rule5-22.md#:~:text=subsurface_b.depth_of_overhang%20!%3D%200))), it includes the ```depth_of_overhang```, but the Python code doesn't include this. Would you mind if you corrected this and updated the TCD as needed?

PR #1207, opened 2023-12-01
RT/JDJ/Rule5-28

JacksonJ-KC:


JacksonJ-KC:
> Case (b) and (c) are correct.
> 
> For Case (a), the `manual_check_required` method's result isn't true. I found that `building_scc_skylight_roof_ratios_dict_b[SCC.EXTERIOR_MIXED] > 0` and `target_shgc_2per_residential == target_shgc_above2_residential` conditions aren't met. Could you modify the TCD ( case (a)) to meet these conditions?

@yunjoonjung-PNNL Thanks for catching this. Are you available to meet briefly to show me how I can test for these conditions when I run rule tests?

JacksonJ-KC:
@yunjoonjung-PNNL See [](https://github.com/pnnl/ruleset-checking-tool/pull/1210)

PR #1210, opened 2023-12-02
RS/JDJ/Rule5-28

JacksonJ-KC:
This pull request also includes the rule test

### DIFF
--- a/rct229/rulesets/ashrae9012019/section5/section5rule28.py
+++ b/rct229/rulesets/ashrae9012019/section5/section5rule28.py
@@ -23,7 +23,6 @@ from rct229.utils.std_comparisons import std_equal
 
 MANUAL_CHECK_MSG = "MANUAL REVIEW IS REQUESTED TO VERIFY SKYLIGHT MEETS SHGC REQUIREMENT AS PER TABLE G3.4."
 
-
 SURFACE_CLASSIFICATION = SchemaEnums.schema_enums["SubsurfaceClassificationOptions"]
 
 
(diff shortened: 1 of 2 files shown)


yunjoonjung-PNNL (approved):
Good work! Verified all the cases passed as expected. Thanks!

rct229/rulesets/ashrae9012019/section5/section5rule28.py
```diff
@@ -101,10 +100,16 @@ def manual_check_required(self, context, calc_vals=None, data=None):
             return all(
```
weilixu, line 100: This is one of the convention.
In RDS, the `Rule Assertion - Component` case 1 says For each subsurface, if roof is exterior mixed type and the baseline requirements for residential and non-residential type SHGC for different skylight-roof-ratio are different.
In that case, we perform the `manual_check_required` at each subsurface level. 

So we will need to create a `manual_review_flag` in `create_data` follows this logic at the building level.
Then in the sub-surface level, we take out the `manual_review_flag` and combine with the roof conditioned type for manual check.

One example is [5-20](https://github.com/pnnl/ruleset-checking-tool/blob/develop/rct229/rulesets/ashrae9012019/section5/section5rule20.py). I think in 5-20, it put the manual check logic at AboveGradeWall level. I will move that to subsurface level as well.

weilixu (commented):
One comment that need to follow with the RDS logic.

weilixu:
Reviewed and approved.

PR #1211, opened 2023-12-02
RS/JDJ/Rule5-22

JacksonJ-KC:
Addresses comment in [https://github.com/pnnl/ruleset-checking-tool/pull/1206](https://github.com/pnnl/ruleset-checking-tool/pull/1206)

### DIFF
--- a/rct229/rulesets/ashrae9012019/section5/section5rule22.py
+++ b/rct229/rulesets/ashrae9012019/section5/section5rule22.py
@@ -104,7 +104,10 @@ class Section5Rule22(RuleDefinitionListIndexedBase):
                     }
 
                 def rule_check(self, context, calc_vals=None, data=None):
+                    has_shading_overhang_b = calc_vals["has_shading_overhang"]
+                    depth_of_overhang_b = calc_vals["depth_of_overhang"]
+                    has_shading_sidefins_b = calc_vals["has_shading_sidefins"]
                     return (
-                        calc_vals["has_shading_overhang"] is False
-                        or calc_vals["depth_of_overhang"] == ZERO.LENGTH
-                    ) and calc_vals["has_shading_sidefins"] is False
+                        has_shading_overhang_b is False
+                        or depth_of_overhang_b == ZERO.LENGTH
+                    ) and has_shading_sidefins_b is False
--- a/rct229/rulesets/ashrae9012019/section5/section5rule22.py
+++ b/rct229/rulesets/ashrae9012019/section5/section5rule22.py
@@ -14,6 +14,7 @@ from rct229.rulesets.ashrae9012019.ruleset_functions.get_surface_conditioning_ca
 from rct229.rulesets.ashrae9012019.ruleset_functions.get_surface_conditioning_category_dict import (
     get_surface_conditioning_category_dict,
 )
+from rct229.utils.pint_utils import ZERO
 
 FAIL_MSG = "Baseline fenestration was modeled with shading projections and/or overhangs, which is incorrect."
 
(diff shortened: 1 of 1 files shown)


rct229/rulesets/ashrae9012019/section5/section5rule22.py
```diff
@@ -99,10 +100,11 @@ def get_calc_vals(self, context, data=None):
...
                     }
 
                 def rule_check(self, context, calc_vals=None, data=None):
-                    return not (
-                        calc_vals["has_shading_overhang"]
-                        or calc_vals["has_shading_sidefins"]
-                    )
+                    return (
+                        calc_vals["has_shading_overhang"] is False
+                        or calc_vals["depth_of_overhang"] == ZERO.LENGTH
```
weilixu, line 109: The logic is quite twisted here. So in RDS it says ` Not(subsurface_b.has_shading_overhang OR subsurface_b.depth_of_overhang !=0)`
Apply De Morgan's Law:
`Not subsurface_b.has_shading_overhang AND Not subsurface_b.depth_of_overhang != 0`
So in this context. it is:
`calc_vals["has_shading_overhang"] is False and calc_vals["depth_of_overhang"] = ZERO.LENGTH.`

rct229/rulesets/ashrae9012019/section5/section5rule22.py
```diff
@@ -99,10 +100,11 @@ def get_calc_vals(self, context, data=None):
...
+                    ) and calc_vals["has_shading_sidefins"] is False
```
weilixu, line 110: This is a style comment.
Usually we'd like to take out the variable in the `rule_check` function so that we can correct label the data source.
```python
has_shading_overhang_b = calc_vals["has_shading_overhang"]
depth_of_overhang_b = calc_vals["depth_of_overhang"]
has_shading_sidefins_b = calc_vals["has_shading_sidefins"]
return not has_shading_overhang_b and depth_of_overhang_b == ZERO.LENGTH and not has_shading_side_fins_b
```

weilixu (approved):
Approved with one logic issue needs address.

rct229/rulesets/ashrae9012019/section5/section5rule22.py
JacksonJ-KC: Weili, I saw this in the RDS as well and it seemed like it was written incorrectly because if a window does not have a shading overhang then the overhang depth will not exist. 

I believe the intent was to test if the window has an overhang specified but the depth is 0, then it can be considered as flush with the wall.

weilixu: > h the pseudo-python logic.
> 
> I believe the intent was to test if the window has an overhang specified but the depth is 0, then it can be considered as flush with the wall. If you agree then I will submit a PR to update that line in the RDS.

Interesting, I never would consider a flush is a overhang but it is probably hard to tell from simulation software? In that case, please update the RDS to reflect the actual intent. Thanks for the clarification.

rct229/rulesets/ashrae9012019/section5/section5rule22.py
JacksonJ-KC: @claperle will you please confirm if you agree with my interpretation of this rule, "it seemed like it was a mistake in the way that the pseudo-python was written because the logic written in English does not match the pseudo-python logic"

I am suggesting changing the python logic to:
                        `(has_shading_overhang_b is False or depth_of_overhang_b == ZERO.LENGTH) and has_shading_sidefins_b is False`


while the RDS has written: if subsurface is flush with the exterior wall, and no shading projections are modeled or projection depth is 0. Then described in pseudo-python: `NOT (subsurface_b.has_shading_overhang OR subsurface_b.depth_of_overhang != 0)) AND ( NOT subsurface_b.has_shading_sidefins )`

rct229/rulesets/ashrae9012019/section5/section5rule22.py
JacksonJ-KC: @claperle will you please confirm if you agree with my interpretation of this rule, "it seemed like it was a mistake in the way that the pseudo-python was written because the logic written in English does not match the pseudo-python logic"

I am suggesting changing the python logic to:
(has_shading_overhang_b is False or depth_of_overhang_b == ZERO.LENGTH) and has_shading_sidefins_b is False

while the RDS has written: if subsurface is flush with the exterior wall, and no shading projections are modeled or projection depth is 0. Then described in pseudo-python: NOT (subsurface_b.has_shading_overhang OR subsurface_b.depth_of_overhang != 0)) AND ( NOT subsurface_b.has_shading_sidefins )

claperle: @JacksonJ-KC I agree.

PR #1216, opened 2023-12-06
RDS/JDJ/Rule1-3

JacksonJ-KC:
Rule ID is my suggested new numbering that has not yet been incorporated into MasterListofRules. See [Performance Calculations TCD Spreadsheet](https://onedrive.live.com/edit.aspx?resid=796769604E905084!1713&cid=796769604e905084&authkey=!ACtFTkgcGDgG1Qk&CT=1701831975786&OR=ItemsView) for my suggested changes to rules and ID.


docs/section1/Rule1-3.md
```diff
@@ -0,0 +1,33 @@
...
+**Function Calls:** None
+
+## Rule Logic:   
+
+- Get the Performance Cost Index-Target from the output of the proposed model: `output_pci_target = P_RMD.output.performance_cost_index_target`
+- Get the project area weighted average building performance factor (BPF) from the output of the proposed model: `output_bpf = P_RMD.output.total_area_weighted_building_performance_factor`
+- Get the baseline building performance (BBP) from the output of the proposed model: `output_bbp = P_RMD.output.baseline_building_performance_energy_cost`
+- Get the baseline building regulated energy cost (BBREC): `output_bbrec = P_RMD.output.baseline_building_regulated_energy_cost`
+- Get the baseline building unregulated energy cost (BBUEC): `output_bbuec = P_RMD.output.baseline_building_unregulated_energy_cost`
+
```
claperle, line 25: 1. Should the path for each output parameter take the form of: P_RMD.output.Output2019ASHRAE901.data_element?
2. I see you are using the P_RMD to get the parameters, will the baseline building performance information be included in the P_RMD? Perhaps a discussion was had about this?

JacksonJ-KC: Per Jason, 
Some vendors may take the approach of generating multiple RMD files without combining them into an RPD.  Consequently, we need to test:
outputs appropriate for the overall project exist in at least 1 RMD
if outputs appropriate for the overall project exist in more than one RMD, they are equivalent for all RMDs

Applied to your question (2) above, we apparently cannot assume that every RMD will have the parameters as I was thinking we could. I will add logic to all section 1 rules where I made this assumption.

JacksonJ-KC: @weilixu Can we use the same technique of bringing this to the schema validation level? So a json that has elements from the output schema where the note states: "output appropriate for the overall project" cannot have different values in different instances of `Output` - if it does then the json is considered invalid?

JacksonJ-KC: Reworded, To be considered a valid json: all output schema elements where note is "appropriate for the overall project" must have the same value if they exist in multiple RMDs

JacksonJ-KC: Discussed on the call today:
Schema validation will check that elements appropriate for the overall project exist in AT LEAST 1 output. I will send note to Jason to suggest that these be noted as required in the output schema.

We will write a new function that gets called by all rules in section 1 to verify that applicable elements are the same value (or not specified) in all RMDs (those present in the RPD). This function will not include checking the quantity of RMDs or that the appropriate instances of RMD exist (PROPOSED, BASELINE_0, BASELINE_90, etc.) - whichever RMDs are present will be checked for consistent values where the elements are present

JacksonJ-KC: Hi Weili, I saw this note and thought it might apply to what we are considering for these section1 rules. Was this "umbrella rule" ever implemented? [19-1 Umbrella Rule Note](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section19/Rule19-1.md?plain=1#L48)

docs/section1/Rule1-3.md
```diff
@@ -0,0 +1,44 @@
+# Section 1 - Rule 1-3
+**Schema Version:** 0.0.36  
+**Mandatory Rule:** True  
+**Rule ID:** 1-3
+**Rule Description:** The Performance Cost Index-Target (PCIt) shall be calculated using the procedures defined in Section 4.2.1.1. The PCIt shall be equal to baseline building unregulated energy cost (BBUEC) + BPF x baseline building regulated energy cost (BBREC)/ BBP
```
KarenWGard, line 5: add brackets to make the formula more clear here:

`PCIt = [BBUEC + (BPF x BBREC) – PRE]/ BBP`

KarenWGard, line 5: Rule 1-5 deals with the case where on-site renewable energy generation exceeds the thresholds defined in Section 4.2.1.1.  Are projects expected to always use the formula in this rule to generate PCIt and then subtract the 5% afterwards for rule 1-5?  We may need to clarify in the schema what calculation we expect for `performance_cost_index_target` for projects that exceed the threshold defined in 4.2.1.1

JacksonJ-KC: `performance_cost_index_target` is always calculated the same way. The formula in Rule 1-5 handles verifying the PCI is at least 5% less than PCIt, excluding renewables, there is no expectation that 5% is subtracted out at any point.

claperle (approved):

KarenWGard (approved):

PR #1219, opened 2023-12-07
Rds/kjw/rule 23 2

KarenWGard:
updates to 23-2 based on conflicts between G3.1.3.12 and G3.1.3.17


docs/section23/Rule23-2.md
```diff
@@ -28,23 +28,32 @@
...
+        - check if it is one of the system 11 system types & if true, set sys_11_types to TRUE: `if any(baseline_system_type_compare(baseline_system_type, sys_11_types, false) for target_system_type in target_system_types): sys_11_types = TRUE`
+  
 
     - get the hvac system: `hvac_b = get_component_by_id(B_RMI, hvac_id)`
     
       - For each fan system in HVAC system: `for fan_system_b in hvac_b.fan_systems:`
         **Rule Assertion:**
 
-        - Case 1: For each HVAC system that is Type-5, 6, 7, 8, 11.1, 11.2, 7a, 8a, 11.1a, 11.2a, 5b, 6b, 7b, 8b, 11b, 7c, 11.1c, if supply air temperature is reset higher by 5F under minimum cooling load condition: `if ( fan_system_b.temperature_control == "ZONE_RESET" ) AND ( fan_system_b.reset_differential_temperature == 5 ): PASS`
+        - Case 1: if the system is a type 11 system, return UNDETERMINED: `if is_sys_11: UNDETERMINED`
```
claperle, line 49: Per email exchange with Maria, allow it to pass if it meets G3.1.3.12 or G3.1.3.17. I wonder how this should be handled since it could pass this rule and then be undetermined for G3.1.3.17. 

We would need to make this so if sec 11s fail this rule it does not return a fail.

I am not sure handle the G3.1.3.17 rule to achieve what Maria recommended, does the group think it makes sense test if it passes G3.1.3.12 within the G3.1.3.17 rule and return as pass if it does and otherwise return UNDETERMINED? We would need to make it clear in the rule description that this is the case.

PR #1220, opened 2023-12-08
Ruletest JSON- 23-12

jugonzal07:
First cut at non-mandatory 23-12 rule test JSON

PR #1221, opened 2023-12-08
Section 23 rule updates

jugonzal07:
Section 23's Rule tab got updated and this affected all the ruletest JSONs for section 23.


yunjoonjung-PNNL (approved):
Looks great. Thanks Juan.

I believe the 23-12 JSON file is duplicated with #1220. Could we close it?

PR #1226, opened 2023-12-12
RT/JDJ/Rule5-40

JacksonJ-KC:


Jiarongx-Xie:
Thank you @JacksonJ-KC . Can you double-check test b? I checked the table on ASHRAE 90.1. The three available options for baseline U-factor are not the same. Also, the RCT is ready to test, feel free to use it to test the json file. The branch is [](https://github.com/pnnl/ruleset-checking-tool/tree/RCT/JX/Rule5-40/rct229/rulesets/ashrae9012019/section5)
Thanks!

JacksonJ-KC:
> Thank you @JacksonJ-KC . Can you double-check test b? I checked the table on ASHRAE 90.1. The three available options for baseline U-factor are not the same. Also, the RCT is ready to test, feel free to use it to test the json file. The branch is [https://github.com/pnnl/ruleset-checking-tool/tree/RCT/JX/Rule5-40/rct229/rulesets/ashrae9012019/section5](url) Thanks!

Hi @Jiarongx-Xie because the surface is exterior mixed, only the nonresidential and residential U-factors should be considered as options for the target U-factor [Link to case in RDS](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section5/Rule5-40.md?plain=1#L44)

In this case the nonresidential and residential are the same so we can determine the target U-factor, outcome should not be UNDETERMINED.
image
Because the baseline U-factor is lower we also want to include the conservative outcome message.

Jiarongx-Xie:
> > Thank you @JacksonJ-KC . Can you double-check test b? I checked the table on ASHRAE 90.1. The three available options for baseline U-factor are not the same. Also, the RCT is ready to test, feel free to use it to test the json file. The branch is [https://github.com/pnnl/ruleset-checking-tool/tree/RCT/JX/Rule5-40/rct229/rulesets/ashrae9012019/section5](url) Thanks!
> 
> Hi @Jiarongx-Xie because the surface is exterior mixed, only the nonresidential and residential U-factors should be considered as options for the target U-factor [Link to case in RDS](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section5/Rule5-40.md?plain=1#L44) I briefly reviewed the rule development in your branch and I think you may need to keep working on the conditions which populate the available options for target U-factor, as well as the handling of UNEXPECTED_DOOR.
> 
> In this test case, even if I had specified OTHER (UNEXPECTED_DOOR) for the subclassification (swinging or nonswinging is unknown) all of the options should be U=0.5 and would still evaluate to a FAIL outcome, outcome should not be UNDETERMINED. image

That's super clear. I thought the MIXED type include the semiheated. Thanks for the detailed clarification and double-checking! @JacksonJ-KC

weilixu:
The test json is merged. Close this Pull request now.

PR #1228, opened 2023-12-13
RDS/JDJ/Rule5-39

JacksonJ-KC:


docs/section5/Rule5-39.md
```diff
@@ -1,32 +1,35 @@
...
+  - If the surface is a floor and adjacent to ground, rule is applicable and continue to rule assertion: `if surface.classification == FLOOR and surface.adjacent_to == GROUND: is_applicable = true; break`
 
-**Rule Assertion:**  
-
-- Case 1: If B-RMR includes annual average or monthly average ground temperature: `if B_RMR.ASHRAE229.weather.monthly_ground_temperature: PASS`
-
-- Case 2: Else: `else: FAIL`
+**Rule Assertion:**
+- Case 1: If B_RMD has no floor surfaces adjacent to ground: `if not is_applicable: NOT_APPLICABLE`
+- Case 2: If RPD includes a ground temperature schedule: `if RPD.weather.ground_temperature_schedule: UNDETERMINED`
```
weilixu, line 25: One quick observation is - do we want to add data association verification here?
My thinking is we may elevate the association to a schema validation level so before we reaching this rule checking logic, the RCT already checked whether the `ground_temperature_schedule` is actually in the `Schedule` data group.

weilixu (approved):
Approved with one question.

docs/section5/Rule5-39.md
JacksonJ-KC: Is the ruleset still evaluated if a json does not pass the schema validation level?

I believe we would still want the schema to be considered valid, but only this rule would result in UNDETERMINED outcome.

JacksonJ-KC: I see, so that the schema would be considered invalid if the `id` for `weather.ground_temperature_schedule` does not exist in the `Schedule` data group. That makes sense to me.

JacksonJ-KC: Do you need me to make any changes to the RDS to specify this?

docs/section5/Rule5-39.md
```diff
@@ -1,32 +1,35 @@
...
+- Case 3: Else: `else: FAIL`
 
 
 **Notes:**
-1. Do we need to check if RMR has basement floors and only apply this rule to models with basement floors?
-2. Update Rule ID from 5-52 to 5-39 on 10/26/2023
+1. Are slab-on-grade floors different from basement floors - Can we tell the difference from the schema?
+2. Why was below grade wall removed from rule? G3.1-14(b) - It is acceptable to use either an annual average ground temperature or monthly average ground temperatures for calculation of heat loss through ***below-grade walls and*** basement floors
```
weilixu, line 31: I am not sure too. We should add below-grade walls to the list.

docs/section5/Rule5-39.md
```diff
@@ -1,32 +1,35 @@
...
+**Rule Assertion:**
+- Case 1: If B_RMD has no floor surfaces adjacent to ground: `if not is_applicable: NOT_APPLICABLE`
+- Case 2: If RPD includes a ground temperature schedule: `if RPD.weather.ground_temperature_schedule: UNDETERMINED`
+- Case 3: Else: `else: FAIL`
 
 
 **Notes:**
-1. Do we need to check if RMR has basement floors and only apply this rule to models with basement floors?
-2. Update Rule ID from 5-52 to 5-39 on 10/26/2023
+1. Are slab-on-grade floors different from basement floors - Can we tell the difference from the schema?
```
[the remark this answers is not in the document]
claperle: @JacksonJ-KC They are different.

slab-on-grade floor: that portion of a slab floor of the building envelope that is in contact with the ground and that is either above grade or is less than or equal to 24 in. below the final elevation of the nearest exterior grade. 

Below grade floors are unregulated in 90.1 as I understand it.

Based on the language of this rule in 90.1 it appears to only apply to below grade floors (although basement floor is not explicitly defined in 90.1 but seems reasonable to assume this means below grade floor). 

Can you check the elevation of the floor using the schema to determine whether it meets the definition of slab-on-grade versus below grade floor? I would think this would have to be possible since we need to check other rules related to slab on grade floors.

docs/section5/Rule5-39.md
claperle: I agree, unless there is another separate rule that covers the below grade walls, I think they shoudl be included.

docs/section5/Rule5-39.md
weilixu: I just wondering if it make sense for RDS to perform such check vs. RCT's validation module checks it. The current validation module performs schema and non-schema validation functions. The non-schema validation currently include only unique ID check. We can add the associations in the non-schema validation function too. 
@claperle @JacksonJ-KC @mkarpman what do you think?

docs/section5/Rule5-39.md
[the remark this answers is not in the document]
JacksonJ-KC: If a zone has a below-grade wall surface would that then mean that the floor surface must be a basement floor? This could be a good solution! @weilixu @claperle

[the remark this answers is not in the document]
weilixu: It sounded like it needs to function such as `is_floor_basement_floor()`.

I do not think building segment can work. This data group includes many floors.
What could possibly work is to group all zones in a floor and then check if any of the walls are below grade wall, then all floor surface adjacent to ground in those zones are considered basement floor.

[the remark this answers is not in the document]
claperle: I agree with @weilixu suggestion.

docs/section5/Rule5-39.md
JacksonJ-KC: Discussed on the RDS call, 12/14, this validation makes sense for the RCT level. However, many rule tests would then become invalid so there also needs to be a way for rule tests to bypass this validation

docs/section5/Rule5-39.md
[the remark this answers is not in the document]
JacksonJ-KC: Since we are only checking for applicability, if we find that a RMD has no below-grade wall surfaces then this means there are also no basement floor surfaces and the rule is not applicable. And since we agreed to add below-grade walls to also be included in this rule, then finding that a below-grade wall in an RMD means the rule is applicable if we are evaluating at the RMD level. I wrote the .md for a new function `get_basement_floors` to return a list of basement floor surface ids, but I don't believe we need it unless we evaluate at the surface level so I did not include in the PR.

docs/section5/Rule5-39.md
```diff
@@ -1,32 +1,34 @@
...
+- For each surface in the B_RMD: `for surface in B_RMD...Surface:`  
+  - If the surface is a below-grade wall rule is applicable and continue to rule assertion: `if get_opaque_surface_type(surface) == "BELOW-GRADE WALL": is_applicable = true; break`  
 
 **Rule Assertion:**  
-
-- Case 1: If B-RMR includes annual average or monthly average ground temperature: `if B_RMR.ASHRAE229.weather.monthly_ground_temperature: PASS`
-
-- Case 2: Else: `else: FAIL`
+- Case 1: If B_RMD has no below-grade wall surfaces and therefore also no basement floor surfaces: `if not is_applicable: NOT_APPLICABLE`  
+- Case 2: If RPD includes a ground temperature schedule: `if RPD.weather.ground_temperature_schedule: UNDETERMINED`
```
claperle, line 26: Should we include a message to tell the review what to check? I think this is useful.

docs/section5/Rule5-39.md
```diff
@@ -1,32 +1,34 @@
...
+  - If the surface is a below-grade wall rule is applicable and continue to rule assertion: `if get_opaque_surface_type(surface) == "BELOW-GRADE WALL": is_applicable = true; break`  
 
 **Rule Assertion:**  
-
-- Case 1: If B-RMR includes annual average or monthly average ground temperature: `if B_RMR.ASHRAE229.weather.monthly_ground_temperature: PASS`
-
-- Case 2: Else: `else: FAIL`
+- Case 1: If B_RMD has no below-grade wall surfaces and therefore also no basement floor surfaces: `if not is_applicable: NOT_APPLICABLE`  
+- Case 2: If RPD includes a ground temperature schedule: `if RPD.weather.ground_temperature_schedule: UNDETERMINED`  
+- Case 3: Else: `else: FAIL`
```
claperle, line 27: I recommend we explain why it is fail (i.e., lack of schedule).

docs/section5/Rule5-39.md
JacksonJ-KC: Is this undetermined because:
1) we can't tell if the ground temperature was used to calculate heat loss? 
2) we don't know if the ground temperature used is representative of project climate?
3) both?

docs/section5/Rule5-39.md
weilixu: We just added a way to bypass the association checks in rule tests.

PR #1233, opened 2023-12-18
Rule test JSON for 23-4

jugonzal07:
First cut at rule test JSON for 23-4.


yunjoonjung-PNNL (commented):
Juan, could you address the following parts?

1) Could you add proposed part?
2) could you add the ```climate_zone``` and ```is_leap_year``` keys?

These are needed in the ```get_lab_zone_hvac_systems``` function. However, the RDS doesn't describe these required arguments.

yunjoonjung-PNNL (approved):
Looks great. Thank you Juan!

PR #1240, opened 2023-12-26
RS/YJ/Rule 23-4

yunjoonjung-PNNL:
This rule hasn't been passed yet. I'll work with Juan to make this rule pass. Thanks!


weilixu:
@yunjoonjung-PNNL @jugonzal07  I think this rule is ready to merge.
The error is because the rule requires proposed model but in the test json, it only contains Baseline.

PR #1234, opened 2023-12-22
Section 5 numbering update

jugonzal07:
Section 5 renumbering. 

- Old 5-16 removed.
- Old 5-17 now 5-16
- Old 5-18 now 5-17
- Old 5-29 now 5-18. 
- Old 5-30 and up are all one value lower (e.g., old 5-30 is new 5-29)

Also includes updates for changes in naming convention (e.g., RMR now RMD, RMD now RPD).

PR #1242, opened 2023-12-27
RT/JDJ/Rule5-30

JacksonJ-KC:
~~run_ruletests yields "rct229.utils.assertions.RCTFailureException: RPDs provided are invalid. See error messages in terminal." But I struggled to trace the source of the error. Is there a new way that I should run ruletests which bypasses the association validation?~~

Issue resolved by testing on the branch with section 5 renumbering


yunjoonjung-PNNL:
Hi Jackson,

This is because your testing JSON file doesn't have the baseline type and this caused the error. And, according to the [5-30 RDS](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section5/Rule5-30.md), we don't need the proposed and user types. 

Further, are you aware that section 5 was renumbered? Please refer to PR #1234.

JacksonJ-KC:
> Hi Jackson,
> 
> This is because your testing JSON file doesn't have the baseline type and this caused the error. And, according to the [5-30 RDS](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section5/Rule5-30.md), we don't need the proposed and user types.
> 
> Further, are you aware that section 5 was renumbered? Please refer to PR #1234.

Got it, thanks @yunjoonjung-PNNL . In that case, maybe I was correct in changing the base branch to Juan's branch with the renumbering. I will checkout that branch and test there to make sure all outcomes are as expected, then make this PR merge into the renumbered branch instead of develop

JacksonJ-KC:
Tested and all outcomes are as expected

PR #1248, opened 2023-12-29
RCT/JX/Rule5-40

Jiarongx-Xie:
this rule has been tested


rct229/rulesets/ashrae9012019/section5/section5rule40.py
```diff
@@ -0,0 +1,238 @@
...
+                        target_u_factor_b = target_u_factor_res_b
+                    elif scc_dict_b == SCC.SEMI_EXTERIOR:
+                        target_u_factor_b = target_u_factor_semiheated_b
+                    else:
+                        assert scc_dict_b == SCC.EXTERIOR_MIXED, (
+                            f"Surface conditioning category is not one of the five types: EXTERIOR_NON_RESIDENTIAL, "
+                            f"EXTERIOR_RESIDENTIAL, SEMI_EXTERIOR, EXTERIOR_MIXED, UNREGULATED, "
+                            f"got {scc_dict_b} instead "
+                        )
+                        if target_u_factor_nonres_b == target_u_factor_res_b:
```
weilixu, line 168: Should this if else be a separate statement instead of under the 
```python
else:
                        assert scc_dict_b == SCC.EXTERIOR_MIXED, (
                            f"Surface conditioning category is not one of the five types: EXTERIOR_NON_RESIDENTIAL, "
                            f"EXTERIOR_RESIDENTIAL, SEMI_EXTERIOR, EXTERIOR_MIXED, UNREGULATED, "
                            f"got {scc_dict_b} instead "
                        )
```

rct229/rulesets/ashrae9012019/section5/section5rule40.py
```diff
@@ -0,0 +1,238 @@
...
+                            target_u_factor_b = target_u_factor_res_b
```
weilixu, line 169: Usually, in this scenario, we set the `target_u_factor_b = target_u_factor_nonres_b`. Nothing wrong here but I think change to nonres to keep the code consistent.

weilixu (approved):
Great work. 
It is a tough rule and I am glad you navigated through the challenge swiftly.
Approved with two comments. One of them needs to be addressed.

rct229/rulesets/ashrae9012019/section5/section5rule40.py
Jiarongx-Xie: Good point! Addressed.

rct229/rulesets/ashrae9012019/section5/section5rule40.py
Jiarongx-Xie: Addressed! Thank you for the clarification!

Jiarongx-Xie:
> Great work. It is a tough rule and I am glad you navigated through the challenge swiftly. Approved with two comments. One of them needs to be addressed.

Thanks Weili! These comments have been addressed.

PR #1229, opened 2023-12-13
RT/JDJ/Rule5-38 (new numbering)

JacksonJ-KC:

PR #1254, opened 2024-01-04
RDS/JDJ/Rule5-39

JacksonJ-KC:
This RDS has already been reviewed/approved. However, per discussion, the fail case is not appropriate for secondary rules so is now adjusted to result in UNDETERMINED.


weilixu:
Thank you @JacksonJ-KC, could you update this to #1229 ? We will merge these to the renumbering branch that uses the new rule IDs.

PR #1255, opened 2024-01-04
RDS/JDJ/5_38

JacksonJ-KC:
This RDS was already reviewed/approved but, per discussion, cannot have a fail case because it is a secondary rule.


JacksonJ-KC:
@weilixu I see now I misread your previous directions to merge into the branch with the ruletest, but I think this is sufficient

PR #1277, opened 2024-02-22
RS/YJ/Rule4-14

yunjoonjung-PNNL:
This PR hasn't been tested. Thanks!


rct229/rulesets/ashrae9012019/section4/section4rule14.py
```diff
@@ -0,0 +1,103 @@
...
+        sch_ids_p = find_all(
+            f'$.buildings[*].building_segments[*].zones[*].spaces[*].miscellaneous_equipment[*][?(@.energy_type="ELECTRICITY")].multiplier_schedule',
+            rmd_p,
+        )
+        schedule_p = []
+        for scd_id_p in sch_ids_p:
+            schedule_p += find_all(f'$.schedules[*][?(@id = "{scd_id_p}")]', rmd_p)
+
+        return {"schedule_p": schedule_p}
+
```
weilixu, line 49: I think you may need to add the `list_filter` here to remove spaces that are not labeled as `LIGHTING_SPACE.COMPUTER_ROOM` from processing, otherwise, this rule will generate many failures.

And add an `is_applicable()` function here to check whether the project has spaces labeled as `LIGHTING_SPACE.COMPUTER_ROOM`. If not, return `NOT_APPLICABLE` outcome. This part is not in the RDS, but I think it is important.

rct229/rulesets/ashrae9012019/section4/section4rule14.py
```diff
@@ -0,0 +1,103 @@
...
+            )
+
+        def get_calc_vals(self, context, data=None):
+            space_p = context.proposed
+
+            schedule_p = data["schedule_p"]
+            floor_area_p = space_p["floor_area"]
+
+            total_space_misc_Wattage_including_multiplier_p = ZERO.POWER
+            if space_p["lighting_space_type"] == LIGHTING_SPACE.COMPUTER_ROOM:
```
weilixu, line 73: follow the above comment, we do not need to do a filter here. Any space reached here shall be guaranteed as a computer room.

rct229/rulesets/ashrae9012019/section4/section4rule14.py
```diff
@@ -0,0 +1,103 @@
...
+                for misc_equip_p in space_p["miscellaneous_equipment"]:
+                    if misc_equip_p["energy_type"] == ENERGY_SOURCE.ELECTRICITY:
+                        misc_equip_power_p = getattr_(
+                            misc_equip_p, "miscellaneous_equipment", "power"
+                        )
+                        mis_multiplier_id_p = getattr_(
+                            misc_equip_p,
+                            "miscellaneous_equipment",
+                            "multiplier_schedule",
+                        )
```
weilixu, line 83: If this schedule is not present, then the multiplier is always one. In that case, we should not use `getattr_` function.

rct229/rulesets/ashrae9012019/section4/section4rule14.py
```diff
@@ -0,0 +1,103 @@
...
+                        total_space_misc_Wattage_including_multiplier_p += min(
+                            1,
+                            max(
+                                misc_equip_power_p
+                                * find(
+                                    schedule_p,
+                                    lambda d: d.get("id") == mis_multiplier_id_p,
+                                )
+                            ),
+                        )
```
weilixu, line 94: In RDS, it says `MAX(1, MAX(misc_p.multiplier_schedule.hourly_values)` and then `misc_p_total_Wattage = misc_p_power * misc_p_multiplier_value`.
The expression used about seems wrong.
This shall be revised to:
```python
misc_equip_power_p * max(1.0, max(getattr_(find(schedule_p, lambda d: d["id"] == mis_multiplier_id_p), "schedule", "hourly_values")))
```

You can also break it down into multiple expressions if this is getting too long to understand.

weilixu (commented):
Comments need to address

PR #1278, opened 2024-02-23
Try schema processor

weilixu:


weilixu:
@JacksonJ-KC I realized in the extra schema, all of the "AppG B_RMDs Same" are "unknown". In the function, the logic will set comparison to false if the data  is "unknow".
For baseline equal to baseline, we will need to update the extra schema to include equality check.

rct229/rulesets/ashrae9012019/data_fns/extra_schema_fns.py
```diff
@@ -0,0 +1,195 @@
...
+) -> bool:
+    """
+    Perform equality comparison between two RPD context
+    based on the value in the extra schema specification
+
+    Parameters
+    ----------
+    index_context: dict|list|str|None, the indexed context (RPD)
+    compare_context: dict|list|str|None, the compare context (RPD)
+    element_json_path: str, The json path of the compare context
```
jugonzal07, line 80: An example JSON path might help folks that don't know the convention:

e.g., $.ruleset_model_descriptions[0].buildings[0].building_open_schedule

rct229/rulesets/ashrae9012019/data_fns/extra_schema_fns.py
```diff
@@ -0,0 +1,195 @@
...
+        if (
+            compare_context
+            and compare_context.get("id")
+            and required_equal
+            and index_context["id"] != compare_context["id"]
+        ):
+            error_msg_list.append(
+                f'path: {element_json_path}: data object {index_context["id"]} in index context does not match the one {compare_context["id"]} in compare context'
+            )
+            matched = False
```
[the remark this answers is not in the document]
weilixu: I am debating whether to provide a whole picture of comparison or stop at key determination path.

One example is the [optical_properties](https://github.com/pnnl/ruleset-checking-tool/blob/develop/rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section5/rule_5_31.json#L61-L64) under surface object.
For this object, proposed and user should be equal and so does the `absorptance_thermal_exterior` and `absorptance_solar_exterior` inside this object.

So the question is if the object's id is different, should we stop processing the object and return the error message, or continue scanning through the object and identify any mismatches in the data?

Example:

```python
#user:
"optical_properties": {
    "id": "Surface Optical Properties 1",
    "absorptance_solar_exterior": 0.7},

#proposed:
"optical_properties": {
    "id": "Surface Optical Properties",
    "absorptance_solar_exterior": 0.8
 },
```
1. If we stopped at id checking, then the error message will be:
```
path: $....optical_properties: data object Surface Optical Properties in index context does not match the one Surface Optical Properties in compare context.
```

2. If we continue scanning the object:
```
path: $.....optical_properties: data object Surface Optical Properties in index context does not match the one Surface Optical Properties in compare context.
path: $.....optical_properties.absorptance_solar_exterior: index context data: 0.8 does not equal to compare context data: 0.7
```

Would like to get some inputs from the team @supriyagoel @KarenWGard @JacksonJ-KC @mkarpman @claperle

rct229/rulesets/ashrae9012019/data_fns/extra_schema_fns.py
weilixu: Good point, I can add an example to the comment.

JacksonJ-KC:
> @JacksonJ-KC I realized in the extra schema, all of the "AppG B_RMDs Same" are "unknown". In the function, the logic will set comparison to false if the data is "unknown". For baseline equal to baseline, we will need to update the extra schema to include equality check.

@weilixu Would it be best for us to make a pull request to the RCT project, or to the open229 project with these changes?

weilixu:
@JacksonJ-KC I suggest making changes to the [open229 project](https://github.com/open229/ruleset-model-description-schema/blob/main/schema-source/ASHRAE229_extra.schema.yaml). Thanks for checking.

PR #1287, opened 2024-03-12
RCT/JX/Rule4-2

Jiarongx-Xie:
This rule has not been tested.


rct229/rulesets/ashrae9012019/section4/section4rule2.py
```diff
@@ -0,0 +1,269 @@
...
+                elif (
+                    not minimum_humidity_stpt_sch_id_b
+                    and minimum_humidity_stpt_sch_id_p
+                ) or (
+                    not maximum_humidity_stpt_sch_id_b
+                    and maximum_humidity_stpt_sch_id_p
+                ):
+                    return FAIL_MSG_P
+                else:
+                    return ""
```
weilixu, line 269: There is no need for two ifs here. This function is called when RCT already determined it is failed.
So the message is only B or P.

weilixu (commented):
Good work, thank you for getting it done. I made several comments below.

rct229/rulesets/ashrae9012019/section4/section4rule2.py
Jiarongx-Xie: Good point!

Jiarongx-Xie:
Thanks for the comments @weilixu. I have addressed them.

rct229/rulesets/ashrae9012019/section4/section4rule2.py
```diff
@@ -0,0 +1,269 @@
...
+                    getattr_(
+                        find_exactly_one(
+                            f'$[*][?(@.id="{minimum_humidity_stpt_sch_id_b}")]',
+                            schedules_b,
+                        ),
+                        "schedules",
+                        "hourly_values",
+                    )
+                    if schedules_b and minimum_humidity_stpt_sch_id_b
+                    else [] * number_of_hours
```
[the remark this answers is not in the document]
weilixu: correct. The difference is that if the `thermostat_heating_setpoint_schedule` is not present, then it assumes to be a constant schedule with design thermostat heating setpoint.
However, the `minimum_humidity_setpoint_schedule` does not have that. If it is missing, then it is not specified.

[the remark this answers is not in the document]
Jiarongx-Xie: Thanks for the clarification! It makes sense to me.

rct229/rulesets/ashrae9012019/section4/section4rule2.py
```diff
@@ -117,100 +112,96 @@ def get_calc_vals(self, context, data=None):
...
                 )
+
+                # ^ comparison for data type.
                 minimum_humidity_schedule_type_matched = (
                     minimum_humidity_stpt_hourly_values_b is None
-                    and minimum_humidity_stpt_hourly_values_p is None
-                ) or (
-                    minimum_humidity_stpt_hourly_values_b is not None
-                    and minimum_humidity_stpt_hourly_values_p is not None
+                    == minimum_humidity_stpt_hourly_values_p is None
```
Jiarongx-Xie: Here we will need a parenthesis. Otherwise, it will return false if both `minimum_humidity_stpt_hourly_values_b ` and `minimum_humidity_stpt_hourly_values_p` exist. That' what happened when I tested locally, but I'm not sure why this does not influence the testing result. 
```
 minimum_humidity_schedule_type_matched = (
                    (minimum_humidity_stpt_hourly_values_b is None)
                    == (minimum_humidity_stpt_hourly_values_p is None)
)
```
Could you double-check? @weilixu Thanks

rct229/rulesets/ashrae9012019/section4/section4rule2.py
```diff
@@ -278,8 +263,10 @@ def get_fail_msg(self, context, calc_vals=None, data=None):
...
                     "maximum_humidity_stpt_sch_id_b"
                 ]
+
+                # guaranteed that either b or p has Null
                 return (
-                    FAIL_MSG_B
-                    if minimum_humidity_stpt_sch_id_b or maximum_humidity_stpt_sch_id_b
-                    else FAIL_MSG_P
+                    FAIL_MSG_P
+                    if (minimum_humidity_stpt_sch_id_b and maximum_humidity_stpt_sch_id_b)
```
Jiarongx-Xie: I think this should be or. Suppose `minimum_humidity_stpt_sch_id_b` exists and `minimum_humidity_stpt_sch_id_p` does not exist. Meanwhile, `minimum_humidity_stpt_sch`does not exist in both baseline and proposed. This is a failure case where `FAIL_MSG_B = (
    "Fail because a humidity schedule is defined in the B_RMR but not in the P_RMR."
)` applies. Could you double-check? @weilixu Thanks

Jiarongx-Xie: The naming of FAIL_MSG_B and FAIL_MSG_P maybe misleading, we may switch them

rct229/rulesets/ashrae9012019/section4/section4rule2.py
weilixu: You are right, we need to add parenthesis.

rct229/rulesets/ashrae9012019/section4/section4rule2.py
weilixu: I think the parameter name is confusing. I got this swapped but I think It is `and`.
Note this is failed message - preconditioned needs to take into consideration from `manual_check_required` and `rule_check`.
At this step, we already know schedule id data type are mismatched because:
1. manual check returns True if schedule data type matched and schedule value mismatched
2. rule check returns True if schedule data type matched and schedule value matched.

The only failure that comes this far is schedule data type mismatched. Basically, it means the `humidity_schedule_id_b` and `humidity_schedule_id_p` have different data type (one is None, one has value). 
Note - if a schedule id is not exist, then the schedule won't exist, this is another pre-condition checked by non-schema validation.

So in that case, what I need to know at this point is simple which model does not have the schedule. Is it baseline or proposed?
So in this case, I will say either `minimum_humidity_stpt_sch_id_b` or `maximum_humidity_stpt_sch_id_b` is None (`(minimum_humidity_stpt_sch_id_b is None) or (maximum_humidity_stpt_sch_id_b is None)` = `minimum_humidity_stpt_sch_id_b  and maximum_humidity_stpt_sch_id_b`), then we return the `The schedule is defined in proposed, but not in baseline.`

weilixu: Oh, I don't think the message is swapped.
`if minimum_humidity_stpt_sch_id_b  and maximum_humidity_stpt_sch_id_b` means proposed is missing but baseline has. So the message is correct `FAIL_MSG_B`

Jiarongx-Xie: `So in this case, I will say either minimum_humidity_stpt_sch_id_b or maximum_humidity_stpt_sch_id_b is None ((minimum_humidity_stpt_sch_id_b is None) or (maximum_humidity_stpt_sch_id_b is None) = minimum_humidity_stpt_sch_id_b  and maximum_humidity_stpt_sch_id_b), then we return the The schedule is defined in proposed, but not in baseline.`, everything makes sense before this. I tested locally and if we use `and`, case 4 and 5 would return the same msg. It took me quite a while to realize that there could be some complex case, such as `minimum_humidity_stpt_sch` exists in baseline but not in proposed, plus `maximum_humidity_stpt_sch` exists in proposed but not in baseline, then the msg can be `FAIL_MSG_B` or `FAIL_MSG_P`

weilixu: I see. The confusing part is RDS is trying to condense four logic branches (maximum/minimum and baseline/proposed) into two logic branches. Reading RDS languages again, I think it uses `or` which means as long as there is one type of schedule existing in baseline, then returning `FAIL_MSG_G`.
The `and` takes another approach (both types of schedules exist in baseline to return `FAIL_MSG_G`), which is indeed different from the RDS.

I will update the rules to reflect the RDS. thanks.

PR #1290, opened 2024-03-13
Ruletest 4-2

jugonzal07:
first cut at 4-2 ruletest


Jiarongx-Xie:
@jugonzal07 Hi Juan, for 4-2-b, it seems that the schedules are missing in the rmr. Could you double-check?

PR #1292, opened 2024-03-13
Ruletest 4-14

jugonzal07:
first cut at 4-14 ruletest


weilixu:
`miscellaneous_equipment` is missing `energy_type`
This needs to be added in order for a correct EPD calculation in the computer room.

jugonzal07:
> `miscellaneous_equipment` is missing `energy_type` This needs to be added in order for a correct EPD calculation in the computer room.

Addressed.

PR #1293, opened 2024-03-13
Ruletest 10-7

jugonzal07:
First cut at ruletests for section 10-7


JacksonJ-KC:
It looks like all spaces will need `floor_area` defined for all test cases

jugonzal07:
> It looks like all spaces will need `floor_area` defined for all test cases

Addressed.

PR #1296, opened 2024-03-15
RT/JDJ/19-13

JacksonJ-KC:
Corrected rule test 19-13-e primary_airflow and rule definition summation of primary_airflow for all zones. 

I am also noticing some odd behavior when I insert breakpoints at lines 127 & 128 which may require more investigation.


weilixu (approved):
The change makes sense to me, thank you Jackson for finding this bug.

PR #1298, opened 2024-03-18
RS/JDJ/Rule 10-14

JacksonJ-KC:
~~This is marked as a draft because the RDS is not yet merged and there may still be minor changes made to the RDS that I will need to update here as well.~~

@weilixu I am done making updates, and we are passing all test cases once again


JacksonJ-KC:
I am realizing the data tables that I created for this implementation were not necessary. data tables already exist in various "ashrae_90_1_prm_" tables.

rct229/rulesets/ashrae9012019/section10/section10rule14.py
```diff
@@ -0,0 +1,510 @@
...
+                len(modeled_effs_b) > 0
+                and all(
+                    modeled_eff_b[0] == expected_eff_b[0]
+                    for modeled_eff_b, expected_eff_b in zip(
+                        modeled_effs_b, expected_effs_b
+                    )
+                )
+                and is_zone_agg_factor_undefined_and_needed
+            ):
+                return True
```
weilixu, line 392: I think we can condense the logic more but I do appreciate the clarity in the code so it is OK to keep as it is.
If you think this can be improved a little more without losing the clarify, I suggest considering:
```python
if (
                len(modeled_effs_b) > 0
                and all(
                    modeled_eff_b[0] == expected_eff_b[0]
                    for modeled_eff_b, expected_eff_b in zip(
                        modeled_effs_b, expected_effs_b
                    )
                )
                and (
                        #Case 3 and 7 both satisfied here
                        total_capacity_b is None 
                        or
                        # Case 4 and 10 both satisfied here
                        is_zone_agg_factor_undefined_and_needed
                )
```

rct229/rulesets/ashrae9012019/section10/section10rule14.py
```diff
@@ -0,0 +1,510 @@
...
+
+            else:  # HVAC_SYS.SYS_4
+                expected_high_temp_eff_b = calc_vals["expected_high_temp_eff_b"]
+                modeled_high_temp_eff_b = calc_vals["modeled_high_temp_eff_b"]
+                expected_low_temp_eff_b = calc_vals["expected_low_temp_eff_b"]
+                modeled_low_temp_eff_b = calc_vals["modeled_low_temp_eff_b"]
+
+                # Case 5
+                return (
+                    modeled_high_temp_eff_b == expected_high_temp_eff_b
```
weilixu, line 506: It seems the `expected_high_temp_eff_b` is guaranteed not NONE value if the HVAC system is type SYS4. Could you confirm if this is correct?
I ran through the G3_5_2 functions and think that maybe the case. If so, this logic is perfect.

weilixu (approved):
I tried following most of the logics and think this is done nicely.
Many places can be simplified but given how complex this rule is, I like the implementation being clarity driven instead of following convention. I got a few questions just for confirmation. Otherwise, this looks good to me.

JacksonJ-KC:
> I tried following most of the logics and think this is done nicely. Many places can be simplified but given how complex this rule is, I like the implementation being clarity driven instead of following convention. I got a few questions just for confirmation. Otherwise, this looks good to me.

Thanks @weilixu. I am planning to address your comments and test with Juan's new test cases before we merge to develop

rct229/rulesets/ashrae9012019/section10/section10rule14.py
JacksonJ-KC: Confirmed

rct229/rulesets/ashrae9012019/section10/section10rule14.py
```diff
@@ -0,0 +1,510 @@
...
+
+                # Filter expected efficiencies to only include those with the modeled efficiency metrics
+                expected_effs_b = [
+                    expected_eff_b
+                    for expected_eff_b in expected_effs_b
+                    if expected_eff_b[1] in [eff[1] for eff in modeled_effs_b]
+                ]
+                assert len(modeled_effs_b) == len(
+                    expected_effs_b
+                ), "The number of modeled efficiencies does not match the number of expected efficiencies. Check for duplicated efficiency metrics in the modeled efficiencies."
```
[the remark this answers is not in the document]
JacksonJ-KC: Actually I agree with adding this check to RPD validation which would make sure that if an efficiency metric is defined for a system that it only exists once for that system.

rct229/rulesets/ashrae9012019/section10/section10rule14.py
```diff
@@ -170,7 +170,7 @@ def get_calc_vals(self, context, data=None):
                     )
 
             if total_capacity_b is not None:
-                total_capacity_b = total_capacity_b.magnitude
+                total_capacity_b = total_capacity_b.to("Btu/h").magnitude
```
JacksonJ-KC, line 173: @weilixu Do you agree with this update?

weilixu: I see, the HEATPUMP_CAPACITY_LOW_THRESHOLD etc does not have unit assignment. We typically want to keep those constants as quantity objects (see example; https://github.com/pnnl/ruleset-checking-tool/blob/master/rct229/rulesets/ashrae9012019/section19/section19rule13.py#L46-L47)

The pint internally handles unit conversion when comparing for example "w" and "btu/h". So if you add units to these constants, then you do not need to extract mag from the quantity object.

JacksonJ-KC: I used the magnitude because of a previous decision to keep the lookup tables using integers only

PR #1299, opened 2024-03-25
RS/JDJ/Zone Association Check

JacksonJ-KC:


rct229/schema/validate.py
```diff
@@ -21,7 +21,112 @@
...
+            "$.buildings[*].building_segments[*].zones[*].transfer_airflow_source_zone",
+            rmd,
+        )
+    )
+    referenced_id_list.extend(
+        find_all(
+            "$.service_water_heating_distribution_systems[*].service_water_piping[*].location_zone",
+            rmd,
+        )
+    )
```
weilixu, line 123: I wonder if we can create a utility function called:
`find_all_by_jsonpaths` which takes in a list of jsonpaths and return a list of matching results. Will that be too confusing?

weilixu (approved):

rct229/schema/validate.py
weilixu: Also, please run black format before merging.

JacksonJ-KC: The `find_all_by_jsonpaths` utility makes sense to me. 

The black format test is failing from Rule 4-2, I don't believe it is originating from the code updated in this PR.

PR #1300, opened 2024-03-25
RS/JDJ/FluidLoop Association Check

JacksonJ-KC:


rct229/schema/validate.py
```diff
@@ -19,7 +19,111 @@
...
+    referenced_id_list.extend(
+        find_all(
+            "$.external_fluid_sources[*].loop",
+            rmd,
+        )
+    )
+
+    for fluid_loop_id in referenced_id_list:
+        if fluid_loop_id not in fluid_loop_id_list:
+            mismatch_list.append(fluid_loop_id)
```
weilixu, line 123: One thing I was thinking is whether we allow orphan fluid loops in the project.
I can see zone does not have this issue but not sure about fluid.
If that is the logic we want to add, then the HVAC association check will also need update.

weilixu (approved):

rct229/schema/validate.py
weilixu: Need to run black before merging.

JacksonJ-KC: A particular BEM tool might have errors arise from orphan model elements, but I don't think it will be universal. I don't think we want to add logic that checks if there are orphans.

PR #1301, opened 2024-03-25
RS/JDJ/Schedule Association Check

JacksonJ-KC:
I am aware of 1 schedule reference for `Weather.ground_temperature_schedule` which is not included in the check because I am not sure if the association checks are built in a way that allows checking the RPD, or if it has to be the RMD.

### DIFF
--- a/examples/baseline_rmr.json
+++ b/examples/baseline_rmr.json
@@ -121,6 +121,13 @@
                 ]
             }
         ],
+        "schedules": [
+            {"id":  "tcs_sched_1"},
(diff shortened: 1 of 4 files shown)


weilixu:
> I am aware of 1 schedule reference for `Weather.ground_temperature_schedule` which is not included in the check because I am not sure if the association checks are built in a way that allows checking the RPD, or if it has to be the RMD.

The scope is RPD, not RMD. We should update the terms there to avoid confusion.

PR #1302, opened 2024-03-25
RS/JDJ/FluidLoop or ServiceWaterPiping Association Check

JacksonJ-KC:


rct229/schema/validate.py
```diff
@@ -25,7 +25,36 @@
...
+    )
+
+    referenced_fluid_loop_or_piping_id_list = find_all(
+        "$.pumps[*].loop_or_piping",
+        rmd,
+    )
+    for fluid_loop_or_piping_id in referenced_fluid_loop_or_piping_id_list:
+        if fluid_loop_or_piping_id not in fluid_loop_or_piping_id_list:
+            mismatch_list.append(fluid_loop_or_piping_id)
+    return mismatch_list
```
weilixu, line 56: Similarly, I am not sure if we should allow orphan water heater loops - that sounds OK to me though

weilixu (approved):

rct229/schema/validate.py
JacksonJ-KC: In this case it would mean there are no pumps in the model which reference the FluidLoop/ServiceWaterPiping. I think there could be cases where a loop may not have a pump so in this case orphans are ok.

PR #1307, opened 2024-04-02
RS/YJ/get_zone_BPF_BAT Function

yunjoonjung-PNNL:
This function is required for section 1 rules. Thanks!


rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Returns
+    -------
+    zone_BPF_BAT_dict: A dict for the zone that saves the BPF_BAT as keys and the areas as the values.
+                       Example: {SCHOOL: 50000, UNDETERMINED: 2000}
+
+    """
+
+    zone_BPF_BAT_dict = {}
+    for space in find_all(
```
JacksonJ-KC, line 132: We realized that `find_all()` actually takes `rpd`, instead of `rmd` as the argument. So the search path must begin with `$.ruleset_model_descriptions[*]`

JacksonJ-KC: Actually maybe that was only in the context of the schema validation. Do you think this is correct to pass `rmd` here?

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
```diff
@@ -0,0 +1,144 @@
...
+        f'$.buildings[*].building_segments[*].zones[*][?(@.id = "{zone_id}")].spaces[*]',
+        rmd,
+    ):
+        lighting_space_type = getattr_(space, "spaces", "lighting_space_type")
```
JacksonJ-KC, line 136: I don't think we want an undetermined outcome for the rule that calls this function if a space does not have a `lighting_space_type` defined. Instead, I think this should map the space area to the `"UNDETERMINED"` building area

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
JacksonJ-KC: Disregard my comments here. 

We just need to make sure that `get_BPF_building_area_types_and_zones()` is provided with the `rmd` object when it is called, rather than the `rpd` object.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
yunjoonjung-PNNL: The RDS says ```space_BPF_BAT = building_area_lookup[space.lighting_space_type]```, which assumes the ```lighting_space_type``` key is always available. 

I can modify RDS when the ```lighting_space_type``` key isn't defined, the space type is mapped to ```UNDETERMINED```, but I just wanted to double-check with you.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
yunjoonjung-PNNL: Sounds great! Thanks for the clarification.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
yunjoonjung-PNNL: This issue was fixed!

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
```diff
@@ -0,0 +1,148 @@
...
+    LIGHTING_SPACE.SPORTS_ARENA_PLAYING_AREA_CLASS_I_FACILITY: "ALL_OTHER",
+    LIGHTING_SPACE.SPORTS_ARENA_PLAYING_AREA_CLASS_II_FACILITY: "ALL_OTHER",
+    LIGHTING_SPACE.SPORTS_ARENA_PLAYING_AREA_CLASS_III_FACILITY: "ALL_OTHER",
+    LIGHTING_SPACE.SPORTS_ARENA_PLAYING_AREA_CLASS_IV_FACILITY: "UNDETERMINED",
+    LIGHTING_SPACE.TRANSPORTATION_FACILITY_BAGGAGE_CAROUSEL_AREA: "ALL_OTHER",
+    LIGHTING_SPACE.TRANSPORTATION_FACILITY_AIRPORT_CONCOURSE: "ALL_OTHER",
+    LIGHTING_SPACE.TRANSPORTATION_FACILITY_TICKET_COUNTER: "ALL_OTHER",
+    LIGHTING_SPACE.WAREHOUSE_STORAGE_AREA_MEDIUM_TO_BULKY_PALLETIZED_ITEMS: "WAREHOUSE",
+    LIGHTING_SPACE.WAREHOUSE_STORAGE_AREA_SMALLER_HAND_CARRIED_ITEMS: "WAREHOUSE",
+}
```
weilixu, line 112: See the example lighting to hvac map. To be consistent, we can move this map to a json file.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
```diff
@@ -0,0 +1,148 @@
...
+            if lighting_space_type is None
+            else BUILDING_AREA_LOOKUP[lighting_space_type]
+        )
+
+        if space_BPF_BAT in zone_BPF_BAT_dict:
+            zone_BPF_BAT_dict[space_BPF_BAT] += space.get("floor_area", ZERO.AREA)
+        else:
+            zone_BPF_BAT_dict[space_BPF_BAT] = space.get("floor_area", ZERO.AREA)
+
+    return zone_BPF_BAT_dict
```
weilixu, line 148: This may return an empty dict. So the type hint is incorrect.
I suggest changing it to either
1. `-> dict` and provide context in the documentation that his function may return empty dictionary if zone does not have space.
2. Add an RCT assertion, `assert_(zone_BPF_BAT_dict, f"No spaces have been found in zone {zone_id}. Check the RPD inputs."`

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT_test.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test__TEST_RMD__is_valid():
+    schema_validation_result = schema_validate_rmr(TEST_RPD)
+    assert schema_validation_result[
+        "passed"
+    ], f"Schema error: {schema_validation_result['error']}"
+
+
+def test__get_zone_BPF_BAT__all_undetermined():
+    assert get_zone_BPF_BAT(TEST_RMD, "zone_1") == {"UNDETERMINED": 300 * ureg("m2")}
```
weilixu, line 94: Should add one more test, a zone has no spaces scenario.

weilixu (approved):
Approved with comments need to be addressed.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT_test.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
yunjoonjung-PNNL: 1. Since we added the point 2 (added assert_ - right below), we no longer return an empty dictionary. Therefore, I didn't do anything for this comment. Please let me know if I'm missing something here!
2. Addressed.

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
yunjoonjung-PNNL: Good point. Addressed

rct229/rulesets/ashrae9012019/ruleset_functions/get_zone_BPF_BAT.py
weilixu: Thanks Yun Joon. You can choose one over another. That is ok.

weilixu (approved):

PR #1308, opened 2024-04-03
RS/YJ/get_BPF_building_area_types_and_zones function

yunjoonjung-PNNL:
I modified the implementation logic slightly because I believe the current logic for RDS doesn't yield the expected results as outlined in the RDS. (I'd appreciate it if @JacksonJ-KC could review this PR to verify the implementation.)

To see the as-is implementation, please refer to [this commit](https://github.com/pnnl/ruleset-checking-tool/commit/3cc2afb701f29795bb63c65f155b13c03f11e547). Based on our discussion, I'll further modify the RDS. Thanks!


rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,142 @@
...
+    LIGHTING_BUILDING_AREA.TOWN_HALL: "OFFICE",
+    LIGHTING_BUILDING_AREA.TRANSPORTATION: "ALL_OTHER",
+    LIGHTING_BUILDING_AREA.WAREHOUSE: "WAREHOUSE",
+    LIGHTING_BUILDING_AREA.WORKSHOP: "WAREHOUSE",
+    LIGHTING_BUILDING_AREA.NONE: "NONE",
+}
+
+
+class BuildingAreaTypeWithTotalAreaAndZone(TypedDict):
+    zone_id: list
```
JacksonJ-KC, line 53: `zone_id: List[str]`

JacksonJ-KC:
> I modified the implementation logic slightly because I believe the current logic for RDS doesn't yield the expected results as outlined in the RDS. (I'd appreciate it if @JacksonJ-KC could review this PR to verify the implementation.)
> 
> To see the as-is implementation, please refer to [this commit](https://github.com/pnnl/ruleset-checking-tool/commit/3cc2afb701f29795bb63c65f155b13c03f11e547). Based on our discussion, I'll further modify the RDS. Thanks!

Hi @yunjoonjung-PNNL From what I see so far, your implementation looks like it follows what I had in mind. Can you elaborate for me the parts of the RDS that might be incorrect?

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
yunjoonjung-PNNL: Good catch. Addressed. I used ```list``` instead of ```List``` because Python 3.10 has this built-in ```list```. If I remember correctly, earlier versions didn't have ```list```.

JacksonJ-KC:
@yunjoonjung-PNNL An overview of the intent here. The main objective is to assign __one__ of the area types from 90.1 Table 4.2.1.1 to each building segment using either the lighting building area type or the lighting space type. 

If the `building_segment.lighting_building_area_type` is defined, it applies to the area of all spaces in the building segment. 

If it is not defined, we attempt to guess the best matching area type from 4.2.1.1 based on the `space.lighting_space_type` for each space, and we keep track of the associated area for each unique guess. After iterating through every space, whichever guess has the most area associated with it will be the building area type that is inherited by ALL spaces area in that building segment.

yunjoonjung-PNNL:
> classification_source

Thanks for the clarification. I believe your 1st and 2nd paragraphs were implemented correctly and 3rd paragraph wasn't. 

Question to @weilixu) Do we need to consider a multiple ```building_segment```s case? I vaguely remember that (I might be wrong) RCT only deals with one building and thus one building_segment.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,142 @@
...
+            and building_segment["lighting_building_area_type"]
+            != LIGHTING_BUILDING_AREA.NONE
+        ):
+            building_segment_BPF_BAT = LIGHTING_BUILDING_AREA_LOOKUP[
+                building_segment["lighting_building_area_type"]
+            ]
+            classification_source = "BUILDING_SEGMENT_LIGHTING"
+        else:
+            building_segment_space_types_areas_dict = {}
+            building_segment_BPF_BAT = {}
```
JacksonJ-KC, line 87: A building segment can only have 1 BPF BAT assigned to it so I don't see why it is a dict in the space lighting classification method.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,142 @@
...
+                    {
+                        key: building_segment_space_types_areas_dict.get(key, ZERO.AREA)
+                        + zone_BPF_BAT_dict.get(key, ZERO.AREA)
+                        for key in set(building_segment_space_types_areas_dict)
+                        | set(zone_BPF_BAT_dict)
+                    }
+                )
+
+                # find the key that has the greatest value
+                building_segment_BPF_BAT[zone_id] = max(
```
JacksonJ-KC, line 104: I don't follow this part. We don't need to keep track of the max value for each zone. We only need to determine the max in `building_segment_space_types_areas_dict` after completing the iteration through all zones.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,142 @@
...
+                    building_segment_space_types_areas_dict,
+                    key=lambda k: building_segment_space_types_areas_dict[k],
+                )
+            classification_source = "SPACE_LIGHTING"
+
+        for zone in find_all("$.zones[*]", building_segment):
```
JacksonJ-KC, line 110: I know that the RDS has iteration through all zones twice separately but I wonder if we should have a more efficient implementation.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
JacksonJ-KC: In other words, the get_zone_BPF_BAT makes a guess at the associated BPF BAT for every space in a zone and returns a dict with the BPF BAT as the key, and the total space area as the value, e.g. `{SCHOOL: 50000, UNDETERMINED: 2000}`. These keys and values get updated into the `building_segment_space_types_areas_dict` by summing the area values throughout the iteration through zones. After the iteration through all zones in a building segment, get the key associated with the max value in `building_segment_space_types_areas_dict`, and update `building_area_types_with_total_area_and_zones_dict` with that key and the sum of all space area in that building segment.

JacksonJ-KC:
> > classification_source
> 
> Thanks for the clarification. I believe your 1st and 2nd paragraphs were implemented correctly and 3rd paragraph wasn't.
> 
> Question to @weilixu) Do we need to consider a multiple `building_segment`s case? I vaguely remember that (I might be wrong) RCT only deals with one building and thus one building_segment.

@yunjoonjung-PNNL I can weigh in here and say we 100% need to consider multiple building_segment. A single building can have multiple building_segments. The intent of the return dict of this function is to track the unique BPF BAT's from each building_segment, the area associated with those building_segments, and if any building_segment has classification_source "SPACE_LIGHTING" for a particular BPF BAT, to prioritize that as the classification_source for that BPF BAT

weilixu:
> > classification_source
> 
> Thanks for the clarification. I believe your 1st and 2nd paragraphs were implemented correctly and 3rd paragraph wasn't.
> 
> Question to @weilixu) Do we need to consider a multiple `building_segment`s case? I vaguely remember that (I might be wrong) RCT only deals with one building and thus one building_segment.

Yes, we need to consider multiple `building_segment` case. RCT engine can only deal with one `RMD` per time. `Buildings` and Building_segments` shall be fine.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,142 @@
...
+                # merge zone_BPF_BAT_dict and building_segment_space_types_areas_dict
+                # if the two dicts have the same keys, add the two values
+                building_segment_space_types_areas_dict.update(
+                    {
+                        key: building_segment_space_types_areas_dict.get(key, ZERO.AREA)
+                        + zone_BPF_BAT_dict.get(key, ZERO.AREA)
+                        for key in set(building_segment_space_types_areas_dict)
+                        | set(zone_BPF_BAT_dict)
+                    }
+                )
```
weilixu, line 101: I am a bit confused with the implementation here. Is it more straightforward following the RDS description?
```python
for key in zone_BPF_BAT_dict:
    building_segment_space_types_areas_dict[key] = building_segment_space_types_areas_dict.get(key, ZERO.AREA) + zone_BPF_BAT_dict[key]
```

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,142 @@
...
+                building_segment_BPF_BAT_zone
+                not in building_area_types_with_total_area_and_zones_dict
+            ):
+                building_area_types_with_total_area_and_zones_dict[
+                    building_segment_BPF_BAT_zone
+                ] = {
+                    "zone_id": [],
+                    "area": ZERO.AREA,
+                    "classification_source": classification_source,
+                }
```
weilixu, line 127: Isn't this part on the same level as the `for building_segment`?

weilixu (commented):
There are something not correct in this PR. The logic to create `building_area_types_with_total_area_and_zones_dict` needs revisit.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,142 @@
...
+    for building_segment in find_all("$.buildings[*].building_segments[*]", rmd):
+        if (
+            building_segment.get("lighting_building_area_type") is not None
+            and building_segment["lighting_building_area_type"]
+            != LIGHTING_BUILDING_AREA.NONE
+        ):
+            building_segment_BPF_BAT = LIGHTING_BUILDING_AREA_LOOKUP[
+                building_segment["lighting_building_area_type"]
+            ]
+            classification_source = "BUILDING_SEGMENT_LIGHTING"
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

yunjoonjung-PNNL:
@JacksonJ-KC @weilixu 

Could you review this func one more time? I made big updates to make the implementation simpler than the original RDS.

@Jiarongx-Xie 
You can pull this branch for your rule 1-1 testing. Please let me know if anything came out unexpectedly.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,124 @@
...
+        classification_source: str,
+    ) -> dict:
+
+        building_area_types_with_total_area_and_zones_dict[building_segment_BPF_BAT] = {
+            "zone_id": [],
+            "area": ZERO.AREA,
+            "classification_source": classification_source,
+        }
+
+        return building_area_types_with_total_area_and_zones_dict
```
weilixu, line 52: This function is not a best practice in functioning programming concept.
It passes in a variable, modifies the variable, and returns the same variable.
It is not ideal from the memory pointer view and could cause memory leaks. But in Python, the memory is protected so it is OK.

If you want something like this, consider a function just simply initialize a dictionary such as:
```python
def _initialize_bldg_segment_bpf_bat() -> dict:
    return {
        "zone_id": [],
        "area": ZERO.AREA
        "classification_source: ClassificationType. BUILDING_SEGMENT_LIGHTING
    }
```
Then in the body, you can modify the parameters accordingly.

weilixu (approved):
Approved with one comment for review.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
yunjoonjung-PNNL: Good point and thank you for letting me know about this! 

I removed the ```_initialize_bldg_segment_bpf_bat``` function and defined the ```building_area_types_with_total_area_and_zones_dict``` below. The main reason why I created the function was I wanted to get rid of redundancy.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,113 @@
...
+                        "area": ZERO.AREA,
+                        "classification_source": classification_source,
+                    }
+
+                building_area_types_with_total_area_and_zones_dict[
+                    building_segment_BPF_BAT
+                ]["zone_id"].append(zone["id"])
+                building_area_types_with_total_area_and_zones_dict[
+                    building_segment_BPF_BAT
+                ]["area"] += sum(area for area in zone_BPF_BAT_dict.values())
```
weilixu, line 86: I think you want to make sure that this `building_BPF_BAT` source is SPACE_LIGHTING. So in here, you should also override the `classification_source`
```python
building_area_types_with_total_area_and_zones_dict[building_segment_BPF_BAT]["classification_source"] = ClassificationType.SPACE_LIGHTING
```

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,113 @@
...
+                    building_segment_BPF_BAT
+                    not in building_area_types_with_total_area_and_zones_dict
+                ):
+                    building_area_types_with_total_area_and_zones_dict[
+                        building_segment_BPF_BAT
+                    ] = {
+                        "zone_id": [],
+                        "area": ZERO.AREA,
+                        "classification_source": classification_source,
+                    }
```
weilixu, line 79: Can create a typedict for this enum.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
```diff
@@ -0,0 +1,113 @@
...
+                if (
+                    building_segment_BPF_BAT
+                    not in building_area_types_with_total_area_and_zones_dict
+                ):
+                    building_area_types_with_total_area_and_zones_dict[
+                        building_segment_BPF_BAT
+                    ] = {
+                        "zone_id": [],
+                        "area": ZERO.AREA,
+                        "classification_source": classification_source,
```
weilixu, line 78: To make it clear, I think you do not need the line 53 and line 56. You just assign `ClassificationType.SPACE_LIGHTING` here.
Same as the later portion of the code.

weilixu (approved):
Approved with one comment needs to address

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/ruleset_functions/get_BPF_building_area_types_and_zones.py
yunjoonjung-PNNL: Addressed.

PR #1309, opened 2024-04-04
RS/YJ/Rule 1-2

yunjoonjung-PNNL:
~~This rule hasn't been tested yet. Thanks!~~

This rule has been tested. Thanks!

### DIFF
--- a/rct229/rulesets/ashrae9012019/section1/section1rule2.py
+++ b/rct229/rulesets/ashrae9012019/section1/section1rule2.py
@@ -91,9 +91,9 @@ class Section1Rule2(RuleDefinitionBase):
 
         FAIL_MSG = ""
         if len(pci_set) != 1:
-            FAIL_MSG = "uleset expects exactly one PCI value to be used in the project"
+            FAIL_MSG = "Ruleset expects exactly one PCI value to be used in the project"
         elif len(pbp_set) != 1:
-            FAIL_MSG = "uleset expects exactly one PBP value to be used in the project"
+            FAIL_MSG = "Ruleset expects exactly one PBP value to be used in the project"
         elif len(bbp_set) != 1:
             FAIL_MSG = "Ruleset expects exactly one BBP value to be used in the project"
 
--- a/rct229/rulesets/ashrae9012019/section1/section1rule2.py
+++ b/rct229/rulesets/ashrae9012019/section1/section1rule2.py
@@ -65,14 +65,10 @@ class Section1Rule2(RuleDefinitionBase):
                     find_one("$.output.baseline_building_performance_energy_cost", rmd)
                 )
 
-        pci_set = set(pci_set)
-        pbp_set = set(pbp_set)
-        bbp_set = set(bbp_set)
-
         return {
-            "pci_set": pci_set,
-            "pbp_set": pbp_set,
-            "bbp_set": bbp_set,
+            "pci_set": set(pci_set),
+            "pbp_set": set(pbp_set),
+            "bbp_set": set(bbp_set),
         }
 
     def rule_check(self, context, calc_vals=None, data=None):
--- a/rct229/rulesets/ashrae9012019/section1/section1rule2.py
+++ b/rct229/rulesets/ashrae9012019/section1/section1rule2.py
@@ -1,3 +1,4 @@
+from pydash import filter_
 from rct229.rule_engine.rule_base import RuleDefinitionBase
 from rct229.rule_engine.ruleset_model_factory import produce_ruleset_model_instance
 from rct229.rulesets.ashrae9012019 import (
@@ -8,6 +9,7 @@ from rct229.rulesets.ashrae9012019 import (
     PROPOSED,
     USER,
 )
+from rct229.utils.assertions import assert_
 from rct229.utils.jsonpath_utils import find_one
 from rct229.utils.std_comparisons import std_equal
 
@@ -35,14 +37,6 @@ class Section1Rule2(RuleDefinitionBase):
             standard_section="Section G1.2.2",
             is_primary_rule=True,
             rmr_context="ruleset_model_descriptions/0",
-            required_fields={
-                "$": ["output"],
-                "output": [
-                    "performance_cost_index",
-                    "total_proposed_building_energy_cost_including_renewable_energy",
-                    "baseline_building_performance_energy_cost",
-                ],
-            },
         )
 
     def get_calc_vals(self, context, data=None):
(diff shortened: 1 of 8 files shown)


Jiarongx-Xie:
Hi Yun Joon, I only have one question, why not including the user model and the proposed model as indicated in the RDS? Could you clarify that? Thanks! @yunjoonjung-PNNL

yunjoonjung-PNNL:
> Hi Yun Joon, I only have one question, why not including the user model and the proposed model as indicated in the RDS? Could you clarify that? Thanks! @yunjoonjung-PNNL

That's because I made a mistake :). Addressed. Good catch!

Jiarongx-Xie:
> > Hi Yun Joon, I only have one question, why not including the user model and the proposed model as indicated in the RDS? Could you clarify that? Thanks! @yunjoonjung-PNNL
> 
> That's because I made a mistake :). Addressed. Good catch!

Thank you for addressing that! Shall we also update other parts? Please see my comments. @yunjoonjung-PNNL

rct229/rulesets/ashrae9012019/section1/section1rule2.py
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+    def rule_check(self, context, calc_vals=None, data=None):
+        pci_set = calc_vals["pci_set"]
+        pbp_set = calc_vals["pbp_set"]
+        bbp_set = calc_vals["bbp_set"]
+
+        return len(pci_set) == len(pbp_set) == len(bbp_set) == 1 and std_equal(
+            pbp_set[0] / bbp_set[0], pci_set[0]
+        )  # TODO: what if  bbp_set[0] == 0?
```
JacksonJ-KC, line 100: Good catch. My suggestion is to add bbp_set[0] != 0 in the return conditions and also add a fail message for this case to say "Ruleset expects `baseline_building_performance_energy_cost` to be greater than 0"

yunjoonjung-PNNL: Good point. Addressed

JacksonJ-KC (approved):

Jiarongx-Xie (approved):

rct229/rulesets/ashrae9012019/section1/section1rule2.py
```diff
@@ -0,0 +1,125 @@
...
+                PROPOSED=True,
+            ),
+            rmrs_used_optional=produce_ruleset_model_instance(
+                USER=True,
+                BASELINE_0=True,
+                BASELINE_90=True,
+                BASELINE_180=True,
+                BASELINE_270=True,
+                PROPOSED=True,
+            ),
```
weilixu, line 37: In the optional dict, you do not need to set `USER`, `BASELINE_0`, and `PROPOSED` because these are required in the models.

rct229/rulesets/ashrae9012019/section1/section1rule2.py
```diff
@@ -0,0 +1,125 @@
...
+                        rmd,
+                    )
+                )
+                bbp_set.append(
+                    find_one("$.output.baseline_building_performance_energy_cost", rmd)
+                )
+
+        pci_set = filter_(pci_set, lambda x: x is not None)
+        pbp_set = filter_(pbp_set, lambda x: x is not None)
+        bbp_set = filter_(bbp_set, lambda x: x is not None)
```
weilixu, line 73: Switch to the Python native filter function due to performance issues with Pydash packages.

weilixu (approved):
Approved with 2 comments need to be addressed.

rct229/rulesets/ashrae9012019/section1/section1rule2.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section1/section1rule2.py
yunjoonjung-PNNL: Good point. Addressed.

rct229/rulesets/ashrae9012019/section1/section1rule2.py
weilixu: @yunjoonjung-PNNL 
I see the reason why set `USER` to optional. It is because the test case does not have a user RMDs.
@yunjoonjung-PNNL could you test if just keep `USER` to be optional but make `BASELINE_0` and `PROPOSED` to be mandatory, will it pass the CI?
@jugonzal07, is it possible to make the spreadsheet generate 3 RMDs? My guess is it could require a lot effort but is it possible to add a script to copy a proposed case for user case in the post processing script?

yunjoonjung-PNNL: @weilixu When I kept ```USER``` optional, and made ```BASELINE_0``` and ```PROPOSED``` mandatory, this ruleset fails.

Question) do ```USER```, ```BASELINE_0```, and ```PROPOSED``` have to always exist in this ruleset? My understanding is it's ok to have just one of ```USER```, ```BASELINE_0```, ```BASELIN_90```, ```BASELINE_180```, ```BASELINE_270```, AND ```PROPOSED``` because of [this line in the RDS](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section1/Rule1-2.md#:~:text=the%20various%20outputs-,(it%20only%20needs%20to%20exist%20in%20at%20least%201%20output),-%3A%20pci_set%20%3D%20set).

weilixu: Ok, this line is a bit misleading.
Because the way how schema structures the output, so the information in the output is repeated in every RMD model.
That is why it says it only needs exist in one model.

My understanding is since we are adding these into the rule definition as part of evaluation. The RMDs themselves shall exist. This will include the mandatory set (`USER`, `PROPOSED`, `BASELINE_0`). The line is saying the output data in the RMDs can be optional and only need one exist among the 3 or 6 RMDs.

yunjoonjung-PNNL: Thank you for this Weili.

I made a PR to address this discussion ([HERE](https://github.com/pnnl/ruleset-checking-tool/pull/1320)). 

I believe RMDs in other sections don't always include all the ```USER```, ```BASELINE_0```, and ```PROPOSED```. Should we include all of these in RMDs in other sections as well?

jugonzal07: > @yunjoonjung-PNNL I see the reason why set `USER` to optional. It is because the test case does not have a user RMDs. @yunjoonjung-PNNL could you test if just keep `USER` to be optional but make `BASELINE_0` and `PROPOSED` to be mandatory, will it pass the CI? @jugonzal07, is it possible to make the spreadsheet generate 3 RMDs? My guess is it could require a lot effort but is it possible to add a script to copy a proposed case for user case in the post processing script?

Generating `USER`, `BASELINE_0`, and `PROPOSED` is no problem, we can do that easily with the spreadsheet. We just overwrite a formula. The challenge only comes about if we need to generate different versions of `BASELINE_X`. As those stand, they appear to be duplicated when ingesting ruletest JSONs:
https://github.com/pnnl/ruleset-checking-tool/blob/develop/rct229/ruletest_engine/ruletest_rmd_factory.py#L44-L72

PR #1313, opened 2024-04-09
Ruletest 1-6

jugonzal07:
First cut at ruletest for 1-6


yunjoonjung-PNNL:
Hi Juan, Could you point me to where 1-6 RDS is located? I don't see it in the develop branch now..

weilixu:
I thought 1-6 is the user compare to proposed and 1-7 is proposed compare to baseline.
@JacksonJ-KC could you confirm?

I am getting failures:

FAILURE: Test rule-1-6-b passed unexpectedly. The following condition was not identified: A one story building's baseline and proposed RMDs  fail to have matching list lengths for value determined by Table G3.1(1) Baseline Building Performance and https://github.com/open229/ruleset-model-description-schema/blob/main/schema-source/ASHRAE229_extra.schema.yaml.
Unknown context: Calculated values - {'comparison_result': True, 'error_msg_list': []}
Processing Rule 1-6
FAILURE: Test rule-1-6-c passed unexpectedly. The following condition was not identified: A one story building's baseline and proposed RMDs  fail to have matching IDs in their lists for value determined by Table G3.1(1) Baseline Building Performance and https://github.com/open229/ruleset-model-description-schema/blob/main/schema-source/ASHRAE229_extra.schema.yaml.
Unknown context: Calculated values - {'comparison_result': True, 'error_msg_list': []}
Processing Rule 1-6
FAILURE: Test rule-1-6-d passed unexpectedly. The following condition was not identified: A one story building's baseline and proposed RMDs  fail to have matching IDs for a non-array value that should match as determined by Table G3.1(1) Baseline Building Performance and https://github.com/open229/ruleset-model-description-schema/blob/main/schema-source/ASHRAE229_extra.schema.yaml.
Unknown context: Calculated values - {'comparison_result': True, 'error_msg_list': []}

Could you provide a list of mismatches in these test cases so I can look into the issue? The file is very large so it might be more efficient if you can let me know the mismatch portion.

weilixu:
@jugonzal07 from the [master list of rules](https://onedrive.live.com/edit?id=796769604E905084!1705&resid=796769604E905084!1705&ithint=file%2cxlsx&authkey=!ACQq5kwrZXzWJVg&wdo=2&cid=796769604e905084), 1-6 is baseline equals to proposed and 1-7 is proposed equals to user. Your interpretation is correct.

jugonzal07:
> @jugonzal07 I did a comparison in 1-6 and it seems for case b, both proposed and baseline have `Thermal Zone 2` as an empty thermal zone. case c: both proposed and baseline have `Terminal One` id. case d: both proposed and baseline have climate zone `CZ3A`.
> 
> It seems case b, c, d all have identical baseline and proposed. Could you verify the json generation script?

Sorry about that! I'm not sure what copy/paste mistake I must have made. Corrected those now.

PR #1316, opened 2024-04-10
Ruletest 1-2

jugonzal07:
First cut at ruletest 1-2


yunjoonjung-PNNL (approved):
Great work! Thanks Juan!

JacksonJ-KC:
Test case 1-2-b fails for 2 different reasons that I think are important to capture separately but not together. Where a value (`performance_cost_index`) is present in more than one RMD but has a different value, and where the `performance_cost_index` does not equal the expected value.

PR #1318, opened 2024-04-10
Ruletest 1-4

jugonzal07:
First cut at ruletest 1-4

### DIFF
--- a/rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section1/rule_1_4.json
+++ b/rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section1/rule_1_4.json
@@ -327,6 +327,114 @@
         "Section": 1,
         "Rule": 4,
         "Test": "d",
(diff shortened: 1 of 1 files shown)


rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section1/rule_1_4.json
```diff
@@ -0,0 +1,538 @@
...
+                                                ]
+                                            }
+                                        ]
+                                    }
+                                ]
+                            }
+                        ],
+                        "output": {
+                            "id": "Output 1",
+                            "performance_cost_index": 1,
```
yunjoonjung-PNNL, line 317: I believe the ```performance_cost_index``` value must be greater than the```performance_cost_index_target``` value to make the case (c) fail.

jugonzal07: Good catch. Fixed!

yunjoonjung-PNNL (approved):

JacksonJ-KC:
suggest adding one test case that fails because an output value differs between RMDs

PR #1324, opened 2024-04-23
Ruletest 10-14 Updates

jugonzal07:
Added many test cases and corrected some errors in old one as pointed out by Jackson.


jugonzal07:
Looks like I may need to address merge conflicts.

JacksonJ-KC:
@jugonzal07 Hi Juan, it looks like we are getting unexpected fails for many of the new cases. When I started looking into it I saw in case E that there is no cooling system capacity defined. It is not intuitive but the heat pump system capacities in table G3.5.2 for heating mode are actually the cooling capacities. This might be causing issues for other cases as well e, f, g, h, i, j, k, m all fail their test currently. If you upload the spreadsheet it might help me to provide more feedback at once, instead of addressing 1 issue at a time

jugonzal07:
> @jugonzal07 Hi Juan, it looks like we are getting unexpected fails for many of the new cases. A couple of things I am noticing:
> 
> 1. the heat pump systems in heating mode actually use the cooling system capacity for the table lookup. Can you switch the capacities you have defined from the heating system to the cooling system?
> 2. the zone.aggregation_factor plays an important role in the logic but does not appear to be incorporated in the test cases. Can you make sure this gets added where it is needed?
> 
> I know the number of cases is getting out of hand but I would like to have test cases that correspond to each of the outcome cases in the RDS. I have not looked in detail enough at the cases you have to know if we are close to this already.
> 
> Test cases e, f, g, h, i, j, k, m all fail their test currently.

No sweat, Jackson-- it's an involved rule! Sorry to hear that it's not working out as I had hoped.

I'll look into updating things based on your #1 and #2 comments.

In case it helps, here is the corresponding Excel file:

https://onedrive.live.com/edit.aspx?resid=796769604E905084!1855&cid=796769604e905084&authkey=!ACtFTkgcGDgG1Qk&CT=1714509769478&OR=ItemsView

jugonzal07:
> @jugonzal07 Hi Juan, it looks like we are getting unexpected fails for many of the new cases. A couple of things I am noticing:
> 
> 1. the heat pump systems in heating mode actually use the cooling system capacity for the table lookup. Can you switch the capacities you have defined from the heating system to the cooling system?
> 2. the zone.aggregation_factor plays an important role in the logic but does not appear to be incorporated in the test cases. Can you make sure this gets added where it is needed?
> 
> I know the number of cases is getting out of hand but I would like to have test cases that correspond to each of the outcome cases in the RDS. I have not looked in detail enough at the cases you have to know if we are close to this already.
> 
> Test cases e, f, g, h, i, j, k, m all fail their test currently.

Ok @JacksonJ-KC ,

I attempted to add your feedback to the tests. 

1. Changed heating capacity over to cooling capacity for System 4 test cases
2. Added test case "o" that incorporates zone aggregation factors for a System 3 test case. I'm less certain I did this one exactly right, so let me know if I goofed anywhere.

JacksonJ-KC:
Thanks for the updates Juan! We do also need the aggregation factor specified as 1 in a lot of these cases - anytime the capacity is larger than the smallest capacity threshold and you are not trying to catch a case where `zone_aggregation_factor_is_undefined_and_needed = True`

jugonzal07:
> Thanks for the updates Juan! We do also need the aggregation factor specified as 1 in a lot of these cases - anytime the capacity is larger than the smallest capacity threshold and you are not intentionally trying to catch a case where `zone_aggregation_factor_is_undefined_and_needed = True`

Hi Jackson,

I added zone aggregation factors for all system 4 test cases greater than their minimum capacity threshold (just made them 1) then added an additional test 10-14p that checks for the scenario where a system 4 requires a zone aggregation factor, doesn't have one (i.e., zone_aggregation_factor_is_undefined_and_needed = True) and returns undetermined.

Let me know if you think I missed any! The OneDrive spreadsheet has the latest updates.

JacksonJ-KC:
Juan, any idea why this is failing validation?
"schema invalid: Additional properties are not allowed ('aggregation_factor ' was unexpected)"

JacksonJ-KC:
Case G is expecting the low temperature efficiency metric to be "HEAT_PUMP_COEFFICIENT_OF_PERFORMANCE_LOW_TEMPERATURE_NO_FAN"

JacksonJ-KC:
Test case m falls under Case 7 of the RDS and expects an UNDETERMINED outcome

JacksonJ-KC:
All test cases will pass the rule test after the above comments are addressed! Confirmed locally. 🎉

jugonzal07:
> All test cases will pass the rule test after the above comments are addressed! Confirmed locally. 🎉 @jugonzal07 I updated the TCD spreadsheet on OneDrive

ahh didn't read that you updated the TCD spreadsheet and I did it locally. I implemented your fixes manually. (I copied what I saw on the other PR: https://github.com/pnnl/ruleset-checking-tool/pull/1298/) 

This should be nearly the same. The only difference is this should include the updates I made to test descriptions.

jugonzal07:
For the sake of keeping things orderly, perhaps we close this PR and use the spreadsheet on the OneDrive as the new standard. The test description differences were very minor.

PR #1330, opened 2024-04-24
RDS/JDJ/Rule 16-5

JacksonJ-KC:
Pull request open for feedback


docs/section16/Rule 16-5.md
```diff
@@ -0,0 +1,34 @@
+# Elevators – Rule 16-5  
+**Schema Version:** 0.0.36        
+**Mandatory Rule:** True          
+**Rule ID:** 16-5  
+**Rule Description:** When included in the proposed design, the baseline elevator cab ventilation fan and lights shall operate continuously  
+**Rule Assertion:** B-RMD = Expected Value                                           
+**Appendix G Section:** G3.1  
+**Appendix G Section Reference:** Table G3.1-16
```
claperle, line 8: recommend a change to "Table G3.1-16 Baseline Building Performance column"

docs/section16/Rule 16-5.md
```diff
@@ -0,0 +1,34 @@
...
+**Appendix G Section:** G3.1  
+**Appendix G Section Reference:** Table G3.1-16  
+**Data Lookup:** None  
+**Evaluation Context:** Each elevator  
+
+**Applicability Checks:**  
+  1. The proposed building has an elevator.  
+
+**Function Call:**  
+N/A
```
claperle, line 16: add compare_schedules

docs/section16/Rule 16-5.md
```diff
@@ -0,0 +1,34 @@
...
+
+**Applicability Check 1:**
+- Rule is applicable if any building in the proposed RMD has an elevator: `if any(len(building.elevators) > 0 for building in P_RMD...buildings:`
+
+## Rule Logic:
+- For each elevator in the baseline RMD: `for elevator in B_RMD...elevators`
+  - Get the modeled schedule for the baseline elevator cab ventilation fan: `cab_ventilation_schedule_b = elevator.cab_ventilation_fan_multiplier_schedule`
+  - Get the modeled schedule for the baseline elevator cab lighting: `cab_lighting_schedule_b = elevator.cab_lighting_multiplier_schedule`
+  - Compare the schedules: `comparison_data = compare_schedules(cab_lighting_schedule_b, cab_ventilation_schedule_b):`
```
claperle, line 25: I see that this RDSs checks that the ventilation and cab lighting schedules were modeled the same but not that it checks that they were modeled as continuously operating. I think additional logic is needed to check that the schedules are equal to 1 for all hours as well.

docs/section16/Rule 16-5.md
```diff
@@ -0,0 +1,34 @@
...
+  - Get the modeled schedule for the baseline elevator cab ventilation fan: `cab_ventilation_schedule_b = elevator.cab_ventilation_fan_multiplier_schedule`
+  - Get the modeled schedule for the baseline elevator cab lighting: `cab_lighting_schedule_b = elevator.cab_lighting_multiplier_schedule`
+  - Compare the schedules: `comparison_data = compare_schedules(cab_lighting_schedule_b, cab_ventilation_schedule_b):`  
+
+  **Rule Assertion:**  
+    - Case 1: If the schedules for lighting and ventilation are identical and continuous for all hours: `if comparison_data['SCHEDULE_1_EFLH'] == len(cab_lighting_schedule_b) == len(cab_ventilation_schedule_b): PASS`
+    - Case 2: Else: `FAIL`
+
+**Notes/Questions:**
+1. Can we create additional keys in the return dictionary of the compare_schedules ruleset function to also include the schedule 1 EFLH and schedule 2 EFLH? Is there a more preferred way to do this?
```
claperle, line 32: I am not following what you are trying to do with this additional key.

I think you can just check that the schedules match and the check that SUM(cab_ventilation_schedule_b) = 8760?

JacksonJ-KC: You're right I was overthinking this. I will update.

docs/section16/Rule 16-5.md
```diff
@@ -0,0 +1,34 @@
...
+- Rule is applicable if any building in the proposed RMD has an elevator: `if any(len(building.elevators) > 0 for building in P_RMD...buildings:`
+
+## Rule Logic:
+- For each elevator in the baseline RMD: `for elevator in B_RMD...elevators`
+  - Get the modeled schedule for the baseline elevator cab ventilation fan: `cab_ventilation_schedule_b = elevator.cab_ventilation_fan_multiplier_schedule`
+  - Get the modeled schedule for the baseline elevator cab lighting: `cab_lighting_schedule_b = elevator.cab_lighting_multiplier_schedule`
+  - Compare the schedules: `comparison_data = compare_schedules(cab_lighting_schedule_b, cab_ventilation_schedule_b):`  
+
+  **Rule Assertion:**  
+    - Case 1: If the schedules for lighting and ventilation are identical and continuous for all hours: `if comparison_data['SCHEDULE_1_EFLH'] == len(cab_lighting_schedule_b) == len(cab_ventilation_schedule_b): PASS`
```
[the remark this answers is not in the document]
JacksonJ-KC: I will change this to no longer use the compare_schedule function, but the check for continuous operation is accomplished here by making sure the EFLH is equal to the length of the schedule

docs/section16/Rule 16-5.md
```diff
@@ -22,10 +22,9 @@ N/A
 - For each elevator in the baseline RMD: `for elevator in B_RMD...elevators`
   - Get the modeled schedule for the baseline elevator cab ventilation fan: `cab_ventilation_schedule_b = elevator.cab_ventilation_fan_multiplier_schedule`
   - Get the modeled schedule for the baseline elevator cab lighting: `cab_lighting_schedule_b = elevator.cab_lighting_multiplier_schedule`
-  - Compare the schedules: `comparison_data = compare_schedules(cab_lighting_schedule_b, cab_ventilation_schedule_b):`  
 
   **Rule Assertion:**  
-    - Case 1: If the schedules for lighting and ventilation are identical and continuous for all hours: `if comparison_data['SCHEDULE_1_EFLH'] == len(cab_lighting_schedule_b) == len(cab_ventilation_schedule_b): PASS`
+    - Case 1: If the schedules for lighting and ventilation are identical and continuous for all hours: `if sum(cab_ventilation_schedule_b) == sum(cab_lighting_schedule_b) == len(cab_lighting_schedule_b) == len(cab_ventilation_schedule_b): PASS`
```
KarenWGard, line 27: I may be unclear on the definition of “operate continuously”, but it seems that the elevator cab lights & ventilation should operate at 100% during all building operation hours, so would be something like 24hr / 7 for hotel. I think there’s a building operation schedule that these schedules can be checked against

JacksonJ-KC: Operate continuously is assumed to mean 8760/8784 EFLH here. I don't think there is a definition that we can point to that includes "during operating hours" or anything to that effect but let me know if I am wrong.

KarenWGard: Is the length of the schedule always 8760 or 8784?

JacksonJ-KC: Any time the compare_schedules function is used, it includes an assertion that the schedules are the same length as the total year hours. (That function is not used here, but I am using that as an example where that assumption is worked into the code) I don't believe there is a validation check for this, so the RPD file would still be valid if schedules are not 8760/8784, but they would fail rule evaluations where they are expected to be 8760/8784

KarenWGard: Ok. I just want to make sure that if we’re expecting it to be 8760/8784, that it actually is. If you’re confident that it is based on the current logic, then I’m ok with that.

JacksonJ-KC:
Maybe we should be more strict in this evaluation of "continuous" by using the compare_schedule function with all 1 values to make sure that every hour = 1. Right now a multiplier schedule with values [4380, 4380, 0, 0, 0.... 0] would pass

claperle:
> Maybe we should be more strict in this evaluation of "continuous" by using the compare_schedule function with all 1 values to make sure that every hour = 1. Right now a multiplier schedule with values [4380, 4380, 0, 0, 0.... 0] would pass

@JacksonJ-KC  I agree

supriyagoel (approved):

docs/section16/Rule 16-5.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Rule Logic:
+- Determine if the project schedule uses a leap year: `is_leap_year = RPD.schedule.is_leap_year`
+- If the project schedule uses a leap year, set total hours to 8784, else 8760: `total_hours = 8784 if is_leap_year else 8760`
+- Create a continuous schedule to compare against, and also to use as the mask schedule for the function: `continuous_schedule = [1]*total_hours`
+- For each elevator in the baseline RMD: `for elevator in B_RMD...elevators`
+  - Get the modeled schedule for the baseline elevator cab ventilation fan: `cab_ventilation_schedule_b = elevator.cab_ventilation_fan_multiplier_schedule`
+  - Get the modeled schedule for the baseline elevator cab lighting: `cab_lighting_schedule_b = elevator.cab_lighting_multiplier_schedule`
+  - Compare the cab ventilation schedule with the continuous schedule: `vent_sched_compare_data = compare_schedules(cab_ventilation_schedule_b, continuous_schedule, continuous_schedule, is_leap_year)`
+  - Compare the cab lighting schedule with the continuous schedule: `light_sched_compare_data = compare_schedules(cab_lighting_schedule_b, continuous_schedule, continuous_schedule, is_leap_year)`
```
supriyagoel, line 29: not sure what the 'continuous schedule' is .. I believe thee intent is to verify that ventilation is provided when a zone is expected to be occupied .. Maybe we can look at the building open/close hrs and verify that the ventilation schedule is set to 1 for the open/close hrs?

supriyagoel (commented):
need one additional step in the check for 'continuous'

docs/section16/Rule 16-5.md
JacksonJ-KC: Appendix G just says the ventilation fan and lighting "both operate continuously" and I was not comfortable interpreting that to mean during building operating hours only in the implementation

claperle (approved):

docs/section16/Rule 16-5.md
supriyagoel: This makes sense

supriyagoel (approved):

JacksonJ-KC:
I believe this is ready to merge. Agree @supriyagoel ?

KarenWGard (approved):

PR #1338, opened 2024-04-30
Rds/kjw/rule 11 8

KarenWGard:


docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,46 @@
...
+  - get a dictionary of all SHW equipment connected to the SHW space type in the baseline model: `b_shw_equipment_dict = get_SHW_equipment_connected_to_use_type(B_RMD)[shw_space_type]`
+  - create a variable num_shw_systems = `len(b_shw_equip_dict)["SHWDistribution"]`
+
+    ## Rule Assertion: 
+    - Case1: there is exactly one system: PASS: `if num_shw_systems == 1: PASS`
+    - Case2: all else, FAIL: `else: FAIL`
+
+  
+  **Notes:**
+  1.  Are there configurations in which a complete SHW system could exist without SHWDistribution?
```
claperle, line 44: It shows the distrbution_system as a required data element in the ServiceWaterHeatingEquipment data table so I don't think so

JacksonJ-KC:

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,46 @@
...
+
+**Rule ID:** 11-8
+
+**Rule Description:** "One system per building area type shall be modeled in the baseline."
+
+**Rule Assertion:** Options are PASS/FAIL
+
+**Appendix G Section Reference:** Table G3.1 #11, baseline column, a + b
+
+**Evaluation Context:** B-RMD each SHW type
```
claperle, line 14: Alternatively it may be better to do by ServiceWaterHeatingDistributionSystem since this would encompass a SHW system as opposed to SHWHeatingEq where there could be mulitple SHWHeatingEq IDs per system. So it would be {"ServiceWaterHeatingDistributionSystem":{"SHWHeatingEq":["shwe1","shwe2"], "Building area types served":["Dormitory","Multifamlily"], "Pumps":["p1"], "Tanks":["t1"], "Piping":["piping1"], "SolarThermal":[], "Service_water_heating_uses":["swhuse1","shwuse2"]}} so all this information would be returned for each SHWHeatingEq ID.

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,46 @@
...
+  - look at each space: `for space_id in shw_and_spaces_dict[shw_space_type]:`
+    - get the space using get_component_by_id: `space = get_component_by_id(P_RMD, space_id)`
+    - look for the ServiceWaterHeatingUse in the space.  If even one space has a ServiceWaterHeatingUse, continue to rule logic: `if len(space.service_water_heating_uses) > 0: CONTINUE TO RULE LOGIC`
+  - if the program reaches this line without going to the rule logic, the project is not applicable for this SHW heating use: `NOT_APPLICABLE`
+
+    ## Rule Logic: 
+  - create a boolean to keep track of whether everything matches: `all_match = TRUE`
+  - create an error string: `error_str = ""`
+  - get a dictionary of all SHW equipment connected to the SHW space type in the baseline model: `b_shw_equipment_dict = get_SHW_equipment_connected_to_use_type(B_RMD)[shw_space_type]`
+  - create a variable num_shw_systems = `len(b_shw_equip_dict)["SHWDistribution"]`
```
claperle, line 36: I think this may neglect to test whether the same distribution system is attached to multiple BATs.

docs/section11/Rule11-8.md
KarenWGard: discussed 5/20/2024 - Christina is OK with current logic

claperle: Recommend updatign so the meaning of SHW type is clearer.

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,54 @@
...
+
+**Rule Assertion:** Options are PASS/FAIL
+
+**Appendix G Section Reference:** Table G3.1 #11, baseline column, a + b
+
+**Evaluation Context:** B-RMD each SHW type
+**Data Lookup:**   
+**Function Call:** 
+- **get_SHW_types_and_spaces**
+- **get_SHW_equipment_connected_to_use_type**
```
claperle, line 18: Update function names if any have changed.

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,54 @@
...
+**Appendix G Section Reference:** Table G3.1 #11, baseline column, a + b
+
+**Evaluation Context:** B-RMD each SHW type
+**Data Lookup:**   
+**Function Call:** 
+- **get_SHW_types_and_spaces**
+- **get_SHW_equipment_connected_to_use_type**  
+
+**Applicability Checks:**
+- check that the SHW type in the P_RMD has SHW loads - **is P_RMD appropriate for the applicability check, or should it be B_RMD?**
```
claperle, line 21: Recommend avoiding "SHW type" language as it is not clear what it means.

claperle: To your question, I think checking both makes sense. Although, maybe it is better to check the B_RMD in case there is an allowed measure in the proposed that eliminate loads.

claperle: To your question, I Don't feel strongly but generally I think you woudl check the proposed in instances like this.

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,54 @@
...
+**Applicability Checks:**
+- check that the SHW type in the P_RMD has SHW loads - **is P_RMD appropriate for the applicability check, or should it be B_RMD?**  
+
+## Applicability Checks:
+- only projects with SHW for the SHW space type in the proposed model are expected to have a SHW system in the baseline model.  If there is no SHW in the proposed model, we assume no SHW system is needed.
+- use the function get_SHW_types_and_spaces to get a list of spaces for each SHW type: `shw_and_spaces_dict = get_SHW_types_and_spaces(P_RMD)`
+- look at each SHW bat: `for shw_bat in shw_and_spaces_dict:`
+  - look at each space: `for space_id in shw_and_spaces_dict[shw_bat]:`
+    - get the space using get_component_by_id: `space = get_component_by_id(P_RMD, space_id)`
+    - look for the ServiceWaterHeatingUse in the space.  If even one space has a ServiceWaterHeatingUse, continue to rule logic: `if len(space.service_water_heating_uses) > 0: CONTINUE TO RULE LOGIC`
```
claperle, line 29: would this then keep repeating the logic below everytime there was a load encountered?

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,54 @@
...
+  - if the program reaches this line without going to the rule logic, the project is not applicable for this SHW heating use: `NOT_APPLICABLE`
+
+    ## Rule Logic: 
+  - get a dictionary of all SHW equipment connected to the all SHW space bats in the baseline model: `b_shw_equipment_dict = get_SHW_equipment_connected_to_use_type(B_RMD)[shw_bat]`
```
claperle, line 33: I think the name of this function has changed.

docs/section11/Rule11-8.md
KarenWGard: yes

docs/section11/Rule11-8.md
KarenWGard: discussed on call 6/5/2024 - OK to keep reference to proposed

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,54 @@
...
+
+    ## Rule Logic: 
+  - get a dictionary of all SHW equipment connected to the all SHW space bats in the baseline model: `b_shw_equipment_dict = get_SHW_equipment_associated_with_each_SWH_bat(B_RMD)[shw_bat]`
+  - get a dictionary of all SHW equipment connected to the current SHW space bats in the baseline model: `b_shw_equipment_dict_this_use = b_shw_equipment_dict[shw_bat]
+  - create a variable num_swh_systems_this_use: `num_swh_systems = len(b_shw_equipment_dict_this_use["SHWDistribution"])`
+  - create a variable num_swh_equipment_this_use: `num_swh_equipment_this_use = len(b_shw_equipment_dict_this_use["SHWHeatingEq"])`
+  - create a boolean is_referenced_in_other_bats and set it to false: `is_referenced_in_other_bats = false`
+  
+  - if there is exactly one system, check that the system is not used in any of the other BATs: `if num_swh_systems == 1:`
+    - create a variable shw_dist_id and set it to the id of the shw distribution equipment for this bat: `shw_dist_id = b_shw_equipment_dict_this_use["SHWDistribution"][0]`
```
JacksonJ-KC, line 40: inconsistent use of the words "system" and "equipment" is confusing. please change "equipment" to system when referring to ServiceWaterHeatingDistributionSystem

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,54 @@
...
+    - loop through each bat: `for other_shw_bat in b_shw_equipment_dict:`
+      - if other_shw_bat is not equal to the current shw_bat: `if other_shw_bat != shw_bat:`
+        - check whether the shw_dist_id is in the list of SHW Distribution equipment for the other SHW_bat: `if shw_dist_id in? b_shw_equipment_dict[other_shw_bat]["SHWDistribution"]:`
```
JacksonJ-KC, line 43: typo `shw_dist_id in` ~~?~~

JacksonJ-KC (changes requested):

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,54 @@
...
+  - if the program reaches this line without going to the rule logic, the project is not applicable for this SHW heating use: `NOT_APPLICABLE`
+
+    ## Rule Logic: 
+  - get a dictionary of all SHW equipment connected to the all SHW space bats in the baseline model: `b_shw_equipment_dict = get_SHW_equipment_associated_with_each_SWH_bat(B_RMD)[shw_bat]`
+  - get a dictionary of all SHW equipment connected to the current SHW space bats in the baseline model: `b_shw_equipment_dict_this_use = b_shw_equipment_dict[shw_bat]
+  - create a variable num_swh_systems_this_use: `num_swh_systems = len(b_shw_equipment_dict_this_use["SHWDistribution"])`
+  - create a variable num_swh_equipment_this_use: `num_swh_equipment_this_use = len(b_shw_equipment_dict_this_use["SHWHeatingEq"])`
+  - create a boolean is_referenced_in_other_bats and set it to false: `is_referenced_in_other_bats = false`
+  
+  - if there is exactly one system, check that the system is not used in any of the other BATs: `if num_swh_systems == 1:`
```
JacksonJ-KC, line 39: What if the BAT is `ALL_OTHERS` and the building has more than 1, different BAT that fall under `ALL_OTHERS`?

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,65 @@
+# Service_Water_Heating - Rule 11-8
+**Schema Version:** 0.0.37  
+
+**Mandatory Rule:** TRUE
+
+**Rule ID:** 11-8
+
+**Rule Description:** "One system per building area type shall be modeled in the baseline."
+
+**Rule Assertion:** Options are PASS/FAIL
```
claperle, line 10: I think this should also ahve IN_APPLICABLE for if there are no SWH loads

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,65 @@
...
+    - get the quantity of service water heating distribution systems serving this building area type: `num_swh_systems = len(swh_bats_and_equip_dict_this_use["SWHDistribution"])`
+    - get the quantity of service water heating equipment serving this building area type: `num_swh_equipment_this_use = len(swh_bats_and_equip_dict_this_use["SWHHeatingEq"])`
+    - create a boolean is_referenced_in_other_bats and set it to false: `is_referenced_in_other_bats = false`
+    
+    - if there is exactly one system, check that the system is not used in any of the other BATs: `if num_swh_systems == 1:`
+        - create a variable swh_dist_id and set it to the id of the swh distribution system for this bat: `swh_dist_id = swh_bats_and_equip_dict_this_use["SWHDistribution"][0]`
+        - loop through each bat: `for other_swh_bat in swh_bats_and_equip_dict:`
+        - if other_swh_bat is not equal to the current swh_bat: `if other_swh_bat != swh_bat:`
+            - check whether the swh_dist_id is in the list of SWH Distribution system for the other SWH_bat: `if swh_dist_id in? swh_bats_and_equip_dict[other_swh_bat]["SWHDistribution"]:`
```
claperle, line 43: Ruby

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,65 @@
...
+    - otherwise, rule is not applicable for this swh_bat: `else: NOT_APPLICABLE`
+
+        ## Rule Logic: 
+    - use get_SWH_equipment_associated_with_each_SWH_bat to get the SWH use types and SWH use in the building: `swh_bats_and_equip_dict = get_SWH_equipment_associated_with_each_SWH_bat(B_RMD)`
+    - get a dictionary of all SWH equipment connected to the current SWH bat in the baseline model: `swh_bats_and_equip_dict_this_use = swh_bats_and_equip_dict[swh_bat]`
+    - get the quantity of service water heating distribution systems serving this building area type: `num_swh_systems = len(swh_bats_and_equip_dict_this_use["SWHDistribution"])`
+    - get the quantity of service water heating equipment serving this building area type: `num_swh_equipment_this_use = len(swh_bats_and_equip_dict_this_use["SWHHeatingEq"])`
+    - create a boolean is_referenced_in_other_bats and set it to false: `is_referenced_in_other_bats = false`
+    
+    - if there is exactly one system, check that the system is not used in any of the other BATs: `if num_swh_systems == 1:`
```
claperle, line 39: Just wrappign my head around this in writing. The reason that there is no need to also check SWHEquip is because equip can only reference one distribution system and if there is only one distribution system associated with the BAT and there is only one Equip object associated with the BAT then it guarantees that Equip will not be referenced by other BATs as long as the distribution system isn't referenced by other BATs.

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,65 @@
...
+    - check whether the swh_bat is "OTHER": `if swh_bat == "OTHER":`
+        - now create a variable num_segments and set it to 0.  This variable will be used to count the number of building segments with swh_bat == "OTHER": `num_segments = 0`
+        - now look at each building segment in the building: `for building_segment in B_RMD:`
+            - get the building segment swh_bat.  If it's "OTHER" increment, num_segments: `if (get_building_segment_SWH_bat(B_RMD, building_segment) == "OTHER"): num_segments += 1`
+        - check whether there are multiple building segments with swh_bat "OTHER", if so, set the boolean other equal to true: `if num_segments > 1: multiple_segments_with_bat_other = true`
+
+    ## Rule Assertion: 
+    - Case1: there is more than one swh system in the building segment: FAIL: `if num_swh_systems > 1: FAIL`
+    - Case2: The SWH BAT is "UNDETERMINED", there is only one building segment in the model: PASS: `elif shw_bat == "UNDETERMINED" && len(B_RMD.building.building_segments) == 1: PASS`
```
claperle, line 54: I think this should also check that num_swh_equipment_this_use == 1

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,65 @@
...
+    - create a boolean called multiple_segments_with_bat_other and set it to false.  This boolean will tell us if the a) the swh_bat is "OTHER"; b) there are multiple building segments with swh_bat equal to "OTHER": `multiple_segments_with_bat_other = false` 
+    - check whether the swh_bat is "OTHER": `if swh_bat == "OTHER":`
+        - now create a variable num_segments and set it to 0.  This variable will be used to count the number of building segments with swh_bat == "OTHER": `num_segments = 0`
+        - now look at each building segment in the building: `for building_segment in B_RMD:`
+            - get the building segment swh_bat.  If it's "OTHER" increment, num_segments: `if (get_building_segment_SWH_bat(B_RMD, building_segment) == "OTHER"): num_segments += 1`
+        - check whether there are multiple building segments with swh_bat "OTHER", if so, set the boolean other equal to true: `if num_segments > 1: multiple_segments_with_bat_other = true`
+
+    ## Rule Assertion: 
+    - Case1: there is more than one swh system in the building segment: FAIL: `if num_swh_systems > 1: FAIL`
```
claperle, line 53: I think this shoudl be an or I think this should also check that num_swh_equipment_this_use > 1, right? Or am I missing something.

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,65 @@
...
+    - Case2: The SWH BAT is "UNDETERMINED", there is only one building segment in the model: PASS: `elif shw_bat == "UNDETERMINED" && len(B_RMD.building.building_segments) == 1: PASS`
+    - Case3: The SWH BAT is "UNDETERMINED" and there is more than one building segment: UNDETERMINED & provide a note: `elif swh_bat == "UNDETERMINED": UNDETERMINED note = "The Service Water Heating Building Area type for this building segment is undetermined, and there are multiple building segments in the project.  Therefore it cannot be determined whether this building segment shares a service water heating building area type with one of the other building segments."`
```
claperle, line 55: I think len(B_RMD.building.building_segments) > 1 needs to be added

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,65 @@
...
+    - Case4: the swh_bat is "OTHER" and is applied to more than one building segment - there could be multiple types of "OTHER" SWH BAT's - return UNDETERMINED: `elif multiple_segments_with_bat_other: UNDETERMINED; note = "The Service Water Heating Building Area Type is 'OTHER' and is applied to multiple building segments.  'OTHER' can cover multiple undetermined Service Water Heating Building Area Types.  Confirm that Service Water Heating Building Area Type is provided with one and only one service water heating system.`
```
claperle, line 56: Recommend rewording. Not sure what 'Other" can cover multiple undetermined SWH BATs means. Maybe just remove this part?

docs/section11/Rule11-8.md
```diff
@@ -0,0 +1,65 @@
...
+        - if other_swh_bat is not equal to the current swh_bat: `if other_swh_bat != swh_bat:`
+            - check whether the swh_dist_id is in the list of SWH Distribution system for the other SWH_bat: `if swh_dist_id in? swh_bats_and_equip_dict[other_swh_bat]["SWHDistribution"]:`
+            - set the boolean is_referenced_in_other_bats to true: `is_referenced_in_other_bats = true`
+    - create a boolean called multiple_segments_with_bat_other and set it to false.  This boolean will tell us if the a) the swh_bat is "OTHER"; b) there are multiple building segments with swh_bat equal to "OTHER": `multiple_segments_with_bat_other = false` 
+    - check whether the swh_bat is "OTHER": `if swh_bat == "OTHER":`
+        - now create a variable num_segments and set it to 0.  This variable will be used to count the number of building segments with swh_bat == "OTHER": `num_segments = 0`
+        - now look at each building segment in the building: `for building_segment in B_RMD:`
+            - get the building segment swh_bat.  If it's "OTHER" increment, num_segments: `if (get_building_segment_SWH_bat(B_RMD, building_segment) == "OTHER"): num_segments += 1`
```
claperle, line 49: I would need to go back and look at the get_building_segment_SWH_bat function but maybe you know, would this function have separate BATs if they have the same designation "OTHER", "UNDETERMINED" or would it be grouped together?

docs/section11/Rule11-8.md
KarenWGard: So, my understanding of this rule is that there can only be one distribution system per BAT.  There is no mention of only one waterheater - does this seem correct?

**ServiceWaterHeatingEquipment** can only be connected to one **ServiceWaterHeatingDistributionSystem**, so if the distribution system is only connected to one BAT, the equipment must also be only connected to one BAT, even if there are multiple.

docs/section11/Rule11-8.md
KarenWGard: the line above checks for swh_bat == "UNDETERMINED" and len(B_RMD.building_segments) == 1, so it's impossible to get here with 1 building segment.

docs/section11/Rule11-8.md
KarenWGard: 7/18 - discussed on call - conclusion that the requirement is for one "Service Water Heating System" though the definition is vague, and we should allow multiple Service Water Heating Equipment.

docs/section11/Rule11-8.md
KarenWGard: as discussed on 7/18 call - one distribution system is required, but multiple equipment is allowed

claperle (approved):

JacksonJ-KC (approved):

weilixu:
Addressed in the commit: https://github.com/pnnl/ruleset-checking-tool/commit/577ff4ce7cc8a3a51478afca563b40d141dd33be close this PR.

PR #1342, opened 2024-04-30
RS/YJ/Rule 16-2, 3, and 4

yunjoonjung-PNNL:
I included 3 rulesets here because these are very similar to each other. Thanks!


rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,85 @@
...
+                PROPOSED=True,
+            ),
+            each_rule=Section16Rule2.BuildingRule(),
+            index_rmd="baseline",
+            id="16-2",
+            description="The baseline elevator motor use shall be modeled with the same schedule as the proposed design. Rule Assertion: B-RMD = P-RMD",
+            ruleset_section_title="Elevators",
+            standard_section="Section G3.1",
+            is_primary_rule=True,
+            list_path="ruleset_model_descriptions[0].buildings[*]",
```
JacksonJ-KC, line 28: I believe list_path should be only "ruleset_model_descriptions[0]"

rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,85 @@
...
+
+        def get_calc_vals(self, context, data=None):
+            building_b = context.BASELINE_0
+            building_p = context.PROPOSED
+
+            is_leap_year_b = data["is_leap_year_b"]
+            mask_schedule = [1] * 8784 if is_leap_year_b else [1] * 8760
+
+            motor_use_schedule_b = getattr_(
+                building_b, "building", "elevators", "cab_motor_multiplier_schedule"
```
JacksonJ-KC, line 57: elevators is a list of {Elevator} objects so there can be more than one. The rule should evaluate for each elevator

rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,85 @@
...
+                    BASELINE_0=True,
+                    PROPOSED=True,
+                ),
+            )
+
+        def is_applicable(self, context, data=None):
+            building_p = context.PROPOSED
+            elevators_p = building_p.get("elevators", [])
+
+            return len(elevators_p) > 0
```
JacksonJ-KC, line 47: is_applicable looks to be checking if the rule is applicable for each building in the RMD. The RDS was written so that the rule is only not applicable if none of the buildings in the RMD have an elevator (If there are 3 buildings in the RMD and only 1 building has an elevator then the rule is still applicable)

I don't think there is a strong preference on this so if there is a good reason to do it differently please let me know.

JacksonJ-KC (commented):
Hi Yun Joon, I made a few comments on 16-2 that might also apply to the other rules too

rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,85 @@
...
+    """Rule 2 of ASHRAE 90.1-2019 Appendix G Section 16 (Elevators)"""
+
+    def __init__(self):
+        super(Section16Rule2, self).__init__(
+            rmds_used=produce_ruleset_model_instance(
+                USER=False,
+                BASELINE_0=True,
+                PROPOSED=True,
+            ),
+            each_rule=Section16Rule2.BuildingRule(),
```
JacksonJ-KC, line 21: rule should evaluate for each elevator in building.elevators

yunjoonjung-PNNL:
@JacksonJ-KC All comments make sense! For some reason, I think I confused myself that the elevator key is just one string value instead of list of dicts. I'll address your comments when the testing JSONs are available. Thanks!

rct229/rulesets/ashrae9012019/section16/section16rule2.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section16/section16rule2.py
yunjoonjung-PNNL: Good point. Addressed.

rct229/rulesets/ashrae9012019/section16/section16rule2.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section16/section16rule2.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section16/section16rule3.py
```diff
@@ -0,0 +1,99 @@
...
+    def is_applicable(self, context, data=None):
+        rmd_p = context.PROPOSED
+
+        return (
+            len(
+                find_all(
+                    "$.ruleset_model_descriptions[0].buildings[*].elevators", rmd_p
+                )
+            )
+            > 0
```
weilixu, line 44: I wonder if you can just do `return find_all("$.ruleset_model_descriptions[0].buildings[*].elevators", rmd_p)`

rct229/rulesets/ashrae9012019/section16/section16rule3.py
```diff
@@ -0,0 +1,99 @@
...
+                cab_ventilation_schedule_p,
+                mask_schedule,
+                is_leap_year_b,
+            )
+
+            return {
+                "cab_ventilation_schedule_b": cab_ventilation_schedule_b,
+                "cab_ventilation_schedule_p": cab_ventilation_schedule_p,
+                "compare_schedules": compare_schedules,
+            }
```
weilixu, line 88: Think the issue here is the same as the `16-2`

rct229/rulesets/ashrae9012019/section16/section16rule4.py
```diff
@@ -0,0 +1,103 @@
...
+            ]
+            cab_lighting_multiplier_schedule_p = calc_vals[
+                "cab_lighting_multiplier_schedule_p"
+            ]
+
+            return (
+                compare_schedules["total_hours_matched"]
+                == len(cab_lighting_multiplier_schedule_b)
+                == len(cab_lighting_multiplier_schedule_p)
+            )
```
weilixu, line 103: Same comments.

rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,93 @@
...
+                    PROPOSED=True,
+                ),
+            )
+
+        def get_calc_vals(self, context, data=None):
+            building_b = context.BASELINE_0
+            building_p = context.PROPOSED
+
+            is_leap_year_b = data["is_leap_year_b"]
+            mask_schedule = [1] * 8784 if is_leap_year_b else [1] * 8760
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed

rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,93 @@
...
+                rmds_used=produce_ruleset_model_instance(
+                    USER=False,
+                    BASELINE_0=True,
+                    PROPOSED=True,
+                ),
+            )
+
+        def get_calc_vals(self, context, data=None):
+            building_b = context.BASELINE_0
+            building_p = context.PROPOSED
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed

rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,93 @@
...
+
+            is_leap_year_b = data["is_leap_year_b"]
+            mask_schedule = [1] * 8784 if is_leap_year_b else [1] * 8760
+
+            motor_use_schedule_b = getattr_(
+                building_b, "building", "elevators", "cab_motor_multiplier_schedule"
+            )
+            motor_use_schedule_p = getattr_(
+                building_p, "building", "elevators", "cab_motor_multiplier_schedule"
+            )
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed

rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,93 @@
...
+                motor_use_schedule_b,
+                motor_use_schedule_p,
+                mask_schedule,
+                is_leap_year_b,
+            )
+
+            return {
+                "motor_use_schedule_b": motor_use_schedule_b,
+                "motor_use_schedule_p": motor_use_schedule_p,
+                "compare_schedules": compare_schedules,
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed

rct229/rulesets/ashrae9012019/section16/section16rule3.py
yunjoonjung-PNNL: Addressed

rct229/rulesets/ashrae9012019/section16/section16rule4.py
yunjoonjung-PNNL: Addressed

rct229/rulesets/ashrae9012019/section16/section16rule3.py
yunjoonjung-PNNL: Addressed

rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,127 @@
...
+                        PROPOSED=True,
+                    ),
+                )
+
+            def get_calc_vals(self, context, data=None):
+                elevator_b = context.BASELINE_0
+                elevator_p = context.PROPOSED
+
+                motor_use_schedule_b = elevator_b["cab_motor_multiplier_schedule"]
+                motor_use_schedule_p = elevator_p["cab_motor_multiplier_schedule"]
```
weilixu, line 89: use `getattr_`

rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,127 @@
...
+    class RuleSetModelDescriptionRule(RuleDefinitionListIndexedBase):
+        def __init__(self):
+            super(Section16Rule2.RuleSetModelDescriptionRule, self).__init__(
+                rmds_used=produce_ruleset_model_instance(
+                    USER=False, BASELINE_0=True, PROPOSED=True
+                ),
+                each_rule=Section16Rule2.RuleSetModelDescriptionRule.ElevatorRule(),
+                index_rmd=BASELINE_0,
+                list_path="buildings[*].elevators[*]",
+            )
```
weilixu, line 44: @JacksonJ-KC , could you check what outcome should it be if the baseline has no elevator but proposed has?
The rule passes if no elevator in baseline model but proposed has.

rct229/rulesets/ashrae9012019/section16/section16rule2.py
```diff
@@ -0,0 +1,127 @@
...
+
+        def is_applicable(self, context, data=None):
+            rmd_p = context.PROPOSED
+
+            return find_all("$.buildings[*].elevators", rmd_p)
```
weilixu, line 49: This should be ` find_all("$.buildings[*].elevators[*]", rmd_p)`

yunjoonjung-PNNL: Good point! Addressed

rct229/rulesets/ashrae9012019/section16/section16rule2.py
yunjoonjung-PNNL: Addressed.

rct229/cli.py
```diff
@@ -29,7 +29,7 @@ def cli():
     Software test workflow, add sections to do test. \n
     --ruleset or -rs: default is ashrae9012019, available: ashrae9012019\n
     argument (optional): section string, \n
-    currently available: section1, section4, section5, section6, section10, section18, section19, section21, section22 and section23"""
+    currently available: section1, section4, section5, section6, section16, section18, section19, section21, section22 and section23"""
```
weilixu, line 32: Should this be adding section16 instead of replacing section10 with 16?

weilixu:
@yunjoonjung-PNNL these rules have a fundamental issue that has not been resolved. Please reach out to Jackson for answers.
The issue is, that if baseline has no elevators, the rule evaluation will be silenced and return in pass outcome.

rct229/cli.py
yunjoonjung-PNNL: Good catch! Addressed

rct229/rulesets/ashrae9012019/section16/section16rule2.py
JacksonJ-KC: This should not be a Pass. This would result in a fail for rule 1-6 already. My only questions is: do we want to provide another Fail outcome for this rule, or have it be not applicable?

weilixu: > This should not be a Pass. This would result in a fail for rule 1-6 already. My only questions is: do we want to provide another Fail outcome for this rule, or have it be not applicable?

It seems like an outcome at RPD level instead of elevator level. However, if 1-6 captures the mismatch, I am considering a not applicable outcome here as well. Since the issue is highlighting the missing elevators in baseline?

yunjoonjung-PNNL: For record purpose.

Email: May 31, 2024, Title: RCT, Question about section 16-2 rule

Weili:
"""
The key issue is not elevator key doesn’t exist but if the elevator is an empty list in the baseline model.
In that case, currently, the RCT will loop in an empty list and return nothing.

We will need to flag the issue if the baseline has empty list. There are two questions here:
if baseline has no elevator key or an empty elevator list but proposed has elevator(s), do we flag the rule as undermined or failed?
if proposed has no elevator but baseline has elevator(s), do we flag it not applicable, or failed, or undermined?
"""

Jackson:
"""Personally, I am leaning toward having the rule be not applicable for both scenarios and letting the Baseline == Proposed rule (1-6) handle the failed outcome that would result.
"""

Weili:
"""
I agree with that decision. I also left the same comment following your quote.
In that case, we should update all 16 RDSs to include this logic in the applicable evaluation. Jackson, could you update them all in one PR?

Yun Joon, Jiarong, please update section 16 rule applicable checks following the logic:

If proposed has no elevators OR (proposed has elevators AND (baseline has no elevators OR empty elevator list)), return NOT_APPLICABLE
"""

yunjoonjung-PNNL: @Jiarongx-Xie FYI

yunjoonjung-PNNL: Logic derivation:

P: elevator in proposed, B: elevator in baseline

Original  logic from Weili: (P is False) or (P is True and (B is None or B is False))

we should negate the original logic b/c we have the is_applicable (positive).

Applying De Morgan's laws
not (P is False) and not (P is True and (B is None or B is False))
= (P is True) and (P is False or (B is not None and B is True))
= P is True and (B is not None and B is True)

yunjoonjung-PNNL: @jugonzal07 Should we add a not_applicable case in the rpd?

weilixu: I do not think the De Morgan's law applied correctly here.
In the second step: `= (P is True) and (P is False or (B is not None and B is True))` it is impossible

Technically, the original logic should be:
`(P is False or (B is None or B is False))`

The logic after the first `or` has a pre-condition that P is True.

You can keep the original logic as it is for readability.

yunjoonjung-PNNL: Good point! For readability, I put ```not``` at the beginning of the logic.
When I tested the following 3 scenarios, the logic detects the not applicable cases correctly.
1) When the ```elevator_p``` doesn't exist
2) When the ```elevator_b``` doesn't exist
3) When the ```elevator_b``` is an empty list

Please let me know if any further changes need to be made.

weilixu: make sense to me.

weilixu (approved):

PR #1343, opened 2024-04-30
RCT/YJ/Fixed the wrong Rankin temp in report

yunjoonjung-PNNL:
This PR fixes the following things
- Instead of using the Rankin temp unit , the ```delta F``` is used in the detailed report.
- Fixed the missing schedules, wrong case-sensitive strings in the Chicago demo USER/PROPOSED/BASELINE RPDs.

Remaining part:
When I ran the ```poetry run rct229 evaluate -rs ashrae9012019 -f examples\chicago_demo\baseline_model.json -f examples\chicago_demo\proposed_model.json -f examples\chicago_demo\user_model.json -r ASHRAE9
012019DetailReport``` command, I had to exclude rules 1-6 and 1-7 because there is an error from the ```extra_schema_find.py``` file. I'll fix this in a different PR. Thanks!


rct229/reports/utils.py
```diff
@@ -71,7 +74,7 @@ def _count_results(outcomes):
             elif type(result) is dict and any(
                 [
                     key.startswith("INVALID_") and key.endswith("_CONTEXT")
-                    for key in result.keys()
```
weilixu, line 74: Why do we need to update this line?

weilixu (approved):
Seems a good approach. I can look into the 1-6 and 1-7 error

rct229/reports/utils.py
yunjoonjung-PNNL: This is because we don't need ```.keys()``` to iterate a dictionary. I believe we removed all the ```.keys()``` in the past, but for some reason this file still had ```.keys()```. So, I deleted it.

PR #1348, opened 2024-05-07
RDS/JDJ/Rule 12-1

JacksonJ-KC:


supriyagoel (commented):
@JacksonJ-KC since this check is a straight up comparison of P and B, could we make this part of the umbrella check?

supriyagoel:
also, should P not equal to B result in fail outcome, in case of this scenario- equipment, shall be estimated based on the building area type or space type category and shall be assumed to be identical in the proposed design and baseline building design, except as specifically approved by the rating authority only when quantifying performance that exceeds the requirements of Standard 90.1 but not when the Performance Rating. Method is used as an alternative path for minimum standard compliance in accordance with Section 4.2.1.1.

JacksonJ-KC:
"except as specifically approved by the rating authority only when quantifying performance that exceeds the requirements of Standard 90.1 but not when the Performance Rating. Method is used as an alternative path for minimum standard compliance in accordance with Section 4.2.1.1."

Supriya, how would you like to handle this part? I agree if it is a straightforward comparison of B == P then we could use the rule in Section 1 and change the associated Boolean in the extra schema to TRUE. If we want an undetermined case or another way of handling the above scenario then it should stay as its own rule.

claperle:
> > "except as specifically approved by the rating authority only when quantifying performance that exceeds the requirements of Standard 90.1 but not when the Performance Rating. Method is used as an alternative path for minimum standard compliance in accordance with Section 4.2.1.1."
> > Supriya, how would you like to handle this part? I agree if it is a straightforward comparison of B == P then we could use the rule in Section 1 and change the associated Boolean in the extra schema to TRUE. If we want an undetermined case or another way of handling the above scenario then it should stay as its own rule.
> 
> Maybe we can keep this as a separate RDS and in the case with P is less than B, we could provide an undetermined outcome and provide a outcome message which informs the reviewer than the proposed building equipment load is less than the baseline, which is only permitted when the model is being used to quantify performance that exceeds the requirements of Standard 90.1. I believe we could safely provide a Fail outcome if P > B.
> 
> Does that sound like a good approach?

I agree with Supriya's suggestion.

docs/section12/Rule12-1.md
```diff
@@ -1,36 +1,37 @@
...
+    - Append the miscellaneous equipment info for reporting: `reduced_misc_equipment_power.append({"id": misc_equipment_b.id, "baseline_power": misc_equipment_power_b, "proposed_power": misc_equipment_power_p})`
+  - If the miscellaneous equipment power in the baseline RMD is less than the miscellaneous equipment power in the proposed RMD: `elif misc_equipment_power_b < misc_equipment_power_p:`
+    - Append the miscellaneous equipment info for reporting: `unexpected_misc_equipment_power.append({"id": misc_equipment_b.id, "baseline_power": misc_equipment_power_b, "proposed_power": misc_equipment_power_p})`
+**Rule Assertion:**  
+  - Case 1: If all equipment power was modeled identically: PASS `if len(unexpected_misc_equipment_power) == 0 and len(reduced_misc_equipment_power) == 0: PASS`
+  - Case 2: Else if any equipment power is reduced in the proposed: UNDETERMINED: `elif len(unexpected_misc_equipment_power) == 0: UNDETERMINED and raise_message="The proposed building miscellaneous equipment load is less than the baseline, which is only permitted when the model is being used to quantify performance that exceeds the requirements of Standard 90.1."`
```
claperle, line 31: This could also incorporate RPD.compliance_path. Based on the labeling it appears that whether it is for minimum code compliance or beyond code performance would be indicated by the data element. If it says minimum code compliance that could have a fail outcome.

JacksonJ-KC: Good point!

docs/section12/Rule12-1.md
```diff
@@ -1,36 +1,37 @@
...
+  - Get the miscellaneous equipment power in the baseline RMD: `misc_equipment_power_b = misc_equipment_b.power`
+  - If the miscellaneous equipment power in the baseline RMD is greater than the miscellaneous equipment power in the proposed RMD: `if misc_equipment_power_b > misc_equipment_power_p:`
+    - Append the miscellaneous equipment info for reporting: `reduced_misc_equipment_power.append({"id": misc_equipment_b.id, "baseline_power": misc_equipment_power_b, "proposed_power": misc_equipment_power_p})`
+  - If the miscellaneous equipment power in the baseline RMD is less than the miscellaneous equipment power in the proposed RMD: `elif misc_equipment_power_b < misc_equipment_power_p:`
+    - Append the miscellaneous equipment info for reporting: `unexpected_misc_equipment_power.append({"id": misc_equipment_b.id, "baseline_power": misc_equipment_power_b, "proposed_power": misc_equipment_power_p})`
+**Rule Assertion:**  
+  - Case 1: If all equipment power was modeled identically: PASS `if len(unexpected_misc_equipment_power) == 0 and len(reduced_misc_equipment_power) == 0: PASS`
```
weilixu, line 30: My understanding is Rule 1-7 will check and highlight these outcome so this rule essentially, provides more detail messages about the failures?

JacksonJ-KC: MiscellaneousEquipment.power is set to FALSE for the AppG B_RMD Equals P_RMD column in the extra schema so 1-6 will not check this (1-7 is for proposed vs. user). False is appropriate for 1-6 and we check in this rule instead because when the project is for a beyond-code program the proposed is allowed to be less than the baseline, but we are unable to tell if the reduction was correctly applied for the beyond code program so outcome is UNDETERMINED. 

If proposed is greater than the baseline, FAIL. Also, as Christina pointed out, `if RPD.compliance_path == CODE_COMPLIANT`, any difference between proposed and baseline is a FAIL.

docs/section12/Rule12-1.md
```diff
@@ -1,36 +1,38 @@
...
-
-  - For each zone in thermal block: ```zone_baseline in thermal_block_baseline.zones:```
-
-  - For each space in zone, add to the total number of spaces in the Baseline model: ```for space_baseline in zone_baseline.spaces: num_of_spaces_baseline += 1```  
-
-**Rule Assertion:** ```num_of_spaces_user == num_of_spaces_baseline```  
-
-**[Back](../_toc.md)**
+**Evaluation Context:** Each RPD  
+
```
KarenWGard, line 11: I'm curious about the choice of evaluation context here.

I've had many projects where credit was allowed for Energy Star equipment in kitchens or laundries (or other allowed misc equipment reductions), but most of the equipment in the building was the same.  The logic here looks great, just want to make sure the end-user gets the information about which equipment is compliant vs not.  We apply rules about surfaces to each surface, so why not here?

JacksonJ-KC: The choice was made to minimize the number of evaluations and messages that are reported - I am anticipating that we may be at a point where we are providing so many messages that they no longer will be useful because there are so many. If the implementation is done in a way that shows the space ID's which cause an UNDETERMINED outcome would that address your concern?

KarenWGard: I definitely see the concern, and I think this has a lot to do with how the output report is displayed.  For example, a user should be able to filter by rule, evaluation result (PASS, FAIL, UNDETERMINED), or even by the raise_message string.

What you're suggesting works for this rule, but it means that this rule is different from the other rules, so it'll make creating a concise tool tool to show data more difficult.

The issue you've highlighted is much bigger than will be solved by changing the approach with this one rule.  I think it's better handled with some filtering capabilities in a result viewer tool

JacksonJ-KC: There are rules that exist that do it both ways. Here is an example of a rule that is evaluated once and lists the ID's that caused the outcome.

image

claperle (approved):

KarenWGard (approved):

PR #1355, opened 2024-05-09
RS/JDJ/10-7-CalcQ

JacksonJ-KC:
I forgot to use CalcQ when this was first implemented


yunjoonjung-PNNL (approved):
Good catch! Thanks Jackson.

weilixu (approved):

PR #1357, opened 2024-05-13
Rule Test 16-1

jugonzal07:
First cut at rule test 16-1


yunjoonjung-PNNL:
Juan,

We need ```proposed``` for all the cases because of the not applicable check. Can you add it to the RPD?

yunjoonjung-PNNL:
Can you add the ```number_of_floors_served``` key in case (a)?

weilixu:
Need revisit @jugonzal07 , this test json is not completed.

yunjoonjung-PNNL:
Juan,

For case (c), could you check if the ```motor_power``` in baseline should be 46508.553 W (=62.369 hp)? Other than this case, all cases passed.

yunjoonjung-PNNL:
> `rct229/rulesets/ashrae9012019/section16/__init__.py`

Hi Juan, Weili and I found that your testing json file is correct. So, please disregard my previous comment. My apologies!

PR #1358, opened 2024-05-13
Rule Test 16-2

jugonzal07:
First cut at Rule Test 16-2


yunjoonjung-PNNL:
Juan,

For case (b), the baseline and proposed have the same schedules which makes the case pass unexpectedly. Could you modify one of the schedules?

yunjoonjung-PNNL (approved):
Looks great to me. Thanks Juan!

PR #1359, opened 2024-05-13
Rule Test 16-3

jugonzal07:
First cut at Rule Test 16-3


yunjoonjung-PNNL:
Juan,

For case (b), the baseline and proposed have the same schedules. Could you modify one of them to make the case fail?

jugonzal07:
> Juan,
> 
> For case (b), the baseline and proposed have the same schedules. Could you modify one of them to make the case fail?

Thanks Yun Joon, this should be addressed now.

yunjoonjung-PNNL (approved):
Looks great. Thanks Juan!

PR #1360, opened 2024-05-13
Rule Test 16-4

jugonzal07:
First cut at Rule Test 16-4


yunjoonjung-PNNL:
Juan,

Case (b) passed unexpectedly. Could you modify the schedules to make the case fail?

jugonzal07:
> Juan,
> 
> Case (b) passed unexpectedly. Could you modify the schedules to make the case fail?

Thanks Yun Joon, i did goof this. They should be fixed now.

PR #1368, opened 2024-05-28
Rt/jdj/rpd test correct answers

JacksonJ-KC:
Creating a draft pull request so that files can begin to be reviewed.


weilixu:
@JacksonJ-KC once this is ready for review, could you please add `readyForReview` tag? Thanks!

JacksonJ-KC:
@weilixu All files that I have committed are ready for review. I am still working on 3 additional test cases: Appendix E Test Case 3 and Appendix F Test Case 230 & 240.

jugonzal07:
> @jugonzal07 I am noticing now that excel_to_test_json is behaving unexpectedly. I am seeing it is creating 5 HVAC systems (1 for each zone) even though the spreadsheet only has 1 baseline_system defined and all zones are defined as children to systems[0].
> 
> Can you help me look into this?

I can take a look into it. Feel free to email me the spreadsheet or setup that is giving you issues.

It depends on the system type. Systems that are single zone (e.g., zone equipment, PSZ systems) will duplicate themselves if they are assigned to multiple zones, even if they're all on the same `systems[x]` list. If it's one of those, then this is expected behavior-- at least with how the script was designed.

JacksonJ-KC:
Thanks Juan, that was enough info for me to search for and find the "determine_system_classification" function and debug to figure out what was happening. All set!

JacksonJ-KC:
@weilixu @jugonzal07 all test case RPDs are ready for review.. only 863,000 new lines to approve! 😂Most of it is schedules.

All RPDs are currently failing validation because the Infiltration Schedule is referenced but does not exist yet. It still needs to be defined in Table 12 of Standard 229.

jugonzal07 (approved):

PR #1381, opened 2024-06-09
Rds/kjw/rule 11 15

KarenWGard:


docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,48 @@
+# ServiceWaterHeating - Rule 11-15
+
+**Schema Version:** 0.0.36  
+**Mandatory Rule:** True  
+**Rule ID:** 11-15  
+
+**Rule Description:** "Service water loads and use shall be the same for both the proposed design and baseline building design and shall be documented by the calculation procedures described in Section 7.4.1.
+Exceptions: 
+(1) ECMs are used to reduce SWH use..
```
claperle, line 9: Suggest using language more in line with the standard

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,48 @@
...
+- Each SHW use is applicable
+- look at each space in the proposed building: `for space in B_RMD...spaces:`
+  - look at each SWH use in the space, continue to rule logic: `for swh_use_p in space.service_water_heating_uses: CONTINUE TO RULE LOGIC`
+
+    ## Rule Logic:
+    - get the baseline equivalent swh use: `swh_use_b = get_component_by_id(swh_use_p.id, B_RMD)`
+    - create a value to track the rule status: `rule_status = "PASS"`
+    - create a variable to track any notes: `rule_note = """`
+    - check that swh_use_b exists, if it does, continue with rule logic: `if swh_use_b != null:`
+      - check that the use matches by checking that the use, use_units, use_multiplier_schedule, and temperature_at_fixture all match (ARE THERE OTHERS THAT NEED TO MATCH TOO?)
```
claperle, line 31: the temperature of the distributed water shoudl also match

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,48 @@
...
+      - check use_units, if not equal, set to fail and add a note indicating that the use units are different: `if swh_use_p.use_units != swh_use_b.use_units: rule_status = "FAIL"; rule_note += "Service water heating use units are inconsistent between proposed and baseline models.  "`
```
claperle, line 32: i wonder if this should be undetermined. I think low flow fixture measures are very common and I think it would annoy a lot of projects to always have this fail outcome.

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,48 @@
...
+        - check whether the proposed use is less than the baseline, this will result in UNDETERMINED as an ECM could reduce swh use due to low-flow fixtures.  Change rule_status and append note: `if swh_use_p.use < swh_use_b.use: rule_status = "UNDETERMINED"; rule_note += "Proposed Service Water Heating Use is less than the baseline.  Manually verify that reduction is due to an ECM that reduces service water heating use, such as low-flow fixtures.  "`
+        - otherwise, if the proposed use is greater than the baseline, the result is fail.  Append a note: `if swh_use_p.use > swh_use_b.use: rule_status = "FAIL"; rule_note += "Proposed Service Water Heating Use is greater than the baseline.  "`
+
+     **Rule Assertion - Zone:**
+    - Case1: rule_status is PASS: `if rule_status == "PASS": PASS`.
+    - Case2: rule_status is UNDETERMINED, UNDETERMINED & return rule_note: `elsif rule_status == "UNDETERMINED": UNDETERMINED; rule_note`
+    - Case3: rule_status is FAIL, FAIL & return rule_note: `elsif rule_status == "FAIL": FAIL; rule_note`
+
+
```
claperle, line 43: Note to self, this RDS makes sense to me

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,48 @@
+# ServiceWaterHeating - Rule 11-15
+
+**Schema Version:** 0.0.36
```
JacksonJ-KC, line 3: most recent version is 0.0.37

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,48 @@
...
+**Appendix G Section Reference:** Table G3.1 #11, baseline column, (g)  
+
+**Evaluation Context:** Each SWH use  
+**Data Lookup:**   
+**Function Call:**
+
+1. get_component_by_id()
+
+**Applicability Checks:**
+- Each SHW use is applicable
```
JacksonJ-KC, line 22: Seems like these `for` loops should be a part of the rule logic. The applicability check should be that the building has at least 1 SHW use (where ServiceWaterHeatingUse.use > 0?)

docs/section11/Rule11-15.md
JacksonJ-KC: Christina, are you thinking we need to trace up to the parent `ServiceWaterHeatingDistributionSystem.design_supply_temperature`?

claperle: Yeah

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,67 @@
...
+- **ge_component_by_id**  
+
+**Applicability Checks:**
+- create a list of swh use ids that need to be checked - we'll create the list from both the proposed and baseline models: `swh_use_ids = []`
+- look at each building in the B_RMD: `for building in B_RMD:`
+    - create a boolean is_applicable and set it to false: `is_applicable = false`
+    - look at each building segment in the building: `for building_segment in building.building_segments:`
+    - get the service water heating uses in the building segment `service_water_heating_use_ids = get_SWH_uses_associated_with_each_building_segment(P_RMD, building_segment.id)`
+    - look at each service water heating use id: `for swh_use_id in service_water_heating_use_ids:`
+        - append the swh_use_id to the list of swh use ids: `swh_use_ids.append(swh_use_id)`
```
claperle, line 28: Not following why this step is needed, looks like you can just use service_water_heating_use_ids

claperle: Maybe this is missing an if statement?

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,67 @@
...
+- look at each building in the P_RMD: `for building in P_RMD:`
+    - create a boolean is_applicable and set it to false: `is_applicable = false`
+    - look at each building segment in the building: `for building_segment in building.building_segments:`
+    - get the service water heating uses in the building segment `service_water_heating_use_ids = get_SWH_uses_associated_with_each_building_segment(P_RMD, building_segment.id)`
+    - look at each service water heating use id: `for swh_use_id in service_water_heating_use_ids:`
+        - append the swh_use_id to the list of swh use ids if it's not already in the list: `if !swh_use_id.in? sw_use_ids: swh_use_ids.append(swh_use_id)`
```
claperle, line 34: Ruby syntax

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,67 @@
...
+
+- look at each swh_use_id in the swh_use_ids list.  Each of these is applicable (not sure how this works, when an ID is applicable instead of an object): `for swh_use_id in swh_use_ids: CONTINUE TO RULE LOGIC`
```
claperle, line 36: Suggest writing that if len(swh_use_id) == 0 then the rule is not_applicable

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,67 @@
...
+
+    ## Rule Logic:
+    - get the proposed swh use: `swh_use_p = get_component_by_id(swh_use_id, P_RMD)`
```
claperle, line 39: I am a little confused about the context here. I don't see the for statement to loop through the IDs in the lists of IDs that you created in the applicability check section.

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,67 @@
...
+    - get the proposed swh use: `swh_use_b = get_component_by_id(swh_use_id, B_RMD)`
+    - create a value to track the rule status: `rule_status = "PASS"`
+    - create a variable to track any notes: `rule_note = ""`
+    - if swh_use_b is null: `if swh_use_b == nil:`
+        - set the rule status to FAIL and provide note: `rule_status = "FAIL"; note = swh_use_id + "exists in the Proposed model, but not in the baseline.`
+        - continue to rule assertion: `CONTINUE TO RULE ASSERTION`
+    - if swh_use_P is null: `if swh_use_P == nil:`
+        - set the rule status to FAIL and provide note: `rule_status = "FAIL"; note = swh_use_id + "exists in the Baseline model, but not in the proposed.`
+        - continue to rule assertion: `CONTINUE TO RULE ASSERTION`
+    - check that the use matches by checking that the use, use_units, use_multiplier_schedule, and temperature_at_fixture all match (ARE THERE OTHERS THAT NEED TO MATCH TOO?)
```
claperle, line 49: Note to self, stopped here

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,67 @@
...
+    - now, if the rule_status is still "PASS", we'll check if the use is equal or not between proposed and baseline: `if rule_status == "PASS":`
+      - check whether the proposed use is less than the baseline, this will result in UNDETERMINED as an ECM could reduce swh use due to low-flow fixtures.  Change rule_status and append note: `if swh_use_p.use < swh_use_b.use: rule_status = "UNDETERMINED"; rule_note = "Proposed Service Water Heating Use is less than the baseline.  Manually verify that reduction is due to an ECM that reduces service water heating use, such as low-flow fixtures.  "`
+      - otherwise, if the proposed use is greater than the baseline, the result is fail.  Provide a note: `if swh_use_p.use > swh_use_b.use: rule_status = "FAIL"; rule_note += "Proposed Service Water Heating Use is greater than the baseline.  "`
+
+     **Rule Assertion - Zone:**
+    - Case1: rule_status is PASS: `if rule_status == "PASS": PASS`.
```
claperle, line 58: I think this rule also needs to address cpmpliance with 7.4.1. Perhaps it shoudl be two separate rules. One that checks baselein and proposed and nother that jsut checks if there are loads and then directs the reviewer toensure compliance with 7.4.1 calculations procedures
.

KarenWGard: 11-9 already states that the baseline needs to be sized according to 7.4.1.  So perhaps we can remove the language about 7.4.1 from this rule?

claperle: @KarenWGard I agree with that approach.

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,68 @@
...
+**Evaluation Context:** Each SWH use  
+**Data Lookup:**   
+**Function Call:**
+- **get_SWH_uses_associated_with_each_building_segment**  
+- **get_component_by_id**  
+
+**Applicability Checks:**
+- create a list of swh use ids that need to be checked - we'll create the list from both the proposed and baseline models: `swh_use_ids = []`
+- look at each building in the B_RMD: `for building in B_RMD:`
+    - create a boolean is_applicable and set it to false: `is_applicable = false`
```
claperle, line 24: not seeing where this is used in the rest of the RDS

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,68 @@
...
+    - look at each building segment in the building: `for building_segment in building.building_segments:`
+    - get the service water heating uses in the building segment `service_water_heating_use_ids = get_SWH_uses_associated_with_each_building_segment(P_RMD, building_segment.id)`
+    - look at each service water heating use id: `for swh_use_id in service_water_heating_use_ids:`
+        - append the swh_use_id to the list of swh use ids: `swh_use_ids.append(swh_use_id)`
+- look at each building in the P_RMD: `for building in P_RMD:`
+    - create a boolean is_applicable and set it to false: `is_applicable = false`
```
claperle, line 30: not seeing where this is used in the rest of the RDS

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,68 @@
...
+    - get the proposed swh use: `swh_use_b = get_component_by_id(swh_use_id, B_RMD)`
+    - create a value to track the rule status: `rule_status = "PASS"`
+    - create a variable to track any notes: `rule_note = ""`
+    - if swh_use_b is null: `if swh_use_b == nil:`
+        - set the rule status to FAIL and provide note: `rule_status = "FAIL"; note = swh_use_id + "exists in the Proposed model, but not in the baseline.`
+        - continue to rule assertion: `CONTINUE TO RULE ASSERTION`
+    - if swh_use_P is null: `if swh_use_P == nil:`
+        - set the rule status to FAIL and provide note: `rule_status = "FAIL"; note = swh_use_id + "exists in the Baseline model, but not in the proposed.`
+        - continue to rule assertion: `CONTINUE TO RULE ASSERTION`
+    - check that the use matches by checking that the use, use_units, use_multiplier_schedule, and temperature_at_fixture all match (ARE THERE OTHERS THAT NEED TO MATCH TOO?)
```
claperle, line 50: if we want to check whether they are identical I think we would also need tocheck that the 
 ServiceWaterHeatingDistributionSystem.entering_water_mains_temperature_schedule and the ServiceWaterHeatingDistributionSystem..design_supply_water_temperature, I recall making this comment previously so maybe we discussed it and decide not to?

docs/section11/Rule11-15.md
KarenWGard: removed

docs/section11/Rule11-15.md
KarenWGard: removed

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,70 @@
...
+    - look at each service water heating use id: `for swh_use_id in service_water_heating_use_ids:`
+        - append the swh_use_id to the list of swh use ids: `swh_use_ids.append(swh_use_id)`
+- look at each building in the P_RMD: `for building in P_RMD:`
+    - look at each building segment in the building: `for building_segment in building.building_segments:`
+    - get the service water heating uses in the building segment `service_water_heating_use_ids = get_SWH_uses_associated_with_each_building_segment(P_RMD, building_segment.id)`
+    - look at each service water heating use id: `for swh_use_id in service_water_heating_use_ids:`
+        - append the swh_use_id to the list of swh use ids if it's not already in the list: `if !swh_use_id.in sw_use_ids: swh_use_ids.append(swh_use_id)`
+
+- look at each swh_use_id in the swh_use_ids list.  Each of these is applicable (not sure how this works, when an ID is applicable instead of an object): `for swh_use_id in swh_use_ids: CONTINUE TO RULE LOGIC`
+- if there are no swh_use_ids in the list, the rule is not applicable: `if len(swh_use_ids) == 0: NOT_APPLICABLE`
```
JacksonJ-KC, line 35: this belongs above the `for` loop

JacksonJ-KC: am I right that the length of swh_use_ids will be greater than 0 if the actual energy consumption of every SWH use is 0?

docs/section11/Rule11-15.md
```diff
@@ -0,0 +1,70 @@
...
+
+    ## Rule Logic:
+    - get the proposed swh use: `swh_use_p = get_component_by_id(swh_use_id, P_RMD)`
+    - get the baseline swh use: `swh_use_b = get_component_by_id(swh_use_id, B_RMD)`
+    - get the proposed distribution system that serves the swh use: `swh_dist_sys_p = swh_use_p.served_by_distribution_system`
+    - get the baseline distribution system that serves the swh use: `swh_dist_sys_b = swh_use_b.served_by_distribution_system`
+    - create a value to track the rule status: `rule_status = "PASS"`
+    - create a variable to track any notes: `rule_note = ""`
+    - if swh_use_b is null: `if swh_use_b == nil:`
+        - set the rule status to FAIL and provide note: `rule_status = "FAIL"; note = swh_use_id + "exists in the Proposed model, but not in the baseline.`
```
JacksonJ-KC, line 45: seems like if the notes are collecting a list of potentially multiple failures then we should include the use id in each note.

docs/section11/Rule11-15.md
JacksonJ-KC: could set rule_status to "NOT_APPLICABLE" and CONTINUE TO RULE ASSERTION to stay consistent with the rest of the rule

KarenWGard: > am I right that the length of swh_use_ids will be greater than 0 if the actual energy consumption of every SWH use is 0?

I don't completely follow - for the applicability, we're only looking for the existence of swh uses

docs/section11/Rule11-15.md
KarenWGard: the rule is currently evaluate for each swh_use_id.  Not sure how it works for Weili's team, but the reason I flagged it here and not for the others is that in this case, there is a mismatch between the two RMDs

JacksonJ-KC: I see! Thanks for clarifying

docs/section11/Rule11-15.md
KarenWGard: This rule just checks that the uses are the SWH uses are the same between proposed and baseline.  If a tool by default creates swh uses, I'd imagine that they are the same between models.

However, SWH uses are going to be required to be connected to distribution systems, so I doubt a tool would create all of that as a default approach.

JacksonJ-KC (approved):

PR #1383, opened 2024-06-11
Rule Test JSON 16-5

jugonzal07:
Added rule test for 16-5


yunjoonjung-PNNL:
Hi Juan,

This is great! A few comments are
1) the case (c) falls into ```undetermined``` because ```baseline``` doesn't have the ```cab_lighting_multiplier_schedule``` key. 
2) Could you add test cases for not applicable?

yunjoonjung-PNNL:
Juan,
Thanks for the update. However, for the not applicable case (c), I believe we should check the following 3 scenarios  ,rather than checking whether the ```cab_ventilation_fan_multiplier_schedule``` key exists.
i.	When the ```elevator_p``` doesn't exist
ii.	When the ```elevator_b``` doesn't exist
iii.	When the ```elevator_b``` is an empty list

jugonzal07:
Thanks Yun Joon, sorry I goofed this one. I belive I cannot provide suggestion `iii` as it violates the schema.

jugonzal07:
Updated.

yunjoonjung-PNNL:
Thank you Juan. Looks great!

yunjoonjung-PNNL (approved):

PR #1393, opened 2024-06-25
Rule test 1-3 calculation update

jugonzal07:
Fix for section 1-3 calculations.


weilixu:
The updates in this PR conflicts with the updates in #1391
#1391 json test are updated by @JacksonJ-KC. @jugonzal07 & @JacksonJ-KC , please coordinate to make sure the spreadsheet has no conflict.

JacksonJ-KC:
@jugonzal07 and @weilixu I have been making updates directly to the version of the spreadsheet on OneDrive. The updates that I made should make the tests easier to understand as I used more realistic numbers for the PBP, BBP, BBREC, BBUEC: which are in units of annual energy cost (so they should not be in the range of 0-3 but rather in the tens/hundreds of thousands).

weilixu:
issue resolved in another PR. Close this now.

PR #1394, opened 2024-06-27
RDS/JDJ/Rule 12-4

JacksonJ-KC:


docs/section12/Rule12-4.md
```diff
@@ -0,0 +1,49 @@
...
+**Function Call:**
+1) find_all
+2) is_space_a_computer_room
+3) compare_schedules
+
+## Applicability Checks:  
+- Iterate through the spaces in the baseline RMD: `for space_b in find_all("$.buildings[*].building_segments[*].zones[*].spaces[*]", B_RMD):`
+  - Check if the space is a computer room: `if is_space_a_computer_room(space_b):`
+    - Iterate through the miscelleneous equipment loads: `for misc_equip_b in space_b.miscellaneous_equipment:`
+      - Check if the equipment is IT equipment: `if misc_equip_b.type == "INFORMATION_TECHNOLOGY_EQUIPMENT" and misc_equip_b.power > 0:`
```
supriyagoel, line 25: @JacksonJ-KC do we need to check the misc equipment type? is the space type check and equip power >0 sufficient?
Just concerned about a scenario where the mis equip type might not be set- is that likely?

docs/section12/Rule12-4.md
```diff
@@ -0,0 +1,49 @@
...
+        - Rule is applicable: `return True`
+
+## Rule Logic:
+- Determine if it is a leap year: `is_leap_year = RPD.calendar.is_leap_year`
+- Create a dictionary to map the month numbers to their respective fractions: `month_fractions = {1: 0.25, 2: 0.5, 3: 0.75, 4: 1, 5: 0.25, 6: 0.5, 7: 0.75, 8: 1, 9: 0.25, 10: 0.5, 11: 0.75, 12: 1}`
+- Create a dictionary to map the number of days in each month: `days_in_month = {1: 31, 2: 29 if is_leap_year else 28, 3: 31, 4: 30, 5: 31, 6: 30, 7: 31, 8: 31, 9: 30, 10: 31, 11: 30, 12: 31}`
+- Create an empty list to store the expected hourly values of the schedule: `expected_hourly_values = []`
+- Iterate through the months of the year: `for month in range(1, 13):`
```
supriyagoel, line 33: should this be 1,12?

docs/section12/Rule12-4.md
```diff
@@ -0,0 +1,49 @@
...
+  - Get the fraction for the month: `fraction = month_fractions[month]`
+  - Get the number of days in the month: `days = days_in_month[month]`
+  - Append the hourly fraction to the list: `expected_hourly_values.extend([fraction] * days * 24)`
+- Create the mask schedule to be used for comparison: `mask_schedule = [1] * 8784 if is_leap_year else [1] * 8760`
+- Get the multiplier schedule that was modeled for the equipment: `multiplier_schedule_b = misc_equip_b.multiplier_schedule.hourly_values`
+- Compare the modeled schedule with the expected schedule: `comparison_data = compare_schedules(multiplier_schedule_b, expected_hourly_values, mask_schedule, is_leap_year)`
```
supriyagoel, line 39: I am not sure I followed steps 36-39. Maybe we can wait for Christina and Karen's review on this

docs/section12/Rule12-4.md
JacksonJ-KC: My thought in doing in this way was that even a computer room might have regular plug loads that are not "computer room equipment". I am glad you brought this up though because I was thinking this would be good to get additional input on

docs/section12/Rule12-4.md
JacksonJ-KC: upper bound is not inclusive in python's range()

docs/section12/Rule12-4.md
JacksonJ-KC: Sure, in plain english these lines are creating the expected hourly values of the annual schedule and comparing every hour to make sure the modeled schedule matches for all hours.

docs/section12/Rule12-4.md
claperle: Based on the definition in 90.1 I don't think we need to check the equipment type. 

computer room: a room whose primary function is to house equipment for the processing
and storage of electronic data and that has a design electronic data equipment power density
exceeding 20 W/ft2 of conditioned floor area.

G3.1.3.16. _Computer room equipment_ schedules shall be modeled as a constant fraction of the
peak design load per the following monthly schedule:

Because it does not differentiate the different types of equipment in a computer room I think it is safe to apply the schedule fraction requirement to all the equipment in the room.  Note that computer room and equipment are two separate defined terms. There is no computer room equipment definition.

docs/section12/Rule12-4.md
claperle: I don't see the is_leap_year term in the compare_schedules function

docs/section12/Rule12-4.md
```diff
@@ -0,0 +1,49 @@
...
+**Evaluation Context:** Each Miscellaneous Equipment object in the baseline model    
+**Data Lookup:** None  
+**Function Call:**
+1) find_all
+2) is_space_a_computer_room
+3) compare_schedules
+
+## Applicability Checks:  
+- Iterate through the spaces in the baseline RMD: `for space_b in find_all("$.buildings[*].building_segments[*].zones[*].spaces[*]", B_RMD):`
+  - Check if the space is a computer room: `if is_space_a_computer_room(space_b):`
```
weilixu, line 23: This function requires an `RMD` as one of the argument.

docs/section12/Rule12-4.md
weilixu: The space type check includes power check - so you will be fine just using the `is_space_a_computer_room` for this applicability check.

docs/section12/Rule12-4.md
weilixu: @JacksonJ-KC I think you can do it in a more generic way in the RDS to avoid confusion.
Put something such as:
`Generate an annual schedule based on the prescribed value for each hour in the standard: "`
The detail implementation can leave to the RCT team.

docs/section12/Rule12-4.md
```diff
@@ -0,0 +1,49 @@
...
+- Determine if it is a leap year: `is_leap_year = RPD.calendar.is_leap_year`
+- Create a dictionary to map the month numbers to their respective fractions: `month_fractions = {1: 0.25, 2: 0.5, 3: 0.75, 4: 1, 5: 0.25, 6: 0.5, 7: 0.75, 8: 1, 9: 0.25, 10: 0.5, 11: 0.75, 12: 1}`
+- Create a dictionary to map the number of days in each month: `days_in_month = {1: 31, 2: 29 if is_leap_year else 28, 3: 31, 4: 30, 5: 31, 6: 30, 7: 31, 8: 31, 9: 30, 10: 31, 11: 30, 12: 31}`
+- Create an empty list to store the expected hourly values of the schedule: `expected_hourly_values = []`
+- Iterate through the months of the year: `for month in range(1, 13):`
+  - Get the fraction for the month: `fraction = month_fractions[month]`
+  - Get the number of days in the month: `days = days_in_month[month]`
+  - Append the hourly fraction to the list: `expected_hourly_values.extend([fraction] * days * 24)`
+- Create the mask schedule to be used for comparison: `mask_schedule = [1] * 8784 if is_leap_year else [1] * 8760`
+- Get the multiplier schedule that was modeled for the equipment: `multiplier_schedule_b = misc_equip_b.multiplier_schedule.hourly_values`
```
weilixu, line 38: Is this checking per equipment or per space?
The answer determines whether to enforce all equipment in the computer room to running on the same schedule (per equipment) or the sum of the equipment running on the prescribed schedule (per space).

docs/section12/Rule12-4.md
JacksonJ-KC: This seems like an area where semantics could change our approach. 

Is "Computer room equipment schedules" referring to schedules for computer room equipment, or is it referring to equipment schedules for computer rooms?

In my eyes, it seems reasonable to have computer room equipment, or INFORMATION_TECHNOLOGY_EQUIPMENT, as a separate MiscellaneousEquipment object from the typical plug loads for the space and have them on independent schedules.

JacksonJ-KC: @weilixu in `is_space_a_computer_room` it seems like if the lighting space type is COMPUTER_ROOM then the power check is bypassed

claperle: Jackson, see my comment. I think if you look at 90.1 language it means the equipment (all of it) in a computer room. You disagree with my logic?

docs/section12/Rule12-4.md
JacksonJ-KC: @weilixu I often struggle with this because the convention of having pseudo code associated with it often makes me take a more complicated approach of describing the RDS. I understand the RCT dev team may use a different approach than what is described here.

JacksonJ-KC: > I don't see the is_leap_year term in the compare_schedules function



> I don't see the is_leap_year term in the compare_schedules function

https://github.com/pnnl/ruleset-checking-tool/blob/develop/rct229/rulesets/ashrae9012019/ruleset_functions/compare_schedules.py#L17

claperle: I guess the RDS needs updating then.

docs/section12/Rule12-4.md
JacksonJ-KC: I see your argument Christina but I am not sure it is a slam dunk. Hoping to get input from more people on this and go with the majority consensus

docs/section12/Rule12-4.md
JacksonJ-KC: I think it should be evaluated for each MiscellaneousEquipment

docs/section12/Rule12-4.md
KarenWGard: > @weilixu in `is_space_a_computer_room` it seems like if the lighting space type is COMPUTER_ROOM then the power check is bypassed

This function might need to be revisited - it basically checks for one or the other - lighting or power, but thinking about it now (I wrote the function), it should probably check for both in order to qualify.  Also, the 20W is supposed to apply to data equipment per ASHRAE:

```
**computer room**: a room whose primary function is to house equipment for the processing
and storage of electronic data and that has a design electronic data equipment power density
exceeding 20 W/ft2 of conditioned floor area.
```

KarenWGard (commented):
logic looks good.  Approval pending resolution of space vs equipment debate

claperle (approved):

docs/section12/Rule12-4.md
JacksonJ-KC: On second thought maybe the better solution is to remove is_leap_year from the arguments in this rule RDS to keep RDS's consistent with each other.

We will find many places where the RDS differs from the python implementation.

KarenWGard (approved):

PR #1406, opened 2024-07-10
Create get_energy_required_to_heat_swh_use.md

KarenWGard:


docs/ruleset_functions/get_energy_required_to_heat_swh_use.md
```diff
@@ -0,0 +1,71 @@
...
+- get_spaces_served_by_SWH_use
+- get_obj_by_id
+
+Data Lookup: None
+
+Logic:
+
+- get the schedule: `hourly_schedule = swh_use.use_multiplier_schedule`
+- get the temperature at fixture: `supply_temperature = swh_use.distribution_system.design_supply_temperature`
+- get the distribution system: `distribution_system = swh_use.served_by_distribution_system`
```
claperle, line 23: Suggest moving this line up one since you need to use the distribution_system object to get the design_supply_temperature. Right now line 22 does not appear to be referencing the correct data element which woould be .served_by_distribution_system instead of distrubrion_system.

docs/ruleset_functions/get_energy_required_to_heat_swh_use.md
```diff
@@ -0,0 +1,71 @@
...
+  - in this case, swh_use_value is assumed to have units of btu/hr/person, so calculate energy_required by multiplying power (btu/hr) by the sum of the hourly_schedule: `energy_required = power * sum(hourly_schedule)`
+  - return energy_required: `return energy_required`
+- otherwise if use_units is POWER_PER_AREA: `elsif use_units == "POWER_PER_AREA":`
+  - set power equal to the swh_use_value * the sum of the floor area in all of the spaces: `power = swh_use_value * sum(space.floor_area for space in spaces)`
+  - in this case, swh_use_value is assumed to have units of btu/hr/ft2, so calculate energy_required by multiplying power (btu/hr) by the sum of the hourly_schedule: `energy_required = power * sum(hourly_schedule)`
+  - return energy_required: `return energy_required`
+- otherwise if use_units is POWER: `elsif use_units == "POWER":`
+  - in this case, swh_use_value is assumed to have units of btu/hr, so calculate energy_required by multiplying swh_use_value by the sum of the hourly_schedule: `energy_required = swh_use_value * sum(hourly_schedule)`
```
claperle, line 43: Do we agree that it is power across all the spaces as opposed to power per space?

docs/ruleset_functions/get_energy_required_to_heat_swh_use.md
```diff
@@ -0,0 +1,71 @@
...
+  - in this case swh_use_value is assumed to have the units of gallons/hr/ft2:
+  - set volume_flow_rate equal to swh_use_value * the sum of the floor area in all of the spaces: `volume_flow_rate = swh_use_value * sum(space.floor_area for space in spaces)`
+- otherwise if use_units is VOLUME: `elsif use_units == "VOLUME":`
+  - in this case swh_use_value is assumed to have the units gallons/hr
+  - set volume_flow_rate equal to swh_use_value: `volume_flow_rate = swh_use_value`
+
+- set energy_required to 0: `energy_required = 0`
+- iterate through each hourly value of the hourly_schedule: `for index, hourly_value in enumerate(hourly_schedule):`
+  - set volume_this_hour equal to volume_flow_rate * hourly_value: `volume_this_hour = volume_flow_rate * hourly_value`
+  - reduce volume_this hour by the recovery efficiency: `volume_this_hour = volume_this_hour * (1-drain_heat_recovery_efficiency)`
```
claperle, line 58: Should this also be applied above to the other units as well?

docs/ruleset_functions/get_energy_required_to_heat_swh_use.md
KarenWGard: I think we should discuss this on the call - this would apply to both POWER and VOLUME.  The general rule that we agreed on a few weeks ago was:

**The rule is that if spaces reference the swh_use, then the use applies to only those spaces.  If the use references no spaces, it applies to all spaces in the building segment.**

However, for VOLUME and POWER, it doesn't make as much sense to me for it to get repeated for each space in a building segment, or for each applied space.

claperle: Sounds good to discuss on the call

KarenWGard: discussed on call 7/18 - if it's power or volume, it it's applied to no spaces, only count it once.  If it's explicitly applied to any spaces, then count it once for each space.

docs/ruleset_functions/get_energy_required_to_heat_swh_use.md
KarenWGard: this is applied regardless of the use units.  volume_flow_rate is calculated for each use_unit type and then this calculation is done.

docs/ruleset_functions/get_energy_required_to_heat_swh_use.md
```diff
@@ -0,0 +1,89 @@
...
+- **swh_use**
+- **RMD**
+
+Returns:
+- **energy_required_by_space**: A dict where they keys are space_ids and values are the total energy required to heat the swh_use for that space.  If a swh_use is not assigned to any spaces, the key will be "NO_SPACES_ASSIGNED"
+
+Function Call:
+
+- get_spaces_served_by_SWH_use
+- get_obj_by_id
```
JacksonJ-KC, line 15: I think get_obj_by_id is what I am used to seeing in RDSs but I see a lot of get_component_by_ID used in the other section 11 functions. Recommend consistency, or please correct me if they are different.

docs/ruleset_functions/get_energy_required_to_heat_swh_use.md
```diff
@@ -0,0 +1,89 @@
...
+- get the schedule: `hourly_schedule = swh_use.use_multiplier_schedule`
+- get the distribution system: `distribution_system = swh_use.served_by_distribution_system`
+- get the temperature at fixture: `supply_temperature = distribution_system.design_supply_temperature`
+- get the water used: `swh_use_value = swh_use.use`
+- get the inlet water temperature: `inlet_T_sched = distribution_system.entering_water_mains_temperature_schedule`
+- set the drain heat recovery efficiency to 0: `drain_heat_recovery_efficiency = 0`
+- get the drain heat recovery efficiency if heat is recovered by drain: `if(swh_use.is_heat_recovered_by_drain): drain_heat_recovery_efficiency = distribution_system.drain_heat_recovery_efficiency`
+- get the use_units: `use_units = swh_use.use_units`
+
+- now we need to get all of the spaces that the swh_use is applied to.  The rule is that if spaces reference the swh_use, then the use applies to only those spaces.  If the use references no spaces, it applies to all spaces in the building segment.
```
JacksonJ-KC, line 30: The convention is that if any spaces reference the swh_use, then the service water heating use applies to only those spaces. If no spaces reference the service water heating use, it applies to all spaces in the building segment.

JacksonJ-KC (approved):
Thanks for the updates Karen, and for putting up with my comments!

claperle (approved):

PR #1407, opened 2024-07-10
Create get_spaces_served_by_SWH_use.md

KarenWGard:


docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,41 @@
...
+
+Function Call:
+
+
+Data Lookup: None
+
+Logic:
+
+- find the building_segment that contains this swh_use by looking through each building segment: `for building_segment in RMD.building_segments:`
+    - check if the swh_use is in this building segment: `if swh_use in? building_segment.service_water_heating_uses:`
```
claperle, line 21: Looks like there may be a typo on this line

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,41 @@
...
+        - create a list of spaces: `spaces = []`
+        - look eat each building segment: `for building_segment in RMD.building_segments`
+            - check if the swh_use is in this building_segment: `if swh_use in? building_segment.service_water_heating_uses:`
+                - look at each space in the building segment to see which of them reference the swh_use: `for space in building_segment.spaces:`
+                    - if the swh_use ID is in the list of space ServiceWaterHeatingUses, then the space is one of the spaces and we add it to the list: `if swh_use.id in? space.service_water_heating_uses: spaces << space`
```
claperle, line 26: Looks like there may be a typo on this line (i.e., not understanding why the ? is included after in)

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,41 @@
...
+        - here, we check whether there are any spaces applied to the swh_use.  If no spaces are applied, then this use applies to all spaces in the building_segment: `if len(spaces) == 0: spaces = building_segment.spaces`
+        - convert the list of spaces to a list of space ids: `space_ids = [space.id for space in spaces]
+        - return space_ids: `return space_ids`
+
+
+
+- return result: `return: building_segment_swh_bat`
```
claperle, line 33: Ths does not appear correct. return: space_ids

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+Data Lookup: None
+
+Logic:
+
+- find the building_segment that contains this swh_use by looking through each building segment: `for building_segment in RMD.building_segments:`
+    - check if the swh_use is in this building segment: `if swh_use in building_segment.service_water_heating_uses:`
+        - create a list of spaces: `spaces = []`
+        - look at each building segment: `for building_segment in RMD.building_segments`
```
JacksonJ-KC, line 23: this loop through building segments is nested inside a loop through building segments

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,39 @@
...
+            - check if the swh_use is in this building_segment: `if swh_use in building_segment.service_water_heating_uses:`
```
JacksonJ-KC, line 24: we just did this?

JacksonJ-KC, line 24: remove this line

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,39 @@
...
+                - look at each space in the building segment to see which of them reference the swh_use: `for space in building_segment.spaces:`
+                    - if the swh_use ID is in the list of space ServiceWaterHeatingUses, then the space is one of the spaces and we add it to the list: `if swh_use.id in space.service_water_heating_uses: spaces << space`
```
JacksonJ-KC, line 26: `<<` is not python syntax

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,39 @@
...
+Function Call:
+
+
+Data Lookup: None
+
+Logic:
+
+- find the building_segment that contains this swh_use by looking through each building segment: `for building_segment in RMD.building_segments:`
+    - check if the swh_use is in this building segment: `if swh_use in building_segment.service_water_heating_uses:`
+        - create a list of spaces: `spaces = []`
```
JacksonJ-KC, line 22: suggest more specific variable name like `spaces_served` to not confuse with `building_segment.zones.spaces`

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,39 @@
...
+        - look at each building segment: `for building_segment in RMD.building_segments`
+            - check if the swh_use is in this building_segment: `if swh_use in building_segment.service_water_heating_uses:`
+                - look at each space in the building segment to see which of them reference the swh_use: `for space in building_segment.spaces:`
```
JacksonJ-KC, line 25: `for space in building_segment...spaces:` add extra dots for clarity to indicate that spaces are not a direct child of building segment

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,39 @@
...
+                    - if the swh_use ID is in the list of space ServiceWaterHeatingUses, then the space is one of the spaces and we add it to the list: `if swh_use.id in space.service_water_heating_uses: spaces << space`
+        - here, we check whether there are any spaces applied to the swh_use.  If no spaces are applied, and the service water heating use units are not POWER or VOLUME, then this use applies to all spaces in the building_segment: `if len(spaces) == 0:`
```
JacksonJ-KC, line 27: check whether any spaces reference the service water heating use. If the service water heating use is not referenced by any spaces....

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,39 @@
...
+            - check if the swh use use_units are not POWER or VOLUME.  Service water heating uses with use units power or volume are only counted once, unless they are expicitly assigned to multiple spaces: `if swh_use.use_units not in ["POWER","VOLUME"]: spaces = building_segment.spaces`
```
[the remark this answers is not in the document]
JacksonJ-KC: This is my understanding of how this is currently working:
1) Spaces reference service water heating use, with units of POWER or VOLUME --> return list of spaces that reference the service water heating use
2) Spaces reference service water heating use, with units of POWER_PER_AREA --> return list of spaces that reference the service water heating use
3) Spaces do not reference the service water heating use, with units of POWER or VOLUME --> return empty list
4) Spaces do not reference the service water heating use, with units of POWER_PER_AREA --> return list of all building segment spaces

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
```diff
@@ -0,0 +1,39 @@
...
+        - convert the list of spaces to a list of space ids: `space_ids = [space.id for space in spaces]
```
JacksonJ-KC, line 29: we return a list of IDs here, but the first place I am seeing this function called then uses the get_obj_by_id() to go and get the object again. Consider returning the list of objects.

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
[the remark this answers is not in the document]
KarenWGard: Hmmm… interesting. My knee-jerk reaction is that this would make it challenging to maintain the encapsulation of potential schema changes that this function and get_SWH_uses_associated_with_each_building_segment are intended to achieve, but I think if we always return an empty list when no spaces are served, this will work. Putting a pin /placeholder in this - I’ll keep this in mind as I go through the rest of your comments

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
KarenWGard: I thought the convention for the RCT was to always pass ids and not objects? I’d be happy to pass objects instead! Can you confirm?

claperle: My thought on this is that it is something the RCT team would handle during implementation and that we did not have to worry too much about it as long as the intent was communicated.

docs/ruleset_functions/get_spaces_served_by_SWH_use.md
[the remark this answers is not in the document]
KarenWGard: @JacksonJ-KC - making the change you suggest would break encapsulation.  The function `get_energy_required_to_heat_swh_use `relies on `get_spaces_served_by_SWH_use`.  This function doesn't know which building segment the swh_use is located in.  We might be able to pass the building segment to that function, but that makes `get_energy_required_to_heat_swh_use` unnecessarily more complicated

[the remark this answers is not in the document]
JacksonJ-KC: `get_energy_required_to_heat_swh_use` is only called from `get_SWH_equipment_associated_with_each_SWH_bat`. It is called from within a loop through building segments so it will be easy to pass the building segment. If the list returned from `get_energy_required_to_heat_swh_use` is empty, use all spaces in the building segment as the list. This does not add complication, it significantly simplifies `get_spaces_served_by_SWH_use` and keeps power and volume out of the function because they are not related to the function's own description:

> This function determines the spaces served by a given SWH use.  The convention is that if any spaces reference the swh_use, then the service water heating use applies to only those spaces. If no spaces reference the service water heating use, it applies to all spaces in the building segment.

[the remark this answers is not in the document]
KarenWGard: what you're suggesting works much better given our proposed schema changes, and it will work if the schema doesn't change, but in that case it's a bit messier because if the schema stays the same as it is now, SWH_use is a child of space, so a SWH_use will never be applied to zero or multiple spaces, and `get_energy_required_to_heat_swh_use` has to know something about the structure of how swh_uses are stored.

I've changed it to do as you suggested because I think it is a better solution, but I wanted to highlight the above so you understand my reasoning and so we can make sure these functions will work no matter which way the schema goes.

JacksonJ-KC (approved):

claperle (approved):

PR #1412, opened 2024-07-23
RS/JX/Rule12-1 

Jiarongx-Xie:
This rule has not been tested.


rct229/rulesets/ashrae9012019/section12/section12rule1.py
```diff
@@ -11,35 +14,88 @@ class Section12Rule1(RuleDefinitionListIndexedBase):
...
+            )
+
+            for (
+                misc_equipment_b,
+                misc_equipment_p,
+            ) in proposed_baseline_misc_equipment_pairs:
+                misc_equipment_power_b = misc_equipment_b["power"]
+                misc_equipment_power_p = misc_equipment_p["power"]
+                if misc_equipment_power_b > misc_equipment_power_p:
+                    reduced_misc_equipment_power.append(misc_equipment_power_b)
```
weilixu, line 74: The RDS says appending an object following the format below:
```python
{"id": misc_equipment_b.id, "baseline_power": misc_equipment_power_b, "proposed_power": misc_equipment_power_p}
```

rct229/rulesets/ashrae9012019/section12/section12rule1.py
```diff
@@ -11,35 +14,88 @@ class Section12Rule1(RuleDefinitionListIndexedBase):
...
+                "compliance_path": data["compliance_path"],
             }
 
+        def manual_check_required(self, context, calc_vals=None, data=None):
+            unexpected_misc_equipment_power = calc_vals[
+                "unexpected_misc_equipment_power"
+            ]
+            compliance_path = calc_vals["compliance_path"]
+            return len(unexpected_misc_equipment_power) == 0 and (
+                compliance_path is None or compliance_path != "CODE_COMPLIANT"
```
weilixu, line 90: ` compliance_path is None` is part of `compliance_path != "CODE_COMPLIANT"` logic.
So `compliance_path = None, compliance_path != "CODE_COMPLIANCE" is True`.

weilixu, line 90: The `CODE_COMPLIANCE` shall be a schema enum.

At the top of this script,
```python
COMPLIANCE_PATH_TYPE = SchemaEnums.schema_enums["CompliancePathOptions2019ASHRAE901"]
```
Then in here,  you can do:
`COMPLIANCE_PATH_TYPE.CODE_COMPLIANCE`

weilixu:
@Jiarongx-Xie Thanks for addressing the comments.
There is one more comment left and a conflicting files need to be addressed.

Jiarongx-Xie:
@weilixu Thanks for the review. I have addressed all comments and all tests passed when tested locally. It seems that the CI failed due to some problem not related to this rule.

rct229/rulesets/ashrae9012019/section12/section12rule1.py
```diff
@@ -11,35 +17,95 @@ class Section12Rule1(RuleDefinitionListIndexedBase):
...
+            proposed_baseline_misc_equipment_pairs = zip(
+                misc_equipment_list_b, matched_misc_equipment_list_p
+            )
+
+            for (
+                misc_equipment_b,
+                misc_equipment_p,
+            ) in proposed_baseline_misc_equipment_pairs:
+                misc_equipment_power_b = misc_equipment_b.get("power", ZERO.POWER)
+                misc_equipment_power_p = misc_equipment_p.get("power", ZERO.POWER)
```
weilixu, line 70: @JacksonJ-KC After reading this code, I wonder if we ever considered what means equal between `misc_equipment_b` and `misc_equipment_p`. Have we ever discussed whether the rule's scope is limited in `power` or the object? What if the `misc_equipment_b` has a different energy type but same power as `misc_equipment_p`, do they consider equal in this rule?

weilixu (approved):
Good work.
I have one question for RDS logic - but others look great. Thank you.

rct229/rulesets/ashrae9012019/section12/section12rule1.py
JacksonJ-KC: Good point Weili, I don't think this came up during discussion in the RDS meetings. I think the energy type should also need to match for the rule to pass.

PR #1422, opened 2024-07-27
Rule test engine bug fix - sometimes fail to flag not_applicable and undetermined

jugonzal07:
Addressed issue where `undetermined` and `not_applicable` tests would not get flagged if a ruletest section had exclusively failures of those type. NOTE-- this will result in needing to revisit test cases that passed when they should have failed.

### DIFF
--- a/rct229/ruletest_engine/ruletest_engine.py
+++ b/rct229/ruletest_engine/ruletest_engine.py
@@ -158,13 +158,20 @@ def process_test_result(test_result, test_dict, test_id):
             outcome_text = "NOT_APPLICABLE"
 
     else:
+
+        expected_result = test_dict["expected_rule_outcome"]
+
         if test_result == "pass":
-            outcome_text = f"FAILURE: Test {test_id} passed unexpectedly. The following condition was not identified: {description}"
+            outcome_text = f"FAILURE: Test {test_id} passed unexpectedly. Expected '{expected_result}'. The following condition was not identified: {description}"
         elif test_result == "fail":
-            outcome_text = f"FAILURE: Test {test_id} failed unexpectedly. The following condition was not identified: {description}"
+            outcome_text = f"FAILURE: Test {test_id} failed unexpectedly. Expected '{expected_result}'. The following condition was not identified: {description}"
         elif test_result == "undetermined":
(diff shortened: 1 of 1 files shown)


jugonzal07:
Very weird! I will definitely take a look. thanks @JacksonJ-KC

jugonzal07:
I found the issue and added a correction. In short, it looks like these tests are evaluating to nothing at all and return a blank list. This raises failures thankfully, but results in an empty log. My new commit will now correctly log this error. Also, added additional check in case some other combination of factors results in a failure without a log describing why it fails-- at least some log message will let us know.

JacksonJ-KC:
Perfect, thanks!

JacksonJ-KC:
Many rules in sections 5 and 22 use list filter to figure out if a rule is applicable. If list filter returns an empty list, then the rule is not evaluated at all and the NOT_APPLICABLE outcome is not actually being received. 

I think if context_list is empty then we should continue to provide the FAILURE message that you added, but if filtered_context_list is empty we should provide NOT_APPLICABLE outcome. If it's possible for this only to happen in rule tests then I think that would be preferable. I like that the rules simply don't evaluate when they don't apply to real projects.

Rules I have noticed that behave this way:
5-5
5-7
5-9
5-11
22-17
22-18
22-19
22-28

jugonzal07:
Cool, we should talk through this a bit. I'm less familiar with the Rule section of the repo and the contexts for `context_list` and `filtered_context_list` is unfamiliar to me outside them being used by a `RuleDefinitionListBase`. The ruletest engine entirely relies on the outcome dictionary from [evaluate_rule](https://github.com/pnnl/ruleset-checking-tool/blob/RT/JDJ/resolve-na-undetermined-failures/rct229/rule_engine/engine.py#L117-L131) and the `context_list` and `filtered_context_list` are not easily available. Perhaps if they could be added to that dictionary we'd have more context for the rule test engine portion of this.

JacksonJ-KC:
Sounds good! I have addressed many of the issues already and for the others I have a plan in place. See this comment. https://github.com/pnnl/ruleset-checking-tool/pull/1453#issue-2441359672

PR #1432, opened 2024-07-30
RS/JDJ/Rule19-18 bug fix

JacksonJ-KC:
The only part of the issue that I did not address is the RCTFailureException when ZERO.POWER is passed to table_G3_9_1_lookup if there are no supply fans. `RCTFailureException: The 'shaft_input_power' must be greater than 0.0 hp.` 

I'm not sure if this needs to be addressed, but wanted to bring attention to it.

### DIFF
--- a/rct229/rulesets/ashrae9012019/section19/section19rule18.py
+++ b/rct229/rulesets/ashrae9012019/section19/section19rule18.py
@@ -22,11 +22,13 @@ from rct229.rulesets.ashrae9012019.ruleset_functions.get_fan_system_object_suppl
     get_fan_system_object_supply_return_exhaust_relief_total_power_flow,
 )
 from rct229.schema.config import ureg
+from rct229.schema.schema_enums import SchemaEnums
 from rct229.utils.assertions import getattr_
 from rct229.utils.jsonpath_utils import find_all, find_one
 from rct229.utils.pint_utils import ZERO
 from rct229.utils.std_comparisons import std_equal
 
+ENERGY_RECOVERY = SchemaEnums.schema_enums["EnergyRecoveryOptions"]
 APPLICABLE_SYS_TYPES = [
     HVAC_SYS.SYS_3,
     HVAC_SYS.SYS_4,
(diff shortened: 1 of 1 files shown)


rct229/rulesets/ashrae9012019/section19/section19rule18.py
```diff
@@ -183,15 +183,16 @@ def get_calc_vals(self, context, data=None):
...
+                if exactly_one_supply_fan
                 else ZERO.POWER
             ).to(ureg.hp)
 
             A = 0.0
             more_than_one_exhaust_fan_and_energy_rec_is_relevant_b = False
             if (
                 fan_sys_b.get("air_energy_recovery")
-                and fan_sys_info_b["exhaust_fans_qty"] == 1
+                and fan_sys_b["air_energy_recovery"].get("type") != ENERGY_RECOVERY.NONE
```
weilixu, line 194: None does not equal to `ENERGY_RECOVERY.NONE`. Is that allowed here?

rct229/rulesets/ashrae9012019/section19/section19rule18.py
```diff
@@ -172,9 +174,7 @@ def get_calc_vals(self, context, data=None):
             )
             supply_flow_b = fan_sys_info_b["supply_fans_airflow"].to(ureg.cfm)
 
-            more_than_one_supply_fan_b = (
-                fan_sys_info_b["supply_fans_qty"] != 1
-            )  # the reason why > 1 isn't used is to prevent an error when no supply fan exists
+            exactly_one_supply_fan = fan_sys_info_b["supply_fans_qty"] == 1
```
weilixu, line 177: Does no supply fan a concern in this rule? I saw it is captured by manual check so we should be fine with this logic.

weilixu (approved):
Thanks, I have two comments but generally this is a bug fix.

rct229/rulesets/ashrae9012019/section19/section19rule18.py
JacksonJ-KC: This is related to my initial comment on the PR. I noticed in our test case for this scenario that the logic goes to `table_G3_9_1_lookup` and results in  `RCTFailureException: The 'shaft_input_power' must be greater than 0.0 hp.`

JacksonJ-KC: I am not sure if we want to handle this scenario more explicitly

weilixu: I see, the failure comes from missing supply flows - this is a typical data errors we should capture in the RCT. I do think it makes sense to leave it as is so the data look up function captures the issue.

Issue #1429, opened 2024-07-29
Rule 23-1 logic does not align with Rule Description


JacksonJ-KC:
Rule description: 

> System 2 and 4 - Electric air-source heat pumps shall be modeled with electric auxiliary heat and an outdoor air thermostat. The systems shall be controlled to energize auxiliary heat only when the outdoor air temperature is less than 40°F. The air-source heat pump shall be modeled to continue to operate while auxiliary heat is energized.

The rule logic does not include any logic pertaining to auxiliary heat. Data elements `heatpump_auxilliary_heat_type`, and `heatpump_auxilliary_heat_high_shutoff_temperature` are not used.

KarenWGard:
good catch.  Updated the logic

JacksonJ-KC:
Can you please point to the update?

KarenWGard:
> Can you please point to the update?

 [created new pull request](https://github.com/pnnl/ruleset-checking-tool/pull/1434)

PR #1434, opened 2024-07-30
Update Rule23-1.md

KarenWGard:
[updates to address issue 1429](https://github.com/pnnl/ruleset-checking-tool/issues/1429#issuecomment-2258556141)

### DIFF
--- a/docs/section23/Rule23-1.md
+++ b/docs/section23/Rule23-1.md
@@ -36,15 +36,17 @@
 - get the hvac system from the hvac_system_id: `hvac_system = get_component_by_id(hvac_system_id)`
 - get the heating system: `heating_system = hvac_system.heating_system`
 - get the heatpump_low_shutoff_temperature = `heatpump_low_shutoff = heating_system.heatpump_low_shutoff_temperature`
+- get the auxiliary heat high temperature shutoff = 'aux_heat_high_temp_shutoff = heating_system.heatpump_auxilliary_heat_high_shutoff_temperature`
+- get the auxiliary heat energy source: `aux_heat_energy_source = heatpump_auxilliary_heat_type`
 
 **Rule Assertion:**
-- Case 1: The system type is 2 and the heatpump_low_shutoff is <= 17F, PASS: `if system_type == HVAC_SYS.2: if heatpump_low_shutoff <= 17: PASS`
-- Case 2: The system type is 2 and the heatpump_low_shutoff is >25F, FAIL, provide a message: `elif system_type == HVAC_SYS.2:`
-    - `if heatpump_low_shutoff > 25: FAIL; note = "Fail because low temperature heat pump shutoff is above 25F for system 2. The modeled low temperature heat pump shutoff value is " + heatpump_low_shutoff 
-+ "."`
-- Case 3: the system type is 2 and the heatpump_low_shutoff is between 17F and 25: UNDETERMINED; provide a message: `else: UNDETERMINED; note = "Undetermined because the low temperature shutoff is between 17F and 25F for System Type 2.  Check with the rating authority to ensure correct shutoff temperature.  Low shutoff temperature is currently modeled at: " heatpump_low_shutoff = "."`
-- Case 4: the system type is 4 (we already know it's 4, because we've checked the 3 SYS 2 cases above), and the low shutoff temperature is <=10F: PASS: `if heatpump_low_shutoff <= 10F: PASS`
-- Case 5: anything else, FAIL: `ELSE: Fail`
+- Case 1: The auxiliary heat high temperature shutoff is greater than 40F, or the auxiliary heat energy source is anything other than electricity - FAIL: `if (aux_heat_high_temp_shutoff > 40) or (aux_heat_energy_source != "ELECTRIC_RESISTANCE"): FAIL`
+- Case 2: The system type is 2 and the heatpump_low_shutoff is <= 17F, PASS: `if system_type == HVAC_SYS.2: if heatpump_low_shutoff <= 17: PASS`
+- Case 3: The system type is 2 and the heatpump_low_shutoff is >25F, FAIL, provide a message: `elif system_type == HVAC_SYS.2 && heatpump_low_shutoff > 25:`
+    - `FAIL; note = "Fail because low temperature heat pump shutoff is above 25F for system 2. The modeled low temperature heat pump shutoff value is " + heatpump_low_shutoff + "."`
+- Case 4: the system type is 2 and the heatpump_low_shutoff is between 17F and 25: UNDETERMINED; provide a message: `else: UNDETERMINED; note = "Undetermined because the low temperature shutoff is between 17F and 25F for System Type 2.  Check with the rating authority to ensure correct shutoff temperature.  Low shutoff temperature is currently modeled at: " heatpump_low_shutoff = "."`
+- Case 5: the system type is 4 (we already know it's 4, because we've checked the 3 SYS 2 cases above), and the low shutoff temperature is <=10F: PASS: `if heatpump_low_shutoff <= 10F: PASS`
+- Case 6: anything else, FAIL: `ELSE: Fail`
 
 **Notes:**
 


JacksonJ-KC:
I think rule assertion cases can be reduced to this:
1) Aux heat energy source != ELECTRIC_RESISTANCE,
 OR Aux heat high temperature shutoff != 40 F
--> FAIL
2) System type 2 heat pump low shutoff temperature <= 17 F
OR System type 4 heat pump low shutoff temperature <= 10 F
--> PASS
3) System type 2 heat pump low shutoff temperature between 17 F and 25 F
--> UNDETERMINED
4) Anything else
--> FAIL

JacksonJ-KC:
If a model has more than one thing incorrect, can we structure this in a way that provides all the feedback they need to get it fixed in a single run of the RCT?

KarenWGard:
I'm not super familiar with calc_vals, but I could create some variables instead of hard-coding numbers into the rule assertion, then the user could look at this in the detailed RCT report:

`expected_heat_energy_source
expected_heatpump_low_shutoff`... etc

docs/section23/Rule23-1.md
```diff
@@ -36,15 +36,15 @@
 - get the hvac system from the hvac_system_id: `hvac_system = get_component_by_id(hvac_system_id)`
 - get the heating system: `heating_system = hvac_system.heating_system`
 - get the heatpump_low_shutoff_temperature = `heatpump_low_shutoff = heating_system.heatpump_low_shutoff_temperature`
+- get the auxiliary heat high temperature shutoff = 'aux_heat_high_temp_shutoff = heating_system.heatpump_auxilliary_heat_high_shutoff_temperature`
```
claperle, line 39: I think the last character needs to be corrected. Looks like a character is off making it so the python is not blue

docs/section23/Rule23-1.md
```diff
@@ -36,15 +36,15 @@
...
+- get the auxiliary heat energy source: `aux_heat_energy_source = heatpump_auxilliary_heat_type`
```
claperle, line 40: I think this needs to be modified to `aux_heat_energy_source = **heating_system.**heatpump_auxilliary_heat_type`

docs/section23/Rule23-1.md
```diff
@@ -36,15 +36,15 @@
...
-+ "."`
-- Case 3: the system type is 2 and the heatpump_low_shutoff is between 17F and 25: UNDETERMINED; provide a message: `else: UNDETERMINED; note = "Undetermined because the low temperature shutoff is between 17F and 25F for System Type 2.  Check with the rating authority to ensure correct shutoff temperature.  Low shutoff temperature is currently modeled at: " heatpump_low_shutoff = "."`
-- Case 4: the system type is 4 (we already know it's 4, because we've checked the 3 SYS 2 cases above), and the low shutoff temperature is <=10F: PASS: `if heatpump_low_shutoff <= 10F: PASS`
-- Case 5: anything else, FAIL: `ELSE: Fail`
+- Case 1: The auxiliary heat high temperature shutoff is greater than 40F, or the auxiliary heat energy source is anything other than electricity - FAIL: `if (aux_heat_high_temp_shutoff > 40) or (aux_heat_energy_source != "ELECTRIC_RESISTANCE"): FAIL`
+- Case 2: The system type is 2 and the heatpump_low_shutoff is <= 17F OR tye system type is 4 and the heatpump_low_shutoff is <=10F, PASS: `if(system_type == HVAC_SYS.2 and heatpump_low_shutoff <= 17) or (system_type == HVAC_SYS.4 and heatpump_low_shutoff <= 10): PASS`
```
claperle, line 44: Need to revisit as I read the remainder of this RDS, but I think we want a fail outcome also if the low shut off is greater than these values because it would mean that it operates with the electric resistance heating more often in the baseline thereby increasing consumption.

docs/section23/Rule23-1.md
```diff
@@ -36,15 +36,15 @@
...
+- Case 3: The system type is 2 and the heatpump_low_shutoff is >25F, FAIL, provide a message: `elif system_type == HVAC_SYS.2 && heatpump_low_shutoff > 25:`
```
claperle, line 45: Use of Ruby instead of python with the &&?

docs/section23/Rule23-1.md
```diff
@@ -36,15 +36,15 @@
...
+    - `FAIL; note = "Fail because low temperature heat pump shutoff is above 25F for system 2. The modeled low temperature heat pump shutoff value is " + heatpump_low_shutoff + "."`
```
claperle, line 46: There is no Appendix G basis for this, correct? Based on recent discussions within the commitee, I am not sure we should be making these assumptions :(

docs/section23/Rule23-1.md
claperle, line 44: There is no Appendix G basis for this, correct? Based on recent discussions within the commitee, I am not sure we should be making these assumptions :( I know we decided on this approach previously but wonder if it is still appropriate. It does seem like a no brainer to use 17F but not sure about the 10F since there were comments about us "interpreting" 90.1 requirements which this kind of falls into.

KarenWGard: I think that these minimum numbers come from the minimum temperature at which the heat pumps are rated based on the AHRI conditions from Table 6.8.1-2

KarenWGard: ... but open to discussion.

claperle: I don't think 10F is in those tables. 17F is though.

docs/section23/Rule23-1.md
```diff
@@ -36,15 +36,15 @@
...
-- Case 4: the system type is 4 (we already know it's 4, because we've checked the 3 SYS 2 cases above), and the low shutoff temperature is <=10F: PASS: `if heatpump_low_shutoff <= 10F: PASS`
-- Case 5: anything else, FAIL: `ELSE: Fail`
+- Case 1: The auxiliary heat high temperature shutoff is greater than 40F, or the auxiliary heat energy source is anything other than electricity - FAIL: `if (aux_heat_high_temp_shutoff > 40) or (aux_heat_energy_source != "ELECTRIC_RESISTANCE"): FAIL`
+- Case 2: The system type is 2 and the heatpump_low_shutoff is <= 17F OR tye system type is 4 and the heatpump_low_shutoff is <=10F, PASS: `if(system_type == HVAC_SYS.2 and heatpump_low_shutoff <= 17) or (system_type == HVAC_SYS.4 and heatpump_low_shutoff <= 10): PASS`
+- Case 3: The system type is 2 and the heatpump_low_shutoff is >25F, FAIL, provide a message: `elif system_type == HVAC_SYS.2 and heatpump_low_shutoff > 25:`
+    - `FAIL; note = "Fail because low temperature heat pump shutoff is above 25F for system 2. The modeled low temperature heat pump shutoff value is " + heatpump_low_shutoff + "."`
```
supriyagoel, line 46: Same question of the 25F heat pump low shut off for System 2. Is this an interpretation from our side or is there a requirement that states this that I'm missing?

KarenWGard:
we decided on the call 12/4/2024 the following:

1.  Split this rule into two rules, with 23-1 covering: "System 2 and 4 - Electric air-source heat pumps shall be modeled with electric auxiliary heat and an outdoor air thermostat. The systems shall be controlled to energize auxiliary heat only when the outdoor air temperature is less than 40°F"

2.  A new rule (23-17) will cover: "System 2 and 4 - Electric air-source heat pumps shall be modeled to continue to operate while auxiliary heat is energized."

23-17 will look for a "very low" heatpump_low_shutoff_temperature.  The language of ASHRAE 90.1 G3.1.3.1 indicates that the heat pumps are expected to operate while the auxiliary heat is operating.  They can operate at low capacity, or low efficiency, but they still operate.  On the call, we discussed using -40F as the cutoff temperature because this is low enough that the heat pump is always operating.  Another option is to compare the heatpump_low_shutoff_temperature to the heating design day drybulb temperature

claperle (approved):

PR #1436, opened 2024-07-30
Ruletest 12-2

jugonzal07:
First cut at rule test 12-2 JSON


Jiarongx-Xie:
I think test case b should return undetermined (corresponding to Case 4)instead of fail. Could you double-check? Thanks @jugonzal07

Jiarongx-Xie (changes requested):
I think test case b should return undetermined (corresponding to Case 4 in the RDS) instead of fail. Could you double-check? Thanks @jugonzal07

jugonzal07:
> I think test case b should return undetermined (corresponding to Case 4 in the RDS) instead of fail. Could you double-check? Thanks @jugonzal07

Addressed! I made the baseline case's EFLH < proposed cases EFLH. I think that should address this. I would argue that we should update the [RDS Case 2](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section12/Rule12-2.md#:~:text=and%20comparison_data%5B%22eflh_difference%22%5D-,!%3D,-0%3A%20FAIL%20and) to specify that this should be such that eflh_difference **< 0** instead of **!=0**. If it is > 0, this description would meet the requirements of Case 4.

Jiarongx-Xie:
> > I think test case b should return undetermined (corresponding to Case 4 in the RDS) instead of fail. Could you double-check? Thanks @jugonzal07
> 
> Addressed! I made the baseline case's EFLH < proposed cases EFLH. I think that should address this. I would argue that we should update the [RDS Case 2](https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section12/Rule12-2.md#:~:text=and%20comparison_data%5B%22eflh_difference%22%5D-,!%3D,-0%3A%20FAIL%20and) to specify that this should be such that eflh_difference **< 0** instead of **!=0**. If it is > 0, this description would meet the requirements of Case 4.

Thanks Juan. I further looked into the RDS. I think instead of changing Case 2 condition, we need to add `not auto_receptacle_controls_b` to Case 4 condition. Please let me know your thoughts. @jugonzal07 @weilixu

Jiarongx-Xie:
I changed the logic and now test case d failed where it is supposed to fail (Case 2 in RDS) instead of returning UNDETERMINED (Case 4 in RDS). Could you double-check? @jugonzal07

Jiarongx-Xie (approved):

Jiarongx-Xie:
Thank you for making the change. @jugonzal07 Everything looks good to me now.

PR #1438, opened 2024-07-30
Rule test 12-4

jugonzal07:
First cut at rule test 12-4 JSON. Also includes adding a few schedules to the schedule library. In particular, this one required a monthly fractional schedule for computer room operations.


yunjoonjung-PNNL (approved):
Looks great to me. Thanks Juan!

PR #1452, opened 2024-08-01
RS/JDJ/Rule1-1 Bug Fix

JacksonJ-KC:
This PR will be needed for https://github.com/pnnl/ruleset-checking-tool/pull/1422

### DIFF
--- a/rct229/rulesets/ashrae9012019/section1/section1rule1.py
+++ b/rct229/rulesets/ashrae9012019/section1/section1rule1.py
@@ -110,6 +110,8 @@ class Section1Rule1(RuleDefinitionListIndexedBase):
             climate_zone = data["climate_zone"]
             for bpf_bat in bpf_building_area_type_dict:
                 if bpf_bat == "UNDETERMINED":
+                    bpf_bat_dict_area = bpf_building_area_type_dict[bpf_bat]["area"]
+                    total_area += bpf_bat_dict_area
                     continue
                 expected_bpf = table_4_2_1_1_lookup(bpf_bat, climate_zone)[
                     "building_performance_factor"


rct229/rulesets/ashrae9012019/section1/section1rule1.py
```diff
@@ -110,6 +110,8 @@ def get_calc_vals(self, context, data=None):
             climate_zone = data["climate_zone"]
             for bpf_bat in bpf_building_area_type_dict:
                 if bpf_bat == "UNDETERMINED":
+                    bpf_bat_dict_area = bpf_building_area_type_dict[bpf_bat]["area"]
+                    total_area += bpf_bat_dict_area
                     continue
```
weilixu, line 115: If that is the case, why do we need this `if` statement?
line 117 does the same

JacksonJ-KC: The if statement is needed to avoid updating bpf_bat_sum_product. `total_area` must be greater than 0 and this will fix an issue where total_area is 0 when the whole building's area is associated with the "UNDETERMINED" `bpf_bat`

weilixu: The `bpf_bat_sum_product` won't be called when there is `'UNDETERMINED" bpf_bat` because of the `manual_check_required` - that is minor, and is a stylist comment.

I prefer avoiding `continue` statement because it truncates a logic, making it harder to read. It was there because of earlier implementation. If refactor it, I would recommend just do 
```python
if bpf_bat != "UNDERTERMINED":
    expected_bpf = table_4_2_1_1_lookup(bpf_bat, climate_zone)[
                        "building_performance_factor"
                    ]
    bpf_bat_sum_prod += expected_bpf * bpf_bat_dict_area
total_area += bpf_building_area_type_dict[bpf_bat]["area"]
```

JacksonJ-KC: I saw the solution you pushed to Juan's branch and it looks better

PR #1473, opened 2024-08-06
fix test josn 23-4

weilixu:


jugonzal07:
Addressed in https://github.com/pnnl/ruleset-checking-tool/pull/1483

PR #1475, opened 2024-08-06
fix json 5-3

weilixu:


JacksonJ-KC:
This was giving undetermined because when adjacent_to is set to INTERIOR, the adjacent_zone must be defined. I tried fixing this by changing `adjacent_to` from `INTERIOR` to `IDENTICAL` which led to an unfortunate discovery in `get_surface_conditioning_category_dict` not being able to handle the `IDENTICAL` key. This seems like an issue. Do you agree @weilixu ?

JacksonJ-KC:
Changed to EXTERIOR per your original suggestion to bypass the issue for now. https://github.com/pnnl/ruleset-checking-tool/pull/1478

jugonzal07:
Hi all. I see this has already been merged and the spreadsheet was up-to-date as well.

jugonzal07:
> Changed to EXTERIOR per your original suggestion to bypass the issue for now. #1478

Per Jackson's comment, this has been addressed

PR #1485, opened 2024-08-06
RS/JDJ/22-17, 22-18, 22-28 Applicability

JacksonJ-KC:


rct229/rulesets/ashrae9012019/section22/section22rule17.py
```diff
@@ -40,6 +41,10 @@ def create_data(self, context, data):
 
         return {"heat_rejection_loop_ids_b": heat_rejection_loop_ids_b}
 
+    def is_applicable(self, context, data=None):
```
weilixu, line 44: `is_applicable` is called before the `create_data`
Move this function before the `create_data`

The same applies to the other two rules.

JacksonJ-KC: All set

weilixu (approved):

PR #1489, opened 2024-08-07
update the rule logic to pass the test unit.

weilixu:


JacksonJ-KC:
Will calc_vals display the incorrect value for `num_primary_loops` since primary loops are only counted if they have a secondary loop attached?

Issue #1455, opened 2024-08-01
Rule 19-3 Applicability


JacksonJ-KC:
> Weather conditions used in sizing runs to determine baseline equipment capacities shall be based ~~either~~ on design days developed using 99.6% heating design temperatures and 1% dry-bulb and 1% wet-bulb cooling design temperatures.

We are currently returning NOT_APPLICABLE when either the cooling design day type does not equal 1% OR the heating design day type does not equal 99.6%. I'm struggling to understand this because it seems like we are discovering that something is incorrect but we are returning NOT_APPLICABLE.

claperle:
@JacksonJ-KC We discussed this one earlier today 8/6/202 and for some reason the RDS got changed after it was approved and merged. There are no circumstances where this would not be applicable which violates the requirements in the standard requiring that each rule evaluated for applicability only have both an undetermined and not applicable outcome. Since this rule is always applicable does that mean that we should remove it from the list of rules? That doesn't seem to make sense since we want to flag this for the reviewer to conduct a manual check.

PR #1492, opened 2024-08-07
flip the expected outcome.

weilixu:


JacksonJ-KC:
These are not the outcomes that I would expect. The case where "design days developed using 99.0% heating design temperatures and **0.4% dry-bulb and wetbulb cooling design temperatures**." should not result in not_applicable

JacksonJ-KC:
Additional context on the status of this rule from recent email exchanges:
6) [Rule 19-3 Applicability](https://github.com/pnnl/ruleset-checking-tool/issues/1455)
Christina: We discussed this one earlier today 8/6/202 and for some reason the RDS got changed after it was approved and merged. There are no circumstances where this would not be applicable which violates the requirements in the standard requiring that each rule evaluated for applicability only have both an undetermined and not applicable outcome. Since this rule is always applicable does that mean that we should remove it from the list of rules? That doesn't seem to make sense since we want to flag this for the reviewer to conduct a manual check.

JacksonJ-KC:
Can we delete these json files from the repo to pass the CI tests, then recreate them from the spreadsheet when the direction of the rule has been established?

weilixu:
> Additional context on the status of this rule from recent email exchanges: 6) [Rule 19-3 Applicability](https://github.com/pnnl/ruleset-checking-tool/issues/1455) Christina: We discussed this one earlier today 8/6/202 and for some reason the RDS got changed after it was approved and merged. There are no circumstances where this would not be applicable which violates the requirements in the standard requiring that each rule evaluated for applicability only have both an undetermined and not applicable outcome. Since this rule is always applicable does that mean that we should remove it from the list of rules? That doesn't seem to make sense since we want to flag this for the reviewer to conduct a manual check.

I see, checking blame, Yun Joon flipped failed to not_applicable.

The discussion make sense to me as I also have the same question while checking the rule. However, it does not resolve the issue this rule has:
What outcome should `design days developed using 99.0% heating design temperatures and 0.4% dry-bulb and wetbulb cooling design temperatures.` generate? 

A fail is not one of the outcome for non primary rule. If adding a fail outcome, then the rule would become a primary rule and a success outcome pass is required by the standard.

My suggestion is - keep it not_applicable for CI purpose. Bring this back to RDS discussion for modification.

weilixu:
> Can we delete these json files from the repo to pass the CI tests, then recreate them from the spreadsheet when the direction of the rule has been established?

Since most of the work is actually in the spreadsheet, I will leave the decision to you and @jugonzal07 . Whether remove the test cases from the spreadsheet or swap the outcome to pass the CI.
This rule is TBD and require discussion.

JacksonJ-KC:
My preference is to delete the json files temporarily rather than using expected outcomes that we disagree with. Feel free to assign the task of recreating them to me later on.

weilixu:
Addressed in #1493

PR #1496, opened 2024-08-08
Rt/wx/fix rule 18 2

weilixu:
This is not completed.


jugonzal07:
Updates found here:
https://github.com/pnnl/ruleset-checking-tool/pull/1498

PR #1497, opened 2024-08-08
update 18-2 json

weilixu:


jugonzal07:
Addressed here: https://github.com/pnnl/ruleset-checking-tool/pull/1498

PR #1515, opened 2024-08-23
RCT/YJ/get_spaces_served_by_swh_use func

yunjoonjung-PNNL:


Jiarongx-Xie (approved):
It looks good to me. Thank you Yun Joon!

rct229/rulesets/ashrae9012019/ruleset_functions/get_spaces_served_by_swh_use.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    spaces_served = [
+        space["id"]
+        for space in find_all(
+            "$.buildings[*].building_segments[*].zones[*].spaces[*]", rmd
+        )
+        for swh_use in find_all("$.service_water_heating_uses[*]", space)
+        if swh_use_id == swh_use["id"]
+    ]
```
weilixu, line 29: hmmm, I wonder what is the point of a list?
This is because every service water heating use data will be a nested data in a space...

weilixu: Propose to move water heating uses to building segment level so we preserve this logic to minimize future impact.

weilixu (approved):

PR #1521, opened 2024-09-10
RCT/YJ/decimal point update

yunjoonjung-PNNL:
A few notes for review:
From the numerical_assertions_rct.xlsx file, 
- 22-9: Changed from 1 to 0.1 because we compare fractions.
- 22-10: The rule in the spreadsheet doesn't match.
- 22-11: This rule isn't in the spreadsheet, but I added by myself.
- 22-31: The ```building_peak_load_b``` is the unit of power, not kW/ton. For now, I used ```ton``` with the precision 1.
- 18-2, 19-8, and 23-1: The unit should be ```cfm``` instead of dimensionless
- 23-8 and follow rules are empty in the spreadsheet, but I guessed their units/precisions based on other rules.
- All: Even if ```deg F``` is used in the spreadsheet, if the temperature comparison is done in K in the rule, I used ```K``` precision, ```0.1``` because 1 deg F ~= 0.5 deg C. Please let me know if you'd like to change this.


yunjoonjung-PNNL:
19-7-j case passed in my local testing (verified that all the updates were pushed), but failed in the CI/CD testing. I'll revisit this case.

yunjoonjung-PNNL:
Here is what I found for the 19-7-j case.

1. When I ran the case on my laptop, the Fig below shows the comparison values.
image

2. The Fig below shows the comparison values from the CI/CD testing.
image

As we can see the decimal points are treated slightly differently. 

@jugonzal07 Could you update the ```minimum_outdoor_airflow``` values significantly different between ```baseline``` and ```proposed```? This case is supposed to fail, but the reason it passed unexpectedly was two values are too close to each other.

yunjoonjung-PNNL:
The testing case 19-7-j was fixed and this PR is ready to be reviewed!

weilixu:
> A few notes for review: From the numerical_assertions_rct.xlsx file,

> * 22-9: Changed from 1 to 0.1 because we compare fractions.
Ok
> * 22-10: The rule in the spreadsheet doesn't match.
Not sure
> * 22-11: This rule isn't in the spreadsheet, but I added by myself.
Ok.
> * 22-31: The `building_peak_load_b` is the unit of power, not kW/ton. For now, I used `ton` with the precision 1.
Ok.
> * 18-2, 19-8, and 23-1: The unit should be `cfm` instead of dimensionless
Ok.
> * 23-8 and follow rules are empty in the spreadsheet, but I guessed their units/precisions based on other rules.
Yes, those were added after the spreadsheet created.

> * All: Even if `deg F` is used in the spreadsheet, if the temperature comparison is done in K in the rule, I used `K` precision, `0.1` because 1 deg F ~= 0.5 deg C. Please let me know if you'd like to change this.
That works for me.

PR #1523, opened 2024-09-12
RS/YJ/Rule 11-6

yunjoonjung-PNNL:
This rule hasn't been tested as the required function wasn't merged. I'll remove the draft mark when I finished the testing. Thanks!


yunjoonjung-PNNL:
I'm a bit confused by the RDS logic, but my interpretation of it is as follows:
1. if the '``are_thermal_losses_modeled``' key doesn't exist => pass
2. if the '``are_thermal_losses_modeled``' is False => pass
3. if the '``are_thermal_losses_modeled``' is True => fail

Please let me know if my interpretation is wrong. Thank you!

rct229/rulesets/ashrae9012019/section11/section11rule6.py
```diff
@@ -0,0 +1,85 @@
...
+                index_rmd=BASELINE_0,
+                list_path="buildings[*].building_segments[*]",
+            )
+
+        def create_data(self, context, data):
+            rmd_b = context.BASELINE_0
+
+            return {"rmd_b": rmd_b}
+
+        class BuildingSegmentRule(RuleDefinitionBase):
```
weilixu, line 45: My take is this rule should be at `service_water_heating_distribution_systems` level, not building level. (See RDS Evaluation Context)

In RMD Rule, we need to add an `is_applicable` to check if there are `service_water_heating_distribution_systems` or not.
In RMD Rule `create_data`, we should call `get_swh_components_associated_with_each_swh_distribution_system` to get the data.

In the ServiceWaterHeatingDistributionSystem, we need to check every piping, including children, is modeled with thermal losses turned to false.
You can reuse the same code block searches for piping under a ServiceWaterHeatingDistributionSystem from the `get_swh_components_associated_with_each_swh_distribution_system` function.

weilixu (commented):
In general, I feel the evaluation shall be at service water heating system distribution level. The implmenetation currently is in the buildilng level.

rct229/rulesets/ashrae9012019/section11/section11rule6.py
```diff
@@ -0,0 +1,85 @@
...
+    class RMDRule(RuleDefinitionListIndexedBase):
+        def __init__(self):
+            super(Section11Rule6.RMDRule, self).__init__(
+                rmds_used=produce_ruleset_model_description(
+                    USER=False, BASELINE_0=True, PROPOSED=False
+                ),
+                each_rule=Section11Rule6.RMDRule.BuildingSegmentRule(),
+                index_rmd=BASELINE_0,
+                list_path="buildings[*].building_segments[*]",
+            )
```
weilixu, line 38: Need a `is_applicable` here.

weilixu (changes requested):

yunjoonjung-PNNL:
This PR will be replaced with a new 11-6 PR.

PR #1525, opened 2024-09-16
RS/JX/Rule11-11

Jiarongx-Xie:
This rule has been tested and is ready for review.


rct229/rulesets/ashrae9012019/section11/section11rule11.py
```diff
@@ -0,0 +1,89 @@
...
+                        "$.buildings[*].building_segments[*].zones[*].spaces["
+                        "*].service_water_heating_uses[*]",
+                        "id",
+                        swh_use_id,
+                        rmd_b,
+                    )
+                    if swh_use.get("use", 0) > 0:
+                        has_swh_loads = True
+                        break
+            return {"has_swh_loads": has_swh_loads}
```
weilixu, line 85: Similarly, I think the exact logic can be cleaned up a bit.

I also suggest adding the sum of `swh_use` load for reporting instead of a boolean.
```python
return {
    "sum_swh_loads_b": sum_swh_loads
}
```

In the rulecheck, you can simply write:
```python
sum_swh_loads_b = calc_val["sum_swh_loads_b"]
return sum_swh_loads_b <= 0.0
```

weilixu (commented):
I think RDS logics can be simplified in the actual implementation - I provided the suggestions. 
Also, I recommend use swh_use loads instead of flag in the calc_vals for reporting purpose.

Jiarongx-Xie:
> I think RDS logics can be simplified in the actual implementation - I provided the suggestions. Also, I recommend use swh_use loads instead of flag in the calc_vals for reporting purpose.

The implementation you suggested is much more elegant. I have revised the script accordingly. Thank you!

rct229/rulesets/ashrae9012019/section11/section11rule11.py
```diff
@@ -0,0 +1,77 @@
...
+                        "$.zones[*].spaces[*].service_water_heating_uses[*]",
+                        building_segment,
+                    )
+                ]
+            )
+            return {"swh_use_loads_b": swh_use_loads_b}
+
+        def rule_check(self, context, calc_vals=None, data=None):
+            swh_use_loads_b = calc_vals["swh_use_loads_b"]
+            return swh_use_loads_b <= 0.0
```
weilixu, line 77: This implementation is likely making sense. I am not sure if this should be evaluated at building segment level (as indicated in the evaluation context) or RMD level (as indicated in the rule assertion.) We should clarify this with RDS team.
@KarenWGard any insights on this?

rct229/rulesets/ashrae9012019/section11/section11rule11.py
```diff
@@ -0,0 +1,77 @@
...
+                [
+                    swh_use.get("use", 0.0)
+                    # The json path needs to update if schema changes.
+                    for building_segment in find_all(
+                        "$.buildings[*].building_segments[*]", rmd_b
+                    )
+                    for swh_use in find_all(
+                        "$.zones[*].spaces[*].service_water_heating_uses[*]",
+                        building_segment,
+                    )
```
weilixu, line 70: We should follow the RDS way to extract the service water heating uses as the current schema may subject to changes.

Jiarongx-Xie: Sure, then I would suggest creating a function `find_exactly_one_swh_use` in utility_fucntions.py so we only need to update this function if the current schema changes.

weilixu: Sounds good.

Jiarongx-Xie: Addressed.

rct229/rulesets/ashrae9012019/section11/section11rule11.py
KarenWGard: @weilixu - good question.  The rule language references the whole building, but it seems to make more sense to evaluate at the building segment level.  Shall we discuss on our call?

weilixu: @Jiarongx-Xie could you please revise this to building segment level for evaluation?

yunjoonjung-PNNL (approved):
Looks great to me. Thank you Jiarong!

weilixu (approved):
Great work @Jiarongx-Xie

PR #1526, opened 2024-09-16
RS/YJ/Rule 11-7

yunjoonjung-PNNL:
This rule depends on the ```get_swh_bats_and_swh_use ``` function development. Hence, it isn't ready to be reviewed.


rct229/rulesets/ashrae9012019/data_fns/table_G3_1_1_2_fns.py
```diff
@@ -0,0 +1,26 @@
...
+from rct229.rulesets.ashrae9012019.data_fns.table_utils import find_osstd_table_entry
+
+
+def table_g3_1_2_lookup(building_area_type: str) -> dict[str:str]:
+    """
+
+    Parameters
+    ----------
+    building_area_type: str
+        building area type of the service water heating system
```
weilixu, line 11: Specify that this `building_area_type` is ServiceWaterHeatingSpaceOptions2019ASHRAE901

rct229/rulesets/ashrae9012019/data/ashrae_90_1_table_G3_1_1_2.json
```diff
@@ -0,0 +1,148 @@
+{
+    "Baseline_Service_Water_Heating_System": [
+        {
+            "building_area_type": "Automotive facility",
```
weilixu, line 4: I don't think the str would be the description of the `ServiceWaterHeatingSpaceOptions2019ASHRAE901`, but the enums.
So for example, the `Automotive facility` here, should be `AUTOMOTIVE_FACILITY`

rct229/rulesets/ashrae9012019/section11/section11rule7.py
```diff
@@ -0,0 +1,151 @@
...
+                for swh_use_id in swh_bats_and_uses_p[swh_bat]:
+                    swh_uses = find_exactly_one_with_field_value(
+                        "$.buildings[*].building_segments[*].zones[*].spaces[*].service_water_heating_uses[*]",
+                        "id",
+                        swh_use_id,
+                        rmd_p,
+                    )
+                    if (
+                        getattr_(swh_uses, "service_water_heating_uses", "use")
+                        > ZERO.VOLUME
```
weilixu, line 75: The use has no unit - this will cause exception. Just set to 0.0.
Use singular (`swh_use`)
You can also use `swh_use.get("use", 0.0)`

rct229/rulesets/ashrae9012019/section11/section11rule7.py
```diff
@@ -0,0 +1,151 @@
...
+                        or swh_bat != SERVICE_WATER_HEATING_SPACE.PARKING_GARAGE
+                    ):
+                        return True
+                    else:
+                        return False
```
weilixu, line 80: My general feeling is the RDS did not explain where the applicability/evaluation happens. The note says `Applicability is determined at the building segment level, but the rule is evaluated for all SWH equipment serving the building segment.` However, the logic is on `swh_bat`.

In that case, I suggest we use evaluate at `swh_bat` level.
This would require restructure of the context to only include the returned value from `get_SWH_components_associated_with_each_SWH_bat`. See Section5rule14 for the implementation.

weilixu (changes requested):
This is a difficult rule and I think the evaluation shall happens at `SWH_BAT` level, which requires customization.

yunjoonjung-PNNL:
Note: Refer to rule 5-14 for reference

rct229/rulesets/ashrae9012019/section11/section11rule7.py
weilixu: Confirmed with Karen, this rule should be evaluated at building area type level - we should update the note in the RDS to reflect this change.

rct229/rulesets/ashrae9012019/section11/section11rule7.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/data/ashrae_90_1_table_G3_1_1_2.json
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section11/section11rule7.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/data_fns/table_G3_1_1_2_fns.py
yunjoonjung-PNNL: Addressed.

yunjoonjung-PNNL:
Closed. This PR will be replaced with a new 11-7 PR.

PR #1541, opened 2024-10-14
RS/JX/Rule11-15

Jiarongx-Xie:
This rule has been tested (Nov 11).


rct229/rulesets/ashrae9012019/section11/section11rule15.py
```diff
@@ -0,0 +1,223 @@
...
+                swh_use_ids.extend(
+                    [
+                        swh_use_id
+                        for swh_use_id in service_water_heating_use_ids_p
+                        if swh_use_id not in swh_use_ids
+                    ]
+                )
+            return len(swh_use_ids) > 0
+
+        def get_calc_vals(self, context, data=None):
```
weilixu, line 74: I believe this rule should be based at SWH level evaluation as the RDS indicated Evaluation Context is `Each SWH use`.

In this class, you should get `swh_dist_sys_p` and `swh_dist_sys_b` from the RMD.

Then you can start pointing the `list_path="$.buildings[*]"` to `$building[*].building_segments[*].service_water_heating_uses[*]`

The nested class will be `SWHUse` class which does the compare.
Note - the engine takes care of the id matching - so if there is `swh_use_b` exist but no `swh_use_p`, the context.BASELINE_0 is an object but context.PROPOSED = None.

weilixu (changes requested):
This PR needs a complete rewrite - contact me if you are not sure what needs to be done.

Jiarongx-Xie:
@weilixu Thank you for the review. I have updated the implementation based on your feedback. This PR is ready for another round of review.

rct229/rulesets/ashrae9012019/section11/section11rule15.py
```diff
@@ -0,0 +1,213 @@
...
+                elif swh_use_b and not swh_use_p:
+                    rule_status = "fail"
+                    rule_note = (
+                        rule_note
+                        + swh_use_b["id"]
+                        + " exists in the baseline, but not in the proposed model."
+                    )
+                else:
+                    if swh_use_b.get("use_units") != swh_use_p.get("use_units"):
+                        rule_status = "fail"
```
Jiarongx-Xie, line 120: there are two more fields in the latest schema [is_heat_recovered_by_drain](https://github.com/pnnl/ruleset-checking-tool/blob/73427d60cbc3b6a12ee35756ca9d6240defc9e1c/rct229/schema/ASHRAE229.schema.json#L4785) and [is_recovered_heat_used_by_cold_side_feed](https://github.com/pnnl/ruleset-checking-tool/blob/73427d60cbc3b6a12ee35756ca9d6240defc9e1c/rct229/schema/ASHRAE229.schema.json#L4790)

rct229/rulesets/ashrae9012019/section11/section11rule15.py
```diff
@@ -0,0 +1,213 @@
...
+                                + " Proposed Service Water Heating Use is less than the baseline.  Manually verify that reduction is due to an ECM that reduces service water heating use, such as low-flow fixtures. "
+                            )
+                        elif swh_use_p.get("use", 0) > swh_use_b.get("use", 0):
+                            rule_status = "fail"
+                            rule_note = (
+                                rule_note
+                                + swh_use_b["id"]
+                                + " Proposed Service Water Heating Use is greater than the baseline. "
+                            )
+                return {
```
weilixu, line 193: I am not in favor of strictly following the RDS implementation. - `calc_val` function suppose to do calculations and return values for reporting and rule_check/manual_check functions. In here, the rule definition calculation, these are all blended into `calc_val` function.

I suggest focusing on return a few parameters:

```python
return {
    is_swh_use_none_b: swh_use_b is None,
    is_swh_use_none_p: swh_use_p is None,
    swh_use_use_units_b: swh_use_b.get("use_units"),
    swh_use_use_units_p: swh_use_b.get("use_units"),
    swh_use_multiplier_schedule_b: swh_use_b.get("use_multiplier_schedule"),
    swh_use_multiplier_schedule_p: swh_use_p.get("use_multiplier_schedule"),
    swh_use_temperature_at_fixture_b: swh_use_b.get("temperature_at_fixture"),
    swh_use_temperature_at_fixture_p: swh_use_p.get("temperature_at_fixture"),
    swh_use_water_mains_temperature_schedule_b: swh_use_b.get("entering_water_mains_temperature_schedule"),
    swh_use_water_mains_temperature_schedule_p: swh_use_p.get("entering_water_mains_temperature_schedule"),
    swh_use_served_by_distribution_system_b: swh_dist_sys_id_b = swh_use_b.get("served_by_distribution_system"),
    swh_use_served_by_distribution_system_p: swh_dist_sys_id_p = swh_use_b.get("served_by_distribution_system"),
    swh_use_design_supply_water_temperature_b: design_supply_water_temperature_b, # you will need to calculate this value
    swh_use_design_supply_water_temperature_p: design_supply_water_temperature_p, # you will need to calculate this value
    swh_use_use_b: swh_use_b.get("use", 0),
    swh_use_use_p: swh_use_p.get("use", 0)
}
```


Move all the logics including rule_note draft to the rule_check function.
Rule_check function, you can simply `return not rule_note` - if no error, the `rule_note` would be an empty string and it is falsy in python.

Pay extra attention to the manual return function. You will need to make sure the other parameters are all identical (pass the test) before running the manual return. This means you need to compare all the other parameters before checking the equality of the uses.

weilixu (commented):
Comments need to be addressed.

rct229/rulesets/ashrae9012019/section11/section11rule15.py
Jiarongx-Xie: Yes, they are boolean. My understanding is that their equality determines if the swh uses in the baseline and proposed are modeled the same. For example, if [is_heat_recovered_by_drain](https://github.com/pnnl/ruleset-checking-tool/blob/73427d60cbc3b6a12ee35756ca9d6240defc9e1c/rct229/schema/ASHRAE229.schema.json#L4785) is true in the baseline and false in the proposed, the energy required by this swh use will be different in the two models.

rct229/rulesets/ashrae9012019/section11/section11rule15.py
```diff
@@ -0,0 +1,213 @@
...
+
+    class RMDRule(RuleDefinitionListIndexedBase):
+        def __init__(self):
+            super(Section11Rule15.RMDRule, self).__init__(
+                rmds_used=produce_ruleset_model_description(
+                    USER=False, BASELINE_0=True, PROPOSED=True
+                ),
+                each_rule=Section11Rule15.RMDRule.SWHUseRule(),
+                index_rmd=BASELINE_0,
+                list_path="$.buildings[*].building_segments[*].zones[*].spaces[*].service_water_heating_uses[*]",
```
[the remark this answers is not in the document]
Jiarongx-Xie: Added

rct229/rulesets/ashrae9012019/section11/section11rule15.py
Jiarongx-Xie: Addressed

rct229/rulesets/ashrae9012019/section11/section11rule15.py
```diff
@@ -0,0 +1,213 @@
...
+
+            for building_segment in find_all(
+                "$.buildings[*].building_segments[*]", rmd_b
+            ):
+                service_water_heating_use_ids_b = (
+                    get_swh_uses_associated_with_each_building_segment(
+                        rmd_b, building_segment["id"]
+                    )
+                )
+                swh_use_ids.extend(service_water_heating_use_ids_b)
```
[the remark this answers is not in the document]
Jiarongx-Xie: Addressed

rct229/rulesets/ashrae9012019/section11/section11rule15.py
Jiarongx-Xie: Addressed.

yunjoonjung-PNNL:
Could @weilixu review this PR? I made some cosmetic updates. Thank you!

rct229/rulesets/ashrae9012019/section11/section11rule15.py
```diff
@@ -0,0 +1,451 @@
...
+                    else None
+                )
+                swh_use_served_by_distribution_system_id_p = (
+                    swh_use_p.get("served_by_distribution_system")
+                    if swh_use_p
+                    else None
+                )
+                swh_use_served_by_distribution_system_b = swh_dist_sys_dict_b[
+                    swh_use_served_by_distribution_system_id_b
+                ]
```
weilixu, line 135: it seems you will need to safe guard every step.

weilixu (changes requested):
Issues with possible None scenario.

PR #1544, opened 2024-10-16
Rds/kjw/function compare swh dist systems and components

KarenWGard:
rule that contains the functionality for using compare_context_pair for SWH systems - this logic is called twice in the re-written rule 11-1

### DIFF
--- a/docs/ruleset_functions/compare_swh_dist_systems_and_components.md
+++ b/docs/ruleset_functions/compare_swh_dist_systems_and_components.md
@@ -9,7 +9,7 @@ Inputs:
 - **swh_distribution_id**
 
 Returns:
-- **errors**: An array of errors encountered in this function and in compare_context_str.  If errors is an empty array, then all elements that were expected to match do match
+- **errors**: An array of errors encountered in this function and in compare_context_pair.  If errors is an empty array, then all elements that were expected to match do match
 
 Function Call:
 
@@ -34,17 +34,17 @@ Logic:
   - otherwise, add an error and return FALSE: `else: errors << "SWH Distribution System " + swh_distribution_id + " not found in one of the two RMDs"; return FALSE`
 
 
-  - Compare the distribution in the two models using the function compare_context_pair.  compare_context_pair is recursive, so by sending the function the distribution systems, it is also checking the tanks and piping that are child objects of the distribution systems: `compare_context_pair(RMD1_swh_distribution, RMD2_swh_distribution, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, errors)`
+  - Compare the distribution in the two models using the function compare_context_pair.  compare_context_pair is recursive, so by sending the function the distribution systems, it is also checking the tanks and piping that are child objects of the distribution systems.  If the comparison fails, we add an error to the errors list indicating the id of the swh distribution system that failed comparison: `if !compare_context_pair(RMD1_swh_distribution, RMD2_swh_distribution, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, errors): errors << "Comparison of SWH Distribution System: " + swh_distribution_id + " failed."`
   - In addition to the equipment connected to the distribution system, there's also equipment that's part of the service water heating distribution system that are not direct child objects of the distribution system.  We need to check these objects.
   - first, check ServiceWaterHeatingEquipment - when we execute compare_context_pair, this will also check any child objects that exist (SolarThermal and SWH validation point).  Start by checking if there are the same number of objects in models 1 and 2.  We need to do the length check here because it's not checked implicitly as part of compare_context_pair.  For example, if there are more pieces of equipment in the model 1 than model 2, comparing each item found in model 2 could return a false positive: `if len(RMD1_swh_equipment_dict["SWHHeatingEq"]) == len(RMD2_swh_equipment_dict["SWHHeatingEq"]):`
     - look at each SWHEquipment in the model 1: `for swh_eq_id in RMD1_swh_equipment_dict["SWHHeatingEq"]:`
       - get the SWH equipment for models 1 and 2: `swh_eq_1 = get_component_by_id(RMD1, swh_eq_id); swh_eq_2 = get_component_by_id(RMD2, swh_eq_id)`
-      - compare the two SWH equipment using compare_context_pair.  By passing in the same errors list, any new errors get appended to the list: `compare_context_pair(swh_eq_1, swh_eq_2, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, errors)`
+      - compare the two SWH equipment using compare_context_pair.  By passing in the same errors list, any new errors get appended to the list.  If the comparison fails, we add an error to the errors list indicating the id of the swh equipment that failed comparison: `if !compare_context_pair(swh_eq_1, swh_eq_2, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, errors): errors << "Comparison of SWH Equipment: " + swh_eq_id + " failed."`
   - otherwise there is an unequal number of equipment entries.  Add our own error to the errors list: `else: errors << "Unequal numbers of SWH Equipment between the two models for " + swh_distribution_id`
   - next, check Pumps - this will also recursively check PumpOutputValidationPointPumpOutputValidationPoint: `if len(RMD1_swh_equipment_dict["Pumps"]) == len(RMD2_swh_equipment_dict["Pumps"]):`
     - look at each SWHEquipment in the proposed model: `for pump_id in RMD1_swh_equipment_dict["Pumps"]:`
       - get the pumps for models 1 and 2: `pump_1 = get_component_by_id(RMD1, pump_id); pump_2 = get_component_by_id(RMD2, pump_id)`
-      - compare the two pumps using compare_context_pair: `compare_context_pair(pump_1, pump_2, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, error_str)`
+      - compare the two pumps using compare_context_pair.  If the comparison fails, we add an error to the errors list indicating the id of the swh pump that failed comparison: `if !compare_context_pair(pump_1, pump_2, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, error_str): errors << "Comparison of SWH Pump: " + pump_id + " failed."`
   - otherwise, there are an an unequal number of pump entries: `else: errors << "Unequal number of pumps in the two models for SWH Distribution System " + swh_distribution_id`
 
 **Returns** errors
--- a/docs/ruleset_functions/compare_swh_dist_systems_and_components.md
+++ b/docs/ruleset_functions/compare_swh_dist_systems_and_components.md
@@ -7,10 +7,9 @@ Inputs:
 - **RMD2**
 - **compare_context_str**
 - **swh_distribution_id**
-- **errors**
 
 Returns:
-- **all_match**: A boolean indicating whether all of the SWH elements match for the given SWH equipment id in models 1 and 2
+- **errors**: An array of errors encountered in this function and in compare_context_str.  If errors is an empty array, then all elements that were expected to match do match
 
 Function Call:
 
@@ -20,6 +19,7 @@ Function Call:
 Data Lookup: None
 
 Logic:
+  - create the list errors: `errors = []`
   - use the function get_swh_components_associated_with_each_swh_distribution_system to get a dictionary of SWH equipment associated with each Distribution System for model 1 (pumps, tanks, SWH use, distribution system, etc): `RMD1_swh_system_and_equip_dict = get_swh_components_associated_with_each_swh_distribution_system(RMD1)`
   - use the function get_swh_components_associated_with_each_swh_distribution_system to get a dictionary of SWH equipment associated with each Distribution System for model 1 (pumps, tanks, SWH use, distribution system, etc): `RMD2_swh_system_and_equip_dict = get_swh_components_associated_with_each_swh_distribution_system(RMD2)`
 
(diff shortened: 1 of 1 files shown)


docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
+## compare_swh_dist_systems_and_components
+
+Description: This function compares all sub-components of a given SWH Distribution System ID between two models.  The function also accepts   
+
+Inputs:
+- **RMD1**
+- **RMD2**
+- **compare_context_str**
+- **swh_distribution_id**
+- **errors**
```
claperle, line 10: What does this errors do? Suggest explaining that here

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+
+Data Lookup: None
+
+Logic:
+  - use the function get_swh_components_associated_with_each_swh_distribution_system to get a dictionary of SWH equipment associated with each Distribution System for model 1 (pumps, tanks, SWH use, distribution system, etc): `RMD1_swh_system_and_equip_dict = get_swh_components_associated_with_each_swh_distribution_system(RMD1)`
+  - use the function get_swh_components_associated_with_each_swh_distribution_system to get a dictionary of SWH equipment associated with each Distribution System for model 1 (pumps, tanks, SWH use, distribution system, etc): `RMD2_swh_system_and_equip_dict = get_swh_components_associated_with_each_swh_distribution_system(RMD2)`
+
+  - check if this distribution ID is in model 1: `if RMD1_swh_system_and_equip_dict[swh_distribution_id]:`
+    - get the dictionary of all SWH equipment (pumps, tanks, SWH use, dist system, etc) for the given swh_distribution_id for RMD1: `RMD1_swh_equipment_dict = RMD1_swh_system_and_equip_dict[swh_distribution_id]`
+    - get the actual distribution equipment associated with this ID from model 1: `RMD1_swh_distribution = get_component_by_id(RMD1, swh_distribution_id)`
```
claperle, line 28: I am not following this line. What is it doing?

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+  - otherwise, return FALSE: `else: errors << "SWH Distribution System " + swh_distribution_id + " not found in one of the two RMDs"; return FALSE`
+
+  - check if this distribution ID is in model 2: `if RMD2_swh_system_and_equip_dict[swh_distribution_id]:`
+    - get the dictionary of all SWH equipment (pumps, tanks, SWH use, dist system, etc) for the given swh_distribution_id for RMD2: `RMD2_swh_equipment_dict = RMD2_swh_system_and_equip_dict[swh_distribution_id]`
+    - get the actual distribution equipment associated with this ID from model 1: `RMD2_swh_distribution = get_component_by_id(RMD2, swh_distribution_id)`
+  - otherwise, add an error and return FALSE: `else: errors << "SWH Distribution System " + swh_distribution_id + " not found in one of the two RMDs"; return FALSE`
```
claperle, line 34: How is this error message delivered to the reviewer since this function only retyrns a boolean?

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+  - Compare the distribution in the two models using the function compare_context_pair.  compare_context_pair is recursive, so by sending the function the distribution systems, it is also checking the tanks and piping that are child objects of the distribution systems.  The boolean all_match is created and set to the result of the function: `all_match = compare_context_pair(RMD1_swh_distribution, RMD2_swh_distribution, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, errors)`
+  - In addition to the equipment connected to the distribution system, there's also equipment that's part of the service water heating distribution system that are not direct child objects of the distribution system.  We need to check these objects.
```
claperle, line 38: Note to self, need to check from here down

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
KarenWGard: errors is a list that is one of the required inputs for compare_context_str.  Passing it through allows any errors reported by compare_context_str to be accessible by the rule that called this function

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
KarenWGard: because errors is a list that was passed into this function, any items added to the list will be accessible by the rule that called that function.

claperle: I'll have to trust you on that. Not sure my python knowledge is good enough to understand how this can be accessible if the function only returns TRUE or FALSE.

JacksonJ-KC: recommend writing that the function returns the errors and the boolean because this is psuedo code we don't want any assumptions about the scope of variables to be miscommunicated

claperle: Karen going to add note to describe that error is returned

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
weilixu: Can return error list instead of boolean?
An empty list means full match and with message means there are diffs.

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
weilixu: I left the same comment - return a list makes sense.

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
KarenWGard: @weilixu - this is an excellent idea.  I updated the function per your suggestion.  To confirm - does compare_context_str always append an error to the errors list when the result is false?

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+  - first, check ServiceWaterHeatingEquipment - when we execute compare_context_pair, this will also check any child objects that exist (SolarThermal and SWH validation point).  Start by checking if there are the same number of objects in models 1 and 2.  We need to do the length check here because it's not checked implicitly as part of compare_context_pair.  For example, if there are more pieces of equipment in the model 1 than model 2, comparing each item found in model 2 could return a false positive: `if len(RMD1_swh_equipment_dict["SWHHeatingEq"]) == len(RMD2_swh_equipment_dict["SWHHeatingEq"]):`
+    - look at each SWHEquipment in the model 1: `for swh_eq_id in RMD1_swh_equipment_dict["SWHHeatingEq"]:`
+      - get the SWH equipment for models 1 and 2: `swh_eq_1 = get_component_by_id(RMD1, swh_eq_id); swh_eq_2 = get_component_by_id(RMD2, swh_eq_id)`
```
claperle, line 41: I don't think the semi colon is python syntax. if it is not recommend revising.

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+      - compare the two SWH equipment using compare_context_pair.  By passing in the same errors list, any new errors get appended to the list.  If the comparison fails, we add an error to the errors list indicating the id of the swh equipment that failed comparison: `if !compare_context_pair(swh_eq_1, swh_eq_2, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, errors): errors << "Comparison of SWH Equipment: " + swh_eq_id + " failed."`
+  - otherwise there is an unequal number of equipment entries.  Add our own error to the errors list: `else: errors << "Unequal numbers of SWH Equipment between the two models for " + swh_distribution_id`
+  - next, check Pumps - this will also recursively check PumpOutputValidationPointPumpOutputValidationPoint: `if len(RMD1_swh_equipment_dict["Pumps"]) == len(RMD2_swh_equipment_dict["Pumps"]):`
+    - look at each SWHEquipment in the proposed model: `for pump_id in RMD1_swh_equipment_dict["Pumps"]:`
+      - get the pumps for models 1 and 2: `pump_1 = get_component_by_id(RMD1, pump_id); pump_2 = get_component_by_id(RMD2, pump_id)`
```
claperle, line 46: Same comment as above regarding ;

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+
+- **compare_context_pair** - there is no RDS for this function, but it is a function developed for Rule 1-6 that compares two elements  
+- **get_component_by_id**
+
+Data Lookup: None
+
+Logic:
+  - create the list errors: `errors = []`
+  - use the function get_swh_components_associated_with_each_swh_distribution_system to get a dictionary of SWH equipment associated with each Distribution System for model 1 (pumps, tanks, SWH use, distribution system, etc): `RMD1_swh_system_and_equip_dict = get_swh_components_associated_with_each_swh_distribution_system(RMD1)`
+  - use the function get_swh_components_associated_with_each_swh_distribution_system to get a dictionary of SWH equipment associated with each Distribution System for model 1 (pumps, tanks, SWH use, distribution system, etc): `RMD2_swh_system_and_equip_dict = get_swh_components_associated_with_each_swh_distribution_system(RMD2)`
```
JacksonJ-KC, line 24: ~~model 1~~  model 2

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+
+  - check if this distribution ID is in model 1: `if RMD1_swh_system_and_equip_dict[swh_distribution_id]:`
+    - get the dictionary of all SWH equipment (pumps, tanks, SWH use, dist system, etc) for the given swh_distribution_id for RMD1: `RMD1_swh_equipment_dict = RMD1_swh_system_and_equip_dict[swh_distribution_id]`
+    - get the actual distribution equipment associated with this ID from model 1: `RMD1_swh_distribution = get_component_by_id(RMD1, swh_distribution_id)`
+  - otherwise, return FALSE by adding an error to the list: `else: errors << "SWH Distribution System " + swh_distribution_id + " not found in one of the two RMDs"`
+
+  - check if this distribution ID is in model 2: `if RMD2_swh_system_and_equip_dict[swh_distribution_id]:`
+    - get the dictionary of all SWH equipment (pumps, tanks, SWH use, dist system, etc) for the given swh_distribution_id for RMD2: `RMD2_swh_equipment_dict = RMD2_swh_system_and_equip_dict[swh_distribution_id]`
+    - get the actual distribution equipment associated with this ID from model 1: `RMD2_swh_distribution = get_component_by_id(RMD2, swh_distribution_id)`
```
JacksonJ-KC, line 33: ~~model 1~~ model 2

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
JacksonJ-KC: `swh_eq_1, swh_eq_2 = get_component_by_id(RMD1, swh_eq_id), get_component_by_id(RMD2, swh_eq_id)`

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+  - use the function get_swh_components_associated_with_each_swh_distribution_system to get a dictionary of SWH equipment associated with each Distribution System for model 1 (pumps, tanks, SWH use, distribution system, etc): `RMD2_swh_system_and_equip_dict = get_swh_components_associated_with_each_swh_distribution_system(RMD2)`
+
+  - check if this distribution ID is in model 1: `if RMD1_swh_system_and_equip_dict[swh_distribution_id]:`
+    - get the dictionary of all SWH equipment (pumps, tanks, SWH use, dist system, etc) for the given swh_distribution_id for RMD1: `RMD1_swh_equipment_dict = RMD1_swh_system_and_equip_dict[swh_distribution_id]`
+    - get the actual distribution equipment associated with this ID from model 1: `RMD1_swh_distribution = get_component_by_id(RMD1, swh_distribution_id)`
+  - otherwise, return FALSE by adding an error to the list: `else: errors << "SWH Distribution System " + swh_distribution_id + " not found in one of the two RMDs"`
```
JacksonJ-KC, line 29: << is not python syntax

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+    - get the dictionary of all SWH equipment (pumps, tanks, SWH use, dist system, etc) for the given swh_distribution_id for RMD2: `RMD2_swh_equipment_dict = RMD2_swh_system_and_equip_dict[swh_distribution_id]`
+    - get the actual distribution equipment associated with this ID from model 1: `RMD2_swh_distribution = get_component_by_id(RMD2, swh_distribution_id)`
+  - otherwise, return FALSE by adding an error to the list: `else: errors << "SWH Distribution System " + swh_distribution_id + " not found in one of the two RMDs"`
+
+
+  - Compare the distribution in the two models using the function compare_context_pair.  compare_context_pair is recursive, so by sending the function the distribution systems, it is also checking the tanks and piping that are child objects of the distribution systems.  If the comparison fails, we add an error to the errors list indicating the id of the swh distribution system that failed comparison: `if !compare_context_pair(RMD1_swh_distribution, RMD2_swh_distribution, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, errors): errors << "Comparison of SWH Distribution System: " + swh_distribution_id + " failed."`
```
JacksonJ-KC, line 37: << is not python syntax

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+  - otherwise there is an unequal number of equipment entries.  Add our own error to the errors list: `else: errors << "Unequal numbers of SWH Equipment between the two models for " + swh_distribution_id`
+  - next, check Pumps - this will also recursively check PumpOutputValidationPointPumpOutputValidationPoint: `if len(RMD1_swh_equipment_dict["Pumps"]) == len(RMD2_swh_equipment_dict["Pumps"]):`
+    - look at each SWHEquipment in the proposed model: `for pump_id in RMD1_swh_equipment_dict["Pumps"]:`
+      - get the pumps for models 1 and 2: `pump_1 = get_component_by_id(RMD1, pump_id); pump_2 = get_component_by_id(RMD2, pump_id)`
+      - compare the two pumps using compare_context_pair.  If the comparison fails, we add an error to the errors list indicating the id of the swh pump that failed comparison: `if !compare_context_pair(pump_1, pump_2, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, error_str): errors << "Comparison of SWH Pump: " + pump_id + " failed."`
+  - otherwise, there are an an unequal number of pump entries: `else: errors << "Unequal number of pumps in the two models for SWH Distribution System " + swh_distribution_id`
```
JacksonJ-KC, line 48: <<

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
JacksonJ-KC, line 29: I don't think this should say "return FALSE"

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
JacksonJ-KC, line 37: `$`? I suggest ignoring any attempt for proper syntax to call compare_context_pair because there is no RDS for it - I would just describe the arguments that you know you want to pass to it

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+    - look at each SWHEquipment in the model 1: `for swh_eq_id in RMD1_swh_equipment_dict["SWHHeatingEq"]:`
+      - get the SWH equipment for models 1 and 2: `swh_eq_1 = get_component_by_id(RMD1, swh_eq_id); swh_eq_2 = get_component_by_id(RMD2, swh_eq_id)`
+      - compare the two SWH equipment using compare_context_pair.  By passing in the same errors list, any new errors get appended to the list.  If the comparison fails, we add an error to the errors list indicating the id of the swh equipment that failed comparison: `if !compare_context_pair(swh_eq_1, swh_eq_2, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, errors): errors << "Comparison of SWH Equipment: " + swh_eq_id + " failed."`
```
JacksonJ-KC, line 42: same comment for describing arguments and `$`

JacksonJ-KC, line 42: It is important to know that `compare_context_pair()` needs to use one of the items in the pair as the "index". I would suggest indicating that the longer list should always be used as the index

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+  - otherwise there is an unequal number of equipment entries.  Add our own error to the errors list: `else: errors << "Unequal numbers of SWH Equipment between the two models for " + swh_distribution_id`
+  - next, check Pumps - this will also recursively check PumpOutputValidationPointPumpOutputValidationPoint: `if len(RMD1_swh_equipment_dict["Pumps"]) == len(RMD2_swh_equipment_dict["Pumps"]):`
+    - look at each SWHEquipment in the proposed model: `for pump_id in RMD1_swh_equipment_dict["Pumps"]:`
+      - get the pumps for models 1 and 2: `pump_1 = get_component_by_id(RMD1, pump_id); pump_2 = get_component_by_id(RMD2, pump_id)`
+      - compare the two pumps using compare_context_pair.  If the comparison fails, we add an error to the errors list indicating the id of the swh pump that failed comparison: `if !compare_context_pair(pump_1, pump_2, $, extra_schema_for_SWH_comparison.json, true, compare_context_str, error_str): errors << "Comparison of SWH Pump: " + pump_id + " failed."`
```
JacksonJ-KC, line 47: I'm not sure why this is checking `if !compare_context_pair`. If the function returns a list then the list will behave as a truthy value

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
```diff
@@ -0,0 +1,54 @@
...
+    - get the dictionary of all SWH equipment (pumps, tanks, SWH use, dist system, etc) for the given swh_distribution_id for RMD1: `RMD1_swh_equipment_dict = RMD1_swh_system_and_equip_dict[swh_distribution_id]`
+    - get the actual distribution equipment associated with this ID from model 1: `RMD1_swh_distribution = get_component_by_id(RMD1, swh_distribution_id)`
+  - otherwise, return FALSE by adding an error to the list: `else: errors << "SWH Distribution System " + swh_distribution_id + " not found in one of the two RMDs"`
+
+  - check if this distribution ID is in model 2: `if RMD2_swh_system_and_equip_dict[swh_distribution_id]:`
+    - get the dictionary of all SWH equipment (pumps, tanks, SWH use, dist system, etc) for the given swh_distribution_id for RMD2: `RMD2_swh_equipment_dict = RMD2_swh_system_and_equip_dict[swh_distribution_id]`
+    - get the actual distribution equipment associated with this ID from model 1: `RMD2_swh_distribution = get_component_by_id(RMD2, swh_distribution_id)`
+  - otherwise, return FALSE by adding an error to the list: `else: errors << "SWH Distribution System " + swh_distribution_id + " not found in one of the two RMDs"`
+
+
```
JacksonJ-KC, line 36: if the list of errors is not empty at this point I think we should return it and prevent the rest of the logic

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
JacksonJ-KC, line 42: same comment about `if !compare_context_pair`

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
KarenWGard: any suggestions on how to communicate this effectively?

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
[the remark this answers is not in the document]
KarenWGard: thanks!  I've implemented this

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
KarenWGard: old logic - I've removed if !

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
KarenWGard: It's a preference, but I like to provide as many errors as possible to the user at once so they can resolve them at the same time - reduces iterations for the modeler

claperle (approved):
I did not see any issues.

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
weilixu: My take is if the `swh_distribution_id` is not found in one or another, continue processing can become confusing since the rest checks are depending on having these two matched distribution objects.

It would be good to enforce the same length of `RMD1_swh_distribution` and `RMD2_swh_distribution` lists. For example, adding an assert to make sure these two are in the same length. For ones that mismatched (not exist), we can add to the error list, and skip adding them to the `RMD1_swh_distribution` or `RMD2_swh_distribution` since further comparison is just comparing value with `None` anyway.

docs/ruleset_functions/compare_swh_dist_systems_and_components.md
weilixu: > @weilixu - this is an excellent idea. I updated the function per your suggestion. To confirm - does compare_context_str always append an error to the errors list when the result is false?

That is correct.

PR #1557, opened 2024-11-19
Rule 19-7 raised messaged fix

jugonzal07:
Fix for section 19-7 Failing because of of a missing "not" in front of std_equal. Should send this message if the OA CFMs are not equal.


weilixu:

PR #1558, opened 2024-11-19
Rule 19-14 raised messaged fix

jugonzal07:
Missing elements in 19-14 tests needed to run get_fan_system_object_supply_return_exhaust_relief_total_power_flow function. An undetermined test was incorrectly flagged as undetermined for a MissingKeyException, not for raising the right message.


weilixu:

PR #1559, opened 2024-11-19
Rule 19-15 raised messaged fix

jugonzal07:
Fix for 19-15. Function for manual_check_required would never return true if it's checking for both supply_fan_qty_b ==1 AND supply_fan_qty_b !=1. Also fixed issue where keys were wrong in message logic and missing key in rule test.


weilixu:

PR #1560, opened 2024-11-19
Rule 19-19 raised messaged fix

jugonzal07:
Addressed issue in 19-19 where if the W/CFM would pass if it was below the expected value. Now it correctly fails for < 0.3 W/CFM as well, but correctly raises the appropriate message from get_fail_msg. 19-19b is for W/CFM > 0.3 and the new 19-19e is for W/CFM < 0.3


weilixu:

PR #1561, opened 2024-11-19
Rule 19-35 raised messaged fix

jugonzal07:
Error found in the create_data function for 19-35. The parameter all_lighting_space_types_defined is set to true but never evaluated. This logic fails to identify that one of the spaces in the baseline RMI is missing a lighting_space_type.  Likewise, a fix was needed for checking the are_any_lighting_space_types_defined parameter.


weilixu:

PR #1562, opened 2024-11-19
Rule 19-36 raised messaged fix

jugonzal07:
Fix for 19-36. Typos in dict keys (e.g., outdoor_air_flow should have been outdoor_airflow) in applicability test. Additionally the test itself had incorrect values. Updated the test description and unit test to agree with intent of ruletest.


weilixu:

PR #1566, opened 2024-12-02
RCT/YJ/get_fuels_modeled_in_rmd function

yunjoonjung-PNNL:
I updated the RDS heavily because there are lots of redundant codes. Thanks!


weilixu (approved):
nice and clean implementation, good work.

PR #1576, opened 2024-12-13
Rule test 11-9 JSON

jugonzal07:
First cut at 11-9 rule test JSON

### DIFF
--- a/rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section11/rule_11_9.json
+++ b/rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section11/rule_11_9.json
@@ -1,0 +1,202 @@
+{
+    "rule-11-9-a": {
+        "Section": 11,
+        "Rule": 9,
+        "Test": "a",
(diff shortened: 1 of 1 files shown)


yunjoonjung-PNNL (changes requested):
Could @jugonzal07 address the following comments?

1. We only need ```BASELINE``` - the RDS included ```USER```, but Weili confirmed that it's not needed.
2. The case a should has the ```use``` value 0 to be ```not_applicable```.

yunjoonjung-PNNL:
> > Could @jugonzal07 address the following comments?
> > 
> > 1. We only need `BASELINE` - the RDS included `USER`, but Weili confirmed that it's not needed.
> > 2. The case a should has the `use` value 0 to be `not_applicable`.
> 
> Thanks Yun Joon. There was an issue in an equation that created the `PROPOSED`. It should be fixed now along with 0 for `not_applicable`.

Thank you for your quick fix Juan!

yunjoonjung-PNNL (approved):

Issue #1550, opened 2024-11-05
Validation layer - additional logics to include in validating RPD


weilixu:
1. SWH - use and use_units need to be coupled.
2.

JacksonJ-KC:
same with the various efficiency_metric_types/efficiency_metric_values, right?

PR #1578, opened 2024-12-17
RS/JDJ/Rule 11-10

JacksonJ-KC:
I am finished with the rule implementation and TCDs. Ready for review!


rct229/rulesets/ashrae9012019/section11/section11rule10.py
```diff
@@ -0,0 +1,457 @@
...
+
+            def get_fail_msg(self, context, calc_vals=None, data=None):
+                swh_type_b = calc_vals["swh_type_b"]
+
+                if swh_type_b == "OTHER":
+                    return "The water heater type was not recognized, and does not match any of the expected baseline water heater types."
+                else:
+                    return "The modeled efficiency or standby loss for the water heater does not match the expected values."
+
+            def is_tolerance_fail(self, context, calc_vals=None, data=None):
```
JacksonJ-KC, line 436: @weilixu is this method needed?

rct229/rulesets/ashrae9012019/section11/section11rule10.py
```diff
@@ -0,0 +1,508 @@
...
+                                        "gallon"
+                                    ).magnitude,
+                                    "q": swh_input_power_b.to("Btu/h").magnitude,
+                                },
+                            )
+                            standby_loss_target_metric_b = efficiency["metric"]
+                            modeled_standby_loss_b = swh_efficiency_b.get(
+                                standby_loss_target_metric_b
+                            )
+                            to_remove = efficiency
```
weilixu, line 242: I dont quite understand the `remove` here.
The current implement removes the first one meeting the condition, and assign `modeled_standby_loss_b`
Then at line 248, the code requires only one `efficiency_data` remain for `modeled_efficiency_b`

In scenarios 
1. one matched entries - this could return either `modeled_standby_lss_b` is None and `modeled_efficiency_loss_b` not None or `modeled_standby_loss_b` is not None and `modeled_efficiency_loss_b` is None.
2. Two matched entries - depends on condition,`modeled_standby_loss_b` and `modeled_efficiency_b` could be both None or have some values.
3. More than three matched entries, than this will only return `modeled_standby_loss_b` or both `modeled_standby_loss_b` and `modeled_efficiency_b` are None.

Is the bullet point 3 a desired behavior? If bullet point 3 condition would never met, than clearly there is pre-condition for the efficiency_data list.

rct229/rulesets/ashrae9012019/section11/section11rule10.py
weilixu: it is probably good to keep one here since we dont know how the precision works in eval function.

rct229/rulesets/ashrae9012019/section11/section11rule10.py
```diff
@@ -0,0 +1,508 @@
...
+                swh_equip_b = context.BASELINE_0
+                swh_tank_type_b = swh_equip_b["tank"]["type"]
+                swh_tank_storage_volume_b = swh_equip_b["tank"]["storage_capacity"]
+                swh_input_power_b = swh_equip_b["input_power"]
+                swh_efficiency_b = dict(
+                    zip(
+                        swh_equip_b["efficiency_metric_types"],
+                        swh_equip_b["efficiency_metric_values"],
+                    )
+                )
```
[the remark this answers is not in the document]
JacksonJ-KC: https://github.com/pnnl/ruleset-checking-tool/issues/1550

I think this should be included in validation

rct229/rulesets/ashrae9012019/section11/section11rule10.py
JacksonJ-KC: The length of efficiency_data will be a maximum length of 2 based on the usage of the table lookups. This will only occur in the case where it is a gas storage water heater with input rating >105,000 Btu/h where it will include the efficiency rating 80% Et and also the standby loss rating SL >= (Q/800 + 110 ), Btu/h

weilixu: i see, that is the precondition for the `efficiency_data`. Could you add a comment to describe this pre_condition?

rct229/rulesets/ashrae9012019/data_fns/table_7_8_fns.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    # Validate equipment type
+    valid_equipment_types = [
+        "Electric storage water heater",
+        "Gas storage water heater",
+    ]
+    assert_(
+        equipment_type in valid_equipment_types,
+        f"Invalid equipment type. Must be one of {valid_equipment_types}",
+    )
```
weilixu, line 47: I think maybe just use `assert` instead of `assert_`? I check this again and feel like this will only be triggered by software implementation error instead of project file.
@JacksonJ-KC thoughts?

rct229/rulesets/ashrae9012019/section11/section11rule10.py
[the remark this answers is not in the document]
weilixu: got it. We will need to work on the issue then.

PR #1580, opened 2024-12-17
RS/YJ/Rule 11-1

yunjoonjung-PNNL:
This rule needs to be tested


rct229/rulesets/ashrae9012019/section11/section11rule1.py
```diff
@@ -0,0 +1,126 @@
...
+                                swh_distribution_id=swh_dist_id_u,
+                            )
+                        )
+                else:
+                    user_baseline_comparison = compare_swh_dist_systems_and_components(
+                        rmd1=rmd_p,
+                        rmd2=rmd_b,
+                        compare_context_str="AppG 11-1 P_RMD Equals B_RMD",
+                        swh_distribution_id=swh_dist_id_u,
+                    )
```
weilixu, line 87: This logic is when there is distribution system defined but no swh_use load, so we only compare proposed and baseline (typo on the parameter name).

However, one more logic need to add outside of the for loop, which is if there is no distribution system in the user model, then we need to compare proposed and baseline. You will need to start a new loop based on proposed model (if user model has no distribution system) and run through each swh distribution system for comparison.

weilixu (changes requested):

rct229/rulesets/ashrae9012019/section11/section11rule1.py
```diff
@@ -0,0 +1,126 @@
...
+
+            errors = []
+            user_proposed_comparison = []
+            user_baseline_comparison = []
+            for swh_dist_id_u in find_all(
+                "$.service_water_heating_distribution_systems[*].id", rmd_u
+            ):
+                if swh_and_equip_dict_u[swh_dist_id_u].uses:
+                    swh_use_u = find_one(
+                        f'$.buildings[*].building_segments[*].zones[*].spaces[*].service_water_heating_uses[*][?(@.served_by_distribution_system="{swh_dist_id_u}")]',
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section11/section11rule1.py
yunjoonjung-PNNL: Good point. Could you see if my updates align with your explanation?

weilixu (approved):

PR #1581, opened 2024-12-17
RS/YJ/Rule 11-8

yunjoonjung-PNNL:


rct229/rulesets/ashrae9012019/section11/section11rule8.py
```diff
@@ -0,0 +1,248 @@
...
+                        if other_swh_bat != swh_bat_b:
+                            if (
+                                swh_dist_id
+                                in swh_bats_and_uses_b[swh_bat_b].swh_distribution
+                            ):
+                                is_referenced_in_other_bats_b = True
+
+                multiple_segments_with_bat_other_b = (
+                    True if num_other_swh_bat_segment_b > 1 else False
+                )
```
weilixu, line 204: For this one, I think you need to first verify if this bat is OTHER
`multiple_segments_with_bat_other_b = num_other_swh_bat_segment_b > 1 and bat_type == "OTHER"`

Maybe you don't need `multiple_segments_with_bat_other_b`, you can just use num_other_swh_bat_segment_b in the manual check required logic.

rct229/rulesets/ashrae9012019/section11/section11rule8.py
```diff
@@ -0,0 +1,248 @@
...
+                    "multiple_segments_with_bat_other_b": multiple_segments_with_bat_other_b,
+                }
+
+            def manual_check_required(self, context, calc_vals=None, data=None):
+                swh_bat_b = calc_vals["swh_bat_b"]
+                multiple_segments_with_bat_other_b = calc_vals[
+                    "multiple_segments_with_bat_other_b"
+                ]
+
+                return swh_bat_b == "UNDETERMINED" or multiple_segments_with_bat_other_b
```
weilixu, line 220: `(swh_bat_b  == UNDETERMINED and num_of_bldg_segment_b > 1) or (swh_bat_b == OTHER and multiple_segments_with_bat_other_b > 1)`

weilixu (changes requested):

rct229/rulesets/ashrae9012019/section11/section11rule8.py
```diff
@@ -0,0 +1,248 @@
...
+                    )
+                    if get_building_segment_swh_bat(rmd_b, bldg_seg_b["id"])
+                    == SERVICE_WATER_HEATING_SPACE.ALL_OTHERS
+                ]
+            )
+
+            service_water_heating_uses_p = {
+                swh_use["id"]: swh_use.get("use", 0.0)
+                for swh_use in find_all(
+                    "$.buildings[*].building_segments[*].zones[*].spaces[*].service_water_heating_uses[*]",
```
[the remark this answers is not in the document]
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section11/section11rule8.py
yunjoonjung-PNNL: Good point. Addressed.

rct229/rulesets/ashrae9012019/section11/section11rule8.py
yunjoonjung-PNNL: Addressed. We can avoid using ```multiple_segments_with_bat_other_b```, but I used as described in the RDS. Please let me know if you prefer the other way!

weilixu: I see, you are correct. we should keep this parameter

rct229/rulesets/ashrae9012019/section11/section11rule8.py
```diff
@@ -0,0 +1,254 @@
...
+                            if (
+                                swh_dist_id
+                                in swh_bats_and_uses_b[swh_bat_b].swh_distribution
+                            ):
+                                is_referenced_in_other_bats_b = True
+
+                multiple_segments_with_bat_other_b = (
+                    num_other_swh_bat_segment_b > 1
+                    and swh_bat_b == SERVICE_WATER_HEATING_SPACE.ALL_OTHERS
+                )
```
weilixu, line 206: I still don't think this `multiple_segments_with_bat_other_b` is needed. 
See in the `manual_check_required` function below. If you decided to keep the logic here, then in that function, line [225](https://github.com/pnnl/ruleset-checking-tool/pull/1581/files#diff-6326c64cb19fbe9151330d804121974b3801ccd7e057fb6430c6faffe5c29f52R225) should not be > 1 because the data type is boolean, not integer. line 224 is also duplicated.

However, you can just pass `num_other_swh_bat_segment_b` to `manual_check_required` and keep the logic there.

yunjoonjung-PNNL: Good point! Addressed.

weilixu (approved):

PR #1582, opened 2024-12-18
Rule test 11-1 JSON

jugonzal07:
First cut at 11-1 rule test JSON


yunjoonjung-PNNL:
I made the following changes directly the testing schema in PR #1580 because I wasn't sure if the spreadsheet on OneDrive is up-to-date.

(d) changed from fail to pass because ```AppG 11-1 P_RMD Equals B_RMD``` doesn't exist in the extra schema.
(e) changed from fail to pass because the ```USER``` doesn't have ```service_water_heating_distribution_systems``` and nor does the ```PORPOSED```.
(f) i) changed from fail to pass because  ```AppG 11-1 P_RMD Equals B_RMD``` value doesn't exist in the extra schema.
     ii) Added the ```service_water_heating_distribution_systems``` from the PROPOSED``` to the ```BASELINE``` to prevent errors.

PR #1596, opened 2025-01-14
update the search algorithm

weilixu:


jugonzal07 (approved):
Looks good! Both 1-6c and 1-7c now pass successfully without any other adverse behavior for those the rest of the 1-6 and 1-7 rule tests. Good work.

weilixu:
@jugonzal07 i assume all the failed cases are still need more work in the undertmined_text branch?

weilixu:
Fixed in  #1608, close now.

PR #1597, opened 2025-01-14
Rule 1-1 raised message fix

jugonzal07:
Fixed: Issue where section 1 rule 1 would throw where total building area would equal 0 if only UNDETERMINED was returned in check. This happens if BAT's were exclusively oddball ones like restroom. A total area of 0 would throw a different message about the floor area when the issue had to do with the BAT.  Added an exception for these cases to correctly communicate the source of the error.


weilixu:
Fixed in  #1608, close now.

PR #1600, opened 2025-01-15
Create Rule5-44.md

KarenWGard:


docs/section5/Rule5-44.md
```diff
@@ -0,0 +1,50 @@
+
+# Envelope - Rule 5-44  
+
+**Rule ID:** 5-44  
+**Rule Description:** Space Conditioning Categories. Space conditioning categories used to determine applicability of the envelope requirements in Tables G3.4-1 through G3.4-8 shall be the same as in the proposed design.
+
+Exception: Envelope components of the HVAC zones that are semiheated in the proposed design must meet conditioned envelope requirements in Tables G3.4-1 through G3.4-8 if, based on the sizing runs, these zones are served by a baseline system with sensible cooling output capacity 5 Btu/h·ft2 of floor area, or with heating output capacity greater than or equal to the criteria in Table G3.4-9, or that are indirectly conditioned spaces.
```
claperle, line 7: symbol did not come through fomr greater than or equal, maybe write it out instread then

docs/section5/Rule5-44.md
```diff
@@ -0,0 +1,50 @@
+
+# Envelope - Rule 5-44  
+
+**Rule ID:** 5-44  
+**Rule Description:** Space Conditioning Categories. Space conditioning categories used to determine applicability of the envelope requirements in Tables G3.4-1 through G3.4-8 shall be the same as in the proposed design.
+
+Exception: Envelope components of the HVAC zones that are semiheated in the proposed design must meet conditioned envelope requirements in Tables G3.4-1 through G3.4-8 if, based on the sizing runs, these zones are served by a baseline system with sensible cooling output capacity 5 Btu/h·ft2 of floor area, or with heating output capacity greater than or equal to the criteria in Table G3.4-9, or that are indirectly conditioned spaces.  
+
+**Rule Assertion:** Baseline baseline equals proposed
```
claperle, line 9: Not sure if this matters but this is not always the case, should this be something different then?

docs/section5/Rule5-44.md
```diff
@@ -0,0 +1,50 @@
...
+    - CONTINUE TO RULE LOGIC
+- if the applicability logic arrives here without going to rule logic, the rule is not applicable: RULE NOT APPLICABLE
+
+## Rule Logic:  
+
+- create a list of surfaces that do not comply with this rule: ```non_compliant_surface_ids = []```
+  
+- Get surface conditioning category dictionary for P_RMD: ```scc_dictionary_p = get_surface_conditioning_category(P_RMD)```  
+
+- For each building segment in the Baseline model: ```for building_surface_b in B_RMD...surfaces:```
```
claperle, line 37: The description does not appear match the python. This is for each surface in the baseline model.

docs/section5/Rule5-44.md
```diff
@@ -0,0 +1,50 @@
...
+    - according to the exception, if the surface is semi-heated ("SEMI-EXTERIOR") in the proposed design, it might be classified as fully-conditioned in the baseline, if, according to sizing runs, the zone is served by HVAC equipment that meets the requirements of a fully-conditioned space.  The function get_surface_conditioning_category runs these checks, so we just need to check the conditioning types.  First, check whether the proposed conditioning type is SEMI-EXTERIOR: ```if scc_dictionary_p[building_surface_b.id] == "SEMI-EXTERIOR":```
+      - if the baseline is NOT one of the fully-conditioned surface types ("EXTERIOR MIXED", "EXTERIOR RESIDENTIAL", "EXTERIOR NON-RESIDENTIAL"), add this surface ID to the list of non compliant surface ids: ```if !(scc_dictionary_b[building_surface_b.id] in ["EXTERIOR MIXED", "EXTERIOR RESIDENTIAL", "EXTERIOR NON-RESIDENTIAL"]: non_compliant_surface_ids.append(building_surface_b.id)```
+    - Otherwise, the exception is not met, add the id of this surface to the list of non compliant surface ids: ```else: non_compliant_surface_ids.append(building_surface_b.id)```
```
claperle, line 41: I recommend that we discuss this on the RDS call.
I think the logic should be that we cycle through all of the SEMI-EXTERIOR surfaces in the proposed design. Check to see if they are classed as CONDITIONED/INDIRECTLY conditioned in the baseline. If they are classed as CONDITIONED/INDIRECTLY in the baseline then we need to check that the baseline thermal properties meet the conditioned envelope requirements in Tables G3.4-1 through G3.4-8. Perhaps this RDS should really just be an update to the RDS that checks that the building meets baseline building envelope requirements in Tables G3.4-1 through G3.4-8 as opposed to being a separate RDS.

docs/section5/Rule5-44.md
KarenWGard: updated to reference exception, but don't think this is critical

docs/section5/Rule5-44.md
KarenWGard: Discussion on 2/5/2025 - Christina has a valid point.

Other section 5 rules related to the baseline envelope rely on the `get_surface_conditioning_category(B_RMD)` function based to categorize surfaces, so this RMD, as written will not raise any applicable spaces.

As written, the baseline envelope maximum U-values should be based on the more stringent conditioning category between the baseline and proposed.  Most of the time the baseline would be conditioned and the proposed would be semi-conditioned, but there is an edge case where the proposed could be fully conditioned and the baseline is semi-heated.

This rule gives us the guidance that the space conditioning categories should match, except where the proposed is semi-conditioned and the baseline is fully conditioned.

**Approach - for 2022 only**
*create a new function that returns the same dictionary as `get_surface_conditioning_category`, but accepts both the B_RMD and and P_RMD and provides the dictionary based on the more stringent category between the two.

*All Section 5 baseline rules in 2022 will need to reference the new function and pass in the B_RMD and P_RMD.

**This RMD will no longer be needed**

PR #1601, opened 2025-01-15
Rds/kjw/get swh equipment type

KarenWGard:
Updated to account for Propane as a fuel type


claperle (commented):
For some reason I do not see an option to approve this.

PR #1388, opened 2024-06-19
Update Rule22-17.md

KarenWGard:

PR #1610, opened 2025-01-20
RS/YJ/Rule 22-17 Update

yunjoonjung-PNNL:
I've updated Rule 22-17  based on PR #1388. Thanks!

@jugonzal07 - Could you update the RPD?


jugonzal07:
Updated and tests are now passing @yunjoonjung-PNNL @weilixu

PR #1614, opened 2025-01-23
4-1_Update

claperle:
Updated Case 2 outcome from UNDETERMINED to FAIL so that this rule would be have pass and a fail per the requirements of 229 and per the decision to allow a Fail unless

### DIFF
--- a/docs/section4/4-1.md
+++ b/docs/section4/4-1.md
@@ -1,30 +1,30 @@
 # Airside System – Rule 4-1
 
-**Rule ID:** 4-1
+**Rule ID:** 4-1  
  
-**Rule Description:** Temperature Control Setpoints shall be the same for proposed design and baseline building design.
+**Rule Description:** Temperature control setpoints shall be the same for proposed design and baseline building design.  
 **Rule Assertion:** P-RMD = B-RMD                                           
-**Appendix G Section:** Table G3.1 Section 4 proposed and baseline columns including exception #1 and #2. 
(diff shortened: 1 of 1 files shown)


KarenWGard (approved):

supriyagoel (approved):

docs/section4/4-1.md
```diff
@@ -33,7 +33,7 @@
             - Check if schedules match each hour of the year, if there is a mismatch then set the schedule_mismatch boolean variable equal to TRUE: `if thermostat_heating_stpt_sch_b.hourly_values(x) != thermostat_heating_stpt_sch_p.hourly_values(x):schedule_mismatch = TRUE`
     **Rule Assertion:**
     - Case 1: For each zone if the baseline and proposed heating and cooling thermostat schedule match: `if schedule_mismatch == FALSE: PASS`
```
yunjoonjung-PNNL, line 35: Should be  ````if schedule_mismatch == **TRUE**: PASS````? because the text says ```For each zone if the baseline and proposed heating and cooling thermostat schedule match```

claperle: @yunjoonjung-PNNL the boolean name is schedule_**mis**match so what is there is correct.

yunjoonjung-PNNL: I missed this part. Thank you for your confirmation.

claperle: No problem!

weilixu (approved):

PR #1619, opened 2025-01-23
10-15

claperle:
Updated 10-1 to have an undetermined and not_applicable outcome. Split into two rules, 10-15 was created which checks that if humidification was specified in the proposed that it was also modeled in the baseline.


docs/section10/10-15.md
```diff
@@ -0,0 +1,58 @@
...
+If humidification was modeled in the proposed for a specific zone, check if it was modeled in the baseline
+- For each zone_p in the P_RMR: `For zone_p in P_RMR...Zone:`
+    - Reset has_humidification_p in P_RMR boolean variable: `has_humidification_p = FALSE`
+    - Reset has_humidification_b in B_RMR boolean variable: `has_humidification_b = FALSE`
+    - Check if zone has humidification in the P_RMR, if it does then set boolean variable to true: `if zone_p in zones_have_humidification_list_p: has_humidification_p = TRUE`  
+    Check if zone was modeled with humidification in the B_RMR 
+    - Get analogous zone in the B_RMR: `zone_b = match_data_element(B_RMR,Zones,zone_p.id)`
+    - Get list of hvac systems that serve this zone in the B_RMR, should only be one, note that this RDS assumes only one: `hvac_list_b = get_list_hvac_systems_associated_with_zone(B_RMR,zone_b.id)`
+    - Check if the humidification_type does not equal Null or None, if it does then set has_humidification_b to True: `if hvac_b.humidification_type != Null and hvac_b.humidification_type != "None": has_humidification_b = TRUE`
```
KarenWGard, line 46: `hvac_b` is not defined

docs/section10/10-15.md
```diff
@@ -0,0 +1,58 @@
...
+    - Case 3: Else if the baseline zone is not modeled with humidification when the proposed zone has humidification: `elif has_humidification_b == FALSE and has_humidification_p == TRUE: outcome == "FAIL" and raise_message "The baseline was NOT modeled with humidification when humidification was modeled in the proposed design model"`  
+    - Case 4: Else, humidification not modeled for the zone in proposed or baseline: `Else: outcome == "NOT_APPLICABLE"`  
+    
+**Notes**
+1. Based on the wording in Table G3.1 #10 baseline column, it means that if there was a scenario where one zone on a floor was served by a single zone system with humidification in the proposed, and a single system was required to be modeled per floor in the baseline that the whole system would have humidification. Not sure how many software tools would allow you to model humidification for just one zone. Maybe the assumption is that in this case G3.1.1 b would be triggered. The reason I thought about this is that maybe it is odd to be reporting this on the zone level.
```
KarenWGard, line 55: I think zone level is appropriate as there is too much variation between system coverage between proposed and baseline.

Any software tools that allow humidification in just one zone would probably do it through a second system or zone-level humidification, which means that you might need to check multiple HVAC systems in the baseline.  This is one of those situations where complying with one requirement (model humidification in the same zones as proposed) directly conflicts with another baseline requirement (one system per floor, and only one terminal per zone).  Though I think those cases will be rare.

claperle: Agreed. I think this needs to be addressed in App G.

docs/section10/10-15.md
claperle: Fixed, thanks!

supriyagoel (approved):

docs/section10/10-15.md
```diff
@@ -0,0 +1,59 @@
...
+1. match_data_element()
+2. get_baseline_system_types() 
+3. get_list_hvac_systems_associated_with_zone()
+4. get_hvac_zone_list_w_area()
+
+
+**Rule Logic:**  
+**Applicability Check 1:** 
+- Get dictionary list of proposed zones and the associated HVAC systems: `hvac_zone_list_w_area_dict_p = get_hvac_zone_list_w_area (P_RMR)`
+- Get dictionary for finding baseline system types for hvac systems: `baseline_hvac_sys_type_ids_dict_b = get_baseline_system_types(B_RMR)`
```
yunjoonjung-PNNL, line 29: It seems like ```baseline_hvac_sys_type_ids_dict_b``` is not used anywhere below. Could you double-check this?

claperle: I agree. I do not see this used anywhere.

yunjoonjung-PNNL: Thanks! Should we delete the line if not needed?

claperle: Sounds good!

PR #1621, opened 2025-01-23
11-8 code updates and test cases.

weilixu:
json updates included for review.


weilixu:
Condense the change to #1623

PR #1623, opened 2025-01-23
Rt/wx/section 11 debug

weilixu:
@JacksonJ-KC 
It seems there are failure cases in 11-10 once the messages are added.
Could you take a look?


weilixu:
@jugonzal07 rule 11-1, 8 and 16 are fixed in this PR. You can update the json tests according to the changes in this PR. Waiting for 11-10

JacksonJ-KC:
@weilixu I addressed the main issues, the last remaining failure is an interesting case where either the unit conversion or floating point precision is causing an additional 0.00000000000006 to be added to the `storage_capacity` and so it does not fall within the expected range

JacksonJ-KC:
I manually removed the 0.00000000000006 to show passing results, but regenerating the json from the spreadsheet will reintroduce the issue

weilixu:
> @weilixu I addressed the main issues, the last remaining failure in case 11-10-L is an interesting case where either the unit conversion or floating point precision is causing an additional 0.00000000000006 to be added to the `storage_capacity` and so it does not fall within the expected range

Is it ok to round it to a decimal points? Capacity in the code is usually an integer

JacksonJ-KC:
I think this may be a case where we don't want to round at all, but we should trust the exact value that is in the RPD file. The error is only introduced because the Excel->RES generation script introduces these extra digits

JacksonJ-KC:
Maybe we should adjust the test case so that it is not exactly at the threshold, and know that this is how the rule will behave if the threshold is exceed, no matter how small

weilixu:
> Maybe we should adjust the test case so that it is not exactly at the threshold, and know that this is how the rule will behave if the threshold is exceed, no matter how small

I am good with that.

jugonzal07:
Looks great! good work you all.

jugonzal07:
@JacksonJ-KC -- to confirm, it looks like you made these changes in the spreadsheets already, yes? I saw a recent edit and it seems to reflect the changes in this PR. If so, thank you!

JacksonJ-KC:
@jugonzal07 Correct!

jugonzal07 (approved):
Looks good to me! Well done.

weilixu:
@jugonzal07 @JacksonJ-KC Thank you both. I merged this PR.

PR #1625, opened 2025-01-25
Ruletest 23-1 update

jugonzal07:
Updated rule 23-1's rule test JSON to align with the latest RDS changes

### DIFF
--- a/rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section23/rule_23_1.json
+++ b/rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section23/rule_23_1.json
@@ -3,8 +3,8 @@
         "Section": 23,
         "Rule": 1,
         "Test": "a",
-        "test_description": "The project has one building segment with one zone served by baseline system 2.  with a heat pump low shut off of less than 17 F.",
-        "expected_rule_outcome": "pass",
+        "test_description": "The project has one building segment with one zone served by baseline system 2.  The auxiliary heat high temperature shutoff is greater than 40 F.",
+        "expected_rule_outcome": "fail",
         "standard": {
             "rule_id": "23-1",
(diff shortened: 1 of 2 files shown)


rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section23/rule_23_1.json
```diff
@@ -55,7 +52,8 @@
                                                 "heating_system": {
                                                     "id": "HP Heating Coil 1",
                                                     "type": "HEAT_PUMP",
-                                                    "heatpump_low_shutoff_temperature": -8.888888888888857
+                                                    "heatpump_auxilliary_heat_type": "ELECTRIC_RESISTANCE",
+                                                    "heatpump_low_shutoff_temperature": 7.222222222222285
```
weilixu, line 56: This data should be heatpump_auxilliary_heat_high_shutoff_temperature, and the value is 40F

weilixu (changes requested):

rct229/ruletest_engine/ruletest_jsons/ashrae9012019/section23/rule_23_1.json
jugonzal07: ahh my mistake on the key value. And sure, I can lower it to 40 F. I wasn't sure if the `energize auxiliary heat only when the outdoor air temperature is less than 40 F` was inclusive so bumped it up to 45 F bit to make sure it'd flag a failure.

jugonzal07: Actually-- I'm starting to think I got this mixed up (this is why I never do tests right on the line, I get myself all confused). It looks like 40 F should pass. Sorry, let me make this update.

PR #1626, opened 2025-01-25
Ruletest 22-17 update 

jugonzal07:
Updated rule 22-17's rule test JSON to align with the latest RDS changes


yunjoonjung-PNNL:
Hi Juan - Could you address the followings?

1. The value of the ```primary_rule``` key should be "Yes" or "True".
2. Case (d) and (e) falls into the undetermined case, but the messages are not the same. Can you fix them?

jugonzal07:
No worries, I'll grab this branch and do the fix based on our phone call. Should be easy!

yunjoonjung-PNNL:
> No worries, I'll grab this branch and do the fix based on our phone call. Should be easy!

Appreciate it Juan!

PR #1627, opened 2025-01-27
Remove master json when diaggregate json tests

weilixu:


rct229/ruletest_engine/ruletest_jsons/scripts/json_generation_utilities.py
```diff
@@ -521,6 +530,8 @@ def write_ruletest_json(section, rule, ruleset_doc):
 
     # Write out final rule dictionary
     write_ruletest_json(prev_section, prev_rule, ruleset_doc)
+    # Remove the master json file:
+    os.remove(master_json_path)
```
JacksonJ-KC, line 534: I am thinking now that having this master json exist in the repo will be helpful for the rule renumbering process. Because the spreadsheets are not part of the repo, I want to have the renumbering script be able to renumber all test jsons independent of the spreadsheets. I am imagining that the spreadsheets and master json will only use the unique id so that they have no work necessary when renumbering. The disaggregate function will handle the sectionNruleM mapping and renumber the disaggregated rule tests with the appropriate values for N and M.

weilixu: The workflow works good. I can make the code only run in build process (when releasing to PYPI). I think that was one of my intend.

PR #1643, opened 2025-02-12
RCT/JDJ/validation associated list length

JacksonJ-KC:


rct229/schema/validate.py
```diff
@@ -26,6 +26,63 @@
...
+
+        for obj in object_list:
+            for key_1, key_2 in associated_list_data_elements:
+                has_key_1 = key_1 in obj
+                has_key_2 = key_2 in obj
+
+                if has_key_1 or has_key_2:
+                    if not (has_key_1 and has_key_2):  # One key is missing
+                        mismatch_errors.append(
+                            f"'{obj['id']}' has populated '{key_1 if has_key_1 else key_2}' but is missing '{key_1 if not has_key_1 else key_2}'."
```
weilixu, line 68: Is this function run after the schema validation? Just making sure the `["id"]` is well covered.

rct229/schema/validate.py
```diff
@@ -26,6 +26,63 @@
...
+        "$.ruleset_model_descriptions[*].service_water_heating_equipment[*]",
+    ]
+
+    for jsonpath in associated_list_jsonpaths:
+        object_list = find_all(jsonpath, rpd)
+
+        for obj in object_list:
+            for key_1, key_2 in associated_list_data_elements:
+                has_key_1 = key_1 in obj
+                has_key_2 = key_2 in obj
```
weilixu, line 63: Another way to do it - save an if
```python
for obj in object_list:
            for key_1, key_2 in associated_list_data_elements:
               val_1, val_2 = obj.get(key_1, None), obj(key_2, None)
               # XOR operation
               if val_1 is None != val_2 is None:
                  mismatch_errors.append(
                            f"'{obj['id']}' has populated '{key_1 if val_1 else key_2}' but is missing '{key_1 if not val_1 else key_2}'."
```

rct229/schema/validate.py
```diff
@@ -26,6 +26,63 @@
...
+                        )
+                        continue
+
+                    # Ensure when both are lists they have the same length
+                    val_1, val_2 = obj[key_1], obj[key_2]
+                    if (
+                        isinstance(val_1, list)
+                        and isinstance(val_2, list)
+                        and len(val_1) != len(val_2)
+                    ):
```
weilixu, line 78: If schema validation has run before this function, then here you can simply do `val_1 and val_2 and len(val_1) != len(val_2)`
But I do also like to check for list to specify these values shall be list.

weilixu (approved):

rct229/schema/validate.py
JacksonJ-KC: Yes schema validation happens first, then non-schema validation only runs upon validating schema

PR #1645, opened 2025-02-13
19-3_Update

claperle:
New rule 19-37 was added so that 19-3 could be made into a Primary rule with a pass and fail. This is the update to 19-3.


KarenWGard (approved):

supriyagoel (approved):

supriyagoel:
@claperle i think I now understand why 19-3 was split into two and why 19-37 is non mandatory. this split makes sense- appreciate the work on this

PR #1647, opened 2025-02-13
19_23_Update

claperle:
Updated to align with how design day schedules are defined in the 0.1.3 version of the schema and also revised so that there are no "Pass Ifs"

### DIFF
--- a/docs/section19/Rule19-23.md
+++ b/docs/section19/Rule19-23.md
@@ -1,5 +1,5 @@
 # Section 19 - Rule 19-23            
-**Schema Version:** 0.0.23    
+**Schema Version:** 0.1.3    
 **Mandatory Rule:** True    
 **Rule ID:** 19-23               
 **Rule Description:**  For cooling sizing runs, schedules for internal loads, including those used for infiltration, occupants, lighting, gas and electricity using equipment, shall be equal to the highest hourly value used in the annual simulation runs and applied to the entire design day. For heating sizing runs, schedules for internal loads, including those used for occupants, lighting, gas and electricity using equipment, shall be equal to the lowest hourly value used in the annual simulation runs, and schedules for infiltration shall be equal to the highest hourly value used in the annual simulation runs and applied to the entire design day.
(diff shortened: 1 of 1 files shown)


KarenWGard (approved):

JacksonJ-KC (approved):

supriyagoel (approved):

docs/section19/Rule19-23.md
```diff
@@ -87,27 +92,31 @@
...
+    - Case 7: Else if the space type is not defined and the building type is not multifamily and all schedules pass per above then UNDETERMINED: `elif is_dwelling_unit == false and is_space_type_defined == false and building_area_is_MF == false and all(inf_pass_heating, inf_pass_cooling,occ_pass_heating, occ_pass_cooling,int_lgt_pass_heating, int_lgt_pass_cooling,misc_pass_heating, misc_pass_cooling) == true: outcome = "UNDETERMINED" and raise_message "Pass unless the space type is dwelling unit. Dwelling units fall under the exception to Section G3.1.2.2.1.`
```
yunjoonjung-PNNL, line 113: Do we also need to update the Case 7 message?

claperle: I think it is okay because the outcome itself is UNDETERMINED. The message is just saying that it passes unless it is a dwelling unit. Please let me know if you disagree.

claperle: @yunjoonjung-PNNL The issue is that is_space_defined and building_area_is_MF are also equal False so there is no definitive way for us to actually know whether it is a dwelling unit hence the outcome as is. All we know is that is_dwelling_unit was never changed from False to True but we do not know definitively whether it is a dwelling unit or not.

supriyagoel: @claperle I cannot find G3.1.2.2.1

claperle: @supriyagoel See below.
image

PR #1655, opened 2025-02-19
RCT/JDJ/5-23 logic correction

JacksonJ-KC:


rct229/rulesets/ashrae9012019/section5/section5rule23.py
```diff
@@ -37,12 +28,16 @@ def __init__(self):
...
                 each_rule=Section5Rule23.BuildingRule.SurfaceRule(),
                 index_rmd=BASELINE_0,
             )
 
+        def list_filter(self, context_item, data):
+            surface_b = context_item.BASELINE_0
+            subsurfaces_b = surface_b.get("subsurfaces", [])
+            return len(subsurfaces_b) > 0
+
         class SurfaceRule(RuleDefinitionListIndexedBase):
```
weilixu, line 41: If the logic is removed from the surfaces, then you can directly pointing to the subsurface level.
set `list_path = "$.building_segments[*].zones[*].surfaces[*].subsurfaces[*]`

weilixu: A bit of conventions that goes back to early days is regardless whether there is any logics in building level or not, we will add building rule in order to attach the building id to the reporting json (calc_vals).
But for the rest, we usually skip the levels. You can keep the `SurfaceRule` for reporting purpose or remove it as I suggested. I will leave the decision to you.

docs/section5/Rule5-23.md
```diff
@@ -29,21 +29,11 @@
 
         - For each subsurface in surface in P_RMD: `for subsurface_p in surface_p.subsurfaces:`
```
Jiarongx-Xie, line 31: Suggest adding this` Get matching subsurface in B_RMD: subsurface_b = match_data_element(B_RMD, Subsurfaces, subsurface_p.id)`

Jiarongx-Xie (approved):

docs/section5/Rule5-23.md
JacksonJ-KC: Good point! I made this more clear.

weilixu (approved):

PR #1671, opened 2025-03-10
make exception of RMD ID in the match list

weilixu:
This made an exemption on RMD level ID matching by confirming the RMD is matching one of the ruleset model type.


JacksonJ-KC:
I tested this out and it did work as expected. Thanks!

JacksonJ-KC (approved):

PR #1674, opened 2025-03-11
RS/YJ/Rule_4-11_updated_logic

yunjoonjung-PNNL:
This rule hasn't been tested yet.


rct229/rulesets/ashrae9012019/section4/section4rule11.py
```diff
@@ -267,24 +270,42 @@ def manual_check_required(self, context, calc_vals=None, data=None):
                 system_type_match_baseline_proposed = calc_vals[
                     "system_type_match_baseline_proposed"
                 ]
-                return schedule_mismatch or not system_type_match_baseline_proposed
+
+                return not (schedule_mismatch or system_type_match_baseline_proposed)
```
yunjoonjung-PNNL, line 274: This one was derived the following logic.

The pass/fail logic of the rule is as following:

| No. | schedule_mismatch | system_type_match_baseline_proposed | result |
|-----|-------------------|--------------------------------------|--------|
| 1   | FALSE             | TRUE                                 | pass   |
| 2   | TRUE              | TRUE                                 | fail   |
| 3   | TRUE              | FALSE                                | fail   |

The only complimentary case is when ```schedule_mismatch``` is False and ```system_type_match_baseline_proposed```. That logic is ```not schedule_mismatch and not system_type_match_baseline_proposed``` and its simplified version is ```not (schedule_mismatch or system_type_match_baseline_proposed)```

yunjoonjung-PNNL: Please let me know if anything doesn't make sense!

weilixu (approved):

PR #1680, opened 2025-03-14
RCT/JDJ/section19 bugfixes

JacksonJ-KC:
Various bugfixes and improvements to reported values in Section 19


weilixu:
@JacksonJ-KC does the change in 19-8 test json updated in the excelspreadsheet as well?

rct229/utils/pint_utils.py
```diff
@@ -118,6 +120,9 @@ class CalcQ:
 
     def to_str(self, unit_system=UNIT_SYSTEM.IP) -> str:
         units = _UNIT_CONVENTIONS[unit_system][self.q_type]
+        if self.q_type == "area_density" and unit_system == UNIT_SYSTEM.IP:
+            converted_value = self.q.to("1/ft2").magnitude * 1000
+            return f"{converted_value} people/1000 ft²"
```
JacksonJ-KC, line 125: this seems like a hacky solution. Do you see a better way?

JacksonJ-KC: Applies to rule 19-8 FYI

weilixu: see #1696

PR #1683, opened 2025-03-18
RCT/JX/update use of std_equal

### LINKED ISSUE #1659: Do we need to update all use of std_equal?
My understanding is that we had transitioned away from % tolerance and would now be using precision comparisons throughout.

Jiarongx-Xie:
Update numerical comparison in the ruleset and the use of std_equal. Fixes #1659


weilixu:
I did review on the spreadsheet, good work @Jiarongx-Xie , I left my comments but also some questions maybe @JacksonJ-KC can help review.

PR #1688, opened 2025-03-21
Ruletest JSON 4-11 Update

jugonzal07:
First cut at 4-11 update


jugonzal07:
> The following cases failed.
> 
> Cases (c) & (e) - these don't fall into undetermined - Could you see [my logic](https://github.com/pnnl/ruleset-checking-tool/pull/1674#:~:text=This%20one%20was%20derived%20by%20the%20following%20logic.) in my PR? If your understanding is different, we can discuss! Case (d) - Schedules have to be the same.

Case c) has good calculated values so I'm unsure how it's arriving at the wrong message. The calc vals show that they link up with Case 2's expected values.:

- `Calculated values - {'schedule_mismatch': True, 'baseline_served_by_multizone': False, 'proposed_served_by_multizone': True, 'system_type_match_baseline_proposed': False, 'hvac_type_check': True}`
- [Case 2](https://github.com/pnnl/ruleset-checking-tool/blob/26112981218926357b1305076de1912b4c5a9236/docs/section4/4-11.md?plain=1#L63)

Case d) is suppose to have different schedules-- it's testing the failure [Case 3 ](https://github.com/pnnl/ruleset-checking-tool/blob/26112981218926357b1305076de1912b4c5a9236/docs/section4/4-11.md?plain=1#L64)from the RDS.

Case e) did have an issue and I addressed it in an update.

PR #1700, opened 2025-04-11
Rule 19-23 Test JSON Update

jugonzal07:
First cut at 19-23 update


yunjoonjung-PNNL:
Hi Juan - I've tested 19-23 with this testing JSON file.

Could you address the following?
(e) - Case 3, (f) - Case 4, (h) - Case 7
- ```is_heating_schedule_pass``` should be True, but currently it's False

(g) - Case 6
- ```is_space_type_defined_b``` should be True, but currently it's False
- message should include space id at the beginning

jugonzal07:
Addressed these issues and corrected others I found.

PR #1701, opened 2025-04-11
Rule 19-37 Test JSON Update

jugonzal07:
First cut at 19-37 rule test JSON

### DIFF
--- a/rct229/rulesets/ashrae9012019/section19/section19rule37.py
+++ b/rct229/rulesets/ashrae9012019/section19/section19rule37.py
@@ -33,19 +33,19 @@ class Section19Rule37(PartialRuleDefinition):
 
         return {
             "heating_design_day_type_b": weather_b.get(
-                "heating_dry_bulb_temp_design_day_type"
+                "heating_dry_bulb_design_day_type"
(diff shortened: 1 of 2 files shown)


yunjoonjung-PNNL:
HI Juan can you see the following?

Overall, I believe we should have baseline because of Case 2 in RDS. Please let me know if you think differently.

Case (a) ```heating_design_day_type_p``` and ```cooling_design_day_type_p``` are False, which makes the result not applicable
Case  (b) user shouldn't be included
Case  (c) the expected_rule_outcome should be ```not_applicable``` not undetermined

jugonzal07:
Sorry about all these goofy issues with user/baseline/proposed RPDs-- the cells had goofed hardcoded values. Also I made an update to the 19-37 rule itself that corrects what I perceive to be a typo in the RDS. Many relevant design day fields had an extra `temp_` string in there.

Example: `heating_dry_bulb_temp_design_day_type` -> `heating_dry_bulb_design_day_type`

yunjoonjung-PNNL:
Looks great to me. Thank you Juan!

PR #1706, opened 2025-04-22
add CalcQ in processing the outcome data.

weilixu:


JacksonJ-KC:
I tried with the attached RPD file and the Quantity objects are still showing up in the calc vals
[Demo.json](https://github.com/user-attachments/files/19855696/Demo.json)

JacksonJ-KC:
@weilixu The solution for 12-1 works, but not 11-14

rct229/rulesets/ashrae9012019/section11/section11rule14.py
```diff
@@ -120,6 +120,12 @@ def create_data(self, context, data):
                                     "PUMP_POWER"
                                 ] += pump.get("design_electric_power", ZERO.POWER)
 
+            # post-processing the piping info b to convert all quantities to CalcQ.
+            for piping_id in piping_info_b:
+                if piping_info_b[piping_id]["PUMP_POWER"]:
```
JacksonJ-KC: Change this to `if "PUMP_POWER" in piping_info_b[piping_id]:` to make sure the Quantity objects are still converted when the Quantity is 0 and this appears to solve the issue!

PR #1537, opened 2024-10-01
Create Rule5-42.md

KarenWGard:

PR #1716, opened 2025-05-06
Rds/kjw/rule 5 42

KarenWGard:
re-opening pull request.

Original pull request with discussion:

https://github.com/pnnl/ruleset-checking-tool/pull/1537

### DIFF
--- a/docs/section5/Rule5-42.md
+++ b/docs/section5/Rule5-42.md
@@ -9,7 +9,7 @@ a. A separate model of the assembly within the energy simulation model.
 b. Adjustment of the clear-field U-factor in accordance with Section A10.2.    
 
 **Rule Assertion:** B-RMR = expected value                                           
-**Appendix G Section:** Section G3.1-5 Building Envelope Modeling Requirements for the Proposed building
+**Appendix G Section:** Section G3.1-5 Building Envelope Modeling Requirements for the Proposed building  
 **Appendix G Section Reference:** Table G3.1 Section 5(a) Exception #1   
 **Data Lookup:** None  
 **Evaluation Context:** Evaluate each Project   


docs/section5/Rule5-42.md
```diff
@@ -0,0 +1,35 @@
...
+**Mandatory Rule:** False    
+**Rule ID:** 5-42  
+ 
+**Rule Description:** Each linear thermal bridge and point thermal bridge as identified in Section 5.5.5 shall be modeled using either of the following techniques:  
+
+a. A separate model of the assembly within the energy simulation model.  
+b. Adjustment of the clear-field U-factor in accordance with Section A10.2.    
+
+**Rule Assertion:** B-RMR = expected value                                           
+**Appendix G Section:** Table G3.1 Section 5(a)-1 Proposed
```
claperle, line 12: This is technically an exception so recommend Table G3.1 Section 5a Exception #1.

docs/section5/Rule5-42.md
```diff
@@ -0,0 +1,35 @@
...
+**Rule ID:** 5-42  
+ 
+**Rule Description:** Each linear thermal bridge and point thermal bridge as identified in Section 5.5.5 shall be modeled using either of the following techniques:  
+
+a. A separate model of the assembly within the energy simulation model.  
+b. Adjustment of the clear-field U-factor in accordance with Section A10.2.    
+
+**Rule Assertion:** B-RMR = expected value                                           
+**Appendix G Section:** Table G3.1 Section 5(a)-1 Proposed  
+**Appendix G Section Reference:** 5(a)
```
claperle, line 13: Same comment as above

docs/section5/Rule5-42.md
```diff
@@ -0,0 +1,35 @@
...
+
+**Manual Check:** Ensure that linear and point thermal bridges, as identified in Section 5.5.5 are modeled using either of the following techniques:  
+a. A separate model of the assembly within the energy simulation model.  
+b. Adjustment of the clear-field U-factor in accordance with Section A10.2.      
+ 
+**Rule Logic:**  
+**Applicability:**  
+- look at each zone in the RMD: `for zone in P_RMD....zones:`
+    - look at each surface in the zone: `for surface in zone.surfaces:`
+    - if the surface is an exterior surface, then this rule is applicable for the entire project: `if((surface.adjacent_to == "EXTERIOR") or (surface.adjacent_to == "GROUND")): UNDETERMINED`
```
claperle, line 31: This may be too nitpicky, but wouldn't this also apply in the case of an interior surface between an unconditioned and conditioned space? Maybe we should look at the surface conditioning category instead.

Although, I am a little confused. Will this rule produce one outcome for the entire project? If so, I htink it is okay as is. If it is by surface then I think we may need to look at surface conditioning category.

claperle: Per call, Karen (and group) agrees and thinks using the surface conditioning category makes sense since a list of applicable of surfaces will be provided.

KarenWGard: use surface_conditioning_category instead of `.adjacent_to` in order to capture surfaces that separate fully conditioned from partially conditioned spaces

claperle (approved):

weilixu:
This PR is included in the #1886 Close now.

PR #1727, opened 2025-05-16
Rule 6-7 Update: Plenum, crawl-space, interstitial, NONE

### LINKED ISSUE #1724: Rule 6-7 Recommend Checking Space Type
We recently ran the RCT for a multifamily project and received a whole bunch of fails for this daylight dimming control rule because the majority of spaces in the building are dwelling units. I suggest that we update this rule so that it checks if the space type a dwelling unit, Dormitory-Living Quarters, a Fire Station-Sleeping Quarters, Imaging room, or Operating room and that it is classed as not applicable for these space types to avoid false fails.

JacksonJ-KC:


JacksonJ-KC:
Changing to draft because this needs additional thought to handle the fact that multiple spaces can be in 1 zone. I am thinking this would only be not applicable if every space in the zone is in the list of not applicable space types. If only some of the spaces are not applicable then the rule should evaluate as it does currently, agreed?

claperle:
> Changing to draft because this needs additional thought to handle the fact that multiple spaces can be in 1 zone. I am thinking this would only be not applicable if every space in the zone is in the list of not applicable space types. If only some of the spaces are not applicable then the rule should evaluate as it does currently, agreed?

@JacksonJ-KC I agree with your suggested approach.

docs/ashrae_90p1_2019/section6/Rule6-7.md
```diff
@@ -12,9 +12,12 @@
 **Evaluation Context:** Each Data Element  
 **Data Lookup:** None  
 ## Rule Logic: 
+- List lighting space types that have no daylighting control requirements: `not_applicable_space_types = ["DORMITORY_LIVING_QUARTERS", "FIRE_STATION_SLEEPING_QUARTERS", "HEALTHCARE_FACILITY_OPERATING_ROOM", "OUTPATIENT_HEALTH_CARE_FACILITIES_CLASS_1_IMAGING_ROOMS", "DWELLING_UNIT", "GUEST_ROOM", "STORAGE_ROOM_SMALL", "PARKING_AREA_INTERIOR"]`
```
supriyagoel, line 15: @JacksonJ-KC I do not see this space type exception in G3.1.-6(h). What am I missing?

supriyagoel: SG understands why this was done. suggest adding reference to relevant table in Section 9 which states that daylight dimming controls are not required for those space types

supriyagoel (approved):
@JacksonJ-KC small comment to add reference to the Section 9 table which states that those space types are not required to have daylighting control

claperle (approved):

weilixu:
Addressed in https://github.com/pnnl/ruleset-checking-tool/pull/1925

PR #1739, opened 2025-06-17
Updates to allow running rule tests

JacksonJ-KC:
I was not able to run section 1 ruletests without making these changes first. I suspect these updates will be needed for all other section updates too.

### DIFF
--- a/rct229/rule_engine/ruleset_model_factory.py
+++ b/rct229/rule_engine/ruleset_model_factory.py
@@ -51,10 +51,15 @@ def get_rmd_instance():
 
     """
     rmd = RuleSetModels()
-    ruleset_model_types_enum = SchemaEnums.schema_enums[
+    common_model_enum = SchemaEnums.schema_enums["CommonRulesetModelOptions"]
+    ruleset_specific_enum = SchemaEnums.schema_enums[
         ruleset_model_dict[SchemaStore.SELECTED_RULESET]
     ]
-    for ruleset_model in ruleset_model_types_enum.get_list():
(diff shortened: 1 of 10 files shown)


yunjoonjung-PNNL (approved):
Thank you Jackson!

yunjoonjung-PNNL:
> It seems like we also need the 0.1.7 extra schema added to the data folder. Can anyone please share the script that you use to convert from the extra schema yaml to the json format that is being used? @weilixu @yunjoonjung-PNNL @jugonzal07

I'm not aware of the script. Does anyone know of it?

jugonzal07:
> It seems like we also need the 0.1.7 extra schema added to the data folder. Can anyone please share the script that you use to convert from the extra schema yaml to the json format that is being used? @weilixu @yunjoonjung-PNNL @jugonzal07

Hi Jackson,

Thanks for getting this setup for folks-- I noticed these issues as well when trying to run ruletests during the meeting. 

Regarding your question-- I always just grab the raw JSON from:

https://github.com/open229/ruleset-model-description-schema/tree/develop/docs229 

I'll be busy for the afternoon and can't check this out but I hope this file is what you need?

https://github.com/open229/ruleset-model-description-schema/blob/develop/docs229/ASHRAE229_extra.schema.json 

-Juan

JacksonJ-KC:
On second look, the extra schema json file included in the open229 repo is not formatted the same as the extra schema json file used by the RCT. I vaguely remember hearing that you have your own process of generating this file. @weilixu can you point me in the right direction?

PR #1740, opened 2025-06-18
RCT/JDJ/section1 schema update 0.1.7

JacksonJ-KC:
Please review the `create_data` method in rule 1-1. My solution here is likely not the simplest solution available.


weilixu:
@jugonzal07 1-9 test jsons still have is_leap_year and failed validation.

jugonzal07:
> @jugonzal07 1-9 test jsons still have is_leap_year and failed validation.

right, thanks Weili. 1-9 is one of the test JSONs that lives outside the spreadsheets. I'll manually update those few in the coming day or two when i have some time.

PR #1742, opened 2025-06-19
RCT/JDJ/section5 schema update 0.1.7

JacksonJ-KC:
Can likely improve these updates by making constructions an optional argument in utility functions


weilixu:
I think the unit test cases are failing due to the schema updates. I will take a look at it.

PR #1746, opened 2025-06-19
RCT/JDJ/section18 schema update 0.1.7

JacksonJ-KC:


weilixu:
passed the local test, good work @jugonzal07 @JacksonJ-KC

PR #1777, opened 2025-07-15
ASHRAE 90.1-2022 Ruletest Updates

jugonzal07:
First cut at updating the rule test generation process to account for a new rule set. I still need to add a software testing report process for 90.1 2022 but this is more about getting the framework in place to allow for these changes.

This is a draft PR for now since I imagine we'll want to PR into a feature branch for 90.1 2022.


jugonzal07:
Closed due to incorrect base for 90.1-2022 development

PR #1778, opened 2025-07-15
ASHRAE 90.1-2022 Ruletest Updates

jugonzal07:
First cut at updating the rule test generation process to account for a new rule set. I still need to add a software testing report process for 90.1 2022 but this is more about getting the framework in place to allow for these changes.

Complexities to address:

- Do we really want a copy of everything in ashrae90.1-2022 long term? E.g., every python file, every test file? I imagine to do this right, we'd possible want to pull out a lot of the ashrae90.1-2019 logic into a general ashrae90.1 module and have 2019 and 2022 pull from it directly. This will be a lot of work.
- I did not add the ruletest JSONs for the reason described in the previous datapoint. This adds quite a bit of bulk to the codebase that's really just a duplicate of the earlier code except with a 90.1-2022 flag on the JSON.


weilixu:
> First cut at updating the rule test generation process to account for a new rule set. I still need to add a software testing report process for 90.1 2022 but this is more about getting the framework in place to allow for these changes.
> 
> Complexities to address:
> 
> * Do we really want a copy of everything in ashrae90.1-2022 long term? E.g., every python file, every test file? I imagine to do this right, we'd possible want to pull out a lot of the ashrae90.1-2019 logic into a general ashrae90.1 module and have 2019 and 2022 pull from it directly. This will be a lot of work.

- Yes, that would be a lot work - we need to restructure the architecture of the RCT to modularize the installation. I believe each ruleset should still be independent from each other, which means 2022 will have a lot duplications in 2019.

> * I did not add the ruletest JSONs for the reason described in the previous datapoint. This adds quite a bit of bulk to the codebase that's really just a duplicate of the earlier code except with a 90.1-2022 flag on the JSON.

- Ruletest is normative so I think we will need to have two copies. We can move those to a separate report and pull from there?

PR #1784, opened 2025-08-04
RCT/JDJ/simplify skylight evals

### LINKED ISSUE #1783: Building-Level outcomes evaluated for every BuildingSegment
The following 2 rules are performing calculations based on Building-level data, but giving outcomes for every BuildingSegment. The outcome and calc_vals will be the same for every BuildingSegment in a Building. It would be more concise to provide the outcome in a BuildingRule.

5-24 (PRM9012019Rule78j13): If skylight area in the proposed design is 3% or less of the roof surface, the skylight area shall be equal to that in the proposed design.

5-25 (PRM9012019Rule84u02): If the skylight area of the proposed design is greater than 3%, baseline skylight area shall be decreased in all roof components in which skylights are located to reach 3%.

JacksonJ-KC:


yunjoonjung-PNNL (approved):
I checked that the updates changed building segment level to building level and all makes sense. Good work!

PR #1795, opened 2025-08-12
RCT/JDJ/compare context bugfix

### LINKED ISSUE #1785: `compare_context_pair` incorrectly handles the various OperatingPoint data groups
The function `compare_context_pair` expects dicts to contain the 'id' key, however the schema does not allow the 'id' key for these data groups.

```
elif isinstance(index_context, list) and isinstance(compare_context, list):
    if any(isinstance(item, dict) for item in index_context):
        sorted_dict_index = sorted(
            [item for item in index_context if isinstance(item, dict)],
            key=lambda x: x["id"],
        )
        sorted_dict_compare = sorted(
            [item for item in compare_context if isinstance(item, dict)],
            key=lambda x: x["id"],
        )    

```

JacksonJ-KC:


weilixu (approved):
Thanks for catching that.

PR #1808, opened 2025-09-02
Rule Test JSON for 6-10 for ASHRAE 90.1-2022

jugonzal07:
First cut at rule test 6-10

### DIFF
--- a/rct229/rulesets/ashrae9012019/__init__.py
+++ b/rct229/rulesets/ashrae9012019/__init__.py
@@ -86,6 +86,7 @@ rules_dict = {
     "prm9012019rule66m62": "section6rule7",
     "prm9012019rule16x33": "section6rule8",
     "prm9012019rule22c86": "section6rule9",
+    "prm9012019rule95g98": "section6rule10",
     "prm9012019rule86j27": "section10rule1",
     "prm9012019rule34l50": "section10rule7",
     "prm9012019rule73m45": "section10rule10",
--- a/rct229/ruletest_engine/ruletest_jsons/ashrae9012022/LTG/rule_6_10.json
+++ b/rct229/ruletest_engine/ruletest_jsons/ashrae9012022/LTG/rule_6_10.json
@@ -235,7 +235,8 @@
                                                         "interior_lighting": [
                                                             {
                                                                 "id": "Lighting 1",
-                                                                "purpose_type": "RETAIL_DISPLAY"
+                                                                "purpose_type": "RETAIL_DISPLAY",
+                                                                "power_per_area": 5.920150729190349
                                                             }
                                                         ]
                                                     }
@@ -272,7 +273,8 @@
                                                         "interior_lighting": [
                                                             {
                                                                 "id": "Lighting 1",
-                                                                "purpose_type": "RETAIL_DISPLAY"
+                                                                "purpose_type": "RETAIL_DISPLAY",
+                                                                "power_per_area": 5.381955208354862
                                                             }
                                                         ]
                                                     }
--- a/rct229/ruletest_engine/ruletest_jsons/ashrae9012022/section6/rule_6_10.json
+++ b/rct229/ruletest_engine/ruletest_jsons/ashrae9012022/section6/rule_6_10.json
@@ -1,0 +1,292 @@
+{
+    "rule-6-10-a": {
+        "Section": 6,
+        "Rule": 10,
+        "Test": "a",
+        "standard": {
+            "ruleset": "ashrae9012022",
+            "rule_id": "6-10",
+            "ruleset_reference": "(Previously 6-14, Then 6-10) Table G3.1(6) Proposed Building Performance (i)",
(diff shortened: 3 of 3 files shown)


rct229/ruletest_engine/ruletest_jsons/ashrae9012022/LTG/rule_6_10.json
```diff
@@ -0,0 +1,292 @@
+{
+    "rule-6-10-a": {
```
yunjoonjung-PNNL, line 2: Case (a) falls into undetermined because baseline_interior_display_w < min(proposed_interior_display_w, maximum_retail_display_w). Can you slightly adjust the ```proposed_interior_display_w```?

rct229/ruletest_engine/ruletest_jsons/ashrae9012022/section6/rule_6_10.json
```diff
@@ -0,0 +1,292 @@
...
+                                ]
+                            }
+                        ],
+                        "type": "BASELINE_0"
+                    }
+                ]
+            }
+        }
+    },
+    "rule-6-10-c": {
```
yunjoonjung-PNNL: 1. Should this case be (b)? 
2. This case also falls into undetermined because ```baseline_interior_display_w``` and ```proposed_interior_display_w``` aren't the same and ```proposed_interior_display_w``` > ```minimum_retail_display_w```,

rct229/ruletest_engine/ruletest_jsons/ashrae9012022/LTG/rule_6_10.json
jugonzal07: CASE 1 reads as follows in the RDS:

`if((proposed_interior_display_w <= minimum_retail_display_w) and (baseline_interior_display_w == proposed_interior_display_w)): PASS`

Is there something that I missed on this one? The proposed is less than the minimum and baseline is equal to it. 

The area served is 1000 ft2 and minimum retail display wattage is 750 W. That means it's 0.75 W/ft2 for the minimum, we're at 0.65 W/ft2 which is less than the minimum like you point out, but it's also equal to the proposed. Would this not pass?

Let me know if you think I'm missing something.

rct229/ruletest_engine/ruletest_jsons/ashrae9012022/section6/rule_6_10.json
jugonzal07: I'll fix the test cases to a, b, and c and adjust the test! Looks like I must have goofed a column somewhere.

jugonzal07: Hmm the test appears correct however. Here is the RDS I'm following:
https://github.com/pnnl/ruleset-checking-tool/blob/d418aaf5fb1195a96eeb2bd12fdf18207b78857b/docs/section6/Rule6-10.md

`elif baseline_interior_display_w > min(proposed_interior_display_w, maximum_retail_display_w): FAIL` 

Baseline interior display is 1.4 W/ft2, proposed interior is 1.2, and maximum retail space is 1.75 W/ft2 ((1000+750)/1000).

Therefore

`1.4 > min(1.2, 1.75)`
`1.4 > 1.2` 

Am I missing something?

yunjoonjung-PNNL: Thank you for checking Juan!  In our code, we firstly check whether a case falls into undetermined in the ```manual_check_required``` function, and then pass/fail in the ```rule_check```. 

To make the case not undetermined first we should meet this logic (NOT (Case 1 and Case 2)) so, we should have values that meet ```elif baseline_interior_display_w **<** min(proposed_interior_display_w, maximum_retail_display_w)```.

Does this make sense to you?

rct229/rulesets/ashrae9012019/__init__.py
```diff
@@ -86,6 +86,7 @@
     "prm9012019rule66m62": "section6rule7",
     "prm9012019rule16x33": "section6rule8",
     "prm9012019rule22c86": "section6rule9",
+    "prm9012019rule95g98": "section6rule10",
```
yunjoonjung-PNNL, line 89: Juan - Could we remove this change? The line was added to the 2019 ```__init__``` file and I've already added a similar line in the 2022 file.

jugonzal07:
Once you're done with your review, I need to move this to the LTG directory instead of section6 too, FYI. I must have created this ruletest before we made this change.

rct229/ruletest_engine/ruletest_jsons/ashrae9012022/section6/rule_6_10.json
yunjoonjung-PNNL: After a personal call with Juan, I noticed that wrong logic in the rule. Please ignore this comment.

yunjoonjung-PNNL (approved):
Looks good Juan. Thank you!

PR #1810, opened 2025-09-03
RDS/JDJ/22 43

JacksonJ-KC:


docs/section22/Rule22-43.md
```diff
@@ -0,0 +1,30 @@
...
+**Manual Check:** No  
+**Evaluation Context:** Chiller  
+**Data Lookup:** None  
+**Function Call:** 
+None
+
+
+## Applicability Check:  
+- For each chiller in the baseline model: `for chiller in B_RMD.chillers:`
+    - if the chiller has a condensing_loop and the chiller energy source is electricity then the rule is applicable, continue to rule assertion: `if chiller.consensing_loop != NULL and chiller.energy_source_type == "ELECTRICITY": CONTINUE TO RULE LOGIC`
```
supriyagoel: i guess this is checking that the baseline is modeled with a water cooled chiller? if yes, i think that works but is there a reason why we're not checking the baseline system type?

docs/section22/Rule22-43.md
```diff
@@ -0,0 +1,30 @@
...
+
+**Rule Assertion:**
+- Case 1: the chiller's minimum part-load ratio and minimum compressor unloading ratio are both equal to 0.25: PASS `if chiller.minimum_load_ratio == 0.25 AND chiller.minimum_unload_ratio == 0.25: PASS`
+- Case 2: either the chiller's minimum part-load ratio or minimum compressor unloading ratio are not defined: UNDETERMINED `elif chiller.minimum_load_ratio == NULL OR chiller.minimum_unload_ratio == NULL: UNDETERMINED`
+- Case 3: else: FAIL: `else: FAIL`
+
```
supriyagoel: this looks good to me- though do we need to check if the curves from app J are used? the rule states 'when using appendix J'.

docs/section22/Rule22-43.md
JacksonJ-KC: I copied this from the most recent chiller rule, 22-42, but I see other rules follow the method you suggest of checking baseline system types. Do you see any reason this would make sense for 22-42? Otherwise I am thinking of updating 42 and 43 to both check baseline system types instead

docs/section22/Rule22-43.md
JacksonJ-KC: (deleted previous response) I gave this more thought, and I agree with you now. My proposed solution is to write a ruleset function that checks if the modeled performance curve align with appendix J and returns a boolean. Rule 22-42 will be adjusted to use the Baseline System type for the applicability check. This rule will have the subfunction checking appendix J added to the applicability check in addition to the Baseline System type.

docs/section22/Rule22-43.md
claperle: Discussed with Jackson. We also discussed making sure the chiller is not a process chiller.

docs/section22/Rule22-43.md
claperle: App J curves always have to be used for the baseline so why not just update the rule language accordingly to avoid this extra work. As you mention 22-42 already checks if the curves were used so I think that aspect can be excluded from this rule since App J curves are required to always be used it follows logically that the min unload ratio should always be 0.25. To me this just checks a different parameter. Let me know if I am missing something.

JacksonJ-KC: If the simulation tool can't model performance curves, they can model "identical performance" then this minimum ratio requirement would not be applicable.

claperle: Ah, okay

claperle: P.S. I like the function idea.

supriyagoel (approved):

docs/section22/Rule22-43.md
supriyagoel: ok, that makes sense.  i dont yet see the change for adding a function to check if app J is being used and can review again after that is added.

supriyagoel (commented):
need to check again after function has been added

docs/ashrae9012022/section22/Rule22-43.md
```diff
@@ -0,0 +1,40 @@
...
+**Rule Assertion:** Baseline RMD = expected value   
+**Appendix G Section:** Section G3.2.2.1 Equipment Efficiencies      
+**90.1 Section Reference:** None  
+**Data Lookup:** None  
+**Manual Check:** No  
+**Evaluation Context:** Chiller  
+**Data Lookup:** None  
+**Function Call:** 
+1. is_chiller_performance_app_j()
+
```
claperle, line 17: You use other functions as well, get_primary_secondary_loops_dict, get_baseline_system_types

docs/ashrae9012022/section22/Rule22-43.md
```diff
@@ -0,0 +1,40 @@
...
+- Check if B-RMD is modeled with at least one air-side system that is Type-7, 8, 11.1, 11.2, 12, 13, 7b, 8b, 11b, 12b, i.e. with air-side system served by chiller(s), continue to the next applicability check: `if any(sys_type in baseline_hvac_system_dict.keys() for sys_type in ["SYS-7", "SYS-8", "SYS-11.1", "SYS-11.2", "SYS-12", "SYS-13", "SYS-7B", "SYS-8B", "SYS-11B", "SYS-12B"]): NEXT_APPLICABILITY_CHECK`
+
+- Else, rule is not applicable to B-RMD: `else: RULE_NOT_APPLICABLE`
+
+- Get primary and secondary loops for B-RMD: `primary_secondary_loop_dictionary = get_primary_secondary_loops_dict(B_RMD)`
+
+- For each chiller in B_RMD: `for chiller in B_RMD...chillers:`
+
+  - Check if chiller is connected to a primary chilled water loop that serves HVAC systems and has performance that aligns with App J: `if chiller.cooling_loop.id in primary_secondary_loop_dictionary and is_chiller_performance_app_j(chiller): CHECK_RULE_LOGIC`
```
claperle: Just want to confirm it is intentional to use is_chiller_performance_app_j versus does_chiller_performance_match_curve

JacksonJ-KC: I could've sworn I made this update already.. thanks for catching this

supriyagoel (approved):
RDS Looks good!

weilixu:
Closing this PR as the 22-43 RDS is added to the new structure in this PR: https://github.com/pnnl/ruleset-checking-tool/pull/1901/commits/61917036a8d0662673527d81838f5556584fbb37

PR #1813, opened 2025-09-04
RDS/JDJ/12 3 exclude non floor area spaces

JacksonJ-KC:


JacksonJ-KC:
These updates are on hold until further discussion in the RDS group

PR #1815, opened 2025-09-05
RS/YJ/Rule 6-13 (2022) (old: 6-12)

yunjoonjung-PNNL:


JacksonJ-KC:
FYI - This RDS was mistakenly labeled as a 2022 rule, but it should be in both 2019 and 2022.

JacksonJ-KC:
This is to become 6-13

rct229/rulesets/ashrae9012019/__init__.py
```diff
@@ -86,6 +86,7 @@
     "prm9012019rule66m62": "section6rule7",
     "prm9012019rule16x33": "section6rule8",
     "prm9012019rule22c86": "section6rule9",
+    "PRM9012019Rule86d29": "section6rule12",
```
weilixu: Could we set this rule id to lower cases? Does not seem it follows the convention.

yunjoonjung-PNNL: Good catch. Addressed!

yunjoonjung-PNNL:
As discussed during the stand up, I've removed the 6-12-2019 rule.

JacksonJ-KC:
> As discussed during the stand up, I've removed the 6-12-2019 rule.

@weilixu @yunjoonjung-PNNL during the call earlier I did not visit the PR myself to confirm, sorry. Now I am seeing the rule language is in both 2019 and 2022 which explains my previous comment here. Do we want to renumber so that this rule has the same number in both rulesets?

yunjoonjung-PNNL:
> > As discussed during the stand up, I've removed the 6-12-2019 rule.
> 
> @weilixu @yunjoonjung-PNNL during the call earlier I did not visit the PR myself to confirm, sorry. Now I am seeing the rule language is in both 2019 and 2022 which explains my previous comment here. Do we want to renumber so that this rule has the same number in both rulesets?

Thanks for double-checking. I've corrected the rule no. and can you check on this?

weilixu:
@yunjoonjung-PNNL I thought this rule is only in 2022?

yunjoonjung-PNNL:
> FYI - This RDS was mistakenly labeled as a 2022 rule, but it should be in both 2019 and 2022.

@weilixu - Based on this comment, I added the same rule for 2019. Can @JacksonJ-KC confirm this?

JacksonJ-KC:
@weilixu Confirmed, this should be in both 2019 and 2022

weilixu:
@JacksonJ-KC I am confused - this rule number is 6-13, in 2019, we only have up to 6-9, I know there is a new 6-10 rule. If this rule should be in both 2019 and 2022, then should we renumber it to 6-11?

JacksonJ-KC:
I was thinking that too @weilixu which would trigger renumbers for a couple of other rules also under development for 2022. Maybe there is a simpler temporary solution to consider because I think other sections also need renumbers too (section 4, 10, 11)

rct229/rulesets/ashrae9012019/section6/section6rule13.py
```diff
@@ -0,0 +1,120 @@
...
+
+    class BuildingRule(RuleDefinitionListIndexedBase):
+        def __init__(self):
+            super(PRM9012019rule86d29.BuildingRule, self).__init__(
+                rmds_used=produce_ruleset_model_description(
+                    USER=False, BASELINE_0=True, PROPOSED=False
+                ),
+                each_rule=PRM9012019rule86d29.BuildingRule.SpaceRule(),
+                index_rmd=BASELINE_0,
+                list_path="$.building_segments[*].zones[*].spaces[*]",
```
JacksonJ-KC, line 42: add list filter to only include spaces that have interior lighting?

rct229/rulesets/ashrae9012019/section6/section6rule13.py
```diff
@@ -0,0 +1,120 @@
...
+                ) or (
+                    space_function_b == SPACE_FUNCTION.LABORATORY
+                    and lighting_space_type_b
+                    == LIGHTING_SPACE.CLASSROOM_LECTURE_HALL_TRAINING_ROOM_SCHOOL
+                )
+
+            def get_calc_vals(self, context, data=None):
+                space_b = context.BASELINE_0
+
+                occupancy_sensor_controls_b = [
```
JacksonJ-KC, line 91: could use getattr_ here instead of manual_check_required

rct229/rulesets/ashrae9012019/section6/section6rule13.py
```diff
@@ -0,0 +1,120 @@
...
+                    interior_lighting_b.get("occupancy_control_type")
+                    for interior_lighting_b in space_b["interior_lighting"]
+                ]
+                occupancy_sensor_schedules_b = [
```
JacksonJ-KC, line 99: could use getattr_ here instead of manual_check_required

rct229/rulesets/ashrae9012019/section6/section6rule13.py
```diff
@@ -0,0 +1,120 @@
...
+                occupancy_sensor_schedules_b = calc_vals["occupancy_sensor_schedules_b"]
+
+                return any(val is None for val in occupancy_sensor_controls_b) or any(
+                    val is None for val in occupancy_sensor_schedules_b
+                )
+
+            def rule_check(self, context, calc_vals=None, data=None):
+                occupancy_sensor_controls_b = calc_vals["occupancy_sensor_controls_b"]
+
+                return not any(
```
JacksonJ-KC, line 117: need to verify that all values in occupancy_sensor_schedules_b are True.

rct229/rulesets/ashrae9012019/section6/section6rule13.py
weilixu: That is a good suggestion

rct229/rulesets/ashrae9012019/section6/section6rule13.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section6/section6rule13.py
yunjoonjung-PNNL: Addressed.

rct229/rulesets/ashrae9012019/section6/section6rule13.py
yunjoonjung-PNNL: Good point. Addressed.

rct229/rulesets/ashrae9012019/section6/section6rule13.py
yunjoonjung-PNNL: Addressed.

yunjoonjung-PNNL:
@JacksonJ-KC - I've addressed your comments. Can you please check all were addressed correctly? The testing currently fails, but if the rule is correct, I'll reach out to @jugonzal07 for the testing file.

rct229/rulesets/ashrae9012019/section6/section6rule13.py
JacksonJ-KC: Check this again. The `not any` was correct here before. There should be `and` to verify that all values in `occupancy_sensor_schedules_b` are True

yunjoonjung-PNNL: Great - Corrected.

JacksonJ-KC (approved):
Approved!

weilixu:
Approved depending on the conflicts fix.

weilixu (approved):

PR #1817, opened 2025-09-05
RCT/JDJ/12-2 undetermined bugfix

### LINKED ISSUE #1733: Rule 12-2 provides an invalid message when automatic_controlled_percentage is not defined
When the automatic controlled percentage is not defined, it should not raise `MANUAL_CHECK_REQUIRED_MSG_CASE_7`. Case 7 message should only be raised if it was confirmed that the proposed EFLH are less than the baseline EFLH.

https://github.com/pnnl/ruleset-checking-tool/blob/fee2263ffc4b2b3690bb443dd2a686f0a543d916/rct229/rulesets/ashrae9012019/section12/section12rule2.py#L199-L203

This rule logic will also need to be updated as part of the updates to use schema 0.1.7.

JacksonJ-KC:


yunjoonjung-PNNL (approved):
Looks good to me!

PR #1818, opened 2025-09-05
RCT/JDJ/19-16 undetermined bugfix

### LINKED ISSUE #1732: Rule 19-16: Undetermined outcomes for systems 9&10 that do not have cooling systems
System types 9 and 10 will rarely have cooling systems, and this rule only applies to system types 9 & 10 "if the proposed design includes a fan or fans sized and controlled to provide non-mechanical cooling". I don't think this should be giving undetermined outcomes in cases where the proposed RPD does not have a cooling system, instead it should be not applicable.

https://github.com/pnnl/ruleset-checking-tool/blob/fee2263ffc4b2b3690bb443dd2a686f0a543d916/rct229/rulesets/ashrae9012019/section19/section19rule16.py#L84

JacksonJ-KC:


yunjoonjung-PNNL (approved):
Looks good to me. Thank you!

PR #1819, opened 2025-09-05
RS/YJ/Rule_22-42 (2022)

yunjoonjung-PNNL:


yunjoonjung-PNNL:
@weilixu - I added rule 12-5-2019 from the develop branch to prevent an CI/CD error. When we merge the 12-5-2019 rule, we added the rule to the develop branch and added its testing JSON file to this feature branch (ashrae-9012022), which is why I added 12-5-2019 rule in this PR.

@jugonzal07 - While testing, I noticed that 12-5-**2019** and 12-5-**2022** testing JSON files differ in PR #1860, although the two rules are identical. I made the 2019 JSON to match the 2022 version. If this works for you, could you update the spreadsheet?

weilixu (approved):

jugonzal07:
> @weilixu - I added rule 12-5-2019 from the develop branch to prevent an CI/CD error. When we merge the 12-5-2019 rule, we added the rule to the develop branch and added its testing JSON file to this feature branch (ashrae-9012022), which is why I added 12-5-2019 rule in this PR.
> 
> @jugonzal07 - While testing, I noticed that 12-5-**2019** and 12-5-**2022** testing JSON files differ in PR #1860, although the two rules are identical. I made the 2019 JSON to match the 2022 version. If this works for you, could you update the spreadsheet?

Thanks Yun Joon! I have gone ahead and done that.

PR #1825, opened 2025-09-15
RS/YJ/Rule_22-43 (2022)

yunjoonjung-PNNL:
I haven't done with this rule.


weilixu (commented):
This looks good to me, but I think we are waiting for the `is_chiller_performance_app_j`?

yunjoonjung-PNNL:
> This looks good to me, but I think we are waiting for the `is_chiller_performance_app_j`?

Yes - this rule needs the ```is_chiller_performance_app_j``` function and I'll make a PR for it before our next standup.

weilixu (approved):

PR #1827, opened 2025-09-23
Rule Test JSON 5-44 for ASHRAE 90.1-2022

jugonzal07:
First cut at rule 5-44 rule test for ASHRAE 90.1-2022

### DIFF
--- a/rct229/ruletest_engine/ruletest_jsons/ashrae9012022/ENV/rule_5_44.json
+++ b/rct229/ruletest_engine/ruletest_jsons/ashrae9012022/ENV/rule_5_44.json
@@ -14,7 +14,7 @@
             "primary_rule": "Full",
             "schema_version": "0.1.7"
         },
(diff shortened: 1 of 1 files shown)


rct229/ruletest_engine/ruletest_jsons/ashrae9012022/ENV/rule_5_44.json
```diff
@@ -0,0 +1,148 @@
+{
+    "rule-5-44-a": {
```
yunjoonjung-PNNL, line 2: This case failed because Surface 1 is Unregulated type. Can you double-check on this?

yunjoonjung-PNNL: Case (b)  failed as we expected, but the surface 1 was also unregulated type. Should we also fix this in case (b)?

yunjoonjung-PNNL:
Looks Good! Thanks Juan!

PR #1828, opened 2025-09-23
Rule Test JSON 5-45 for ASHRAE 90.1-2022

jugonzal07:
First cut at 5-45 rule test JSON for 90.1-2022


rct229/ruletest_engine/ruletest_jsons/ashrae9012022/ENV/rule_5_45.json
```diff
@@ -0,0 +1,148 @@
+{
+    "rule-5-45-a": {
```
yunjoonjung-PNNL, line 2: Juan - The surface 1 in both case (b) and (b) is unregulated type. Could we make the surface not unregulated?

yunjoonjung-PNNL (approved):
Looks good. Thank you!

PR #1830, opened 2025-09-24
New Rule 6-10

claperle:
Checks if lighting and misc were modeled identically across base and proposed for plenums, crawl spaces, and interstitial spaces. This also needs to be added to 90.1 2019.

PR #1843, opened 2025-10-07
Rule Test JSON 6-10 Update for ASHRAE 90.1-2022

jugonzal07:
Updated rule 6-10 rule test to match latest RDS for ASHRAE 90.1-2022:

https://github.com/pnnl/ruleset-checking-tool/blob/4ce39c397368e68a0e7bd9c502b9c279a81b54a4/docs/section6/Rule6-10.md


yunjoonjung-PNNL:
@jugonzal07  - my understanding is that we have all the keys specified in the RDS (e.g., purpose_type, occupancy_control_type, daylighting_control_type, etc.) and I found that some of these are missing. Could you check if you understanding is the same?

JacksonJ-KC:
I am not sure which rule description these updates were intended for. There have been multiple rules called 6-10 recently. If it is for [this RDS](https://github.com/pnnl/ruleset-checking-tool/pull/1830), we should close this PR.

jugonzal07:
> I am not sure which rule description these updates were intended for. There have been multiple rules called 6-10 recently. If it is for [this RDS](https://github.com/pnnl/ruleset-checking-tool/pull/1830), we should close this PR.

Hi Jackson,

Looking at my comment above, I believe the intent was for this to line up with the RDS you describe here. I must have jumped the gun on this back in October before we had determined how we wanted to handle the plenum/crawlspace space RDSs. Did we arrive at a conclusion for these rules?

weilixu:
Close this PR because 6-10 is no longer needed.

PR #1845, opened 2025-10-07
RS/YJ/Rule 6-10 (2022)

yunjoonjung-PNNL:
I haven't tested this rule yet. Thanks!

### DIFF
--- a/rct229/rulesets/ashrae9012022/section6/section6rule10.py
+++ b/rct229/rulesets/ashrae9012022/section6/section6rule10.py
@@ -35,14 +35,14 @@ class PRM9012022rule12d80(RuleDefinitionListIndexedBase):
         ltg_schedule_b = {
             int_ltg_sch_id_b: find_exactly_one_schedule(rmd_b, int_ltg_sch_id_b)
             for int_ltg_sch_id_b in find_all(
-                "$.buildings[*].building_segments[*].zones[*].spaces[*].interior_lighting[*].ltg_multiplier_schedule",
+                "$.buildings[*].building_segments[*].zones[*].spaces[*].interior_lighting[*].lighting_multiplier_schedule",
                 rmd_b,
             )
         }
         ltg_schedule_p = {
-            int_ltg_sch_id_p: find_exactly_one_schedule(rmd_b, int_ltg_sch_id_p)
+            int_ltg_sch_id_p: find_exactly_one_schedule(rmd_p, int_ltg_sch_id_p)
             for int_ltg_sch_id_p in find_all(
-                "$.buildings[*].building_segments[*].zones[*].spaces[*].interior_lighting[*].ltg_multiplier_schedule",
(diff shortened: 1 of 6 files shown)


rct229/rulesets/ashrae9012022/section6/section6rule10.py
```diff
@@ -0,0 +1,347 @@
...
+            ltg_schedule_p = data["ltg_schedule_p"]
+            equip_schedule_b = data["equip_schedule_b"]
+            equip_schedule_p = data["equip_schedule_p"]
+
+            ltg_purpose_type_match = True
+            ltg_power_per_area_match = True
+            ltg_occupancy_control_type_match = True
+            ltg_daylighting_control_type_match = True
+            ltg_ltg_multiplier_schedule_match = True
+            for int_ltg_p in space_p["interior_lighting"]:
```
weilixu, line 124: I think you need another one to make sure the length matched.

for example:

len(space_p["interior_lighting"]) == len(space_b["interior_lighting"])

yunjoonjung-PNNL: I'd like to double-check - My understanding is that we only compare the keys specified in the RDS (e.g., ```purpose_type```,  ```power_per_area```, etc.) and do not need compare the keys not specified in the RDS (e.g., ```are_schedules_used_for_modeling_occupancy_control```, ```are_schedules_used_for_modeling_daylighting_control```) under the ```interior_lighting```.

yunjoonjung-PNNL: And, thus do the lengths not need to be the same?

weilixu: The rule assertion is at space data group level. I think the first thing need to make sure is the proposed has the same number of lighting, equipment etc as baseline. This is not checked in the [proposed-baseline](https://github.com/pnnl/ruleset-checking-tool/blob/master/rct229/rulesets/ashrae9012019/data/ASHRAE229.9012019.extra.schema.json#L1136). With this code:
```python
for int_ltg_p in space_p["interior_lighting"]:
    int_ltg_b = find_exactly_one_with_field_value(
                    "$.interior_lighting[*]", "id", int_ltg_p["id"], space_b
                )
```
I think it will still pass if baseline just add one more lighting object in its "interior_lighting", which should be incorrect.

rct229/rulesets/ashrae9012022/section6/section6rule10.py
```diff
@@ -0,0 +1,347 @@
+from rct229.rule_engine.rule_base import RuleDefinitionBase
```
weilixu, line 1: Should this rule also included in the 2019?

weilixu:
Copy 6-10 to a separate branch based on develop.
Then close this PR.

yunjoonjung-PNNL:
Closed b/c this rule is for both 2019 and 2022.

PR #1846, opened 2025-10-08
RDS New Rule Section 12 - 2019 & 2022

JacksonJ-KC:
This new rule is for both 2019 and 2022

### DIFF
--- a/docs/section12/Rule12-5.md
+++ b/docs/section12/Rule12-5.md
@@ -1,0 +1,37 @@
+# Receptacle - Rule 12-5
+**Schema Version** 0.1.7  
+**Primary Rule:** True  
+**Rule ID:** 12-5  
+**Rule Description:** These loads shall always be included in simulations of the building. These loads shall be included when calculating the proposed building performance and the baseline building performance as required by Section G1.2.1.   
+**Appendix G Section:** Table G3.1-12 Proposed Building Performance column  
+**Appendix G Section Reference:** None  
+
+**Applicability:** This rule applies to all proposed models.
+**Applicability Checks:**
+1. Always applicable  
+
+**Manual Check:** None  
+**Evaluation Context:** P_RMD    
+**Data Lookup:** None  
+**Function Call:**
+1. get_schedule_multiplier_hourly_value_or_default
+
+
+## Rule Logic:  
+- Iterate through the miscellaneous equipment data groups in the proposed model: `for misc_equipment_p in P_RMD...miscellaneous_equipment:` 
+  - Get the equivalent full load hours (EFLH) for the assigned schedule: `schedule_eflh = sum(get_schedule_multiplier_hourly_value_or_default(P_RMD, misc_equipment_p["multiplier_schedule"], 0))`
+  - Determine if the miscellaneous equipment power is greater than 0 and has either sensible or latent fractions greater than 0, and the assigned schedule has more than 0 EFLH then the loads are included in the building: `loads_included = misc_equipment_p.power > 0 and (misc_equipment_p.sensible_fraction > 0 or misc_equipment_p.latent_fraction > 0) and schedule_eflh > 0`
+  - End iteration early if a load is included: `if loads_included: break`
+- Iterate through the proposed output instance end use results: 
+  - Determine if the proposed output instance has an end use result with type MISC_EQUIPMENT, INDUSTRIAL_PROCESS, OFFICE_EQUIPMENT, COMPUTER_SERVERS, or COMMERCIAL_COOKING and greater than 0 annual site energy use: `has_annual_energy_use = any(result for result in P_RMD.model_output.annual_end_use_results if result.type in ["MISC_EQUIPMENT", "INDUSTRIAL_PROCESS", "OFFICE_EQUIPMENT", "COMPUTER_SERVERS", "COMMERCIAL_COOKING"] and result.annual_site_energy_use > 0)`
+  - End iteration early if a matching end use result is found: `if has_annual_energy_use: break`
+
+  - **Rule Assertion:**  
+      - Case 1: If the proposed model has miscellaneous equipment loads and end use results: PASS `if loads_included and has_annual_energy_use:` PASS`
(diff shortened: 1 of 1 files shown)


docs/section12/Rule12-5.md
```diff
@@ -0,0 +1,37 @@
+# Receptacle - Rule 12-5
+**Schema Version** 0.1.7  
+**Primary Rule:** True  
+**Rule ID:** 12-5  
+**Rule Description:** These loads shall always be included in simulations of the building. These loads shall be included when calculating the proposed building performance and the baseline building performance as required by Section G1.2.1.
```
claperle: Recommend replacing "these" with "receptacle" (at least in the first word) just so there is no confusion about what you're referring to in the rule description.

Another question, do we mean receptacle and process loads? I recommend that we be specific about what we are referring to in this rule.

docs/section12/Rule12-5.md
```diff
@@ -0,0 +1,37 @@
...
+  - End iteration early if a load is included: `if loads_included: break`
+- Iterate through the proposed output instance end use results: 
+  - Determine if the proposed output instance has an end use result with type MISC_EQUIPMENT, INDUSTRIAL_PROCESS, OFFICE_EQUIPMENT, COMPUTER_SERVERS, or COMMERCIAL_COOKING and greater than 0 annual site energy use: `has_annual_energy_use = any(result for result in P_RMD.model_output.annual_end_use_results if result.type in ["MISC_EQUIPMENT", "INDUSTRIAL_PROCESS", "OFFICE_EQUIPMENT", "COMPUTER_SERVERS", "COMMERCIAL_COOKING"] and result.annual_site_energy_use > 0)`
+  - End iteration early if a matching end use result is found: `if has_annual_energy_use: break`
+
+  - **Rule Assertion:**  
+      - Case 1: If the proposed model has miscellaneous equipment loads and end use results: PASS `if loads_included and has_annual_energy_use:` PASS`
```
claperle: So there's a pretty low bar to meeting this check in that as long as there is a single modeled miscellaneous load the project passes? I can't really think of an alternative because sometimes spaces don't have miscellaneous loads so we can't really evaluate this on a space method either.

docs/section12/Rule12-5.md
claperle: Can you confirm that there's another rule that conducts the same check on the baseline? I recall a check that checks whether these are modeled identically across the baseline and proposed perhaps it's accomplished through that RDS.

docs/section12/Rule12-5.md
```diff
@@ -0,0 +1,37 @@
...
+      - Case 2: Else: FAIL and let the user know which part(s) caused the failure: `else: outcome = FAIL and raise_message(f"{'No miscellaneous equipment loads are included. [power: ' + f'{misc_equipment_p.power}, sensible_fraction: {misc_equipment_p.sensible_fraction}, latent_fraction: {misc_equipment_p.latent_fraction}, schedule_eflh: {schedule_eflh}] ' if not loads_included else ''}{'No annual end use energy is reported for the relevant equipment types.' if not has_annual_energy_use else ''}")"`
+
+
+**Notes:**  
+1. Agree on the MiscellaneousEquipment types that are included? MISC_EQUIPMENT, INDUSTRIAL_PROCESS, OFFICE_EQUIPMENT, COMPUTER_SERVERS, or COMMERCIAL_COOKING
+
```
claperle, line 43: I think this depends what we're targeting with this RDS. Are we targeting both receptacle and process loads? I assume we are. If that is the case, I think what you have included makes sense.

I was tempted to suggest we include refrigeration equipment but that is covered by another section so I think makes sense to exclude it here.

docs/section12/Rule12-5.md
```diff
@@ -0,0 +1,37 @@
...
+  - End iteration early if a matching end use result is found: `if has_annual_energy_use: break`
+
+  - **Rule Assertion:**  
+      - Case 1: If the proposed model has miscellaneous equipment loads and end use results: PASS `if loads_included and has_annual_energy_use:` PASS`
+      - Case 2: Else: FAIL and let the user know which part(s) caused the failure: `else: outcome = FAIL and raise_message(f"{'No miscellaneous equipment loads are included. [power: ' + f'{misc_equipment_p.power}, sensible_fraction: {misc_equipment_p.sensible_fraction}, latent_fraction: {misc_equipment_p.latent_fraction}, schedule_eflh: {schedule_eflh}] ' if not loads_included else ''}{'No annual end use energy is reported for the relevant equipment types.' if not has_annual_energy_use else ''}")"`
```
claperle: Recommend being consistent in the nomenclature. If this refers to receptacle and process equipment I recommend that we say this instead of miscellaneous equipment loads.

yunjoonjung-PNNL: I have a confusion on the message. For example, in this part (```if not loads_included else ''```),  ```loads_included``` is a Boolean variable. So, did you mean that if ```loads_included``` is False print out ```False``` in the message?

JacksonJ-KC: The rule can fail if either no loads are included, or no end-uses are included. The message is intended to tell the use which part(s) caused the failure. 

when loads_included is False: 'No miscellaneous equipment loads are included. [power: misc_equipment_p.power, sensible_fraction: misc_equipment_p.sensible_fraction, latent_fraction: misc_equipment_p.latent_fraction, schedule_eflh: schedule_eflh]'

when has_annual_energy_use is False: 'No annual end use energy is reported for the relevant equipment types.'

docs/section12/Rule12-5.md
JacksonJ-KC: Think we should also check baseline in this rule? I don't think this would be covered by B=P rule

JacksonJ-KC: The rule description does also mention the baseline so it makes sense to combine I think

claperle: @JacksonJ-KC I think it is covered by 12-1, https://github.com/pnnl/ruleset-checking-tool/blob/develop/docs/section12/Rule12-1.md

Logic, does proposed have receptacles? Flags if there is a problem.
12-1 checks that baseline and proposed match so the check on the proposed would inherently be covered.

JacksonJ-KC: Ok so the power check is covered but not the end-use check for the baseline.

claperle: That's true.

docs/section12/Rule12-5.md
supriyagoel: I looked up what all should be included and 90.1 is not clear on what in included in process load/energy/application. Anything that is not H/C/lighting could be considered process. In either case, this looks good to me

supriyagoel (approved):
RDS looks good

PR #1860, opened 2025-10-23
Rule Test JSON for 12-5 for ASHRAE 90.1-2019 and 90.1-2022

jugonzal07:
First cut at a Section 12 Rule 5 rule test JSON for both 90.1-2019 and 90.1-2022


yunjoonjung-PNNL:
Juan - Good work. Can you please check the following?

1) The ```annual_site_energy_use``` object shouldn't be under the ```output_instance``` object. Can we remove the ```output_instance``` object?
2) There is a space `at the end of the ``annual_site_energy_use``` object. can we remove the space?
3) We need the ```latent_fraction``` key.

jugonzal07:
> Juan - Good work. Can you please check the following?
> 
> 1. The `annual_site_energy_use` object shouldn't be under the `output_instance` object. Can we remove the `output_instance` object?
> 2. There is a space `at the end of the ``annual_site_energy_use``` object. can we remove the space?
> 3. We need the `latent_fraction` key.

Thanks Yun Joon,

It looks like things changed a bit since I wrote this test. Can you explain point 1? That seems to be where the [RPD suggests it goes](https://github.com/pnnl/ruleset-checking-tool/blob/b8f61c2cd028eac95466a7089de8cbfe95a1633a/docs/section12/Rule12-5.md#:~:text=has_annual_energy_use_p%20%3D%20any(result%20for%20result%20in-,P_RMD.model_output.annual_end_use_results,-if%20result.type%20in%20%5B%22MISC_EQUIPMENT%22%2C%20%22INDUSTRIAL_PROCESS ) 

2) Easy enough to fix
3) I will add that to 0

Also, I believe the updated RDS requires that I add a baseline as well. I'll look to do that.

yunjoonjung-PNNL:
@jugonzal07 - Could you see if the following makes sense to you?

1. Case (b) Proposed - We need the ```multiplier_schedule``` and ```power``` keys under the ```miscellaneous_equipment```.
2. Case (C) Proposed - We need the ```type``` key that isn't one of the ```["MISC_EQUIPMENT", "INDUSTRIAL_PROCESS", "OFFICE_EQUIPMENT", "COMPUTER_SERVERS", "COMMERCIAL_COOKING"]```.

After changing this, all cases passed.

jugonzal07:
> @jugonzal07 - Could you see if the following makes sense to you?
> 
> 1. Case (b) Proposed - We need the `multiplier_schedule` and `power` keys under the `miscellaneous_equipment`.
> 2. Case (C) Proposed - We need the `type` key that isn't one of the `["MISC_EQUIPMENT", "INDUSTRIAL_PROCESS", "OFFICE_EQUIPMENT", "COMPUTER_SERVERS", "COMMERCIAL_COOKING"]`.
> 
> After changing this, all cases passed.

For 1) Ok-- i added a power of 0 then. I left it as nil which i would think would be the same thing but I suppose not? Let me know if you think differently here.

For 2) This one's a bit different. My test was not testing the failure of the wrong use type. It was if `annual_site_energy_use` was set to nil. I updated it to equal 0. I'll create a test (d) that checks the `type` key too.

yunjoonjung-PNNL:
Every case passed. Thank you Juan!

PR #1824, opened 2025-09-12
Rule Test JSON 22-43 for 90.1-2022

jugonzal07:
First cut at rule test 22-43 rule test for 90.1-2022

PR #1863, opened 2025-10-27
RS/YJ/Rule_12-5 (2022)

yunjoonjung-PNNL:


weilixu:
Dependency on #1846

rct229/rulesets/ashrae9012022/section12/section12rule5.py
```diff
@@ -0,0 +1,145 @@
...
+                            misc_equip_p,
+                            "miscellaneous_equipment",
+                            "multiplier_schedule",
+                        ),
+                    )
+                    for misc_equip_p in find_all(
+                        "$.buildings[*].building_segments[*].zones[*].spaces[*].miscellaneous_equipment[*]",
+                        rmd_p,
+                    )
+                ][0],
```
weilixu: Why there is [0] at the end of the list? Also, the sum here will just flatten this 2D array, instead of sum it into a value.

I think this `schedule_eflh_p` shall be a dictionary, in which the key is the schedule ID and the value is the sum of the hourly value?

rct229/rulesets/ashrae9012022/section12/section12rule5.py
```diff
@@ -0,0 +1,145 @@
...
+
+            def get_calc_vals(self, context, data=None):
+                misc_equip_p = context.PROPOSED
+                has_annual_energy_use_p = data["has_annual_energy_use_p"]
+                schedule_eflh_p = data["schedule_eflh_p"]
+
+                loads_included_p = (
+                    misc_equip_p["power"] > 0 * ureg("W")
+                    and misc_equip_p["sensible_fraction"] > 0
+                    and misc_equip_p["latent_fraction"] > 0
```
weilixu: In RDS, it says: `and (misc_equipment_p.sensible_fraction > 0 or misc_equipment_p.latent_fraction > 0) `

weilixu (commented):
Left comments to address

rct229/rulesets/ashrae9012022/section12/section12rule5.py
yunjoonjung-PNNL: Addressed.

weilixu:
Is this ready for review? @yunjoonjung-PNNL

yunjoonjung-PNNL:
> Is this ready for review? @yunjoonjung-PNNL

@weilixu - No not yet because I need to test the rule, but I've got an error running the run_ruletests.py script. I'll ask the issue during our standup today and finalize this rule.

rct229/rulesets/ashrae9012022/section12/section12rule5.py
yunjoonjung-PNNL: Good point. Addressed

yunjoonjung-PNNL:
@weilixu  - This rule has passed the testing JSON files from #1824.

PR #1864, opened 2025-10-28
Rule Test JSON for 5-42 in ASHRAE 90.1-2022

jugonzal07:
First cut at 5-42 rule test JSON

### DIFF
--- a/rct229/ruletest_engine/ruletest_jsons/ashrae9012022/ENV/rule_5_42.json
+++ b/rct229/ruletest_engine/ruletest_jsons/ashrae9012022/ENV/rule_5_42.json
@@ -8,7 +8,7 @@
             "rule_id": "5-42",
             "ruleset_reference": "Section G3.1-5 Building Envelope Modeling Requirements for the Proposed building",
             "rule_description": " Each linear thermal bridge and point thermal bridge as identified in Section 5.5.5 shall be modeled using either of the following techniques:\n\na. A separate model of the assembly within the energy simulation model.\nb. Adjustment of the clear-field U-factor in accordance with Section A10.2.",
-            "applicable_rmr": "Baseline Model",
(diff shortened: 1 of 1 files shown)


yunjoonjung-PNNL:
Could @jugonzal07 check the following comments?

1) I believe we should have ```PROPOSED``` instead of ```BASELINE_0```.
2) case (a) has a unregulated surface and case (b) has an exterior residential. Should case (a) be the ```not_applicable``` and case (b) be the```undetermined```?

jugonzal07:
> Could @jugonzal07 check the following comments?
> 
> 1. I believe we should have `PROPOSED` instead of `BASELINE_0`.
> 2. case (a) has a unregulated surface and case (b) has an exterior residential. Should case (a) be the `not_applicable` and case (b) be the`undetermined`?

Am I looking at the wrong [RDS](https://github.com/pnnl/ruleset-checking-tool/blob/dd6691cfba82672ac216efdb3840c2ace961806d/docs/section5/Rule5-42.md)?

I'm seeing most of this being specified for the proposed RMD. Admittedly, it does seem to suggest it's for the baseline in the Rule Assertion.

You're correct about 2) sorry about that, I misread the RDS the first time.

yunjoonjung-PNNL:
@jugonzal07 - Could you change the baseline to proposed? When I tested with this change, all cases passed.

jugonzal07:
> @jugonzal07 - Could you change the baseline to proposed? When I tested with this change, all cases passed.

Cool! Done. I believe the rule description was goofed, i changed the language in the Rule's applicable RMD to reflect the proposed and not the Baseline.

yunjoonjung-PNNL (approved):
Looks great - Thank you Juan. I'll merge this PR to my 5-42 PR.

PR #1869, opened 2025-10-31
RCT/JDJ/reduce repetitive undetermined

JacksonJ-KC:
This PR adds shortcuts in higher-level rules to reduce repetitive outcomes


weilixu:
Thanks for addressing the 5-11 issue.
For the rest, my interpretation is instead of reporting at component level for missing data, this change reports one general missing data undetermined result to reduce the size of reporting?
It also only applies if all components are missing the required data.

JacksonJ-KC:
Yes all these changes are intended to make the report more easily digestible for rules that have a tendency to repeat outcomes for a large number of objects in a model.

PR #1896, opened 2025-12-15
RCT/JDJ/listpath bugfix

JacksonJ-KC:


rct229/rulesets/ashrae9012019/section22/section22rule15.py
```diff
@@ -29,7 +44,23 @@ def __init__(self):
             standard_section="Section 22 CHW&CW Loop",
             is_primary_rule=True,
             rmd_context="ruleset_model_descriptions/0",
-            list_path="$.heat_rejections[*]",
```
JacksonJ-KC, line 32: I am not sure why, but removing the `$.` allowed the heat rejections to be found

JacksonJ-KC: "List path $.heat_rejections[*] is either incorrect or has no data" was the error previously

weilixu: I am not sure - the low-level function that searches the context with jsonpath is here: 
https://github.com/pnnl/ruleset-checking-tool/blob/master/rct229/utils/jsonpath_utils.py#L14

So regardless you put "$." or not, it is always set to "$.".
If removing this works for you, I am OK to merge it.

weilixu (approved):

PR #1899, opened 2025-12-15
RCT/JDJ/safe get baseline system

### LINKED ISSUE #1876: get_baseline_system_types.py `assert_`
get_baseline_system_types.py `assert_` ends rule evaluation if a single system is not matched to a baseline system type. For rules that give system-level outcomes we want this to be undetermined for the unmatched system(s), but it should continue checking other systems too.

JacksonJ-KC:
This lets calls bypass assertion when a baseline HVAC system is unmatched, allowing all outcomes to evaluate instead of ending evaluation as soon as the unmatched system is encountered.

### DIFF
--- a/rct229/rulesets/ashrae9012019/section18/section18rule2.py
+++ b/rct229/rulesets/ashrae9012019/section18/section18rule2.py
@@ -127,6 +127,29 @@ class PRM9012019Rule51v53(RuleDefinitionListIndexedBase):
                 "$.buildings[*].building_segments[*].heating_ventilating_air_conditioning_systems[*].id",
                 rmd_b,
             ):
+                sys_type = next(
+                    (
+                        sys_type
+                        for sys_type, hvac_id_list in baseline_system_types_dict_b.items()
+                        if hvac_id_b in hvac_id_list
+                    ),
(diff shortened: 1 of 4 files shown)


rct229/rulesets/ashrae9012019/section18/section18rule2.py
```diff
@@ -127,6 +127,29 @@ def create_data(self, context, data):
...
+                sys_type = next(
+                    (
+                        sys_type
+                        for sys_type, hvac_id_list in baseline_system_types_dict_b.items()
+                        if hvac_id_b in hvac_id_list
+                    ),
+                    HVAC_SYS.UNMATCHED,
+                )
+
+                if sys_type == HVAC_SYS.UNMATCHED:
```
weilixu: I am trying to understand this:
If unmatched, we want to get the info on what exact step was failed. Is that correct?
The following line seems a hard coded reporting, so just wondering.

JacksonJ-KC: you're right we should let as many of these populate as possible. I will improve this

weilixu (approved):

PR #1908, opened 2025-12-22
RCT/JDJ/reporting bugfixes

JacksonJ-KC:
This PR includes adjustment to `get_primary_secondary_loops_dict` and the rules that call that function so that the reported number of primary loops and secondary loops can be accurate. 

This removes inferred meaning that all loops returned by the function are in primary/secondary configuration, and is handled accordingly in each rule.

I did not realize how many files I would need to change before starting this work. Let me know if you would prefer I separate out the primary/secondary refactoring to a different PR.


JacksonJ-KC:
Closing to combine with other major PRs

PR #1916, opened 2025-12-23
Rule 6-8 Update: Plenum, crawl-space, interstitial, NONE

JacksonJ-KC:


weilixu:
Addressed in https://github.com/pnnl/ruleset-checking-tool/pull/1926

PR #1921, opened 2025-12-31
RCT/JDJ/design day validation

JacksonJ-KC:


rct229/schema/validate.py
```diff
@@ -434,6 +434,45 @@ def check_annual_schedule_lengths(rpd: dict) -> list[str]:
...
+    )
+    if has_cooling_design_day and has_cooling_design_year:
+        errors.append(
+            "Schedules contain both 'hourly_cooling_design_day' and 'hourly_cooling_design_year'. Only one design day method is allowed."
+        )
+    if has_heating_design_day and has_heating_design_year:
+        errors.append(
+            "Schedules contain both 'hourly_heating_design_day' and 'hourly_heating_design_year'. Only one design day method is allowed."
+        )
+    return errors
```
JacksonJ-KC, line 473: this intentionally does not compare has_cooling_design_day vs. has_heating_design_year, or has_heating_design_day vs. has_cooling_design_year. I think it is ok if the method is different for heating vs. cooling.

weilixu (approved):

PR #1852, opened 2025-10-15
Rule 5-1 Updates

JacksonJ-KC:
These updates incorporate logic to ensure the reported BBP represents the average of baseline rotated RPDs when rotations are expected and otherwise represents the baseline 0 RPD.

PR #1927, opened 2026-02-16
Rule Test JSON 5-1 update

jugonzal07:
Updated the rule test JSON based on this pull request:

https://github.com/pnnl/ruleset-checking-tool/pull/1852/changes#diff-7e5d12351a30a3e888a174ae05e77ad7e3de53bb62306ccbae018f02ede56340

NOTE: This rule test JSON is NOT part of a spreadsheet given it's unique set of orientations. This is created manually.


rct229/ruletest_engine/ruletest_jsons/ashrae9012019/ENV/rule_5_1.json
```diff
@@ -28,7 +28,8 @@
                             "unmet_load_hours_cooling": 0,
                             "unmet_occupied_load_hours_heating": 0,
                             "unmet_occupied_load_hours_cooling": 0,
-                            "building_peak_cooling_load": 25000
+                            "building_peak_cooling_load": 25000,
+                            "baseline_building_performance_energy_cost": 50000
```
yunjoonjung-PNNL: I tested this RPD against the updated 5-1 from Jackson's PR. It turned out that the RPD is invalid due to this key. I believe this key should be under the ```output``` key instead of the ```model_output``` (similar to the ```rule_1_2.json```). Can you double-check this?

jugonzal07: Thanks! I will take a look.

